PR #86, opened 2024-01-18
feat: store 404 result

alinarublea:

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/notfound/handler.js
+++ b/src/notfound/handler.js
@@ -11,10 +11,13 @@
  */
 
 import RUMAPIClient from '@adobe/spacecat-shared-rum-api-client';
-import { internalServerError, noContent } from '@adobe/spacecat-shared-http-utils';
+import { internalServerError, noContent, notFound } from '@adobe/spacecat-shared-http-utils';
+import { retrieveSiteBySiteId } from '../utils/data-access.js';
 import {
   getRUMUrl,
 } from '../support/utils.js';
+
+const AUDIT_TYPE = '404';
 
 export function filter404Data(data) {
   return data.topurl.toLowerCase() !== 'other' && !!data.source; // ignore the combined result and the 404s with no source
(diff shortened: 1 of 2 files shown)


src/notfound/handler.js
```diff
@@ -29,18 +32,50 @@ function process404Response(data) {
...
-  const { log, sqs } = context;
+  const { type, url: siteId, auditContext } = message;
+  const { log, sqs, dataAccess } = context;
   const {
     AUDIT_RESULTS_QUEUE_URL: queueUrl,
   } = context.env;
 
   try {
-    log.info(`Received audit req for domain: ${url}`);
+    log.info(`Received audit req for domain: ${siteId}`);
```
iuliag, line 71: Should it be `for site id`?

src/notfound/handler.js
```diff
@@ -29,18 +32,50 @@ function process404Response(data) {
...
+    const site = await retrieveSiteBySiteId(dataAccess, siteId, log);
+    if (!site) {
+      return notFound('Site not found');
+    }
 
     const rumAPIClient = RUMAPIClient.createFrom(context);
-    const finalUrl = await getRUMUrl(url);
+    const finalUrl = await getRUMUrl(siteId);
```
iuliag, line 78: I think here you have to pass the `site.getBaseURL()` not the site id

src/notfound/handler.js
```diff
@@ -49,15 +84,16 @@ export default async function audit404(message, context) {
 
     const data = await rumAPIClient.get404Sources(params);
     const auditResult = process404Response(data);
+    await processAuditResult(dataAccess, site, auditContext, auditResult);
 
     await sqs.sendMessage(queueUrl, {
       type,
-      url,
+      url: siteId,
```
iuliag, line 91: I think here you have to pass the `site.getBaseURL()` not the site id, judging by the unit test

src/notfound/handler.js
solaris007: also, we should ideally always log both the siteId and the baseURL - so we can more easily extract / filter on coralogix.

src/notfound/handler.js
```diff
@@ -11,11 +11,22 @@
...
+import { createUrl } from '@adobe/fetch';
+import { retrieveSiteByURL } from '../utils/data-access.js';
 import {
   getRUMUrl,
 } from '../support/utils.js';
 
+const AUDIT_TYPE = '404';
+// TODO export from spacecat-shared-rum-api-client
+const RUM_SOURCES = 'https://helix-pages.anywhere.run/helix-services/run-query@v3/rum-sources';
+export const RUM_DEFAULT_PARAMS = {
```
ekremney, line 24: rum-api-client already has `RUM_DEFAULT_PARAMS` and `RUM_SOURCES`; having two sources of truth is not good. Also this logic belongs to rum-api-client. Would you mind moving `create404Url` to rum-api-client?

ekremney (approved):
approved with a comment

src/utils/data-access.js
```diff
@@ -35,3 +35,28 @@ export async function retrieveSiteBySiteId(dataAccess, siteId, log) {
...
+ * @param {string} url - The url of the site to fetch data for.
+ * @param {Object} log - The logging object.
+ * @throws {Error} - Throws an error if the site data cannot be fetched.
+ * @returns {Promise<Object|null>} - Returns the site object if found, otherwise null.
+ */
+export async function retrieveSiteByURL(dataAccess, url, log) {
+  try {
+    const site = await dataAccess.getSiteByBaseURL(url);
+    if (!isObject(site)) {
+      log.warn(`Site not found for site: ${url}`);
```
solaris007, line 55: site for site?

solaris007 (approved):

PR #112, opened 2024-01-30
feat: introduce organic keyword audit (SITES-19317)

dzehnder:
API trigger implemented here: https://github.com/adobe/spacecat-api-service/pull/130
Post Processor implemented here: https://github.com/adobe/spacecat-audit-post-processor/pull/65


src/index.js
```diff
@@ -30,6 +31,7 @@ const HANDLERS = {
   'lhs-desktop': lhs,
   404: notfound,
   'broken-backlinks': backlinks,
+  'organic-keywords': organicKeywords,
```
iuliag, line 34: For all of the keys here, we could probably reuse the types exported from the audit model

src/support/ahrefs-client.js
```diff
@@ -88,4 +88,35 @@ export default class AhrefsAPIClient {
...
+      select: [
+        'keyword',
+        'best_position',
+        'best_position_prev',
+        'best_position_diff',
+        'sum_traffic',
+      ],
+    };
+
+    return this.sendRequest('/site-explorer/organic-keywords', queryParams);
```
iuliag, line 120: What is the cost per row for this query?

src/support/ahrefs-client.js
```diff
@@ -88,4 +88,35 @@ export default class AhrefsAPIClient {
...
+      const year = date.getFullYear();
+      const month = String(date.getMonth() + 1).padStart(2, '0');
+      const day = String(date.getDate()).padStart(2, '0');
+
+      return `${year}-${month}-${day}`;
+    };
+    const today = new Date();
+
+    const queryParams = {
+      country: 'au',
```
iuliag, line 103: Is `au` country applicable to all sites?

src/organic-keywords/handler.js
```diff
@@ -0,0 +1,75 @@
...
+  try {
+    log.info(`Received ${type} audit request for siteId: ${siteId}`);
+
+    const site = await retrieveSiteBySiteId(dataAccess, siteId, log);
+
+    if (!site) {
+      return notFound('Site not found');
+    }
+
+    const auditConfig = site.getAuditConfig();
```
iuliag, line 34: We should run the audit only for sites that are live

src/organic-keywords/handler.js
```diff
@@ -0,0 +1,75 @@
...
+      return ok();
+    }
+
+    if (auditConfig.getAuditTypeConfig(type)?.disabled()) {
+      log.info(`Audit type ${type} disabled for site ${siteId}`);
+      return ok();
+    }
+
+    const ahrefsAPIClient = AhrefsAPIClient.createFrom(context);
+    const response = await ahrefsAPIClient.getOrganicKeywords(site.getBaseURL(), auditContext);
```
iuliag, line 46: How do you query and get results from ahrefs, using the www or non-www version of the site URL?

test/audits/organic-keywords.test.js
```diff
@@ -0,0 +1,168 @@
...
+  const site = createSite(siteData);
+
+  const auditResult = {
+    keywords: [
+      {
+        keyword: 'foo',
+        sum_traffic: 123,
+        best_position: 1,
+        best_position_prev: 1,
+        best_position_diff: 0,
```
iuliag, line 48: This is something we could easily compute ourselves. How much does it cost to get this field for every keyword every time we run the audit for a site?

iuliag (commented):
Looks good so far, with some observations and questions below

src/support/ahrefs-client.js
dzehnder: Base const: 50
keyword: 1
best_position: 1
best_position_prev: 1
best_position_diff: 1
sum_traffic: 10

**Total: 64 units**

Should we document this somewhere?

src/support/ahrefs-client.js
dzehnder: no, this is site specific. @alinarublea just showed me how I can integrate the alert config, I will integrate it there

src/organic-keywords/handler.js
dzehnder: shouldn't it be exactly the same as the saved base url? I tested in ahrefs and if it's not the same it does not return the right results

test/audits/organic-keywords.test.js
dzehnder: it's just 1 unit, so I thought it might save us some performance. Is it worth saving one unit?

src/support/ahrefs-client.js
iuliag: Good idea, maybe add it to the js doc of the function?

src/organic-keywords/handler.js
iuliag: I've checked ahrefs UI with `bamboohr.com` and it returns no results, whereas www.bamboohr.com does.
Screenshot 2024-01-30 at 14 51 32
Screenshot 2024-01-30 at 14 51 04

dzehnder: `bamboohr.com` also forwards to `www.bamboohr.com`. Shouldn't we have `www.bamboohr.com` as a baseURL in that case?

solaris007:
no activity, closing

PR #136, opened 2024-02-15
feat: reusable audits

ekremney:
Based on recent feedback, here's a proposal to improve the audit authoring by providing pluggable reusable parts into separate functions, making it easier for audit authors. With this proposal, authors only need to focus on providing the specific business logic. If necessary, they can easily integrate other components.

In Spacecat, an audit is essentially a review of a site based on its URL, for different purposes. To create an audit, authors would simply need to write a basic function that takes a URL as input and returns an object containing auditResult and fullAuditRef (optional - if needed to reference the audit later on).

Authors won't need to worry about tasks such as handling incoming audit job messages, database operations, or sending audit results to downstream components. These tasks are already set up and can be customized if needed. Also the logging for failures etc. will be unified.

The core Audit::run can be seen here, it's pretty much template pattern with pluggable parts: https://github.com/adobe/spacecat-audit-worker/blob/reusable-audits/src/common/audit.js#L74

Simplest example:

```js
AuditBuilder()
    .withRunner((url, context) => {
      // audit logic here
      return {
        auditResult,
        fullAuditRef,
    })
    .build();
```

Here's an example audit: https://github.com/adobe/spacecat-audit-worker/blob/reusable-audits/src/example/handler.js

Here's how the apex audit will look afterwards: https://github.com/adobe/spacecat-audit-worker/blob/reusable-audits/src/apex/handler.js

Here's how the broken backlink audit will look afterwards: https://github.com/adobe/spacecat-audit-worker/blob/reusable-audits/src/backlinks/handler.js

for special cases like the COGS audit, which doesn't require a URL (or any audit which won't require to be saved in db etc), we offer a special no-operation audit step:

```js
AuditBuilder()
    .withSiteProvider(noopAuditStep)
    .withUrlResolver(noopAuditStep)
    .withRunner(cogsAuditRunner) // a function for COGS audit results
    .build();

```


src/common/audit.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  return site;
+}
+
+export async function defaultUrlResolver(site) {
+  return site.getBaseURL();
+}
+export async function followRedirects(site) {
+  return getRUMUrl(site.getBaseURL());
```
iuliag, line 57: There's a new function in shared `composeAuditURL`

src/common/audit-builder.js
```diff
@@ -0,0 +1,71 @@
...
+      throw Error('"runner" must be a function');
+    }
+
+    return new Audit(
+      this.siteProvider,
+      this.urlResolver,
+      this.runner,
+      this.persister,
+      this.messageSender,
+    ).run;
```
iuliag, line 69: Any reason why you also run the audit it in the `build` function?

iuliag (approved):

Issue #133, opened 2024-02-12
Backlinks audit should validate whether broken backlink is fixed already


solaris007:
Due to the long re-indexing times of the primary broken backlinks source `ahrefs`, results should be validated whether they're already fixed by visiting the broken backlink and excluding from the result if fixed. This is to prevent false positives in the report after people have fixed the links.

7r0u8l3:
Hi, Im new to SC and dev but have talked to Meryll and will work on this ticket but I still need access to SC AWS. Ive requested access in IAM. Need that so I can start testing code. Assigning to me in meantime. Thanks!

PR #165, opened 2024-03-04
fix: broken backlinks filter backlinks returning ok

iuliag:
Fix https://github.com/adobe/spacecat-audit-worker/issues/133

### DIFF
--- a/test/audits/backlinks.test.js
+++ b/test/audits/backlinks.test.js
@@ -53,13 +53,13 @@ describe('Backlinks Tests', () => {
     backlinks: [
       {
         title: 'backlink title',
-        url_from: 'url-from',
-        url_to: 'url-to',
+        url_from: 'https://from.com/from-1',
+        url_to: 'https://www.abcd.com/to-1',
       },
       {
         title: 'backlink title 2',
-        url_from: 'url-from-2',
-        url_to: 'url-to-2',
+        url_from: 'https://from.com/from-2',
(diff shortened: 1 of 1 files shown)


src/backlinks/handler.js
```diff
@@ -16,6 +16,22 @@ import {
 import { composeAuditURL } from '@adobe/spacecat-shared-utils';
 import AhrefsAPIClient from '../support/ahrefs-client.js';
 import { retrieveSiteBySiteId } from '../utils/data-access.js';
+import { fetch } from '../support/utils.js';
+
+async function filterOutValidBacklinks(backlinks, log) {
+  const isStillBrokenBacklink = async (backlink) => {
+    try {
+      const response = await fetch(backlink.url_to);
+      return !response.ok;
```
ekremney, line 25: it might be a good idea to `log.warn` 429s (maybe all 4XXs except 404s) for those sites behind CDN protections because of sending multiple HTTP requests from `Promise.all`.

ekremney (approved):

src/backlinks/handler.js
solaris007: this is a good idea as we then can configure graphs and alerts on coralogix

solaris007 (approved):

src/backlinks/handler.js
iuliag: I incorporated the `log.warn`.

PR #215, opened 2024-04-10
fix: add upper limit to broken backlinks, log costs

iuliag:


iuliag:
@dzehnder, does your question refer to https://github.com/adobe/spacecat-audit-worker/pull/216 or in general?

The audits so far didn't really need caching, because they weren't triggered on demand, rather on a scheduled basis where we were expecting it to hit ahrefs in order to bring the latest state. Since the results of the audit are stored, the reporting/API are consuming them from storage. So we didn't really have where to reduce costs.

For the upcoming SEO assessments, the result of top pages is indeed used by multiple audits and culprit for increased costs. In this case, I'm thinking rather than caching at the http client level, we might prefer to store and consume from the db. Later other SpaceCat workstreams might want to consume this via the API, e.g. experimentation recommendations / powerBI dashboards / UI.

At the same time, if we'd offer SEO assessments to be triggered on demand rather than on a scheduled basis, then for select audits we could just return the latest audit results if triggered several times in a not significant to change interval - e.g. 1 or 7 days for broken backlinks.

PR #266, opened 2024-05-30
feat: verify sitemap urls

AndreiAlexandruParaschiv:
verify whether the URLs listed in the sitemap are accessible


src/sitemap/handler.js
```diff
@@ -276,6 +277,22 @@ export async function findSitemap(inputUrl) {
   );
   const extractedPaths = await getBaseUrlPagesFromSitemaps(inputUrl, filteredSitemapUrls);
 
+  // check if URLs from each sitemap exist and remove entries if none exist
+  if (Object.entries(extractedPaths).length > 0) {
+    const extractedSitemapUrls = Object.keys(extractedPaths);
+    for (const s of extractedSitemapUrls) {
+      const urlsToCheck = extractedPaths[s];
+      // eslint-disable-next-line no-await-in-loop
+      const existingPages = await checkCommonSitemapUrls(urlsToCheck);
```
dzehnder, line 286: I think at this point we should rename this function, it does not really check common sitemap urls, but just make a head requests to check if the response is 200. What about `filterValidUrls`?

dzehnder (approved):

src/sitemap/handler.js
AndreiAlexandruParaschiv: yap, that's a good suggestion.

src/sitemap/handler.js
```diff
@@ -150,12 +150,13 @@ export async function checkSitemap(sitemapUrl) {
  * @param {Array<string>} urls - Array of URLs to check.
  * @returns {Promise<Array<string>>} - List of sitemap URLs that exist.
  */
-async function checkCommonSitemapUrls(urls) {
+async function filterValidUrls(urls) {
   const fetchPromises = urls.map(async (url) => {
     const response = await fetch(url, { method: 'HEAD' });
     return response.ok ? url : null;
   });
   const results = await Promise.all(fetchPromises);
```
iuliag, line 158: A check from one promise rejected will fail all promises. We should use `.allSettled` more.
It would also be a good idea to fetch with a timeout set, like here: https://github.com/adobe/spacecat-audit-worker/blob/main/src/backlinks/handler.js#L25.
And to log the status code, so that we can introduce a Coralogix alert in case we receive status code 429, which would enable us to monitor if the crawling is checking too many URLs at once.

AndreiAlexandruParaschiv: Changed to allSettled and opened a separate PR https://github.com/adobe/spacecat-audit-worker/pull/357 to have fetchWithTimeout fct extensible.

PR #320, opened 2024-07-13
fix: experiment audit persists using post handler

rpapani:
- Added `experimentation-ess-all` audit, which will crunch the RUM data for 6 months. This will not be scheduled, but will be manually triggered when on boarding the customer to get all the historical experiments into persistent store
- Implemented Post handler for experimentation audits to update the experiments entity in persistent store after experiment audit is run
- Added `context` for audit postProcessor logic so that we can access the dataAccess methods in the postHandler and updated documentation

Fixes: [SITES-23404](https://jira.corp.adobe.com/browse/SITES-23404)


ramboz (approved):

src/experimentation-ess/common.js
```diff
@@ -451,4 +453,44 @@ export async function processAudit(auditURL, context, site, days) {
...
+      experiment.id,
+      experiment.url,
+    );
+    if (existingExperiment) {
+      if (new Date(existingExperiment.startDate) < new Date(experiment.startDate)) {
+        experimentData.startDate = existingExperiment.startDate;
+      } else {
+        experimentData.startDate = experiment.startDate;
+      }
+      if (new Date(existingExperiment.endDate) > new Date(experiment.endDate)) {
```
ekremney, line 484: do we have this case?

rpapani: for some reason, if we don't have RUM data for a day or so and end date is missing from experiment page, then we don't want to override the existing end date, that way we'll consider all the RUM data for calculations as intended. This is hypothetical for now, not sure if we really have this in real data, but for completion sake, I've kept this.

src/experimentation-ess/common.js
```diff
@@ -348,11 +351,67 @@ function mergeData(experiment, experimentMetadata, url) {
...
+  const lambdaPayload = {
+    type: 'statsig',
+    payload: {
+      rumData: {
+      },
+    },
+  };
+  for (const experiment of experimentData) {
+    lambdaPayload.payload.rumData[experiment.id] = {};
+    const metric = experiment.conversionEventName || 'click';
```
ekremney, line 378: do we take into the conversion event value into account here? If the conversion is defined as clicks on `.specific-block` or event a `convert` event with a specific source?

rpapani: No, custom conversion events with source and selectors, may be we can do it later

PR #326, opened 2024-07-19
feat: canonical url audit

AndreiAlexandruParaschiv:
Ticket : https://jira.corp.adobe.com/browse/SITES-18823

Canonical Audit 
- gets 200 top url pages from DB
- each url page is verified against diverse canonical checks 
- output results 
```
"auditResult": {
           "canonical-url-lowercased": {
			"success": true,
			"url": [
			"https://example.com",
                                  ...]
                           },
            "canonical-url-same-protocol": {
                        "success": false,
			"url": [
			"https://example.com/testPage",
                                  ...]
                            }
                   ...
            ]
```
Get canonical results: {{ Spacecat_dev }}/{siteId}/latest-audit/canonical


src/canonical/handler.js
```diff
@@ -0,0 +1,494 @@
...
+        check: ChecksAndErrors.URL_UNDEFINED.check,
+        error: ChecksAndErrors.URL_UNDEFINED.error,
+        success: false,
+      }],
+    };
+  }
+
+  try {
+    log.info(`Fetching URL: ${url}`);
+    const response = await fetch(url);
```
dzehnder, line 78: instead of fetching all the URLs here, we could leverage the scraped content stored in S3, so we don't need to make a request for each audit again (canoincal, JSON-LD, hreflang...)

src/canonical/handler.js
```diff
@@ -0,0 +1,494 @@
...
+
+        const urlContentCheck = await validateCanonicalUrlContentsRecursive(canonicalUrl, log);
+        log.info(`validateCanonicalUrlContentsRecursive result for ${canonicalUrl}: ${JSON.stringify(urlContentCheck)}`);
+        checks.push(...urlContentCheck);
+      }
+      log.info(`Checks for URL ${url}: ${JSON.stringify(checks)}`);
+      return { url, checks };
+    });
+
+    const auditResultsArray = await Promise.all(auditPromises);
```
iuliag, line 461: This may result in making 200 calls to fetch the html of the page at the same time.
Also, it one rejects, then the whole is rejected, so you could use Promise.allSettled instead.

src/canonical/handler.js
AndreiAlexandruParaschiv: Yap, but I think this can be tackled in a different PR due to the time constraints currently.

src/canonical/handler.js
```diff
@@ -0,0 +1,494 @@
...
+      log.info(`Canonical URL is valid and accessible: ${canonicalUrl}`);
+      checks.push({
+        check: ChecksAndErrors.CANONICAL_URL_STATUS_OK.check,
+        success: true,
+      });
+
+      // Check for redirection to another URL
+      if (canonicalUrl !== finalUrl) {
+        log.info(`Canonical URL redirects to: ${finalUrl}`);
+        const result = await validateCanonicalUrlContentsRecursive(finalUrl, log, visitedUrls);
```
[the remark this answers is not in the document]
AndreiAlexandruParaschiv: I think we can stop at the first redirect, initially wanted to see if going through the redirects if it returns to itself.

test/audits/canonical.test.js
```diff
@@ -0,0 +1,595 @@
...
+    it('should run canonical audit successfully', async () => {
+      const baseURL = 'http://example.com';
+      const html = `<html><head><link rel="canonical" href="${baseURL}"></head><body></body></html>`;
+      nock('http://example.com/page1').get('').reply(200, html);
+      const context = { log, dataAccess: { getTopPagesForSite: sinon.stub().resolves([{ getURL: () => 'http://example.com/page1' }]) } };
+      const site = { getId: () => 'testSiteId' };
+
+      const result = await canonicalAuditRunner(baseURL, context, site);
+
+      expect(result).to.be.an('object');
```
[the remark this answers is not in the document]
AndreiAlexandruParaschiv: indeed, updated the test to include better checks and removed the last test case.

PR #331, opened 2024-07-24
fix: making the stats lambda ARN configurable for dev and prod

rpapani:


ramboz (approved):

ekremney (approved):
this could also go into `spacecat-shared-utils`. I see this being used by other components as well

PR #333, opened 2024-07-26
fix: incorrect experiment end date showing the experiment status incorrectly

rpapani:
Fixes [SITES-23657](https://jira.corp.adobe.com/browse/SITES-23657)


src/experimentation-ess/common.js
```diff
@@ -418,6 +418,45 @@ function getObjectByProperty(array, name, value) {
...
+ * Updates the experiment start, end dates and status based on the inferred dates
+ * @param {*} experiments
+ */
+function updateExperimentMetadata(experiments) {
+  // get the date 2 days ago to account for timezones
+  const yesterday = getDate(2);
+  for (const experiment of experiments) {
+    experiment.startDate = experiment.startDate || experiment.inferredStartDate;
+    const inferredEndDate = new Date(experiment.inferredEndDate);
+    inferredEndDate.setHours(0, 0, 0, 0);
```
ramboz, line 443: Shouldn't we have the end date "inclusive" and set to 24/59/59/999?

src/experimentation-ess/common.js
```diff
@@ -418,6 +418,45 @@ function getObjectByProperty(array, name, value) {
...
+    if (inferredEndDate >= yesterday) {
+      // found RUM data yesterday, so endDate should be either in the metadata or null
+      experiment.endDate = experiment.endDate || null;
+    } else {
+      experiment.endDate = experiment.endDate || experiment.inferredEndDate;
+    }
+    if (!experiment.status) {
+      if (experiment.endDate && new Date(experiment.endDate) < new Date()) {
+        experiment.status = 'COMPLETE';
```
ramboz, line 452: *nitpicking*: I'm not sure I'd chose "complete" as a status. To me, it kinda means we have reached completion of the experiment which would mean both that it ended and that we reached stats sig. The latter is not guaranteed here though. So from a pure semantics point-of-view, I'd be tempted to go with "inactive"/"stopped".
"Inactive" is what we usually have in the metadata and documented, so maybe that would be the most appropriate?

ramboz (approved):

src/experimentation-ess/common.js
rpapani: since there're some edge cases after considering the timezone differences, I've just used yesterday as 2 days ago date, that should cover all the edge cases. `inferredEndDate` and `yesterday` both will have the same hours as we're trying to look at the date rather than hours, so used the hours as `0 0 0 0`.

PR #346, opened 2024-08-05
fix: gracefully handle broken backlinks suggestion error

dzehnder:
This PR introduces graceful error handling for the broken backlinks auto-suggest.

If the auto-suggest fails, it still returns the audit result.

In the future, auto-detect and auto-suggest should be split anyways.


solaris007 (approved):
should we include the error in the audit result?

AndreiAlexandruParaschiv (approved):

iuliag (approved):

PR #348, opened 2024-08-05
feat: introduce genAI generated broken backlink fixes

dzehnder:
This PR replaces the old logic of obtaining the suggested backlink fixes from the SEO keywords with GenAI based ones.

First, the audit is saved without the suggested fixes, then the Statistics Lambda function is triggered that will execute the generation of the suggested fixes

### DIFF
--- a/src/backlinks/handler.js
+++ b/src/backlinks/handler.js
@@ -18,6 +18,7 @@ import AhrefsAPIClient from '@adobe/spacecat-shared-ahrefs-client';
 import { AbortController, AbortError } from '@adobe/fetch';
 import { retrieveSiteBySiteId } from '../utils/data-access.js';
 import { enhanceBacklinksWithFixes, fetch } from '../support/utils.js';
+import { findSitemap } from '../sitemap/handler.js';
 
 const TIMEOUT = 3000;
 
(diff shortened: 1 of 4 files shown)


src/backlinks/handler.js
```diff
@@ -143,6 +133,24 @@ export default async function auditBrokenBacklinks(message, context) {
       auditContext,
       auditResult,
     };
+
+    try {
```
iuliag, line 137: I would push it even further down after `sqs.sendMessage`, as it would be a sort of post-processor (https://github.com/adobe/spacecat-audit-worker/blob/main/src/common/audit.js#L123-L126)

src/backlinks/handler.js
```diff
@@ -143,6 +133,24 @@ export default async function auditBrokenBacklinks(message, context) {
...
+      const baseUrl = site.getBaseURL();
+      const sitemaps = await findSitemap(baseUrl);
```
iuliag, line 139: The method name is misleading, as it seems `findSitemap` actually returns the page URLs listed in the sitemaps?

iuliag (approved):

AndreiAlexandruParaschiv (approved):

PR #360, opened 2024-08-15
feat: opportunities audit

rpapani:
- Container for opportunities audit.
- experimentation opportunities will be added to `experimentationOpportunities` array
- Added `rageclick` as an experimentation opportunity

Rageclick opportunity - https://github.com/adobe/spacecat-shared/pull/329


src/experimentationopportunities/experimentationopportunities.js
ramboz, line 1: *nitpicking*: it feels more like a completely generic handler, than something specific to experiments.
The fact that you specifically populate a `experimentationOpportunities` sub-property is kinda a hint to me to expect more types down the road here 😉

ramboz (commented):
Looks overall good to me.

ekremney (approved):

PR #367, opened 2024-08-19
feat: introduce structured data audit

dzehnder:
This PR implements the audit for [SITES-24087](https://jira.corp.adobe.com/browse/SITES-24087): Audit: structured data


ekremney:
Wouldn't "rich results audit" be a better name for this audit? I see that it only audits the rich results in the GSC result

package.json
```diff
@@ -58,14 +58,15 @@
     "@adobe/helix-universal-logger": "3.0.18",
     "@adobe/spacecat-shared-ahrefs-client": "1.4.1",
     "@adobe/spacecat-shared-data-access": "1.41.3",
+    "@adobe/spacecat-shared-google-client": "^1.1.18",
```
solaris007, line 61: please pin version

src/url-inspect/pdp-handler.js
```diff
@@ -0,0 +1,116 @@
...
+ * If an error occurs during the inspection of a URL, the object will include an error message.
+ *
+ * @throws {Error} - Throws an error if the audit process fails.
+ */
+async function processAudit(baseURL, context, site) {
+  const { log } = context;
+  const siteId = site.getId();
+
+  const productDetailPages = await site.getConfig().getProductDetailPages('pdp-indexability');
+  if (!productDetailPages || productDetailPages.length === 0) {
```
solaris007, line 35: check isArray?

src/url-inspect/pdp-handler.js
```diff
@@ -0,0 +1,116 @@
...
+    try {
+      const { inspectionResult } = await google.urlInspect(pdp);
+      log.info(`Successfully inspected URL: ${pdp}`);
+
+      const filteredIndexStatusResult = {
+        verdict: inspectionResult.indexStatusResult.verdict,
+        lastCrawlTime: inspectionResult.indexStatusResult.lastCrawlTime,
+      };
+
+      const filteredRichResults = inspectionResult.richResultsResult.detectedItems.map(
```
solaris007, line 52: are these properties always guaranteed to exist?

solaris007 (approved):

src/index.js
```diff
@@ -45,6 +46,7 @@ const HANDLERS = {
   'experimentation-ess-daily': essExperimentationDaily,
   'experimentation-ess-all': essExperimentationAll,
   costs,
+  'gsc-pdp-structured-data': gscPdpStructuredData,
```
iuliag, line 49: I would have seen this as a generic `structured-data` audit, whereas the pages to which it's applied and the source is internal implementation details.

src/url-inspect/pdp-handler.js
```diff
@@ -0,0 +1,46 @@
...
+export async function gscPdpStructuredDataHandler(baseURL, context, site) {
+  const { log } = context;
+  log.info(`Received Product Detail Page indexability audit request for ${baseURL}`);
+  const startTime = process.hrtime();
+
+  const siteId = site.getId();
+
+  const productDetailPages = await site.getConfig().getProductDetailPages('pdp-indexability');
+  if (isArray(productDetailPages) && productDetailPages.length === 0) {
+    log.error(`No top pages found for site ID: ${siteId}`);
```
iuliag, line 25: `No pages configured for structured-data/indexability audit for site: ${siteId/baseURL}` ? Same below

src/url-inspect/pdp-handler.js
```diff
@@ -0,0 +1,46 @@
...
+    throw new Error(`No top pages found for site: ${baseURL}`);
+  }
+
+  const auditResult = await processUrlInspect(baseURL, context, productDetailPages);
+
+  const endTime = process.hrtime(startTime);
+  const elapsedSeconds = endTime[0] + endTime[1] / 1e9;
+  const formattedElapsed = elapsedSeconds.toFixed(2);
+
+  log.info(`URL Inspect Audit completed in ${formattedElapsed} seconds for ${baseURL}`);
```
iuliag, line 35: Seeing url inspect / gsc-pdp-structured-data / pdp-handler / pdp-indexability used to name the same thing is confusing.
Let's settle on a single naming that is open to extensibility in the future and follow it consistently in logs/file/configuration/variable naming?

test/audits/pdp.test.js
```diff
@@ -0,0 +1,282 @@
...
+    sandbox.restore();
+  });
+
+  it('should successfully return a filtered result of the url inspection result', async () => {
+    sandbox.stub(GoogleClient, 'createFrom').returns(googleClientStub);
+    urlInspectStub.resolves(fullUrlInspectionResult);
+
+    const auditData = await gscPdpStructuredDataHandler('https://www.example.com', context, siteStub);
+
+    expect(auditData.auditResult).to.deep.equal(
```
iuliag, line 133: While this expect is somewhat less prone to need updates in case of fullUrlInspectionResult being updated, it doesn't work as documentation on what the audit result looks like.
If we treat unit tests as developer documentation, then having here the explicit JSON would be more helpful.

iuliag (commented):
To my understanding: this is a `structured-data` or `indexability` audit, that's the concept.
The audit result includes both:
- indexability flag
- rich results issues, that are not only PDP specific (you don't filter by schemas specific to PDPs only).
GSC, URL inspect, PDP are just implementation details.

src/index.js
dzehnder: Doesn't the current Audit Builder structure tie the implementation details for the source to the audit itself? As you might want different schedules for different sources for example that could be an issue. But I see your point, we can discuss that another time and build a generic first iteration here

src/structured-data/handler.js
```diff
@@ -0,0 +1,132 @@
...
+    } catch (error) {
+      log.error(`Failed to inspect URL: ${page}. Error: ${error.message}`);
+      return {
+        inspectionUrl: page,
+        error: error.message,
+      };
+    }
+  });
+
+  return Promise.all(urlInspectionResult);
```
AndreiAlexandruParaschiv, line 99: I would use`Promise.allSettled`instead, to ensure that all the promises are handled individually. with Promise.all in case one promise is rejected  it rejects the entire operation

dzehnder: good point, thanks!

solaris007 (approved):

PR #370, opened 2024-08-20
tests(broken-backlinks): reduce time to run tests

iuliag:
Reduce time to run tests.
Before: each test in the broken-backlinks suite that was hitting `filterOutValidBacklinks` was taking >3s to run because the setup included a broken backlink that when called was responding with a delay of 10000ms and fetch was waiting 3s for the response and then timing out.
After this fix: there's a single test that is including this case: `should filter out from audit result broken backlinks the ones that return ok (even with redirection)`.


solaris007 (approved):
thank you, i was noticing that as well

PR #375, opened 2024-08-22
fix: same domain check with www for baseUrl

AndreiAlexandruParaschiv:
Small fix for the usecase : 
- when url.hostname has www but  base.hostname doesn't.
  www.example.com vs  example.com

### DIFF
--- a/src/canonical/handler.js
+++ b/src/canonical/handler.js
@@ -352,7 +352,7 @@ export function validateCanonicalFormat(canonicalUrl, baseUrl, log) {
     }
 
     // Check if the canonical URL has the same domain as the base URL
-    if (url.hostname !== base.hostname) {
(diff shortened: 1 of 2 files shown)


src/canonical/handler.js
```diff
@@ -352,7 +352,7 @@ export function validateCanonicalFormat(canonicalUrl, baseUrl, log) {
     }
 
     // Check if the canonical URL has the same domain as the base URL
-    if (url.hostname !== base.hostname) {
+    if (url.hostname !== base.hostname && url.hostname !== `www.${base.hostname}` && `www.${url.hostname}` !== base.hostname) {
```
solaris007, line 355: i think some of the functions in `spacecat-shared-utils`, like the compose and strip functions may apply here instead.

AndreiAlexandruParaschiv: I've used composeBaseURL

solaris007 (approved):

PR #382, opened 2024-08-27
feat: meta-tags audit

dipratap:
This PR adds an audit meta-tags that audits top-pages tags specifically the title, description and H1 tag. The tags are scraped by content-scraper and stored in S3, the audit worker fetches tags from S3 and performs checks to detect if the specified 3 tags are missing or length wise inoptimal.

JIRA: https://jira.corp.adobe.com/browse/SITES-23343

## Related Docs
https://wiki.corp.adobe.com/display/AEMSites/SEO+%7C+Title%2C+Description+and+H1+Tags+Optimisation

## Related Issues
https://github.com/adobe/spacecat-api-service/pull/470
https://github.com/adobe/spacecat-content-scraper/pull/137

Thanks for contributing!


src/metatags/handler.js
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+import {
+  internalServerError, noContent, notFound, ok,
+} from '@adobe/spacecat-shared-http-utils';
+import { retrieveSiteBySiteId } from '../utils/data-access.js';
+import { getObjectFromKey, getObjectKeysUsingPrefix } from '../utils/s3-utils.js';
+import SeoChecks from './seo-checks.js';
+
+async function fetchAndProcessPageObject(s3Client, bucketName, key, prefix, log) {
```
ssilare-adobe, line 20: Can you please add functional comment to this above the definition?

src/metatags/handler.js
```diff
@@ -0,0 +1,104 @@
...
+  const object = await getObjectFromKey(s3Client, bucketName, key, log);
+  if (!object?.scrapeResult?.tags || typeof object.scrapeResult.tags !== 'object') {
```
ssilare-adobe, line 22: The `?.` optional chaining is great for preventing errors, but after checking for `object?.scrapeResult?.tags`, the `typeof object.scrapeResult.tags !== 'object'` could be unnecessary. 


```
if (!object?.scrapeResult?.tags) 
```

src/metatags/handler.js
```diff
@@ -0,0 +1,104 @@
...
+    log.error(`No Scraped tags found in S3 ${key} object`);
+    return null;
+  }
+  const pageUrl = key.slice(prefix.length - 1).replace('scrape.json', ''); // Remove the prefix and .json suffix
+  log.info(`Scraped tags for ${pageUrl} : ${JSON.stringify(object.scrapeResult.tags)}`);
+  return {
+    [pageUrl]: {
+      title: object.scrapeResult.tags.title,
+      description: object.scrapeResult.tags.description,
+      h1: object.scrapeResult.tags.h1 || [],
```
ssilare-adobe, line 32: Use optional chaining to avoid potential errors if `h1` doesn’t exist in `tags`.

```
h1: object.scrapeResult.tags?.h1 || [],
```

src/metatags/handler.js
```diff
@@ -0,0 +1,104 @@
...
+  try {
+    log.info(`Received ${type} audit request for siteId: ${siteId}`);
+    const site = await retrieveSiteBySiteId(dataAccess, siteId, log);
+    if (!site) {
+      return notFound('Site not found');
+    }
+    if (!site.isLive()) {
+      log.info(`Site ${siteId} is not live`);
+      return ok();
+    }
```
ssilare-adobe, line 52: Can we combine this to a single return?

```
if (!site || !site.isLive()) {
  log.info(`Site ${siteId} not found or not live`);
  return notFound('Site not found or not live');
}

```

src/metatags/handler.js
```diff
@@ -0,0 +1,104 @@
...
+    const prefix = `scrapes/${siteId}/`;
+    const scrapedObjectKeys = await getObjectKeysUsingPrefix(s3Client, bucketName, prefix, log);
+    const extractedTags = {};
+    for (const key of scrapedObjectKeys) {
+      // eslint-disable-next-line no-await-in-loop
+      const pageMetadata = await fetchAndProcessPageObject(s3Client, bucketName, key, prefix, log);
+      if (pageMetadata) {
+        Object.assign(extractedTags, pageMetadata);
+      }
+    }
```
ssilare-adobe, line 69: Using `await` inside a loop can lead to performance issues as it blocks the iteration. Consider refactoring to `Promise.all()` for parallel execution, improving efficiency.

```
const pageMetadataPromises = scrapedObjectKeys.map(key => fetchAndProcessPageObject(s3Client, bucketName, key, prefix, log));
const pageMetadataResults = await Promise.all(pageMetadataPromises);
pageMetadataResults.forEach(pageMetadata => {
  if (pageMetadata) {
    Object.assign(extractedTags, pageMetadata);
  }
});

```

src/metatags/handler.js
```diff
@@ -0,0 +1,104 @@
...
+    const extractedTagsCount = Object.entries(extractedTags).length;
+    if (extractedTagsCount === 0) {
```
ssilare-adobe, line 71: Can directly use `if (Object.keys(extractedTags).length === 0)` without having to create an extra variable

src/utils/s3-utils.js
```diff
@@ -0,0 +1,46 @@
...
+    const params = {
+      Bucket: bucketName,
+      Prefix: prefix,
+      MaxKeys: 1000,
+    };
+    const data = await s3Client.send(new ListObjectsV2Command(params));
+    data?.Contents?.forEach((obj) => {
+      objectKeys.push(obj.Key);
+    });
+    log.info(`Fetched ${objectKeys.length} keys from S3 for bucket ${bucketName} and prefix ${prefix}`);
```
ssilare-adobe, line 26: If more than 1000 objects are expected, you should handle pagination by repeatedly calling `ListObjectsV2` with `ContinuationToken`

```

let continuationToken = null;
do {
  const params = {
    Bucket: bucketName,
    Prefix: prefix,
    MaxKeys: 1000,
    ContinuationToken: continuationToken,
  };
  const data = await s3Client.send(new ListObjectsV2Command(params));
  data?.Contents?.forEach((obj) => {
    objectKeys.push(obj.Key);
  });
  continuationToken = data.IsTruncated ? data.NextContinuationToken : null;
} while (continuationToken);
```

ssilare-adobe (changes requested):

src/metatags/handler.js
dipratap: Since we have already checked if tags object exists or not in line 22, we can safely use object.scrapeResult.tags object here

src/metatags/handler.js
dipratap: In this case, we won't be able to identify what exact error did the audit encountered - if site is not live or it is not found. We need this info in response so that the user knows what exact error we encountered.

src/metatags/handler.js
dipratap: The reason I'm doing sequential invocations instead of parallel is that in parallel execution, all S3 objects would be fetched into memory simultaneously, which could lead to exceeding the allocated memory. With sequential calls, the Nodejs garbage collector has more opportunities to clean up memory after each fetchAndProcessPageObject invocation finishes, reducing the risk of high memory usage. The audits complete in less than 8 seconds, so the execution time remains within acceptable limits.

src/metatags/handler.js
dipratap: The `extractedTagsCount` variable is used in line 75 log as well `log.info(`Performing SEO checks for ${extractedTagsCount} tags`);`. Even if that was not the case, having a variable here improves code readability.

src/utils/s3-utils.js
dipratap: The `ListObject` command doesn't returns the object contents but just the objects metadata like key etc so memory wise it is not much. The listObjectsV2 API has a limit of 1,000 objects per request, and since we are scraping top-pages only which are expected to be < 200, we should be good here in my opinion. Let me know your thoughts.

src/metatags/handler.js
dipratap: Sure

src/metatags/handler.js
solaris007: also, we have `isObject` or `isNonEmptyObject` in spacecat-shared

src/metatags/handler.js
solaris007: more explicit guard statements are preferable

src/metatags/handler.js
solaris007: maybe `Promise.all` is worth a test, as i doubt we'll quickly exceed the allocated 4GB for top 200 pages. this would give a DOM-size per page of 20 kB... additional consideration would be rate limiting towards the S3 API though.

dipratap: Sure, updated the code.

dipratap:
@solaris007 @ssilare-adobe The changes are complete from my end. Can you please review?

ssilare-adobe (approved):

src/metatags/seo-checks.js
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  /**
+   * Checks for missing tags on the page and adds to detected tags array if found lacking.
+   * @param {string} url - The URL of the page.
+   * @param {object} pageTags - An object containing the tags of the page.
+   */
+  checkForMissingTags(url, pageTags) {
+    [TITLE, DESCRIPTION, H1].forEach((tagName) => {
+      if (pageTags[tagName] === undefined
```
solaris007, line 47: should this use `hasText`? what about zero-length strings? this may produce an issue later in the capitalize function. in addition you could make the capitalize function self-guarding.

src/metatags/seo-checks.js
```diff
@@ -0,0 +1,209 @@
...
+    this.allTags[tagName][tagContentLowerCase].pageUrls.add(url);
+  }
+
+  /**
+   * Performs all SEO checks on the provided tags.
+   * @param {string} url - The URL of the page.
+   * @param {object} pageTags - An object containing the tags of the page.
+   */
+  performChecks(url, pageTags) {
+    this.checkForMissingTags(url, pageTags);
```
solaris007, line 187: add early return if
- url is not valid (use `isValidURL`)
- pageTags is not an object (use isObject or isNonEmptyObject)

possibly wrap those checks into class method

src/metatags/seo-checks.js
```diff
@@ -0,0 +1,209 @@
...
+    };
+
+    const checkTag = (tagName, tagContent) => {
+      const capitalizedTagName = SeoChecks.#capitalizeFirstLetter(tagName);
+      let issueDetails;
+      let issueImpact;
+      let issue;
+      let recommendation;
+
+      if (tagContent === '') {
```
solaris007, line 84: use `hasText`

src/utils/s3-utils.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+import { GetObjectCommand, ListObjectsV2Command } from '@aws-sdk/client-s3';
+
+export async function getObjectKeysUsingPrefix(s3Client, bucketName, prefix, log) {
+  const objectKeys = [];
+  try {
+    const params = {
+      Bucket: bucketName,
+      Prefix: prefix,
+      MaxKeys: 1000,
```
solaris007, line 20: should this be configurable?

src/utils/s3-utils.js
```diff
@@ -0,0 +1,46 @@
...
+    };
+    const data = await s3Client.send(new ListObjectsV2Command(params));
```
solaris007, line 22: what happens if there's more than 1000 keys? is pagination needed?

src/support/s3-client.js
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Adds an S3Client instance to the context.
+ *
+ * @returns {function(object, UniversalContext): Promise<Response>}
+ */
+export default function s3Client(fn) {
+  return async (request, context) => {
+    if (!context.s3Client) {
+      context.s3Client = new S3Client();
```
solaris007, line 23: we have an S3 wrapper in shared... if you stay with this one, please instrument with AWS X-ray https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-utils/src/s3.js#L32

src/metatags/constants.js
```diff
@@ -0,0 +1,56 @@
...
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+// Tag Names
+export const TITLE = 'title';
```
solaris007, line 14: this could also be a JSON config file.

src/metatags/seo-checks.js
dipratap: Here we are checking for missing tags, the identifier for which is undefined tag value. If pageTags[tagName] is an empty string then it would be detected in emptyTag check in `checkForTagsLength` method.
The capitalize function is always called with tagName (_not pageTags[tagName]_) which will always be title, description or h1.
As a safety guard, I will add a hasText check to the capitalizeFirstLetter method.

src/metatags/seo-checks.js
dipratap: Here I would like to specifically identify if the string is empty. The hasText method will return false for null, undefined, number, empty string etc.

src/utils/s3-utils.js
dipratap: Although the expected keys shouldn't be more than 200 since we are auditing top-pages, but I have added pagination to accommodate more than 1000 keys.

src/support/s3-client.js
dipratap: Instrumented this one with AWS X-ray

src/metatags/seo-checks.js
dipratap: The url here is actually the endpoint (path) of the complete url. I can use `hasText` to validate it.
Checking the pageTags using isObject.

src/utils/s3-utils.js
dipratap: Sure, added it to function arguments

src/metatags/constants.js
dipratap: Sure, created a json file

src/metatags/seo-checks.js
solaris007: then it should be renamed to make this clear

solaris007 (approved):

src/metatags/seo-checks.js
dipratap: sure, renaming the url

PR #384, opened 2024-08-29
feat: high bounce rate and low ctr opportunities

ekremney:
Implements https://github.com/adobe/spacecat-shared/pull/351

Adds new two experimentation opportunities to the opportunity audit:
1. High inorganic traffic with high bounce rate
2. High organic traffic with low ctr


ramboz (approved):

AndreiAlexandruParaschiv (approved):

src/opportunities/opportunities.js
```diff
@@ -16,6 +16,11 @@ import { AuditBuilder } from '../common/audit-builder.js';
 import { getRUMDomainkey } from '../support/utils.js';
 
 const DAYS = 30;
+const OPPTY_QUERIES = [
```
rpapani, line 19: since there can be other opportunities, which are not experimentation related, so I've kept those as `experimentationHandlers`, so can we still keep the word `experimentation`? like `EXPERIMENTATION_OPPTY_QUERIES` ?

rpapani (approved):

src/opportunities/opportunities.js
ekremney: Actually, more accurate name for this audit should be `experimentation-opportunities` because it only encompasses the experimentation opportunities. Other type opportunities such as broken backlinks or cwv etc. are covered in different audits. Consecutively, I renamed the audit

PR #475, opened 2024-11-21
feat: introduce Opportunity creation for Broken Backlinks

dzehnder:


src/backlinks/handler.js
```diff
@@ -139,13 +139,50 @@ export default async function auditBrokenBacklinks(message, context) {
...
       auditResult,
     };
+
+    const opportunities = await dataAccess.Opportunity.allBySiteIdAndStatus(siteId, 'NEW');
+    let brokenBacklinksOppty = opportunities.find((oppty) => oppty.getType() === 'broken-backlinks');
+
+    if (!brokenBacklinksOppty) {
+      const opportunityData = {
+        siteId: site.getId(),
+        auditId: audit.getId(),
```
habansal, line 156: If the "brokenBacklinksOppty" already exists, I think, we would need to update the auditId in that case also.

src/backlinks/handler.js
```diff
@@ -139,13 +139,50 @@ export default async function auditBrokenBacklinks(message, context) {
...
+            'Paste new entries in your website redirects file.',
+            'Publish the changes.',
+          ],
+        },
+        tags: ['traffic-acquisition'],
+      };
+      brokenBacklinksOppty = await dataAccess.Opportunity.create(opportunityData);
+    }
+
+    const suggestions = await brokenBacklinksOppty.addSuggestions(data.auditResult.brokenBacklinks);
```
habansal, line 176: @dzehnder What about the suggestions that are already existing?

**Example:**
- On 1st Nov. an audit runs and detects 3 broken links: A, B, C and this code creates 3 entries in the Suggestions table.
- On 2nd Nov, a back-office-ui user adds a new broken-backlink D to suggestions table.
- On 3rd Nov, a ESS user marks the backlink B as Ignored.
- On 7th Nov, another audit runs and detects 4 broken backlinks: A', B', E, F

**What is the expected state of Suggestions table, after the audit runs on 7th Nov?**

- A, B(ignored), C, D, A', B', E, F
- A, B(ignored), C, D, E, F
- A', B', C, D, E, F
- A', B', E, F
- A', B'(ignored), E, F
- something else?

src/backlinks/handler.js
```diff
@@ -139,13 +139,50 @@ export default async function auditBrokenBacklinks(message, context) {
...
+          steps: [
+            'Review the list of broken target URLs and the suggested redirects.',
+            'Manually override redirect URLs as needed.',
+            'Copy redirects.',
+            'Paste new entries in your website redirects file.',
+            'Publish the changes.',
+          ],
+        },
+        tags: ['traffic-acquisition'],
+      };
```
habansal, line 172: will we need to add "createdAt" and "updatedAt" properties or would those be auto-added by the data layer?

src/backlinks/handler.js
dzehnder: if a backlink was marked as ignored it should not appear anymore in the ESS UI and stay in the ignored list. But that is handled as part of the site configuration. 
So the state of the suggestions table does not include the ignored property, it only has the data of the current active broken backlinks.
In that case it would look like this:

- A, B, C, D, E, F

src/backlinks/handler.js
dzehnder: they have default values in the data layer (`Date.now()`). Same for the status property (set to 'NEW')

src/backlinks/handler.js
dzehnder: good catch! Yes absolutely

src/backlinks/handler.js
habansal: Also started a thread on slack to clarify the expectations with the PM: https://cq-dev.slack.com/archives/C05A45JBP9N/p1732545991035499

src/backlinks/handler.js
```diff
@@ -139,13 +139,67 @@ export default async function auditBrokenBacklinks(message, context) {
...
+        data.auditResult.brokenBacklinks.map((backlink) => (
+          {
+            opportunityId: brokenBacklinksOppty.getId(),
+            type: 'REDIRECT_UPDATE',
+            rank: backlink.traffic_domain,
+            data: {
+              title: backlink.title,
+              url_from: backlink.url_from,
+              url_to: backlink.url_to,
+            },
```
habansal, line 189: we would also need either of `domain_traffic` or `priority` to show it on the UI.

src/backlinks/handler.js
habansal: Conclusion from slack:

> A', B' (ignored), D, E, F

seems correct.

PR needs to be updated accordingly

src/backlinks/handler.js
```diff
@@ -139,13 +139,67 @@ export default async function auditBrokenBacklinks(message, context) {
...
+        },
+        tags: ['traffic-acquisition'],
+      };
+      brokenBacklinksOppty = await dataAccess.Opportunity.create(opportunityData);
+    } else {
+      brokenBacklinksOppty = brokenBacklinksOppty.setAuditId(audit.getId());
+    }
+
+    if (!data.auditResult.error) {
+      const suggestions = await brokenBacklinksOppty.addSuggestions(
```
habansal, line 179: For recurring weekly audits, I see below issues:

- This will create duplicate entries
- We are not deleting the previous audit suggestions, which didn't surface in the new audit

src/backlinks/handler.js
```diff
@@ -139,13 +139,67 @@ export default async function auditBrokenBacklinks(message, context) {
...
+    const audit = await dataAccess.addAudit(auditData);
     const data = {
       type,
       url: site.getBaseURL(),
       auditContext,
       auditResult,
     };
+
+    const opportunities = await dataAccess.Opportunity.allBySiteIdAndStatus(siteId, 'NEW');
+    let brokenBacklinksOppty = opportunities.find((oppty) => oppty.getType() === 'broken-backlinks');
```
solaris007, line 151: if this is a frequent access pattern i could also introduce a relevant index and a `allBySiteIdAndTypeAndStatus` method. WDYT?

src/backlinks/handler.js
```diff
@@ -139,13 +139,67 @@ export default async function auditBrokenBacklinks(message, context) {
...
+      if (suggestions.errorItems.length > 0) {
+        log.error(`Suggestions for siteId ${siteId} contains ${suggestions.errorItems.length} items with errors`);
+        suggestions.errorItems.forEach((errorItem) => {
+          log.error(`Item ${JSON.stringify(errorItem.item)} failed with error: ${errorItem.error}`);
+        });
+        if (suggestions.createdItems.length <= 0) {
+          return internalServerError(`Failed to create suggestions for siteId ${siteId}`);
+        }
+      }
+    }
```
solaris007, line 202: i don't see any error handling?

src/backlinks/handler.js
dzehnder: done! I also wrote a helper function that everyone can use for their audits, to not have to think about this too much

src/backlinks/handler.js
dzehnder: yes I can add it

README.md
```diff
@@ -216,3 +216,50 @@ export default new AuditBuilder()
   .build();
 
 ```
+
+### How to add Opportunities and Suggestions
```
ekremney, line 220: 💯

src/utils/data-access.js
```diff
@@ -35,3 +35,56 @@ export async function retrieveSiteBySiteId(dataAccess, siteId, log) {
...
+  log,
+}) {
+  const newDataKeys = new Set(newData.map(buildKey));
+  const existingSuggestions = await opportunity.getSuggestions();
+
+  // Remove outdated suggestions
+  await Promise.all(
+    existingSuggestions
+      .filter((existing) => !newDataKeys.has(buildKey(existing)))
+      .map((suggestion) => suggestion.remove()),
```
dipratap, line 65: Hyman's comment from [thread](https://cq-dev.slack.com/archives/C05A45JBP9N/p1732547498674979?thread_ts=1732545991.035499&cid=C05A45JBP9N) :
> If the last audit on Nov 8th does not detect C. It should be removed. I'd expect: A, B (ignored), D, E, F

@dzehnder As per above comment, the suggestions that were present in previous audit, but not in the latest one, also needs to be removed.

src/backlinks/handler.js
dzehnder: still pending, ElectroDB does not allow readme properties to be part of the composite. Solution coming with next PR

ekremney (approved):

src/utils/data-access.js
```diff
@@ -35,3 +35,56 @@ export async function retrieveSiteBySiteId(dataAccess, siteId, log) {
...
+  );
+
+  // Prepare new suggestions
+  const newSuggestions = newData
+    .filter((data) => !existingSuggestions.some(
+      (existing) => buildKey(existing) === buildKey(data),
+    ))
+    .map(mapNewSuggestion);
+
+  // Add new suggestions if any
```
habansal, line 75: We should also update existing suggestions which are not created or modified via Back-Office-UI and but have surfaced in the new audits also.

e.g. 
If following suggestion exists:
```
{
  "url_to": "https://www.lovesac.com/terms_and_conditions",
  "title": "Lovesac returns to its 'Roots' for 20th anniversary",
  "url_from": "https://www.prnewswire.com/news-releases/lovesac-returns-to-its-roots-for-20th-anniversary-300121114.html",
  "traffic_domain": 947754
}
```
and in the new audit, we detect the same broken link again in the same referring page, but the Title or traffic numbers for that referring have changed now, we should update those in the suggestions entry.

src/utils/data-access.js
```diff
@@ -35,3 +35,56 @@ export async function retrieveSiteBySiteId(dataAccess, siteId, log) {
...
+  opportunity,
+  newData,
+  buildKey,
+  mapNewSuggestion,
+  log,
+}) {
+  const newDataKeys = new Set(newData.map(buildKey));
+  const existingSuggestions = await opportunity.getSuggestions();
+
+  // Remove outdated suggestions
```
habansal, line 61: If the existing suggestion was entered via back-office-ui as the source, we should NOT delete/modify it, irrespective of whether the same suggestion exists in the new audit or not

src/backlinks/handler.js
```diff
@@ -139,14 +139,78 @@ export default async function auditBrokenBacklinks(message, context) {
...
+          guidance: {
+            steps: [
+              'Review the list of broken target URLs and the suggested redirects.',
+              'Manually override redirect URLs as needed.',
+              'Copy redirects.',
+              'Paste new entries in your website redirects file.',
+              'Publish the changes.',
+            ],
+          },
+          tags: ['traffic-acquisition'],
```
habansal, line 179: Should we call it "Traffic acquisition", exactly how we want it to show in the UI, else UI will need to keep a mapping of all possible tags

src/utils/data-access.js
dzehnder: We will handles this as a next step. This will require an additional property in the data model

solaris007 (approved):

habansal (approved):

src/utils/data-access.js
```diff
@@ -35,3 +35,70 @@ export async function retrieveSiteBySiteId(dataAccess, siteId, log) {
...
+  await Promise.all(
+    existingSuggestions
+      .filter((existing) => !newDataKeys.has(buildKey(existing)))
+      .map((suggestion) => suggestion.remove()),
+  );
+
+  // Update existing suggestions
+  await Promise.all(
+    existingSuggestions
+      .filter((existing) => {
```
habansal, line 71: Isn't this filter redundant, since we already removed outdated suggestions above?

PR #492, opened 2024-11-28
feat: introduce opportunity creation for sitemap audit

AndreiAlexandruParaschiv:
Ticket : https://jira.corp.adobe.com/browse/SITES-27083 
Create opportunity for sitemap audit based on the data model


src/sitemap/handler.js
```diff
@@ -393,8 +394,78 @@ export async function sitemapAuditRunner(baseURL, context) {
...
+  if (newData.length > 0) {
+    if (!sitemapOpptyPagesWithIssues) {
+      const opportunityData = {
+        siteId: auditData.siteId,
+        auditId: auditData.id,
+        runbook: 'link-to-runbook',
+        type: 'sitemap-pages-with-issues',
+        origin: 'AUTOMATON',
+        title: 'Opportunity Title',
+        description: 'Opportunity Description',
```
habansal, line 427: this needs to replaced with a better description

src/sitemap/handler.js
```diff
@@ -393,8 +394,78 @@ export async function sitemapAuditRunner(baseURL, context) {
...
+
+  if (newData.length > 0) {
+    if (!sitemapOpptyPagesWithIssues) {
+      const opportunityData = {
+        siteId: auditData.siteId,
+        auditId: auditData.id,
+        runbook: 'link-to-runbook',
+        type: 'sitemap-pages-with-issues',
+        origin: 'AUTOMATON',
+        title: 'Opportunity Title',
```
habansal, line 426: title should be set as:
> Sitemap issues found

Same will be used in the UI also. Refer:
https://cq-dev.slack.com/archives/C05A45JBP9N/p1732609369480039?thread_ts=1732272497.976809&cid=C05A45JBP9N

cc @raghuwan

src/sitemap/handler.js
```diff
@@ -393,8 +394,78 @@ export async function sitemapAuditRunner(baseURL, context) {
...
+        description: 'Opportunity Description',
+        guidance: {
+          steps: [
+            'Step 1',
+            'Step 2',
+          ],
+        },
+        tags: ['indexability', 'sitemap'],
+      };
+      sitemapOpptyPagesWithIssues = await dataAccess.Opportunity.create(opportunityData);
```
habansal, line 436: add error handling as done here:
https://github.com/adobe/spacecat-audit-worker/pull/475/files#diff-212dd58956c947938959c789cc3912af8f1b884de5543486166247e12f8ef215R188

src/sitemap/handler.js
```diff
@@ -393,8 +394,78 @@ export async function sitemapAuditRunner(baseURL, context) {
...
+      });
+    });
+  });
+
+  if (newData.length > 0) {
+    if (!sitemapOpptyPagesWithIssues) {
+      const opportunityData = {
+        siteId: auditData.siteId,
+        auditId: auditData.id,
+        runbook: 'link-to-runbook',
```
habansal, line 423: can you update the link here?

src/sitemap/handler.js
```diff
@@ -393,8 +394,78 @@ export async function sitemapAuditRunner(baseURL, context) {
...
+            'Step 1',
+            'Step 2',
+          ],
+        },
+        tags: ['indexability', 'sitemap'],
+      };
+      sitemapOpptyPagesWithIssues = await dataAccess.Opportunity.create(opportunityData);
+    } else {
+      sitemapOpptyPagesWithIssues = sitemapOpptyPagesWithIssues.setAuditId(auditData.id);
+    }
```
habansal, line 439: await brokenBacklinksOppty.save();

src/sitemap/handler.js
```diff
@@ -393,8 +394,78 @@ export async function sitemapAuditRunner(baseURL, context) {
...
+  if (newData.length > 0 && sitemapOpptyPagesWithIssues) {
+    const buildKey = (auditData1) => `${auditData1.siteId}|sitemap-pages-with-issues|${auditData1.auditResult.reasons.map((a) => a.value)};}`;
+
+    await syncSuggestions({
+      opportunity: sitemapOpptyPagesWithIssues,
+      newData,
+      buildKey,
+      mapNewSuggestion: (issue) => ({
+        opportunityId: sitemapOpptyPagesWithIssues.getId(),
+        type: 'SUGGESTION_TYPE',
```
habansal, line 451: enum values mentioned here:
https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-data-access/src/v2/schema/suggestion.schema.js#L46

src/sitemap/handler.js
```diff
@@ -393,8 +394,78 @@ export async function sitemapAuditRunner(baseURL, context) {
...
+        tags: ['indexability', 'sitemap'],
+      };
+      sitemapOpptyPagesWithIssues = await dataAccess.Opportunity.create(opportunityData);
+    } else {
+      sitemapOpptyPagesWithIssues = sitemapOpptyPagesWithIssues.setAuditId(auditData.id);
+    }
+  }
+
+  if (newData.length > 0 && sitemapOpptyPagesWithIssues) {
+    const buildKey = (auditData1) => `${auditData1.siteId}|sitemap-pages-with-issues|${auditData1.auditResult.reasons.map((a) => a.value)};}`;
```
habansal, line 443: Here, we need to distinguish each suggestion entry (i.e. each issue in the Sitemap)

So, while building the key, we should also use the "sitemap url" and "page url" from the audit result, when they exist (i.e. success=true)

src/sitemap/handler.js
```diff
@@ -393,8 +394,78 @@ export async function sitemapAuditRunner(baseURL, context) {
   };
 }
 
+async function convertToOpportunity(auditUrl, auditData, context) {
+  const { dataAccess, log } = context;
+
+  const opportunities = await dataAccess.Opportunity.allBySiteIdAndStatus(auditData.siteId, 'NEW');
+  let sitemapOpptyPagesWithIssues = opportunities.find((oppty) => oppty.getType() === 'sitemap-pages-with-issues');
```
habansal, line 401: shouldn't the opportunity type be agnostic of whether the sitemap doesn't-exist or sitemap pages have error

src/sitemap/handler.js
```diff
@@ -393,8 +394,78 @@ export async function sitemapAuditRunner(baseURL, context) {
...
+
+  const sitemapsWithPagesWithIssues = auditData.auditResult?.details?.issues
```
habansal, line 403: We also need to create opportunity and suggestion entries for the case, when the sitemap doesn't exist or has any non-page errors (i.e. success:false)

src/sitemap/handler.js
raghuwan: Using below description on UI
`A complete and accurate sitemap helps search engines efficiently crawl and index your website\'s pages, ensuring better visibility in search results. Fixing sitemap issues can improve the discoverability of your content.`

src/sitemap/handler.js
```diff
@@ -393,8 +394,84 @@ export async function sitemapAuditRunner(baseURL, context) {
...
+      mapNewSuggestion: (issue) => ({
+        opportunityId: sitemapOpptyPagesWithIssues.getId(),
+        type: 'CODE_CHANGE',
+        rank: issue.traffic_domain,
+        data: {
+          sourceSitemapUrl: issue.sitemapUrl,
+          pageUrl: issue.url,
+          statusCode: issue.statusCode ?? 500, // to check if this is correctly reported,
+          // as missing status code usually means the page was not reachable
+          recommendationAction: 'remove_page_from_sitemap_or_fix_page_redirect_or_make_it_accessible',
```
habansal, line 464: is this the suggestion, that will need to entered via the backoffice ui after analysing each url with error code?

If yes, we shouldn't send a generic one here, since a success engineer will manually find the reason and store in DB. On recurring audits, mapNewSuggestion overrides all the existing data.

So, I suggest we skip this field "recommendationAction" here

src/sitemap/handler.js
```diff
@@ -393,8 +394,84 @@ export async function sitemapAuditRunner(baseURL, context) {
...
+    const buildKey = (auditData1) => `${auditData1.siteId}|sitemap-pages-with-issues|${auditData1.auditResult.reasons.map((a) => a.value)};}`;
+
+    await syncSuggestions({
+      opportunity: sitemapOpptyPagesWithIssues,
+      newData,
+      buildKey,
+      mapNewSuggestion: (issue) => ({
+        opportunityId: sitemapOpptyPagesWithIssues.getId(),
+        type: 'CODE_CHANGE',
+        rank: issue.traffic_domain,
```
habansal, line 458: Also checking the expected sort order with PM & UX here:
https://www.figma.com/design/lTwm2PWEC7UBTr7OMFPth7?node-id=2255-52468#1050501311

habansal: https://wiki.corp.adobe.com/display/AEMSites/Success+Studio+Projected+Business+Impact+Metrics#SuccessStudioProjectedBusinessImpactMetrics-Invalid/MissingSitemaps

Looking at this, maybe we should consider Organic Traffic to decide the rank, since that is going to influence the impact (Projected Traffic lost)

src/sitemap/handler.js
```diff
@@ -393,8 +394,188 @@ export async function sitemapAuditRunner(baseURL, context) {
...
+    type,
+    title,
+    runbook: 'https://adobe.sharepoint.com/:w:/r/sites/aemsites-engineering/Shared%20Documents/3%20-%20Experience%20Success/SpaceCat/Runbooks/Experience_Success_Studio_Sitemap_Runbook.docx?d=w6e82533ac43841949e64d73d6809dff3&csf=1&web=1&e=FTWy7t',
+    guidance: {
+      steps: [
+        'Verify each URL in the sitemap, identifying any that do not return a 200 (OK) status code.',
+        'Check RUM data to identify any sitemap pages with unresolved 3xx, 4xx or 5xx status codes – it should be none of them.',
+      ],
+    },
+    tags: ['Traffic Acquisition', 'Sitemap'],
```
habansal, line 504: Can you remove "Sitemap" from here, as that is being already hardcoded on UI and other handlers are also not sending type tag

AndreiAlexandruParaschiv: Okey, done

alinarublea (approved):

PR #494, opened 2024-11-28
feat: add opportunity & suggestions handling for Meta-tags

dipratap:
Add opportunity & suggestions creation and management logic for Meta-tags audit.

## Related Issues
https://jira.corp.adobe.com/browse/SITES-27082


src/metatags/opportunityHandler.js
```diff
@@ -0,0 +1,230 @@
...
+ *
+ * @param {Object} params - The parameters for the sync operation.
+ * @param {Object} params.opportunity - The opportunity object to synchronize suggestions for.
+ * @param {Array} params.newData - Array of new data objects to sync.
+ * @param {Function} params.buildKey - Function to generate a unique key for each item.
+ * @param {Function} params.mapNewSuggestion - Function to map new data to suggestion objects.
+ * @param {Object} params.log - Logger object for error reporting.
+ * @returns {Promise<void>} - Resolves when the synchronization is complete.
+ */
+export async function syncSuggestions({
```
dzehnder, line 31: I wrote already a general `syncSuggestions` function as a util function here: https://github.com/adobe/spacecat-audit-worker/blob/main/src/utils/data-access.js#L51

maybe it is possible for you to re-use it? Also feel free to make improvements to it. But since all audits have kind of the same logic for this, it might make sense to have one common function.

src/metatags/opportunityHandler.js
```diff
@@ -0,0 +1,230 @@
...
+      const existing = existingSuggestionsMap.get(buildKey(newSuggestion));
+      if (existing) {
+        return {
+          ...newSuggestion,
+          status: existing.getStatus(),
+          data: {
+            ...newSuggestion.data,
+            ...(existing.data.aiSuggestion && { aiSuggestion: existing.data.aiSuggestion }),
+            ...(existing.data.aiRationale && { aiRationale: existing.data.aiRationale }),
+          },
```
habansal, line 58: 1. how about the case where a ESS UI user has edited the suggestion and provided a custom one ("data.toOverride")

We should retain that property as well.

2. Also, Can you add a TODO comment in the code that once the Suggestions Entity supports createdBy field, we should not delete those existing suggestion entries.

dipratap: > how about the case where a ESS UI user has edited the suggestion and provided a custom one ("data.toOverride"). We should retain that property as well.

Sure, makes sense if we are thinking of using the override field for prioritisation over BO UI edits and llm training.

Added TODO comment as well.

src/metatags/opportunityHandler.js
dipratap: Thank you for pointing that out. I did initially refer to your syncSuggestions function and used parts of it. 
However, there are some customisations done in my method specifically for meta-tags that may not be possible to do in your method while ensuring compatibility with the ongoing work others are doing for audit opportunities, and I decided it would be safer to write a separate function. I’ve also renamed it to avoid any confusion.

dipratap:
@dzehnder I think the syncSuggestions method in data-access folder isn't working as expected when there are existing suggestions. 

`const existingSuggestions = await opportunity.getSuggestions();
  // Update existing suggestions
  await Promise.all(
    existingSuggestions
      .filter((existing) => {
        const existingKey = buildKey(existing);
        return newDataKeys.has(existingKey);
  })`

In above code, the buildKey method tries to get url_from and url_to params from existing suggestion but gets undefined in both fields so the key is built as undefined|undefined. I verified this by adding a [temporary log](https://spacecat.coralogix.com/#/query-new/logs?id=8Fdqksop0ZBUbE4OTaxdC&page=0) to syncSuggestions method in data_access.
I am also facing similar problem in meta-tags. The issue seems to be with `await opportunity.getSuggestions()`. The objects array it is returning can't be used as is to retrieve audit data. I have also logged it in the above coralogix log and it seems to contain a lot of values but not the actual audit data.

dipratap:
@dzehnder  I think the syncSuggestions method in data-access folder isn't working as expected when there are existing suggestions.

```
// Update existing suggestions 
await Promise.all( existingSuggestions .filter((existing) => { 
const existingKey = buildKey(existing); 
return newDataKeys.has(existingKey); 
})
```

In above code, the buildKey method tries to get url_from and url_to params from existing suggestion but gets undefined in both fields so the key is built as undefined|undefined. I verified this by adding a [temporary log](https://spacecat.coralogix.com/#/query-new/logs?id=q7oJIDmHEjLWAeza6eAE7&page=0) to syncSuggestions method in data_access. Here is the added log:

```
log.info(`Suggestion sample backlink: ${existingSuggestions.length},\n ${existingSuggestions[0] ? existingSuggestions[0].url_from : ''}|${existingSuggestions[0] ? existingSuggestions[0].url_to : ''} \n ${existingSuggestions[0] ? existingSuggestions[0].data : ''} \n${JSON.stringify(existingSuggestions[0])}`);
```

I am also facing similar problem in meta-tags. The issue seems to be with await opportunity.getSuggestions(). The objects array it is returning can't be used as is to retrieve audit data. I have also logged it in the above coralogix log and it seems to contain a lot of values but not the actual audit data.
cc @solaris007

src/metatags/opportunityHandler.js
```diff
@@ -0,0 +1,233 @@
...
+    // update new suggestions with data from existing suggestion of same key
+    .map((newSuggestion) => {
+      const existing = existingSuggestionsMap.get(buildKey(newSuggestion));
+      if (existing) {
+        return {
+          ...newSuggestion,
+          status: existing.getStatus(),
+          data: {
+            ...newSuggestion.data,
+            ...(existing.data.aiSuggestion && { aiSuggestion: existing.data.aiSuggestion }),
```
habansal, line 56: existing.getData()

src/metatags/opportunityHandler.js
habansal: As discussed with @dzehnder, we can do this refactoring as a follow-up, once @dipratap is back from vacation next week.

We can merge without this refactoring to unblock the UI work

habansal: Me and @dipratap  discussed this over a call. We will take this up once auto-suggest implementation has been completed by a few opportunities and the common syncSuggestions method is made compatible with the newly introduced aiSuggestions fields since with its current impl it will overwrite any AI suggestions, manually edited in Backoffice UI, with new data.
Currently, metadata has its own method to sync Suggestions and is working fine.

PR #532, opened 2024-12-16
fix: Increase memory for audit worker to avoid OOM with big RUM bundles

rpapani:
- Temporarily increasing the audit worker memory to 6GB, created https://jira.corp.adobe.com/browse/SITES-27905 to provide streaming support in RUM API Client to efficiently handle the memory.

Fixes SITES-27893 - Experimentation opportunities audit keep getting OOM for adobe.com


src/experimentation-opportunities/experimentation-opportunities.js
```diff
@@ -167,6 +167,7 @@ async function updateRecommendations(oppty, context, site) {
   let lambdaResult;
   try {
     // eslint-disable-next-line no-await-in-loop
+    log.info('Invoking lambda function with payload: ', JSON.stringify(lambdaPayload, null, 2));
```
ekremney, line 170: please do not log this big payloads

ekremney (approved):

src/experimentation-opportunities/experimentation-opportunities.js
rpapani: payload isn't big actually, its around 10 lines only

PR #604, opened 2025-01-23
feat: auto suggest for broken internal links

nitinja:
Issue: https://jira.corp.adobe.com/browse/SITES-26545

Uses same prompts as broken backlinks, hence the prompt names. 
Code is based on Damian's code for broken backlinks auto suggestions.

Thanks for contributing!


src/internal-links/handler.js
```diff
@@ -156,9 +157,12 @@ export async function convertToOpportunity(auditUrl, auditData, context) {
       type: 'CONTENT_UPDATE',
       rank: entry.traffic_domain,
       data: {
-        ...entry,
-        /* code commented until implementation of suggested links. TODO: implement suggestions, https://jira.corp.adobe.com/browse/SITES-26545 */
-        // suggestedLink: 'some suggestion here',
+        title: entry.title,
```
naydav, line 160: need to update the OpenAPI schema separately (and approve).

src/internal-links/suggestions-generator.js
```diff
@@ -0,0 +1,151 @@
...
+    const suggestions = [];
+    for (const batch of dataBatches) {
+      // eslint-disable-next-line no-await-in-loop
+      const result = await processBatch(batch, link.url_to);
+      if (result) {
+        suggestions.push(result);
+      }
+    }
+
+    if (totalBatches > 1) {
```
naydav, line 77: I would extract this part of the code or explicitly pass the parameter.

Right now, the function's behavior depends on a variable from the context, introducing **temporal coupling**.

naydav, line 77: I would extract this part of the code or explicitly pass the parameter.

Right now, the function's behavior depends on a variable from the context, introducing **temporal coupling**.

src/internal-links/suggestions-generator.js
```diff
@@ -0,0 +1,151 @@
...
+      ai_rationale:
+        suggestions[0]?.ai_rationale?.length > 0 ? suggestions[0]?.ai_rationale : 'No suitable suggestions found',
+    };
+  };
+
+  const headerSuggestionsResults = [];
+  for (const link of auditData.auditResult.brokenInternalLinks) {
+    try {
+      // eslint-disable-next-line no-await-in-loop
+      const requestBody = await getPrompt({ alternative_urls: headerLinks, broken_url: link.url_to }, 'broken-backlinks', log);
```
naydav, line 117: Does it make sense use Promise.all()?

This also raises a question about how we handle errors within the loop.
Should we explicitly return null in case of an error? → headerSuggestionsResults.push(null);
Do we add a suggestion in cases of partial success?

naydav (changes requested):

src/internal-links/handler.js
nitinja: This [PR](https://github.com/adobe/spacecat-api-service/pull/713/files#diff-c22e97ef49f28d57b260ed547e238893e7086ad2c109c4d4fcff963f870e776bR1670) already has schema updates, its in review 
image

src/internal-links/suggestions-generator.js
```diff
@@ -0,0 +1,151 @@
...
+    { length: totalBatches },
+    (_, i) => siteData.slice(i * BATCH_SIZE, (i + 1) * BATCH_SIZE),
+  );
+
+  log.info(`Processing ${siteData.length} alternative URLs in ${totalBatches} batches of ${BATCH_SIZE}...`);
+
+  const processBatch = async (batch, urlTo) => {
+    try {
+      const requestBody = await getPrompt({ alternative_urls: batch, broken_url: urlTo }, 'broken-backlinks', log);
+      await sleep(1000);
```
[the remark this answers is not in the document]
nitinja: removed sleep, this code was literally copied from broken backlinks, but I have since them amended it.

src/internal-links/suggestions-generator.js
```diff
@@ -0,0 +1,151 @@
...
+
+      if (response.choices?.length >= 1 && response.choices[0].finish_reason !== 'stop') {
+        log.error(`No suggestions found for ${urlTo}`);
+        return null;
+      }
+
+      return JSON.parse(response.choices[0].message.content);
+    } catch (error) {
+      log.error(`Batch processing error: ${error.message}`);
+      return null;
```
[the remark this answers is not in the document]
nitinja: I have created a new function processBatch() and called it in try-catch bloack. Now null returns incase of valid case - no suggestions, but error ias still handled in all other cases.

src/internal-links/suggestions-generator.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+  };
+
+  const processLink = async (link, headerSuggestions) => {
```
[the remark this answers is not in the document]
nitinja: - Created a separate function `processBatches`, that handles batches with only required params.
- No input params being modified
- In all cases, return value is still a link object spreaded, urls_suggested and ai_ratonale. That makes it a uniform interface. Added comment for clarity.

src/internal-links/suggestions-generator.js
nitinja: Since this function is a Inner function inside a larger function, and it is a common practice to access the variable in the outer function scope, so I'm not sure that it's necessary to pass parameters here.

nitinja: duplicate

src/internal-links/suggestions-generator.js
nitinja: I would like to use promise.all but promise.all fails even if a single promise in the list fails which we do not want. 
Inserting null in headerSuggestionResults array is important because we are using this array and location of each result in later processing as `headerSuggestionsResults[index]`.

naydav (approved):

PR #610, opened 2025-01-29
feat: updating the interval for rum query

05rahulsingh:

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/forms-opportunities/handler.js
+++ b/src/forms-opportunities/handler.js
@@ -15,8 +15,8 @@ import { getRUMDomainkey } from '../support/utils.js';
 import { AuditBuilder } from '../common/audit-builder.js';
 import { wwwUrlResolver } from '../common/audit.js';
 
-const DAILY_THRESHOLD = 200;
-const INTERVAL = 7; // days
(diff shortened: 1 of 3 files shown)


src/forms-opportunities/handler.js
```diff
@@ -15,8 +15,8 @@ import { getRUMDomainkey } from '../support/utils.js';
 import { AuditBuilder } from '../common/audit-builder.js';
 import { wwwUrlResolver } from '../common/audit.js';
 
-const DAILY_THRESHOLD = 200;
-const INTERVAL = 7; // days
+const DAILY_THRESHOLD = 0;
```
vdua, line 18: any reason for setting the daily threshold to 0 ?

nit23uec: @05rahulsingh , i think we should keep this threshold.
its just that we should not apply this threshold on form pages but only on the pages having the form CTA for the form oppty: high page views, low form ctr

nit23uec (changes requested):
inline comments.

PR #631, opened 2025-02-05
feat: save rum-traffic as metrics to s3

rpapani:
For calculating the projected traffic lost/value, we need organic traffic from RUM for the given page. Rather than each opportunity crunching the RUM data separately, adding this `rum-traffic` audit, which will save the rum-traffic data to importer metrics. 
So, the opportunities can obtain the rum-traffic data as 
```
import { getStoredMetrics } from '@adobe/spacecat-shared-utils';

const rumTrafficData = await getStoredMetrics(
    { source: 'rum', metric: 'rum-traffic', siteId },
    context,
  );
const pageOrganicTraffic = rumTrafficData['<url>']?.earned || 0;
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -37,6 +37,7 @@ import metaTags from './metatags/handler.js';
 import costs from './costs/handler.js';
 import structuredData from './structured-data/handler.js';
 import siteDetection from './site-detection/handler.js';
+import rumTraffic from './rum-traffic/rum-traffic.js';
 
 const HANDLERS = {
   apex,
@@ -58,6 +59,7 @@ const HANDLERS = {
   'structured-data': structuredData,
   'forms-opportunities': formsOpportunities,
   'site-detection': siteDetection,
+  'rum-traffic': rumTraffic,
   dummy: (message) => ok(message),
 };
 
--- a/src/rum-traffic/rum-traffic.js
+++ b/src/rum-traffic/rum-traffic.js
@@ -50,6 +50,7 @@ export async function handler(auditUrl, context, site) {
     return acc;
   }, {});
   log.info(`Traffic data: ${JSON.stringify(trafficData, null, 2)}`);
+  console.log('site id', site.id);
   const metricsPath = await storeMetrics(
     trafficData,
     { siteId: site.id, source: 'rum', metric: 'rum-traffic' },
--- a/src/rum-traffic/rum-traffic.js
+++ b/src/rum-traffic/rum-traffic.js
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2025 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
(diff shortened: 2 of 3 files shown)


src/index.js
```diff
@@ -58,6 +59,7 @@ const HANDLERS = {
   'structured-data': structuredData,
   'forms-opportunities': formsOpportunities,
   'site-detection': siteDetection,
+  'rum-traffic': rumTraffic,
```
rpapani, line 62: any preference `rum-traffic` vs `rum-metrics` ?

dzehnder: looks like it's only about traffic data, not other metrics. Would keep `rum-traffic` in that case

dzehnder (approved):

ekremney:
@iuliag please review this one: possible overlap with import worker

src/rum-traffic/rum-traffic.js
```diff
@@ -0,0 +1,78 @@
...
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import RUMAPIClient from '@adobe/spacecat-shared-rum-api-client';
+import { storeMetrics } from '@adobe/spacecat-shared-utils';
```
solaris007, line 14: this should not be in utils, rather in the rum-api client package

solaris007 (changes requested):
- this should be an import, not an audit. please move to import worker
- please make sure you add relevant configurations to global config / jobs dispatcher

src/rum-traffic/rum-traffic.js
iuliag: The `storeMetrics` comes from `https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-utils/src/metrics-store.js` and has nothing to do with RUM.
It's a wrapper over the S3 client and put command and makes sure the data is stored in a uniform folder structure for all metrics: `metrics/${siteId}/${source}/${metric}.json`.

rpapani:
@iuliag
> This should be an [import](https://github.com/adobe/spacecat-import-worker/tree/main?tab=readme-ov-file#how-to-create-a-new-import) of type page-organic-traffic with source rum and destination default (meaning S3).

sure, will move the logic to import worker

> ingestor service for unified metrics collection so that auto-identify and auto-suggest have all kinds of metrics available without duplicating data collection implementation and increasing COGS 

is this ingestor service available already? or the future service that we'll implement post GA? If I implement this page-organic-traffic in import worker, will this be ported to ingestor service later?

> timeline for metrics collection to cover needs of the success report
I'm using 30 days as a window for collecting this data. If we make this timeline configurable, then we'll have to add that information in the metrics as well to inform the consumers about the timeline.

PR #632, opened 2025-02-05
fix: error propagation in handlers

dzehnder:
adding cause to the error within a handler, to be able to retrieve the corresponding stack trace in the log

### DIFF
--- a/src/backlinks/handler.js
+++ b/src/backlinks/handler.js
@@ -34,6 +34,8 @@ async function filterOutValidBacklinks(backlinks, log) {
       if (error instanceof AbortError) {
         log.warn(`Request to ${url} timed out after ${timeout}ms`);
         return { ok: false, status: 408 };
+      } else {
+        log.warn(`Request to ${url} failed with error: ${error.message}`);
       }
     } finally {
       clearTimeout(id);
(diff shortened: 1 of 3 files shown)


src/backlinks/handler.js
```diff
@@ -38,21 +38,16 @@ async function filterOutValidBacklinks(backlinks, log) {
...
     }
-    return null;
+    return { ok: false, status: 500 };
   };
 
   const isStillBrokenBacklink = async (backlink) => {
-    try {
-      const response = await fetchWithTimeout(backlink.url_to, TIMEOUT);
-      if (!response.ok && response.status !== 404
+    const response = await fetchWithTimeout(backlink.url_to, TIMEOUT);
```
solaris007, line 45: `fetchWithTimeout` may return `null`

src/backlinks/handler.js
```diff
@@ -38,21 +38,16 @@ async function filterOutValidBacklinks(backlinks, log) {
     } finally {
       clearTimeout(id);
     }
-    return null;
+    return { ok: false, status: 500 };
```
solaris007, line 41: we still should log further up if it is not an `AbortError`

solaris007 (approved):

PR #634, opened 2025-02-06
fix: add fixed status to suggestions

alinarublea:

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/support/utils.js
+++ b/src/support/utils.js
@@ -22,10 +22,11 @@ import URI from 'urijs';
 import { JSDOM } from 'jsdom';
 import { GetSecretValueCommand, SecretsManagerClient } from '@aws-sdk/client-secrets-manager';
 import { ListObjectsV2Command } from '@aws-sdk/client-s3';
+import { AbortController, AbortError } from '@adobe/fetch';
 import { getObjectFromKey } from '../utils/s3-utils.js';
 
 URI.preventInvalidHostname = true;
-
+export const TIMEOUT = 3000;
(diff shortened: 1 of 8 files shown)


src/support/utils.js
```diff
@@ -298,6 +299,36 @@ export const getScrapedDataForSiteId = async (site, context) => {
...
+    return response;
+  } catch (error) {
+    if (error instanceof AbortError) {
+      log.warn(`Request to ${url} timed out after ${timeout}ms`);
+      return { ok: false, status: 408 };
+    }
+  } finally {
+    clearTimeout(id);
+  }
+  return null;
```
dzehnder, line 319: instead of returning `null` we could return `{ ok: false }` to avoid exceptions

src/utils/data-access.js
```diff
@@ -53,16 +54,23 @@ export async function syncSuggestions({
...
   const existingSuggestions = await opportunity.getSuggestions();
-  // Remove outdated suggestions
   await Promise.all(
     existingSuggestions
       .filter((existing) => !newDataKeys.has(buildKey(existing.getData())))
-      .map((suggestion) => suggestion.remove()),
+      .map(async (suggestion) => {
+        const isSuggestionFixed = await isFixed(suggestion);
+        if (isSuggestionFixed) {
+          suggestion.setStatus(Suggestion.STATUSES.FiXED);
```
solaris007, line 69: it would be better to use `Suggestion.bulkUpdateStatus`

PR #641, opened 2025-02-10
fix: finalUrl was removed

solaris007:
Re-add `finalUrl` after it was removed accidentally.

### DIFF
--- a/src/backlinks/handler.js
+++ b/src/backlinks/handler.js
@@ -82,6 +82,7 @@ export async function brokenBacklinksAuditRunner(auditUrl, context, site) {
     log.error(`Broken Backlinks audit for ${siteId} with url ${auditUrl} failed with error: ${e.message}`, e);
     return {
       fullAuditRef: auditUrl,
+      finalUrl: site.getBaseURL(),
       auditResult: {
         error: `Broken Backlinks audit for ${siteId} with url ${auditUrl} failed with error: ${e.message}`,
         success: false,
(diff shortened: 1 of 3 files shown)


src/backlinks/handler.js
```diff
@@ -82,6 +82,7 @@ export async function brokenBacklinksAuditRunner(auditUrl, context, site) {
     log.error(`Broken Backlinks audit for ${siteId} with url ${auditUrl} failed with error: ${e.message}`, e);
     return {
       fullAuditRef: auditUrl,
+      finalUrl: site.getBaseURL(),
```
iuliag, line 85: This should be `auditUrl`, since that's the URL we resolve from the base URL and we use for querying Ahrefs API.

iuliag (changes requested):

iuliag (commented):
In the auditing, it was previously set to:
```auditContext.finalUrl = await composeAuditURL(site.getBaseURL());```
https://github.com/adobe/spacecat-audit-worker/commit/03a1ee2b72d0f11c1fcd017eeea3b0ce35a236ee#diff-212dd58956c947938959c789cc3912af8f1b884de5543486166247e12f8ef215L85

In the Slack message, we display the final URL without the protocol.

For me, it makes sense that the final URL to be used in the reporting to be the actual URL that we use to query our data source to produce the audit results, not the base URL of the site when it was added in SpaceCat.

iuliag (approved):

PR #665, opened 2025-02-17
feat: spacecat user-agent

AndreiAlexandruParaschiv:
WIP


solaris007:
@AndreiAlexandruParaschiv i think the user agent should be baked into `tracingFetch` so that devs don't have to think about it. one can always override with custom headers.

AndreiAlexandruParaschiv:
No longer needed as the fct tracingFetch has user-agent defined (shared-utils)

PR #671, opened 2025-02-18
feat: handleOutdatedSuggestions gracefully

sandsinh:
Problem:
sync suggestions deletes outdated suggestions from the db. 

Proposed solution: 
For fix history phase 1, we are proposing to retain the suggestions and mark them 'OUTDATED' if they are not found in the latest audit run. 

Changes: 
1. After this change, any existing suggestion not found in the audit result will be marked as 'OUTDATED' if their current status is not 'FIXED', 'OUTDATED', 'ERROR', 'SKIPPED'. 
2. Any 'FIXED' or 'OUTDATED' suggestion found in the audit result will be marked as 'NEW'.  (Covers cases for possible regression)

## Related Issues

Thanks for contributing!


src/utils/data-access.js
```diff
@@ -74,18 +108,20 @@ export async function retrieveAuditById(dataAccess, auditId, log) {
 export async function syncSuggestions({
   opportunity,
   newData,
+  context,
   buildKey,
   mapNewSuggestion,
   log,
 }) {
   const newDataKeys = new Set(newData.map(buildKey));
   const existingSuggestions = await opportunity.getSuggestions();
```
solaris007, line 117: while already there, with this change, this will possibly cause performance issues as the list of suggestions grows forevermore. at the very least a status filter should be employed. in fact the handling of outdated suggestions could  be implemented as a bulk-operation as part of the `SuggestionCollection` in data access. We should not have use of data access queries in an unbounded fashion.

src/utils/data-access.js
```diff
@@ -100,6 +136,13 @@ export async function syncSuggestions({
           ...existing.getData(),
           ...newDataItem,
         });
+        if ([
+          SuggestionDataAccess.STATUSES.FIXED,
+          SuggestionDataAccess.STATUSES.OUTDATED,
+        ].includes(existing.getStatus())) {
+          log.warn('Resolved suggestion found in audit. Possible regression.');
+          existing.setStatus(SuggestionDataAccess.STATUSES.NEW);
```
solaris007, line 144: there's already a `Suggestion.bulkUpdateStatus` method...

solaris007 (changes requested):

src/utils/data-access.js
sandsinh: I believe setStatus does not save the suggestion but only updates the entity locally. Save operation is performed later at L146.

src/utils/data-access.js
sandsinh: We have three operations here:
1. handleOutdatedSuggestion
This method targets only 'NEW' and 'IN_PROGRESS' suggestions, having bounded query with status filter makes sense here. 

2. updateExistingSuggestion :
This method targets suggestions of all statuses. No bounded query can be written for status filter. [Based on our assumption that even FIXED or OUTDATED suggestion can be reopened in future]

3. createNewSuggestion : 
This method also requires information for all the suggestions irrespective of the statuses. No bounded query here as well. [Based on our assumption that even FIXED or OUTDATED suggestion can be reopened in future]

Bounding the operations 2 & 3 to 'NEW' and 'IN_PROGRESS' suggestions could work but as a consequence we would have to allow multiple suggestions with the same dataKeys in the db. This means that instead of reopening an existing FIXED or OUTDATED suggestion, we will create 'NEW' suggestions if the audit find the same issue again.

Note: Having the ability to query using the dataKeys would help us write bounded queries very efficiently for all the operations. But i think it is not possible yet. 

@solaris007 Let me know what are you thoughts on this.

src/utils/data-access.js
```diff
@@ -59,6 +60,39 @@ export async function retrieveAuditById(dataAccess, auditId, log) {
...
+    const { Suggestion } = context.dataAccess;
+    const existingOutdatedSuggestions = existingSuggestions
+      .filter((existing) => !newDataKeys.has(buildKey(existing.getData())))
+      .filter((existing) => ![
+        SuggestionDataAccess.STATUSES.OUTDATED,
+        SuggestionDataAccess.STATUSES.FIXED,
+        SuggestionDataAccess.STATUSES.ERROR,
+        SuggestionDataAccess.STATUSES.SKIPPED,
+      ].includes(existing.getStatus()));
+    if (existingOutdatedSuggestions.length > 0) {
```
solaris007, line 87: use `isNonEmptyArray`

src/utils/data-access.js
```diff
@@ -59,6 +60,39 @@ export async function retrieveAuditById(dataAccess, auditId, log) {
...
+ * @param {Set} params.newDataKeys - The set of new data keys to check for outdated suggestions.
+ * @param {Function} params.buildKey - The function to build a unique key for each suggestion.
+ * @param {Object} params.context - The context object containing the data access object.
+ * @returns {Promise<void>} - Resolves when the outdated suggestions are updated.
+ */
+const handleOutdatedSuggestions = async ({
+  existingSuggestions, newDataKeys, buildKey, context,
+}) => {
+  // Check if context is provided
+  if (context) {
```
solaris007, line 77: use early return for readability

solaris007 (approved):

src/utils/data-access.js
sandsinh: We are calling .save on the instance after this already.

src/utils/data-access.js
sandsinh: Change incorporated.

src/utils/data-access.js
sandsinh: Added

PR #674, opened 2025-02-19
fix: audit opportunities not creating

martinst06:
Metatags auditId causing error


dipratap:
@martinst06 I don't think this is related to meta-tags only, but happening for other audits eg. broken backlinks as well.

PR #676, opened 2025-02-20
fix: audits not creating

martinst06:
Audits not creating due to issue in previous refactoring of opportunities and suggestions


solaris007:
superseded by https://github.com/adobe/spacecat-audit-worker/pull/679

PR #682, opened 2025-02-21
feat: unthrottle broken backlinks suggestions

dzehnder:
Due to previous token restrictions from Firefall, the amount of requests where throttled by doing each request sequentially and pause for a second in between.

As the restriction has been lifted, it is now possible to parallelize requests with no timeout in between.


src/backlinks/handler.js
```diff
@@ -191,36 +197,32 @@ export const generateSuggestionData = async (finalUrl, auditData, context, site)
...
-      headerSuggestionsResults.push(JSON.parse(response.choices[0].message.content));
+      return JSON.parse(response.choices[0].message.content);
     } catch (error) {
       log.error(`Header suggestion error: ${error.message}`);
-      headerSuggestionsResults.push(null);
+      return null;
     }
-  }
+  });
+  const headerSuggestionsResults = await Promise.all(headerSuggestionsPromises);
```
AndreiAlexandruParaschiv, line 220: Promise.allSettled instead?

AndreiAlexandruParaschiv (approved):

src/backlinks/handler.js
dzehnder: I am already handling possible errors in each promise individually, there is no need for `allSettled`

PR #707, opened 2025-02-26
fix: [alt-text] reduces image batching and improves logging

igor-grubic:
- Improves logging for better FF debugging
- lowers batching to 3. This will make the audits longer (3 images every 5seconds) but will hopefully reduce the erroring
- includes images that are not from the host as well
- refreshes prompt to avoid caching confusion

## Related Issues
[ASSETS-47969](https://jira.corp.adobe.com/browse/ASSETS-47969)

Thanks for contributing!

### DIFF
--- a/src/image-alt-text/suggestionsEngine.js
+++ b/src/image-alt-text/suggestionsEngine.js
@@ -30,10 +30,11 @@ const filterImages = (imageUrls, auditUrl) => {
   imageUrls.forEach((imageUrl) => {
     if (!SUPPORTED_FORMATS.test(imageUrl)) {
       unsupportedFormatImages.push(imageUrl);
-    } else if (imageUrl.includes(auditUrl)) {
+    } else if (imageUrl.includes(auditUrl.replace('https://', ''))) {
       imagesFromHost.push(imageUrl);
     } else {
       otherImages.push(imageUrl);
+      imagesFromHost.push(imageUrl);
     }
   });
 
--- a/src/image-alt-text/suggestionsEngine.js
+++ b/src/image-alt-text/suggestionsEngine.js
@@ -61,6 +61,7 @@ const generateBatchPromises = (
   };
   const prompt = await getPrompt({ images: batch }, PROMPT_FILE, log);
   try {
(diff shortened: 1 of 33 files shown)


static/prompts/image-alt-text.prompt
```diff
@@ -2,14 +2,15 @@ You are tasked with identifying suitable text for the alt attribute of images.
...
 ### Rules:
-1. The alt text should be helpful for the user, not the search engine.
-2. Consider key elements of why you chose this image, instead of describing every little detail. No need to say image of or picture of. But, do say if its a logo, illustration, painting, or cartoon.
-3. If you can recognize a person in the image, use their name when known.
-4. Follow the industry guidelines for accessibility, https://www.w3.org/WAI/tutorials/images/
-5. Alt-text should reflect how the image relates to the content. Avoid irrelevant descriptions.
-6. Use natural language, ensuring you're not "stuffing" SEO keys.
-7. If an image is purely decorative and adds no functional or informational value, use an empty string as the alt text.
-8. Ideal description length is 50-60 characters.
+
```
igor-grubic, line 5: cache busting "trick"

src/image-alt-text/suggestionsEngine.js
```diff
@@ -61,6 +62,7 @@ const generateBatchPromises = (
   };
   const prompt = await getPrompt({ images: batch }, PROMPT_FILE, log);
   try {
+    log.info(`[${AUDIT_TYPE}]: Batch prompt:`, prompt);
```
solaris007, line 65: we're trying to be mindful of log volume on coralogix. is this one strictly necessary in production? same for info statements below.

solaris007 (approved):

src/image-alt-text/suggestionsEngine.js
igor-grubic: @solaris007 This is more of a temporary situation, as we're getting a lot of bad responses from FF and looking to understand why. I'll bring this forward for these few days as we're doing investigation and remove it afterwards.

PR #709, opened 2025-02-26
fix: handle missing S3 contents and refine scrape file selection

dzehnder:
Fixing issues that came up recently:

- Use optional chaining for listResponse.Contents and default to an empty array if there are no files found.
- Refine index file detection in the scraped content. Previously only worked if it was at the top level.


alinarublea (approved):

dzehnder:
> Should there be any tests for these changes?

first case is covered, I can add a test case for the second one

solaris007 (approved):

iuliag (approved):

PR #713, opened 2025-02-26
fix: retaining slash at Urls end in Meta-tags

dipratap:
Slash at Urls end is not retained in meta-tags audit results

https://jira.corp.adobe.com/browse/SITES-29515

https://cq-dev.slack.com/archives/C085NJRJG2W/p1740428584585269

> One observation I had: volvo site expect a / at the end of many of their urls otherwise they respond with 'Page not found'. For ex. https://www.volvotrucks.us/about-volvo/facilities/parts-distribution-centers
On metadata opportunity UI, many urls on the page are redirecting to 'page not found' due to above issue. Looks like we need to retain the / at the end of these urls. I have created [SITES-29515](https://jira.corp.adobe.com/browse/SITES-29515) for this and am working on it. The detected metadata suggestions are all fine though.

### DIFF
--- a/test/audits/metatags.test.js
+++ b/test/audits/metatags.test.js
@@ -326,6 +326,7 @@ describe('Meta Tags', () => {
         }))).returns({
           Body: {
             transformToString: () => JSON.stringify({
+              finalUrl: 'http://example.com/site-id/',
               scrapeResult: {
                 tags: {
                   title: 'Home Page',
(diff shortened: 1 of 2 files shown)


test/audits/metatags.test.js
```diff
@@ -407,7 +408,7 @@ describe('Meta Tags', () => {
         },
       }));
       expect(addAuditStub.calledOnce).to.be.true;
-    }).timeout(3000);
+    }).timeout(5000);
```
solaris007, line 411: why would this test have to take this long? it's a unit test and should be quick.

dipratap: This test failed once due to timeout so I increased it but looks like it could be due to local lagginess. I removed the timeout and it works fine.

solaris007 (approved):

PR #725, opened 2025-02-28
feat: introduce KPI metrics for broken-backlinks

dzehnder:
Introducing KPI metrics for broken backlinks.
Can only be merged once https://github.com/adobe/spacecat-import-worker/pull/240 has been merged

### DIFF
--- a/src/backlinks/kpi-metrics.js
+++ b/src/backlinks/kpi-metrics.js
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2025 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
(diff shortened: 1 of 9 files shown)


src/backlinks/kpi-metrics.js
```diff
@@ -0,0 +1,72 @@
...
+    log.info(`No RUM traffic data found for site ${siteId}`);
+    return null;
+  }
+
+  const organicTrafficData = await getStoredMetrics(
+    { source: 'ahrefs', metric: 'organic-traffic', siteId },
+    context,
+  );
+
+  if (!isNonEmptyArray(organicTrafficData)) {
```
alinarublea, line 33: I think if there is no organic traffic we can approximate the CPC value to 1

iuliag (approved):
In the future, we'll have to centralize the logic that looks at the latest organic traffic data and computes CPC, as this is now also in https://github.com/adobe/spacecat-api-service/blob/e0e3d1e2310636055b037d24d8ddc038114e31bb/src/controllers/sites.js#L423-L439 and https://github.com/adobe/spacecat-import-worker/blob/e0ae0953ae33aac5775a7e2a9a5d954553f35a36/src/mapper/latest-metrics.js#L30-L35

PR #737, opened 2025-03-04
fix: [alt-text] supports more image mimetypes 

igor-grubic:
Now supports: 
- `webp|png|gif|jpeg|jpg` ootb GPT
- `svg|bmp|tiff|ico` through base64 

Notes: 
- Removes the filtering from `handler.js` and `suggEngine.js`, and moves it to `auditEngine.js` so that it is in a centralized place.
- Obtains base64 blobs for formats that are not supported by FF.  Adding only the more popular ones `svg|bmp|tiff|ico` for control
- Filters by unique image URL
- Filters by unique blob
- runs a slightly different prompt that adds image urls and blobs directly to the prompt rather than using the `"type":"image_url"` feature of FF
- removes some spammy logs that were not necessary

## Related Issues

Thanks for contributing!


src/image-alt-text/auditEngine.js
```diff
@@ -11,39 +11,116 @@
...
     this.auditedTags = {
-      imagesWithoutAltText: [],
+      imagesWithoutAltText: new Map(),
     };
   }
 
   performPageAudit(pageUrl, pageTags) {
     if (!isNonEmptyArray(pageTags?.images)) {
-      this.log.warn(`No images found for page ${pageUrl}`);
+      this.log.debug(`[${AUDIT_TYPE}]: No images found for page ${pageUrl}`);
```
JayKid, line 67: @igor-grubic are `.debug` being shown in Coralogix? 🤔

igor-grubic: @JayKid I changed these to debug as they are `not` shown in coralogix. Reason is that this is run per page, so it leads to a lot of unnecessary spam. I think if we need to debug this we can enable it either there or in the code.
This goes also back to what DJ was saying about reducing spam.

JayKid: Then yes, I agree I'd rather remove it : )

src/image-alt-text/auditEngine.js
```diff
@@ -11,39 +11,125 @@
...
+  const extension = match[1].toLowerCase();
+  return mimeTypesForBase64[extension];
+};
+
+const convertImagesToBase64 = async (imageUrls, auditUrl, log) => {
+  const base64Blobs = [];
+
+  const fetchPromises = imageUrls.map(async (url) => {
+    try {
+      const response = await fetch(new URL(url, auditUrl).toString());
```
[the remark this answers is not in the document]
igor-grubic: d80f2df034ca0b41e97383454a6f61a107a1fca1

PR #744, opened 2025-03-05
feat: Surface structured data errors without auto suggestion

herzog31:
* Fixed a bug where `opportunityAndSuggestions` post processor throws an error if the audit failed. This was detected in the logs.
* If no auto-suggestion can be generated, the actual details of the errors were not surfaced in the UI anymore. With this PR the errors are again surfaced and give customers more insights on how to fix the issue. See attached screenshot.

image


src/structured-data/handler.js
```diff
@@ -125,14 +125,21 @@ export async function processStructuredData(baseURL, context, pages) {
 }
 
 export async function opportunityAndSuggestions(auditUrl, auditData, context) {
+  const { log } = context;
+
+  // Check if audit was successful
+  if (auditData.auditResult.success === false) {
```
solaris007, line 131: this could possibly be applicable to all post processors. should this be handled in the common code?

solaris007 (approved):

iuliag (approved):

src/structured-data/handler.js
herzog31: I agree. This pattern is currently only used by `apex`, `backlinks`, `canonical`, `sitemap` and `structured-data` audits, but can be adopted by all of them.

I'll create a follow up JIRA ticket to update the implementation of the post processor logic and to also update the README to give guidance on how this pattern should be used.

PR #755, opened 2025-03-07
fix: handling site RUM missing for meta-tags

dipratap:
If the site doesn't have RUM enabled then the meta-tags audit fails as it tries to calculate projected traffic value but gets an error in rum query. In this PR, this error has been handled so that failure to calculate projected traffic doesn't fails the meta-tags audit.

### DIFF
--- a/src/metatags/handler.js
+++ b/src/metatags/handler.js
@@ -175,10 +175,9 @@ async function calculateProjectedTraffic(context, auditUrl, siteId, detectedTags
 
     const cpcValue = await calculateCPCValue(context, siteId);
     log.info(`Calculated cpc value: ${cpcValue} for site: ${siteId}`);
-
     const projectedTrafficValue = projectedTrafficLost * cpcValue;
-    // Skip updating projected traffic data if lost traffic value is insignificant
 
+    // Skip updating projected traffic data if lost traffic value is insignificant
     return projectedTrafficValue > 500 ? { projectedTrafficLost, projectedTrafficValue } : {};
   } catch (err) {
     log.warn(`Error while calculating projected traffic for ${auditUrl} : ${siteId}`, err);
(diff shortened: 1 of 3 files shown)


src/metatags/handler.js
```diff
@@ -137,39 +137,52 @@ function getOrganicTrafficForEndpoint(endpoint, rumDataMapMonthly, rumDataMapBiM
...
+        endpoint,
+        rumDataMapMonthly,
+        rumDataMapBiMonthly,
+        log,
+      );
+      Object.values((tags)).forEach((tagIssueDetails) => {
+        // Multiplying by 1% for missing tags, and 0.5% for other tag issues
+        // For duplicate tags, each page's traffic is multiplied by .5% so
+        // it amounts to 0.5% * number of duplicates.
+        const multiplier = tagIssueDetails.issue.includes('Missing') ? 0.01 : 0.005;
```
absarasw, line 171: Variables for multipliers?

ssilare-adobe (approved):
LGTM! 🚀

src/metatags/handler.js
```diff
@@ -137,39 +137,52 @@ function getOrganicTrafficForEndpoint(endpoint, rumDataMapMonthly, rumDataMapBiM
...
     });
-  });
-  return projectedTraffic;
+
+    const cpcValue = await calculateCPCValue(context, siteId);
+    log.info(`Calculated cpc value: ${cpcValue} for site: ${siteId}`);
+    const projectedTrafficValue = projectedTrafficLost * cpcValue;
+
+    // Skip updating projected traffic data if lost traffic value is insignificant
+    return projectedTrafficValue > 500 ? { projectedTrafficLost, projectedTrafficValue } : {};
```
absarasw, line 181: Any discussion on 500 value? Can we make a variable for this as well?

src/metatags/handler.js
dipratap: I tried adding constants but felt that the readability gets impacted, since these multipliers are used at one place only, I feel having these values is fine here. Let me know your thoughts!

PR #757, opened 2025-03-07
fix: generate default guidance for low conv opportunity

nit23uec:

## Related Issues

Thanks for contributing!


src/forms-opportunities/utils.js
```diff
@@ -182,6 +182,14 @@ export async function generateOpptyDataForHighPageViewsLowFormNav(formVitals, co
   );
 }
 
+export function isSearchForm(scrapedFormData) {
```
vdua, line 185: unsubscribe is not a search form. I am assuming we are creating a single API  to identify `isRelevantForm` rather than having isSearchForm, isFilterForm, isCalculatorForm, ...etc.

src/forms-opportunities/opportunityHandler.js
```diff
@@ -10,7 +10,77 @@
...
+  let urlMatches = false;
+  if (isNonEmptyArray(scrapedData?.formData)) {
+    for (const form of scrapedData.formData) {
+      const formUrl = new URL(form.finalUrl);
+      const opportunityUrl = new URL(oppoty.form);
+      let allSearchForms = true;
+      if (formUrl.origin + formUrl.pathname === opportunityUrl.origin + opportunityUrl.pathname) {
+        urlMatches = true;
+        for (const result of form.scrapeResult) {
+          const { visibleFieldCount: fieldCount } = result;
```
vdua, line 28: I am not sure why we are redefining it. It is being used only at a single place and there is no visibleFieldCount variable as well. This should also work
```
const { visibleFieldCount } = result;
```

src/forms-opportunities/utils.js
nit23uec: yeah agree, we can do that as an improvement.

src/forms-opportunities/opportunityHandler.js
nit23uec: incorporated.

src/forms-opportunities/opportunityHandler.js
```diff
@@ -10,7 +10,77 @@
...
+          if (!seachFormDetected && visibleFieldCount > 6) {
+            largeForm = true;
+          }
+          if (!seachFormDetected && !isFormVisibleATF) {
+            isVisibleATF = false;
+          }
+          if (!seachFormDetected) {
+            allSearchForms = false;
+          }
+        }
```
vdua, line 40: The code can be simplified
```
const nonSearchForms = form.scrapeResult.filter(x => !searchForm)
if (nonSearchForms.length !== 0) {
   const {isLargeForm, visibleATF} = nonSearchForms.reduce((acc, {visibleATF, visibleFieldCount}) => {...}, {}) //
}
```

src/forms-opportunities/utils.js
```diff
@@ -190,33 +198,24 @@ export async function generateOpptyDataForHighPageViewsLowFormNav(formVitals, co
...
+      return false; // exclude search pages
+    }
+    if (isNonEmptyArray(scrapedData?.formData)) {
+      for (const form of scrapedData.formData) {
+        const formUrl = new URL(form.finalUrl);
+        const opportunityUrl = new URL(opportunity.form);
+
+        if (formUrl.origin + formUrl.pathname === opportunityUrl.origin + opportunityUrl.pathname) {
+          urlMatches = true;
+          for (const result of form.scrapeResult) {
```
vdua, line 214:
```
hasANonSearchForm = form.scrapeResult.find(x => !isSearchForm(x)) != null
```
or 

```
allSearchForms = form.scrapeResult.find(x => !isSearchForm(x)) == null
```

vdua (approved):

src/forms-opportunities/opportunityHandler.js
```diff
@@ -10,7 +10,77 @@
  * governing permissions and limitations under the License.
  */
 
-import { filterForms, generateOpptyData } from './utils.js';
+import { isNonEmptyArray, isNonEmptyObject } from '@adobe/spacecat-shared-utils';
+import { filterForms, generateOpptyData, isSearchForm } from './utils.js';
+
+function generateDefaultGuidance(scrapedData, oppoty) {
+  let largeForm = false;
+  let isVisibleATF = true;
```
vdua, line 18: to make it readable I would have used isBelowTheFold : true so that the largeForm check and isBelowTheFold check is consistent

src/forms-opportunities/opportunityHandler.js
nit23uec: done.

PR #760, opened 2025-03-07
fix: Group structured data issues before generating suggestions

herzog31:
We observed customers which had the same structured data issue on hundreds of pages. This PR aims to reduce the amount of redundant requests made in these scenarios.

* De-duplicate errors reported from GSC.
    * e.g. breadcrumb with 5 items where all 5 items are missing an attribute previously causes 5 `Missing field "name"` errors. 
    * Multiple errors for the same entity are combined into a single issue.
* Once a suggestion for a specific issue was created, it is cached and can be re-used for other pages that have the exact some issue.
    * e.g. if the same issue for the breadcrumb entity happens on 5 pages, only the first page will generate a suggestion. The other 4 pages will re-use the suggestion from the first page.
* Introduced a hard limit of 50 Firefall requests per audit as a failsafe to prevent timeouts. This can be increased later if needed.
* The scrape for a page is now only retrieved if all conditions are met to generate a suggestion using Firefall.
* Reduced number of debug logs and added page URLs to all log messages.
* Updated unit tests.

## Related Issues
See SITES-29354 for more details.


src/structured-data/handler.js
```diff
@@ -28,6 +28,9 @@ import { generatePlainHtml, getScrapeForPath } from '../support/utils.js';
 const auditType = Audit.AUDIT_TYPES.STRUCTURED_DATA;
 const auditAutoSuggestType = Audit.AUDIT_TYPES.STRUCTURED_DATA_AUTO_SUGGEST;
 
+// Limit to avoid excessive Firefall requests. Can be increased if needed.
+const FIREFALL_REQUEST_LIMIT = 50;
```
solaris007, line 32: should this be made configurable via AWS secrets in case we need to play?

solaris007 (approved):

src/structured-data/handler.js
herzog31: Thanks for the suggestion. I added a new `AUDIT_STRUCTURED_DATA_FIREFALL_REQ_LIMIT` secret to the prod and ci secret manager.

PR #802, opened 2025-04-11
fix: [alt-text] adds locale to suggestions

igor-grubic:
This PR attempts to identify the language of a page, the get the AI suggestions in the specified language.
To identify a language:
1. checks the `lang` attribute on `<html>`
2. looks for any `meta[http-equiv="Content-Language"], meta[name="language"]` with content attribute
3. takes the text of the body and uses [franc](https://github.com/wooorm/franc) to derive the locale

## Related Issues

https://jira.corp.adobe.com/browse/ASSETS-49282

Thanks for contributing!


src/image-alt-text/auditEngine.js
```diff
@@ -85,6 +122,10 @@ export default class AuditEngine {
       return;
     }
 
+    const pageLanguage = getPageLanguage({ document: pageTags.dom?.window?.document });
+
+    this.log.info(`[${AUDIT_TYPE}]: Page language: ${pageLanguage}`);
```
JayKid, line 127: @igor-grubic if we leave this log in place, I think we should probably add which page it refers to. But maybe we don't need this since it will be in the audit's results?

static/prompts/image-alt-text.prompt
```diff
@@ -16,16 +16,15 @@ You are an expert SEO consultant, and your goal is to suggest a description for
...
 ### Response Format:
 Your response must be a valid JSON object using the following structure:
 [{
   "image_url": "string",
   "suggestion": "string",
   "ai_rationale": "string",
   "confidence_score": number,
-  "is_appropriate": boolean
+  "is_appropriate": boolean,
+  "language": "string"
```
JayKid, line 27: Just FYI, since we need to change the prompt, we will need to consider this in the AI Ethics review / report generator so what they approve is the same that we use in the audit.

JayKid (approved):

PR #813, opened 2025-04-16
chore(forms): refactor forms audit worker tests

nit23uec:

## Related Issues

Thanks for contributing!


src/forms-opportunities/formcalc.js
```diff
@@ -0,0 +1,191 @@
+/*
```
05rahulsingh, line 1: will be removing from spacecatshared in subsequent pr, and changing references. Correct.
Idea was to have this at bottom level so that it can be consumed by audit and reporting. Now since reporting is consuming oppties directly, we can move this i believe

nit23uec: correct.

PR #825, opened 2025-04-29
fix: Sites 30031, skip creation of opportunity or update opportunity status incase of no issues found in broken internal links

nitinja:
If broken internal links audit does not return any results,
- If opportunity exists
   -  Update opportunity status as resolved.
- If opportunity does not exist, skip creation 

## Related Issues
https://jira.corp.adobe.com/browse/SITES-30031

Thanks for contributing!


src/internal-links/handler.js
```diff
@@ -146,6 +144,34 @@ export async function opportunityAndSuggestionsStep(context) {
...
   // TODO: skip opportunity creation if no internal link items are found in the audit data
   const kpiDeltas = calculateKpiDeltasForAudit(brokenInternalLinks);
+
+  if (!isNonEmptyArray(brokenInternalLinks)) {
+    // no broken internal links found
+    // fetch opportunity
+    const { Opportunity } = dataAccess;
+    let opportunity;
+    try {
+      // eslint-disable-next-line max-len
```
iuliag, line 154: Could you fix the linting issue instead of disabling the check?

src/internal-links/handler.js
```diff
@@ -146,6 +144,34 @@ export async function opportunityAndSuggestionsStep(context) {
...
+      log.error(`Fetching opportunities for siteId ${site.getId()} failed with error: ${e.message}`);
+      throw new Error(`Failed to fetch opportunities for siteId ${site.getId()}: ${e.message}`);
+    }
+
+    if (!opportunity) {
+      log.info(`[${AUDIT_TYPE}] [Site: ${site.getId()}] no broken internal links found, skipping opportunity creation`);
+    } else {
+      // no broken internal links found, update opportunity status to RESOLVED
+      log.info(`[${AUDIT_TYPE}] [Site: ${site.getId()}] no broken internal links found, but found opportunity, updating status to RESOLVED`);
+      await opportunity.setStatus(Oppty.STATUSES.RESOLVED);
```
iuliag, line 167: Suggestions of an opportunity also have a status. Shouldn't they be updated as well?
Otherwise, you could have an opportunity resolved with suggestions in different statuses like: new, approved, in progress.

test/audits/internal-links/handler.test.js
```diff
@@ -331,12 +332,33 @@ describe('broken-internal-links audit opportunity and suggestions', () => {
...
+        generateSuggestionData: () => [],
+      },
+    });
+
+    await expect(
+      handler.opportunityAndSuggestionsStep(context),
+    ).to.be.rejectedWith('read error happened');
+
+    // make sure that no new suggestions are added
+    expect(opportunity.addSuggestions).to.not.have.been.called;
```
iuliag, line 352: `addSuggestions` wouldn't be called at all if there aren't broken links found, no matter if fetching succeeds or throws.

src/internal-links/handler.js
nitinja: Honestly, I thought about it - but I am not sure if we need to mark all suggestions inside an opportunity as resolved. 
On the second thought, I will resolve all suggestions from that opportunity - just to be sure.

src/internal-links/handler.js
```diff
@@ -146,6 +144,45 @@ export async function opportunityAndSuggestionsStep(context) {
...
```
rpapani, line 167: @nitinja Is there a way to know if the customer have used our suggestions? RESOLVED sounds like customer have used our suggestions and resolved it. If the customer didn't use our auto-suggest/auto-optimize, then this would have been OUTDATED right? Or OUTDATED is handled at Suggestions level where as Opportunity can only be RESOLVED?

rpapani (approved):
Looks good to me, small comment though, please check that.

PR #855, opened 2025-05-13
fix: urlResolver for meta-tags

dipratap:
### **Issue reported** 
The meta-tag audit identified new opportunities for hersheyland and created new suggestions. But some of them were already ignored, e.g. “Duplicate H1” for  https://www.hersheyland.com/milk-duds and https://www.hersheyland.com/products/milk-duds-candy-5-oz-box.html. So now the same suggestion is in “Current” and “Ignored”.

### **RCA**
The meta-tags audit identifies suggestions through key: url | issue | tagContent
The URLs in both the suggestions are different.
From old audit:
"url": "https://hersheyland.com/milk-duds",
From new audit:
"url": "www.hersheyland.com/milk-duds"
This is because there has been a [change](https://github.com/adobe/spacecat-audit-worker/pull/750) to use the new urlResolver, instead of the old noopUrlResolver, which resulted in this issue.

### **Fix**
Use wwwUrlResolver for resolving domain for RUM, for other things use site base url.

### DIFF
--- a/src/metatags/handler.js
+++ b/src/metatags/handler.js
@@ -300,7 +300,7 @@ export async function submitForScraping(context) {
 }
 
 export default new AuditBuilder()
-  .withUrlResolver((site) => site.resolveFinalURL())
+  .withUrlResolver((site) => site.getBaseURL())
   .addStep('submit-for-import-top-pages', importTopPages, AUDIT_STEP_DESTINATIONS.IMPORT_WORKER)
   .addStep('submit-for-scraping', submitForScraping, AUDIT_STEP_DESTINATIONS.CONTENT_SCRAPER)
   .addStep('run-audit-and-generate-suggestions', runAuditAndGenerateSuggestions)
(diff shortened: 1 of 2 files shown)


src/metatags/handler.js
```diff
@@ -143,9 +143,9 @@ function getOrganicTrafficForEndpoint(endpoint, rumDataMapMonthly, rumDataMapBiM
 }
 
 // Calculate the projected traffic lost for a site
-async function calculateProjectedTraffic(context, auditUrl, siteId, detectedTags, log) {
+async function calculateProjectedTraffic(context, site, detectedTags, log) {
   const options = {
-    domain: auditUrl,
+    domain: await wwwUrlResolver(site, context),
```
ekremney, line 148: does `await` here work?

dipratap: Yes, it is working.

ekremney (approved):

PR #859, opened 2025-05-15
feat: add preview page authentication to canonical audit

slitviachenko:

## Related Issues
https://jira.corp.adobe.com/browse/SITES-31664

This PR adds support for authenticating preview pages in the canonical audit functionality. When auditing pages on `.page` domains, the system will now retrieve and use the appropriate authentication token.

### DIFF
--- a/src/canonical/handler.js
+++ b/src/canonical/handler.js
@@ -11,9 +11,13 @@
  */
 
 import { JSDOM } from 'jsdom';
-import { composeBaseURL, tracingFetch as fetch } from '@adobe/spacecat-shared-utils';
+import { composeBaseURL, tracingFetch as fetch, retrievePageAuthentication } from '@adobe/spacecat-shared-utils';
 import { AuditBuilder } from '../common/audit-builder.js';
 import { noopUrlResolver } from '../common/index.js';
+
+/**
+ * @import {type RequestOptions} from "@adobe/fetch"
+*/
 
 export const CANONICAL_CHECKS = Object.freeze({
   CANONICAL_TAG_EXISTS: {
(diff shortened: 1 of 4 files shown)


src/canonical/handler.js
```diff
@@ -459,6 +470,11 @@ export async function validateCanonicalRecursively(canonicalUrl, log, visitedUrl
  * @returns {Promise<Object>} An object containing the audit results.
  */
 export async function canonicalAuditRunner(baseURL, context, site) {
+  function isPreviewPage(url) {
+    const urlObj = new URL(url);
+    return urlObj.hostname.endsWith('.page');
+  }
+
```
dzehnder, line 477: this function could be added to the `utils.js` file

src/canonical/handler.js
```diff
@@ -479,11 +495,29 @@ export async function canonicalAuditRunner(baseURL, context, site) {
       };
     }
 
+    /**
+     * @type {RequestOptions}
+     */
+    const options = {};
+    if (isPreviewPage(baseURL)) {
```
dzehnder, line 502: not sure if all preview pages will require authentication. Maybe we can check if a token can be retrieved through `retrievePageAuthentication`. If no token is found we do not authenticate, otherwise we do. WDYT?

dzehnder (approved):

src/canonical/handler.js
slitviachenko: @dzehnder , Do you mean to unconditionally invoke `retrievePageAuthentication` every time for all pages?
The only issue I see at the moment is that, in addition to storing the page access token, we also need to store the `authentication` mechanism - Bearer, Digest, etc.
For `.page` pages, the `token` scheme is used, which I believe is tied to Helix implementation details and is a `custom` one.
```
        options.headers = {
          Authorization: `token ${token}`,
        };
```
https://github.com/adobe/spacecat-audit-worker/pull/859/files#diff-88d6eda336d671388349471e2f7210bbb88fba4bd0d4fdbbd7b02cc632aaadc7R506-R508

slitviachenko: >If no token is found we do not authenticate, otherwise we do. WDYT?

This is exactly how it works for `.page` pages - if no page access token is provided, we don't set the "Authorization" header and simply proceed to the next line of code.

src/canonical/handler.js
slitviachenko: @dzehnder , extracted into a dedicated `src/utils/url-utils.js` helper file

PR #872, opened 2025-05-19
fix: avoid creating oppty when audit fails to find sitemap path

AndreiAlexandruParaschiv:
- https://jira.corp.adobe.com/browse/SITES-31562

When a sitemap path cannot be retrieved—either from the `robots.txt` file or by appending common paths like `sitemap.xml` or `sitemap_index.xml`, the sitemap audit fails due to the inability to fetch the data. This failure may be intentional, as some customers choose not to expose their sitemaps publicly. Instead of displaying this as an error in the UI, we should handle audit failures gracefully


src/sitemap/handler.js
```diff
@@ -594,10 +594,16 @@ export function generateSuggestions(auditUrl, auditData, context) {
 export async function opportunityAndSuggestions(auditUrl, auditData, context) {
   const { log } = context;
 
+  // Check if audit was successful
+  if (auditData.auditResult.success === false) {
```
dzehnder, line 598: if no valid paths are extracted this is also false. Do you want to skip opportunity creation too in that case?

AndreiAlexandruParaschiv: yes, for the time being, we'll skip the creation of the oppties when audit is marked as fails (due to errors caught). we'll reiterate on them later on it if needed based on the feedback we receive

radhikagpt1208:
@AndreiAlexandruParaschiv 
As discussed [here](https://cq-dev.slack.com/archives/C05A45JBP9N/p1747708289383679?thread_ts=1746723535.354139&cid=C05A45JBP9N), could you also add [sitemap oppty related errors](https://github.com/adobe/spacecat-audit-worker/blob/main/src/sitemap/handler.js#L37-L44) or hints like these as part of Opportunity object itself so that UI can show theses actionable somewhere on the UI.

PR #882, opened 2025-05-20
feat: [Preflight] Split identify and suggest flows

slitviachenko:

## Related Issues
https://jira.corp.adobe.com/browse/SITES-31750

- Add `step: AUDIT_STEP_IDENTIFY | AUDIT_STEP_SUGGEST`, which controls the Preflight execution flow.
- Switch `urls` to an array of strings, based on the recent payload structure discussions.


src/metatags/metatags-auto-suggest.js
```diff
@@ -41,16 +41,19 @@ async function getPresignedUrl(s3Client, log, scrapedData) {
   }
 }
 
-export default async function metatagsAutoSuggest(allTags, context, site) {
+export default async function metatagsAutoSuggest(allTags, context, site, options = {
+  forceAutoSuggest: false,
```
dzehnder, line 45: I think default should be `true` as this is also used by ASO for other purposes

src/metatags/metatags-auto-suggest.js
```diff
@@ -41,16 +41,19 @@ async function getPresignedUrl(s3Client, log, scrapedData) {
...
+}) {
   const { s3Client, dataAccess, log } = context;
   const {
     detectedTags,
     extractedTags,
     healthyTags,
   } = allTags;
+  const { forceAutoSuggest = false } = options;
```
dzehnder, line 53: double default assignment

src/metatags/metatags-auto-suggest.js
slitviachenko: @dzehnder , ^^^ The default ASO mechanism for running or skipping auto-suggest is based on the Site’s configuration [settings](https://github.com/adobe/spacecat-audit-worker/pull/882/files#diff-4aa2786ec8ba548c347d779c5155a0486cb8441ba074fd36c290ff853325059dL53
):
```
if (!configuration.isHandlerEnabledForSite('meta-tags-auto-suggest', site)) {
    log.info('Metatags auto-suggest is disabled for site');
    return detectedTags;
}
```
For Preflight, we need a way to run checks regardless of those settings. Providing override options allows us to control this behavior. That’s why `forceAutoSuggest` is optional and set to `false` by default.

src/metatags/metatags-auto-suggest.js
slitviachenko: UPD: Updated tests to make this change more explicit - https://github.com/adobe/spacecat-audit-worker/pull/882/commits/de0fd0aaae6d36be85b67edcf893b6aa81c0c290

dzehnder (approved):

PR #890, opened 2025-05-22
fix: preflight response according to API spec

dzehnder:
update preflight audit response according to the wiki


slitviachenko:
@dzehnder , The entire `preflightAudit` flow should be wrapped in a `try/catch` block to ensure the job's status is set to `FAILED` with appropriate error messages in case of failure. Otherwise, it will stay in the `IN_PROGRESS` state indefinitely.
Per the agreement:
```
  "status": "FAILED",
  "error": {
    "code": "string",
    "message": "string",
    "details": {}
  },
```

slitviachenko:
@dzehnder , In addition to the _global_ `try/catch`, I think it would be beneficial to wrap each _individual_ check as well. This would allow the other checks to complete even if one of them encounters a critical issue. However, we need to discuss how to reflect this in the response, as up to now we've only considered fully successful check completions:
```
  "result": [
    {
      "pageUrl": "https://example.com/url-1.html",
      "step": "IDENTIFY" | "SUGGEST",
      "audits": [
        ???
      ]
    },
    ...
  ],
```

src/preflight/handler.js
```diff
@@ -19,11 +19,12 @@ import { metatagsAutoDetect } from '../metatags/handler.js';
 import { getObjectKeysUsingPrefix, getObjectFromKey } from '../utils/s3-utils.js';
 import metatagsAutoSuggest from '../metatags/metatags-auto-suggest.js';
 import { runInternalLinkChecks } from './internal-links.js';
+import { validateCanonicalFormat, validateCanonicalRecursively, validateCanonicalTag } from '../canonical/handler.js';
 
 const { AUDIT_STEP_DESTINATIONS } = Audit;
-
 export const AUDIT_STEP_IDENTIFY = 'identify';
 export const AUDIT_STEP_SUGGEST = 'suggest';
+const AUDIT_NAMES = ['canonical', 'links', 'metatags', 'body-size', 'lorem-ipsum', 'h1-count'];
```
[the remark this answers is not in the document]
slitviachenko: Also to extract audit names in separate constants, like:
```
const AUDIT_CANONICAL = 'canonical';
const AUDIT_LINKS = 'links';
const AUDIT_METATAGS = 'metatags';
const AUDIT_BODY_SIZE = 'body-size';
const AUDIT_LOREM_IPSUM = 'lorem-ipsum';
const AUDIT_H1_COUNT = 'h1-count';
const AUDITS = {
  seo: [
    AUDIT_CANONICAL,
    AUDIT_LINKS,
    AUDIT_METATAGS,
    AUDIT_BODY_SIZE,
    AUDIT_LOREM_IPSUM,
    AUDIT_H1_COUNT,
  ],
};
```
And reference those constants wherever needed in the code.

src/preflight/handler.js
```diff
@@ -66,169 +67,141 @@ export const preflightAudit = async (context) => {
...
+    pageUrl: url,
+    step: normalizedStep,
+    audits: AUDIT_NAMES.map((name) => ({ name, type: 'seo', opportunities: [] })),
   }));
+  const resultMap = new Map(result.map((r) => [r.pageUrl, r]));
+
+  // Canonical checks
+  const canonicalResults = await Promise.all(
+    urls.map(async (url) => {
+      const { canonicalUrl, checks: tagChecks } = await validateCanonicalTag(url, log, authHeader);
```
slitviachenko, line 91: It would be really beneficial to annotate the returned values consistently throughout the code.

slitviachenko (changes requested):
@dzehnder , I've added a few comments and suggestions. When you have a moment, please take a look.

src/preflight/handler.js
slitviachenko: @dzehnder , ^^^ I meant adding code annotations to the returned responses - not unifying audit responses at this stage

PR #895, opened 2025-05-23
feat: add accessibility audit

alinalex:
### Overview
This PR introduces a stepped accessibility audit worker in SpaceCat that creates accessibility reports opportunities.

### What's Added

**1. Core Audit Handler (handler.js)**
Two-step audit workflow:
- Content Scraping Step: Initiates accessibility data collection from target URLs
- Processing Step: Aggregates scraped data and generates comprehensive accessibility reports

**2. Data Processing Logic (utils/data-processing.js)**
- Accessibility data aggregation from multiple JSON audit files stored in S3
- Multi-URL audit processing with parallel file processing for efficiency
- Automated cleanup of processed files to maintain storage efficiency
- Report opportunity generation with multiple report types
- URL discovery from existing audit files

**3. Markdown Report Generation (utils/generateMdReports.js)**
- Base Report: General accessibility overview with WCAG compliance scores
- In-Depth Report: Detailed analysis of all accessibility issues
- Enhanced Report: Focused analysis of top 10 most-visited pages
- Fixed vs. New Report: Week-over-week progress tracking

**4. Report Opportunity Definitions (utils/reportOppty.js)**
- Standardized opportunity templates for each report type
- Consistent metadata including runbooks, tags, and descriptions

**5. Constants Needed For The Reports (utils/constants.js)**
- WCAG Success Criteria mappings with direct links to official documentation
- Accessibility issue impact descriptions and user impact explanations
- Standardized solutions and suggestions for common accessibility problems
- Complete reference data for generating actionable reports


src/accessibility/handler.js
```diff
@@ -0,0 +1,125 @@
...
+    status: totalIssues > 0 ? 'OPPORTUNITIES_FOUND' : 'NO_OPPORTUNITIES',
+    opportunitiesFound: totalIssues,
+    urlsProcessed,
+    summary: `Found ${totalIssues} accessibility issues across ${urlsProcessed} URLs`,
+    fullReportUrl: outputKey, // Reference to the full report in S3
+  };
+}
+
+export default new AuditBuilder()
+  .withUrlResolver(wwwUrlResolver) // Keeps the existing URL resolver
```
iuliag, line 120: The `wwwUrlResolver` calls RUM, which I don't think we need for a11y atm.
What is the intention here, in what form do you need the URL to audit?

src/accessibility/handler.js
```diff
@@ -0,0 +1,125 @@
...
+const AUDIT_TYPE_ACCESSIBILITY = Audit.AUDIT_TYPES.ACCESSIBILITY; // Defined audit type
+
+// First step: sends a message to the content scraper to generate accessibility audits
+async function scrapeAccessibilityData(context) {
+  const {
+    site, log, finalUrl, env, s3Client,
+  } = context;
+  const siteId = site.getId();
+  const bucketName = env.S3_SCRAPER_BUCKET_NAME;
+  log.info(`[A11yAudit] Step 1: Preparing content scrape for accessibility audit for ${site.getBaseURL()}`);
```
iuliag, line 28: Could you include the siteId in the logs as well?

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+      const errorMessage = `[A11yAudit] No final result files found for ${siteId}`;
+      log.error(errorMessage);
+      throw new Error(errorMessage);
+    }
+  } catch (error) {
+    log.error(`[A11yAudit] Error getting final result files for ${siteId}: ${error.message}`);
+    throw error;
+  }
+
+  const latestFinalResultFileKey = finalResultFiles[finalResultFiles.length - 1];
```
iuliag, line 388: There's no guarantee that the files are returned in a specific order. So the last one might not be the latest.

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+  }
+
+  delete latestFinalResultFile.overall;
+  const urlsToScrape = [];
+  for (const [key, value] of Object.entries(latestFinalResultFile)) {
+    if (key.includes('https://')) {
+      urlsToScrape.push({
+        url: key,
+        urlId: key.replace('https://', ''),
+        traffic: value.traffic,
```
iuliag, line 410: Where is this traffic value coming from?

src/accessibility/handler.js
```diff
@@ -0,0 +1,125 @@
...
+
+  const urlsToScrape = await getUrlsForAudit(s3Client, bucketName, siteId, log);
+
+  // The first step MUST return auditResult and fullAuditRef.
+  // fullAuditRef could point to where the raw scraped data will be stored (e.g., S3 path).
+  return {
+    auditResult: { status: 'SCRAPING_REQUESTED', message: 'Content scraping for accessibility audit initiated.' },
```
iuliag, line 35: Since the audit result is persisted in DynamoDB, we could include some more information in the JSON, like the URLs that are being audited?

src/accessibility/utils/reportOppty.js
```diff
@@ -0,0 +1,80 @@
...
+      'a11y',
+    ],
+    status: 'IGNORED',
+  };
+}
+
+export function createReportOpportunitySuggestionInstance(suggestionValue) {
+  return [
+    {
+      type: 'CONTENT_UPDATE',
```
iuliag, line 72: Since most of them are code-related, I'd go with `CODE_CHANGE` as a default.

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+
+      if (!data) {
+        log.warn(`Failed to get data from ${key}, skipping`);
+        return null;
+      }
+
+      return { key, data };
+    });
+
+    const results = await Promise.all(processFilePromises);
```
iuliag, line 205: What if one of the promises fails? Should it fail the whole batch, which is the case now?

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+              level: value.level,
+              understandingUrl: value.understandingUrl,
+              successCriteriaNumber: value.successCriteriaNumber,
+            };
+          } else {
+            aggregatedData.overall.violations.critical.items[key].count += value.count;
+          }
+        });
+      }
+      if (violations.serious && violations.serious.items && violations.serious.count) {
```
iuliag, line 237: This looks very similar to the above. The common part could be extracted (and reused for minor, moderate in the future).

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+      runbook: opportunityInstance.runbook,
+      type: opportunityInstance.type,
+      origin: opportunityInstance.origin,
+      title: opportunityInstance.title,
+      description: opportunityInstance.description,
+      tags: opportunityInstance.tags,
+    };
+    const opportunity = await Opportunity.create(opportunityData);
+    return {
+      status: true,
```
iuliag, line 332: Where is this status useful? Since in the invoker you anyways throw an error for a success false.
From my pov, this just makes the code harder to understand with all the statuses and successes of different kinds.

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+  try {
+    const suggestion = await opportunity.addSuggestions(suggestions);
+    return {
+      status: true,
+      suggestion,
+    };
+  } catch (e) {
+    log.error(`Failed to create new suggestion for siteId ${auditData.siteId} and auditId ${auditData.auditId}: ${e.message}`);
+    return {
+      success: false,
```
iuliag, line 361: Same comment as above. + the properties look different: once status, once success.

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+    }
+  } catch (error) {
+    log.error('Failed to create base report opportunity suggestion', error.message);
+    throw new Error(error.message);
+  }
+}
+
+/**
+ * Generates report opportunities for a given site
+ * @param {string} siteId - the site ID to generate report opportunities for
```
iuliag, line 625: Parameter is now site.

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+  const { finalResultFiles } = aggregationResult;
+  const { current, lastWeek } = finalResultFiles;
+
+  // data needed for all reports oppties
+  const week = getWeekNumber(new Date());
+  const year = new Date().getFullYear();
+  // eslint-disable-next-line max-len
+  const latestAudit = await site.getLatestAuditByAuditType('accessibility');
+  const auditData = JSON.parse(JSON.stringify(latestAudit));
+  const envAsoDomain = isProd ? 'experience' : 'experience-stage';
```
iuliag, line 641: This logic is so specific to the setup and naming we currently have for our UI and so deep in the code that when we'll have an edge case, we won't remember where to adjust it.
You could have `linkBuilder` functions for the different links to add in the report, that are initialized with the necessary information (domain, variable path) in the audit-worker and passed down where they should be used.

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+    log.error('Failed to create fixed vs new report opportunity suggestion', error.message);
+    throw new Error(error.message);
+  }
+
+  // 3.4 update status to ignored
+  await fixedVsNewOpportunity.setStatus('IGNORED');
+  await fixedVsNewOpportunity.save();
+
+  // 3.5 construct url for the report
+  return `https://${envAsoDomain}.adobe.com/?organizationId=${orgId}#/@aem-sites-engineering/sites-optimizer/sites/${siteId}/opportunities/${fixedVsNewOpportunity.getId()}`;
```
iuliag, line 575: This link has 2 problems:
1. it contains a static IMS Org "@aem-sites-engineering" that isn't the one of the customer, so it's not a valid link for them (would trigger redirecting and not sure they would end up to the correct page after it).
2. it exposes how we access ASO UI internally for different customers, i.e. with organizationId in the query params.
So the approach needs to be revisited.

src/accessibility/utils/generate-md-reports.js
```diff
@@ -0,0 +1,824 @@
...
+// =============================================
+// Utility Functions
+// =============================================
+
+/**
+ * Format traffic numbers to use K for thousands and M for millions
+ * @param {number} traffic - Traffic number
+ * @returns {string} Formatted traffic value
+ */
+function formatTraffic(traffic) {
```
iuliag, line 30: You could use Intl.NumberFormat with compact - short options.

src/accessibility/utils/generate-md-reports.js
```diff
@@ -0,0 +1,824 @@
...
+  escapedText = escapedText.replace(/<([^>]+)>/g, '`<$1>`');
+  return escapedText.replace(/___BACKTICK___/g, () => `\`${backtickContent.shift()}\``);
+}
+
+/**
+ * Get the ISO week number for a date
+ * @param {Date} date - Date to get week number for
+ * @returns {number} Week number
+ */
+function getWeekNumber(date) {
```
iuliag, line 57: I'd check if Intl.DateTimeFormat has some option to extract the week number. This custom logic is not something we'd want to maintain.

src/accessibility/utils/generate-md-reports.js
```diff
@@ -0,0 +1,824 @@
...
+
+  return diffData;
+}
+
+// =============================================
+// Section Generation Functions
+// =============================================
+
+// WCAG Compliance Sections
+function generateRoadToWCAGSection(wcagData) {
```
iuliag, line 252: I propose we separate the text itself from the data filled into it in all the report generation. As right now, it's hard to:
- understand how the sections look like and if the markdown is correct high-level
- follow the values formatting. 
The text, including some placeholders, can live separately (ideally in files, can be in constants.js as well). Then in these methods, we replace the placeholders with the variable values.
For the table rows, as these are variable in number, you could have a text template for the row itself and replace the placeholders one row at a time.
Alternatively, a proper templating engine could be used.

src/accessibility/utils/data-processing.js
iuliag: Also context already contains log, so no need to be pass it explicitly. Same for all methods that have both context and log as params.

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+    const errorMessage = `[A11yAudit] No URLs found for ${siteId}`;
+    log.error(errorMessage);
+    throw new Error(errorMessage);
+  }
+
+  return urlsToScrape;
+}
+
+// eslint-disable-next-line max-len
+async function generateIndepthReportOpportunity(siteId, log, current, orgId, envAsoDomain, auditData, context, week, year) {
```
iuliag, line 425: Since the steps listed for all these 3 methods (generateIndepthReportOpportunity, generateEnhancedReportOpportunity, generateFixedNewReportOpportunity) are the same, I'd suggest having a single method, to which you can pass down the custom functions to generate markdown and create opportunity, that the common method should invoke.

src/accessibility/handler.js
alinalex: I added this part to preserve the pattern from other handlers. You're right, I don't need RUM atm, will do in future iterations.

src/accessibility/utils/data-processing.js
alinalex: From the JSON  file that I retrieve from S3. The whole premise to get the urls for the content-scraper to consume is to have a pre existing JSON file in S3 for a given siteId. This is a limitation due to time constraints to get a first version merged.

src/accessibility/utils/data-processing.js
alinalex: Good point! The approach will be to have an url without both orgId and IMS Org. The UI will redirect the user to the latest IMS org accessed, the caveat being that for customers with more IMS Orgs this approach could lead to URL errors.

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+    },
+  };
+
+  try {
+    // Prefix for accessibility data for this site
+    const prefix = `accessibility/${siteId}/`;
+    const delimiter = '/';
+    log.info(`Fetching accessibility data for site ${siteId} from bucket ${bucketName}`);
+
+    // Get all subfolders for this site, best case scenario should be 1
```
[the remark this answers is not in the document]
alinalex: Created a separate function with all the documentation in comments

src/accessibility/utils/data-processing.js
alinalex: Good point! I changed the logic to have 1 retry in case it fails and use Promise.allSettled instead of Promise.all, so it won't fail the whole batch in case one of the promises fails.

src/accessibility/utils/data-processing.js
```diff
@@ -0,0 +1,685 @@
...
+  if (urlsToScrape.length === 0) {
+    const errorMessage = `[A11yAudit] No URLs found for ${siteId}`;
+    log.error(errorMessage);
+    throw new Error(errorMessage);
+  }
+
+  return urlsToScrape;
+}
+
+// eslint-disable-next-line max-len
```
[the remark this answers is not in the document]
alinalex: Fixed the linting

src/accessibility/utils/data-processing.js
alinalex: You're right, the status is not very useful. Dropped it and use throw for error handling

src/accessibility/utils/generate-md-reports.js
```diff
@@ -731,20 +712,21 @@ function generateNewIssuesSection(diffData) {
  * @param {Object} lastWeekFile - Last week's data
  * @returns {string} Base report markdown
  */
-function generateBaseReportMarkdown(currentFile, lastWeekFile, relatedReportsUrls) {
+function generateBaseReportMarkdown(mdData) {
```
iuliag, line 715: nitpicking:
Docs need to be updated for all these functions. It'd be good to include the 3 main properties coming in via the single object parameter.
Input param can be called `data` instead of `mdData` and lastWeek could be `previous`.

src/accessibility/utils/data-processing.js
```diff
@@ -421,258 +529,174 @@ export async function getUrlsForAudit(s3Client, bucketName, siteId, log) {
...
-  // 1.1 generate the markdown report for in-depth overview
-  const inDepthOverviewMarkdown = generateInDepthReportMarkdown(current);
+export function linkBuilder(linkData, opptyId) {
+  const { envAsoDomain, siteId } = linkData;
+  return `https://${envAsoDomain}.adobe.com/#/sites-optimizer/sites/${siteId}/opportunities/${opptyId}`;
+}
 
-  if (!inDepthOverviewMarkdown) {
-    throw new Error('Failed to generate in-depth overview markdown');
+export async function generateReportOpportunity(
```
iuliag, line 537: Having docs for the parameters would be helpful.

src/accessibility/utils/data-processing.js
```diff
@@ -421,258 +529,174 @@ export async function getUrlsForAudit(s3Client, bucketName, siteId, log) {
...
+  const { week, year } = opptyData;
+  const { log } = context;
+
+  // 1.1 generate the markdown report
+  const reportMarkdown = genMdFn(mdData);
+
+  if (!reportMarkdown) {
+    // If the markdown is empty, we don't want to create an opportunity
+    // and we don't want to throw an error
+    return '';
```
iuliag, line 560: `undefined` or `null` would be closer to the meaning here. Even if an empty string also evaluates to false, so it will work.

iuliag (commented):
LGTM now, thank you for the refactoring.
Pending tests and removal of coverage exclusion for approval.

test/audits/accessibility/data-processing.test.js
```diff
@@ -3631,4 +3176,1656 @@ describe('data-processing utility functions', () => {
...
+              save: sandbox.stub(),
+              getId: sandbox.stub().returns(`success-opp-${callCount}`),
+              addSuggestions: sandbox.stub().resolves({ id: `success-sugg-${callCount}` }),
+            });
+          }
+          // Fourth call (base report) fails
+          return Promise.reject(new Error('Base report creation failed'));
+        });
+
+        // Act & Assert
```
iuliag, line 4639: For next time:
This try-catch can be alternatively written in a cleaner way as:
```
await expect(generateReportOpportunitiesMocked(...)).to.be.rejectedWith(...);
expect(mockLogForAudit.error).to.have.been.calledWith(...);
```

iuliag (approved):
🚀

PR #905, opened 2025-05-27
fix: filter out valid broken internal links code refactor for accuracy

nitinja:
Refactored code to check broken internal links for validity and remove any valid links before processing further. 
Also added structure to the code. 

## Related Issues
https://jira.corp.adobe.com/browse/SITES-27731

Thanks for contributing!


AndreiAlexandruParaschiv (approved):
LGTM!

PR #915, opened 2025-05-29
fix: preflight internal links identify

dzehnder:
fix preflight internal links identify with wrong request headers


src/preflight/internal-links.js
```diff
@@ -14,12 +14,12 @@ import { JSDOM } from 'jsdom';
...
- * @param scrapedObjects - Array of objects containing the URL and scraped data
- * @param pageAuthToken - Authentication token for the page
-  * @param context - Context object containing the logger
+ * @param {Array<Object>} scrapedObjects - Array of objects containing the URL and scraped data
+ * @param {Object} context - Context object containing the logger
+ * @param {RequestOptions} options - Options for to pass to the fetch request
  * @returns {Promise<Array>} - Array of objects containing the page URL and internal link status
  */
-export async function runInternalLinkChecks(scrapedObjects, pageAuthToken, context) {
+export async function runInternalLinkChecks(scrapedObjects, context, options = {}) {
```
slitviachenko, line 22:
```suggestion
export async function runInternalLinkChecks(scrapedObjects, context, options = {
  pageAuthToken: null,
}) {
```
I would prefer to be more specific here: state that `options` can optionally include `pageAuthToken` and set the Authorization header if the token is provided.

src/preflight/internal-links.js
```diff
@@ -48,9 +48,7 @@ export async function runInternalLinkChecks(scrapedObjects, pageAuthToken, conte
           try {
             const res = await fetch(href, {
               method: 'HEAD',
-              headers: {
-                Authorization: pageAuthToken,
-              },
+              ...options,
```
slitviachenko, line 51: See my above comment - IMO, we should explicitly set the Authorization header if the token is provided via `options`.

src/preflight/handler.js
```diff
@@ -152,7 +152,7 @@ export const preflightAudit = async (context) => {
     );
 
     // Internal link checks
-    const { auditResult } = await runInternalLinkChecks(scrapedObjects, pageAuthToken, context);
+    const { auditResult } = await runInternalLinkChecks(scrapedObjects, context, authHeader);
```
slitviachenko, line 155: @dzehnder , Let's let the concrete check decide how to apply the token. WDYT?

slitviachenko, line 155:
```suggestion
  const { auditResult } = await runInternalLinkChecks(scrapedObjects, context, {
    pageAuthToken,
  });
```

src/preflight/internal-links.js
```diff
@@ -48,7 +51,12 @@ export async function runInternalLinkChecks(scrapedObjects, context, options = {
           try {
             const res = await fetch(href, {
               method: 'HEAD',
-              ...options,
+              redirect: 'manual',
```
slitviachenko, line 54: @dzehnder , Just curious - why is this needed? It effectively prevents following any redirects.

slitviachenko (approved):

PR #936, opened 2025-06-06
feat: introduce auto-suggest for preflight internal links

dzehnder:
Introducing auto-suggest for broken internal links in the preflight audit. The suggestions are based on the pathnames of the live page, with origin swapped to the preview page


package.json
```diff
@@ -20,7 +20,7 @@
     "build": "hedy -v --test-bundle",
     "deploy": "hedy -v --deploy --aws-deploy-bucket=spacecat-prod-deploy --pkgVersion=latest",
     "deploy-stage": "hedy -v --deploy --aws-deploy-bucket=spacecat-stage-deploy --pkgVersion=latest",
-    "deploy-dev": "hedy -v --deploy --pkgVersion=ci$CI_BUILD_NUM -l latest --aws-deploy-bucket=spacecat-dev-deploy --cleanup-ci=24h",
+    "deploy-dev": "hedy -v --deploy --pkgVersion=ci$CI_BUILD_NUM -l dzehnder --aws-deploy-bucket=spacecat-dev-deploy --cleanup-ci=24h",
```
alinarublea, line 23: Don t forget to change it back before merging

src/preflight/handler.js
```diff
@@ -185,22 +186,51 @@ export const preflightAudit = async (context) => {
...
+          status: link.status,
+        }));
+        log.debug(`[preflight-audit] Found ${JSON.stringify(brokenLinks)} broken internal links`);
+        const brokenInternalLinks = await
+        generateSuggestionData(baseURL, brokenLinks, context, site);
+        log.debug(`[preflight-audit] Generated suggestions for broken internal links: ${JSON.stringify(brokenInternalLinks)}`);
+        brokenInternalLinks.forEach(({
+          urlTo, href, status, urlsSuggested, aiRationale,
+        }) => {
+          const audit = resultMap.get(href).audits.find((a) => a.name === AUDIT_LINKS);
```
solaris007, line 207: assumes resultMap.get(href) always exists and .audits.find(...) never returns undefined. this can easily throw.

src/preflight/handler.js
```diff
@@ -185,22 +186,51 @@ export const preflightAudit = async (context) => {
...
+              seoImpact: 'High',
+              seoRecommendation: 'Fix or remove broken links to improve user experience and SEO',
+              urlsSuggested: aiUrls,
+              aiRationale,
+            },
+          });
+        });
+      } else {
+        auditResult.brokenInternalLinks.forEach(({ urlTo, href, status }) => {
+          const audit = resultMap.get(href).audits.find((a) => a.name === AUDIT_LINKS);
```
solaris007, line 223: assumes resultMap.get(href) always exists and .audits.find(...) never returns undefined. this can easily throw.

src/preflight/handler.js
solaris007: you're doing a map lookup followed by a .find() on each iteration. if this is within a large audit set, it becomes a repeated O(n) scan.

fix:

if AUDIT_LINKS audit is known to be unique per href, pre-index or flatten in a Map.

at minimum, cache resultMap.get(href) in a local variable.

solaris007 (approved):

PR #948, opened 2025-06-11
fix: prevent re-scraping of URLs that have already failed an audit

alinalex:
This pull request introduces a mechanism into the accessibility audit worker to prevent the re-scraping of URLs that have already failed an audit within the same day. This is a useful measure in the context of content-scraper audits that result in timeouts due to the fixed duration of the lambda (15 min max).

Key Changes:

**accessibility/handler.js:**
- The scrapeAccessibilityData function has been updated to check for existing failed audits before dispatching a scrape request.
- It now fetches URLs from previously failed audits and filters them out from the list of URLs to be scraped.

**accessibility/utils/scrape-utils.js:**
- A new utility file has been created to house helper functions for the scraping logic.
- getExistingUrlsFromFailedAudits: This new function retrieves the keys of failed audit results from S3 for the current day and reconstructs the original URLs.
- getRemainingUrls: This function efficiently filters the list of URLs to be scraped against the list of existing failed URLs.
- normalizeUrl and reconstructUrlFromS3Key: Helper functions to handle URL manipulation and reconstruction from S3 keys.

By implementing this check, we ensure that the accessibility audit only processes new pages or pages that have not recently failed, making the audit process more efficient and resilient.

### DIFF
--- a/src/accessibility/utils/scrape-utils.js
+++ b/src/accessibility/utils/scrape-utils.js
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2025 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+import { getObjectKeysFromSubfolders } from './data-processing.js';
+
+export async function getExistingUrlsFromFailedAudits(s3Client, bucketName, siteId, log) {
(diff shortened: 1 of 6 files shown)


src/accessibility/utils/scrape-utils.js
```diff
@@ -0,0 +1,114 @@
...
+  const fileName = key.split('/').pop();
+  if (!fileName) {
+    return '';
+  }
+
+  const urlPath = fileName.replace('.json', '');
+  const pieces = urlPath.split('_');
+  const dotIndex = pieces.includes('www') ? 2 : 1;
+
+  const almostFullUrl = pieces.reduce((acc, piece, index) => {
```
AndreiAlexandruParaschiv, line 51: wouldn't this cause issues for sites like: `https://petdrugsonline.co.uk` or `https://footlocker.com.bh`? these won't be reconstructed as before,  I think it will end up as `https://petdrugsonline.co/uk` or  `https://footlocker.com/bh` instead.

alinalex: You're right. I think this approach will still have edge cases here and there. I will use another approach to get the urls that already have been scraped by reading the json files from S3 and getting the url from there.

PR #982, opened 2025-06-20
feat(SITES-32637): add HTML comment parsing for broken internal links…

abhishekgarg18:
… detection

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/internal-links/handler.js
+++ b/src/internal-links/handler.js
@@ -11,19 +11,26 @@
  */
 
 import RUMAPIClient from '@adobe/spacecat-shared-rum-api-client';
-import { Audit, Opportunity as Oppty, Suggestion as SuggestionDataAccess } from '@adobe/spacecat-shared-data-access';
+import {
+  Audit,
+  Site,
+  Opportunity as Oppty,
+  Suggestion as SuggestionDataAccess,
+} from '@adobe/spacecat-shared-data-access';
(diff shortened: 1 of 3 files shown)


AndreiAlexandruParaschiv:
> Thanks Abhishek, Thanks for the PR - it looks reasonable. I have some concerns: The problem with scraped data by spacecat content scraper I think is:
> 
> * Last tim I checked, we couldnt scrape data for entire site, we have to provide URLs for scraping (I could be wrong here).
> * We get these URLs from ahrefs ("top urls") and only those are scraped. So we do not have entire site to get link results from. So this will not be effective strategy.
> * we have already prototyped screaming frog scraping tool to identify broken internal links, which does exactly same thing but more comprehensively on entire site (I believe it can do the comments parsing like you need). This prototype was very effective.
> 
> Because of these reasons, lets discuss this approach with team. Your thought @AndreiAlexandruParaschiv?

The content-scraper is designed to crawl a portion of the site— up to 300 top pages based on traffic. 
- Could we test this feature with a few selected customers (starting with paying ones like Qualcomm, Breville, Crucial, T-Mobile) to see if there's a significant number of broken internal links on their top pages?  
- We should aim to avoid adding unnecessary complexity to the audits, especially if most of the broken links are found on low-traffic or non-essential pages.

PR #984, opened 2025-06-20
feat: auto-suggest for CSP audit

bisswanger:
Changes:
 - CSP audit shall only be triggered if the lighthouse run was successful
   (else it may "remove" / "resolve" existing suggestions as there is no lighthouse report info available)
 - if "nonces or hashes" is returned from lighthouse, remove the reference to "hashes"
 - backwards-compatible warnings are included in the result
 - CSP auto-suggest
   - for now for the "CSP missing" case
   - scans /head.html and /404.html and checks for required changes

### DIFF
--- a/test/audits/csp.test.js
+++ b/test/audits/csp.test.js
@@ -17,6 +17,7 @@ import chaiAsPromised from 'chai-as-promised';
 import sinon from 'sinon';
 import sinonChai from 'sinon-chai';
 import nock from 'nock';
+import { load as cheerioLoad } from 'cheerio';
 
 import { Audit } from '@adobe/spacecat-shared-data-access';
 import { isIsoDate } from '@adobe/spacecat-shared-utils';
@@ -46,6 +47,11 @@ function assertAuditData(auditData) {
     'best-practices': 0.5,
     seo: 0.5,
   });
+}
+
+function normalizeTag(html) {
+  const $ = cheerioLoad(html);
+  return $.html($('script')[0]);
 }
 
 describe('CSP Post-processor', () => {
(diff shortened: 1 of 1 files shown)


andreituicu (approved):

dulvac (approved):

src/csp/csp-auto-suggest.js
```diff
@@ -0,0 +1,125 @@
...
+const AUDIT_TYPE = Audit.AUDIT_TYPES.SECURITY_CSP;
+const scriptRegExp = /<script.*?(?:\/>|<\/script>)/gs;
+const hasNonceRegExp = /<script[^>]*? nonce=/s;
+
+async function determineSuggestionsForPage(url, context, site) {
+  const { log } = context;
+
+  log.debug(`[${AUDIT_TYPE}] [Site: ${site.getId()}]: Fetching page '${url}' for CSP auto-suggest`);
+
+  const response = await fetch(url);
```
herzog31, line 23: You should probably use https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-utils/src/tracing-fetch.js here to automatically add the spacecat user agent to ensure your request can bypass bot detection.

src/csp/csp-auto-suggest.js
```diff
@@ -0,0 +1,125 @@
...
+  const responseBody = await response.text();
+
+  if (!responseBody) {
+    log.warn(`[${AUDIT_TYPE}] [Site: ${site.getId()}] [Url: ${url}]: no script page content found`);
+    return null;
+  }
+
+  let suggestedBody = responseBody;
+  const findings = [];
+  const scriptTags = responseBody.matchAll(scriptRegExp);
```
herzog31, line 37: Instead of regex, you could use `cheerio` for getting the tags and attributes. The library is already installed in the project.

herzog31 (approved):

PR #1007, opened 2025-06-25
fix: remove page auth token from external link checks and fix internal link checks broken bug.

tathagat2241:
This PR includes:
1. removes page auth token for external link checks as external links were incorrectly receiving internal authentication tokens, causing false positives when external domains returned 401/403 responses.
2. Refactored links-check to remove code duplication.
3. Fixed Internal links checks broken bug.
i. Now, checking the links that are only in main body and not in header and footer. Before when we included links that were in header and footer the total links found count was around 150 which was overloading the stream and HEAD request being too restrictive it was failing with error unexpected end of file .
ii. Added a fallback mechanism to GET request when HEAD request fails.

## Related Issues

https://jira.corp.adobe.com/browse/SITES-32254
https://jira.corp.adobe.com/browse/SITES-33323
Thanks for contributing!
Preflight Links-checks test

### DIFF
--- a/src/preflight/links-checks.js
+++ b/src/preflight/links-checks.js
@@ -14,6 +14,68 @@ import { JSDOM } from 'jsdom';
 import { tracingFetch as fetch } from '@adobe/spacecat-shared-utils';
 
 /**
+ * Helper function to check if a link is broken
+ * @param {string} href - The URL to check
+ * @param {string} pageUrl - The source page URL
+ * @param {Object} context - Context object containing the logger
+ * @param {Object} options - Options for the fetch request
+ * @param {string} options.pageAuthToken - Optional authorization token for the page
+ * @param {boolean} options.isInternal - Whether this is an internal link
+ * @returns {Promise<Object|null>} - Object with link details if broken, null otherwise
+ */
+async function checkLinkStatus(href, pageUrl, context, options = {
+  pageAuthToken: null,
+}) {
+  const { log } = context;
+  const { pageAuthToken, isInternal } = options;
+
+  const fetchOptions = {
+    method: 'HEAD',
+    decode: false,
+  };
+
+  // Add Authorization header only for internal links
+  if (isInternal && pageAuthToken) {
+    fetchOptions.headers = {
+      Authorization: pageAuthToken,
+    };
+  }
+
+  try {
+    const res = await fetch(href, fetchOptions);
+
+    if (res.status >= 400) {
+      const linkType = isInternal ? 'internal' : 'external';
+      log.debug(`[preflight-audit] ${linkType} url ${href} returned with status code: %s`, res.status, res.statusText);
+      return { urlTo: href, href: pageUrl, status: res.status };
+    }
+
+    return null;
+  } catch (err) {
+    // Fallback to GET on any error
+    log.warn(`[preflight-audit] HEAD request failed (${err.message}), retrying with GET: ${href}`);
+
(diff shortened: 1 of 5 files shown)


src/preflight/links-checks.js
```diff
@@ -67,10 +67,11 @@ export async function runLinksChecks(urls, scrapedObjects, context, options = {
             try {
               const response = await fetch(href, {
```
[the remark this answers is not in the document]
tathagat2241: Yeah, I will do that, but lets first figure out the problem that we are facing.

src/preflight/links-checks.js
```diff
@@ -62,56 +126,33 @@ export async function runLinksChecks(urls, scrapedObjects, context, options = {
...
-              log.error(`[preflight-audit] Error checking internal link ${href} from ${pageUrl}:`, err.message);
-            }
+        const internalResults = await pMap(
+          Array.from(internalSet),
+          async (href) => checkLinkStatus(href, pageUrl, context, {
+            ...options,
+            isInternal: true,
           }),
+          { concurrency: 5 },
         );
```
dzehnder, line 136: I would do sequential requests to avoid overloads using a reducer function for example:
```suggestion
const internalResults = await Array.from(internalSet).reduce(
  async (accPromise, href) => {
    const acc = await accPromise;

    const result = await checkLinkStatus(
      href,
      pageUrl,
      context,
      {
        ...options,
        isInternal: true,
      },
    );

    return [...acc, result];
  },
  Promise.resolve([])
);
```

src/preflight/links-checks.js
```diff
@@ -62,56 +126,33 @@ export async function runLinksChecks(urls, scrapedObjects, context, options = {
...
-            } catch (err) {
-              log.error(`[preflight-audit] Error checking external link ${href} from ${pageUrl}:`, err.message);
-              brokenExternalLinks.push({
-                urlTo: href,
-                href: pageUrl,
-                status: 'error',
-                error: err.message,
-              });
-            }
+        const externalResults = await pMap(
```
dzehnder, line 139: same here

src/preflight/links-checks.js
```diff
@@ -12,6 +12,70 @@
...
+
+    if (res.status >= 400) {
+      const linkType = isInternal ? 'internal' : 'external';
+      log.debug(`[preflight-audit] ${linkType} url ${href} returned with status code: %s`, res.status, res.statusText);
+      return { urlTo: href, href: pageUrl, status: res.status };
+    }
+
+    return null;
+  } catch (err) {
+    // Fallback to GET on any error
```
dzehnder, line 57: I would avoid doing a second request when the first one failed

src/preflight/links-checks.js
```diff
@@ -12,6 +12,70 @@
...
+    fetchOptions.headers = {
+      Authorization: pageAuthToken,
+    };
+    console.log(`[preflight-audit] Checking internal link ${href} with auth token: ${pageAuthToken}`);
+  } else if (isInternal) {
+    console.log(`[preflight-audit] Checking internal link ${href} without auth token (token is: ${pageAuthToken})`);
+  }
+
+  try {
+    const res = await fetch(href, fetchOptions);
```
dzehnder, line 47: as a workaround to the `unexpected end of file` you can try the `decode: false` header as this might prevent unpacking a non existing body from the head request:
```
 const fetchOptions = {
    method: 'HEAD',
    decode: false,
  };
  ```

PR #1025, opened 2025-07-01
feat [alt-text]: SQS Integration with SpaceCat

jjenscodee:
We are migrating the alt-text audit worker to obtain suggestions from Mystique via SQS. 

For now, we are retaining the existing Firefall implementation in the codebase as a fallback, so we can quickly switch back if any issues arise during the transition. Once the Mystique integration is fully tested and stable, we will remove the Firefall code.

## Related Issues

Thanks for contributing!


src/image-alt-text/guidance-missing-alt-text-handler.js
```diff
@@ -0,0 +1,193 @@
...
+
+  const isGoogleConnected = await checkGoogleConnection(auditUrl, context);
+
+  if (!isGoogleConnected && opportunityData.dataSources) {
+    opportunityData.dataSources = opportunityData.dataSources
+      .filter((source) => source !== DATA_SOURCES.GSC);
+  }
+
+  try {
+    if (!altTextOppty) {
```
alinarublea, line 139: I would keep the opportunity creation as part of the handler itself.

src/image-alt-text/guidance-missing-alt-text-handler.js
```diff
@@ -0,0 +1,193 @@
...
+  //   log,
+  // });
+
+  // const opportunityData = {
+  //   ...projectedMetrics,
+  //   decorativeImagesCount:
+  //   mappedSuggestions.filter((suggestion) => suggestion.isDecorative === true).length,
+  // };
+
+  const batchProjectedMetrics = await getProjectedMetrics({
```
alinarublea, line 91: Could this be computed also in the handler or does Mystique generate something that helps computing this metrics ?

PR #1029, opened 2025-07-02
feat: save paid cookie consent opp

mimchome:
https://jira.corp.adobe.com/browse/SITES-32826

creates or updates opportunity based on data from mystique

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/paid/guidance-opportunity-mapper.js
+++ b/src/paid/guidance-opportunity-mapper.js
@@ -13,11 +13,30 @@
 import { randomUUID } from 'crypto';
 import { DATA_SOURCES } from '../common/constants.js';
 
-function sanitizeMarkup(markup) {
-  if (typeof markup === 'string' && markup.includes('\\n')) {
-    return markup.replace(/\\n/g, '\n');
+function sanitizeMarkdown(markdown) {
+  if (typeof markdown === 'string' && markdown.includes('\\n')) {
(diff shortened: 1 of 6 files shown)


src/paid/guidance-opportunity-mapper.js
```diff
@@ -0,0 +1,111 @@
...
+
+  const desktopScreenshot = `${apiBase}/sites/${siteId}/files?key=scrapes/${siteId}${suffix}/consent-banner-on/screenshot-desktop-viewport.png`;
+  const mobileScreenshot = `${apiBase}/sites/${siteId}/files?key=scrapes/${siteId}${suffix}/consent-banner-on/screenshot-iphone-6-viewport.png`;
+
+  const appendedSection = `
+### Screenshots
+
+| Mobile | Desktop |
+|--------|---------|
+| ![Mobile Screenshot](${mobileScreenshot}) | ![Desktop Screenshot](${desktopScreenshot}) |
```
ekremney, line 36: I was under the impression that we would add only the screenshot of a single device based on the mystique response?

ekremney (approved):

src/paid/guidance-opportunity-mapper.js
mimchome: we havent solved for case of suggestion on both so i think its easier to have as is and if we have two separate suggestions incase of both or one i can update this

PR #1066, opened 2025-07-24
Suppress ImageObject Error 

noruiz:
Fix for:  https://jira.corp.adobe.com/browse/SITES-33564

### DIFF
--- a/src/structured-data/handler.js
+++ b/src/structured-data/handler.js
@@ -185,6 +185,7 @@ export async function generateSuggestionsData(auditUrl, auditData, context, scra
 
 export async function opportunityAndSuggestions(auditUrl, auditData, context) {
   const { log } = context;
+  const isInternalCustomer = context.site.getConfig().isInternalCustomer();
 
   // Check if audit was successful
   if (auditData.auditResult.success === false) {
@@ -194,6 +195,9 @@ export async function opportunityAndSuggestions(auditUrl, auditData, context) {
 
   // Convert suggestions to errors
   for (const issue of auditData.auditResult.issues) {
+    if (isInternalCustomer && issue.rootType === 'ImageObject') {
+      continue;
+    }
     issue.errors = [];
     const fix = generateErrorMarkupForIssue(issue);
     const errorTitle = `${issue.rootType}: ${issue.issueMessage}`;


src/structured-data/handler.js
```diff
@@ -185,6 +185,7 @@ export async function generateSuggestionsData(auditUrl, auditData, context, scra
 
 export async function opportunityAndSuggestions(auditUrl, auditData, context) {
   const { log } = context;
+  const isInternalCustomer = context.site.getConfig().isInternalCustomer();
```
herzog31, line 188: We don't want to limit exclusion of the error message to internal customers, but customers that use AEM.

Please use `deliveryType` from the site config instead.

Example: https://github.com/adobe/spacecat-audit-worker/blob/main/src/preflight/utils.js#L41

src/structured-data/handler.js
```diff
@@ -194,6 +195,9 @@ export async function opportunityAndSuggestions(auditUrl, auditData, context) {
 
   // Convert suggestions to errors
   for (const issue of auditData.auditResult.issues) {
+    if (isInternalCustomer && issue.rootType === 'ImageObject') {
```
herzog31, line 198: This logic should probably happen earlier in the audit. For example in `processStructuredData` so that the problematic issues don't get stored.

Your code excludes all detected issues for `ImageObject`. However, only one specific error should be excluded. Please make the condition more precise.

herzog31 (changes requested):
It's a good start, but needs some more work :) 

Please ensure your changes are covered by a unit test.

test/audits/structured-data/lib.test.js
```diff
@@ -935,4 +936,108 @@ This is an error description
...
+        .withSandbox(sandbox)
+        .withOverrides({
+          log: {
+            info: sinon.stub(),
+            warn: sinon.spy(),
+            error: sinon.stub(),
+            debug: sinon.spy(),
+          },
+          site: {
+            getDeliveryType: sinon.stub(),
```
herzog31, line 960: Your PR has test failures which are mainly caused by this mock being missing in the other tests.

```
TypeError: context.site.getDeliveryType is not a function
```

noruiz: Pushed updates, tests look good

herzog31 (approved):

PR #1077, opened 2025-07-29
fix: Suggestion error IDs are not unique in Structured Data Audit

noruiz:
Fix for https://jira.corp.adobe.com/browse/SITES-33562

### DIFF
--- a/src/structured-data/handler.js
+++ b/src/structured-data/handler.js
@@ -192,13 +192,28 @@ export async function opportunityAndSuggestions(auditUrl, auditData, context) {
     return { ...auditData };
   }
 
-  // Convert suggestions to errors
+  // Convert suggestions to errors, with duplicate removal
   for (const issue of auditData.auditResult.issues) {
     issue.errors = [];
     const fix = generateErrorMarkupForIssue(issue);
     const errorTitle = `${issue.rootType}: ${issue.issueMessage}`;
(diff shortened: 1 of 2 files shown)


src/structured-data/handler.js
```diff
@@ -192,13 +192,28 @@ export async function opportunityAndSuggestions(auditUrl, auditData, context) {
...
     const fix = generateErrorMarkupForIssue(issue);
     const errorTitle = `${issue.rootType}: ${issue.issueMessage}`;
     const errorId = errorTitle.replaceAll(/["\s]/g, '').toLowerCase();
+
+    if (!issue.errors.some((e) => e.id === errorId)) {
+      issue.errors.push({ fix, id: errorId, errorTitle });
+    }
+  }
+
+  // Convert suggestions to errors, keep duplicates and ensure unique ID using index
```
herzog31, line 207: Let's go with this approach.

src/structured-data/handler.js
```diff
@@ -193,11 +193,18 @@ export async function opportunityAndSuggestions(auditUrl, auditData, context) {
   }
 
   // Convert suggestions to errors
+  const errorIdMap = {};
```
herzog31, line 196: You're using a global map across all URLs. So if two pages have the same issue, the issue on the second page will get the number suffix. What we want is the suffix for two or more identical issues on the same page :)

Page is stored in `issue.pageUrl`.

herzog31 (approved):

PR #1082, opened 2025-07-31
fix: change-a11y-suggestion-data-structure-granularity

alinalex:


src/accessibility/guidance-utils/mystique-data-processing.js
```diff
@@ -25,44 +26,46 @@ export function processSuggestionsForMystique(suggestions) {
...
-              target_selector: htmlIssueItem.target_selector || issue.targetSelector || '',
-              issue_description: issue.description || '',
-              issue_id: htmlIssueItem.issue_id || '',
-            });
+        if (issueTypesForMystique.includes(issue.type)) {
+          const { url } = suggestionData;
+          if (!issuesByUrl[url]) {
+            issuesByUrl[url] = [];
           }
+          issuesByUrl[url].push({ ...suggestionData, suggestionId });
```
AndreiAlexandruParaschiv, line 42: this is pushing the entire `suggestionData` object for each issue that matches the `issueTypesForMystique`. So if a suggestion has 3 issues, we loop 3 times and push the same suggestion 3 times, basically duplicating the entries.

PR #1086, opened 2025-08-03
feat : add support on additional urls audit in experimentation opportunity

FentPams:

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/experimentation-opportunities/handler.js
+++ b/src/experimentation-opportunities/handler.js
@@ -88,6 +88,22 @@ function getHighOrganicLowCtrOpportunityUrls(experimentationOpportunities) {
 }
 
 /**
+ * Generates mock experimentation opportunities for URLs that don't have real RUM data
+ * @param {string[]} urlsWithoutData - Array of URLs missing real analytics
+ * @returns {Array} Array of mock experimentation opportunities
+ */
+function generateMockOpportunitiesForMissingUrls(urlsWithoutData) {
+  return urlsWithoutData.map((url) => ({
+    type: HIGH_ORGANIC_LOW_CTR_OPPTY_TYPE,
+    page: url,
+    pageViews: Math.round(1000 + Math.random() * 4000),
+    trackedPageKPIValue: 0.015 + Math.random() * 0.03,
+    trackedKPISiteAverage: 0.035,
+    organicTraffic: Math.round(800 + Math.random() * 1200),
+  }));
+}
+
+/**
  * Audit handler container for all the opportunities
  * @param {*} auditUrl
  * @param {*} context
@@ -95,7 +111,7 @@ function getHighOrganicLowCtrOpportunityUrls(experimentationOpportunities) {
  * @returns
  */
 
-export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
+export async function experimentOpportunitiesAuditRunner(auditUrl, context, staticUrls = null) {
   const { log } = context;
 
   const rumAPIClient = RUMAPIClient.createFrom(context);
@@ -107,6 +123,33 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
   const queryResults = await rumAPIClient.queryMulti(OPPTY_QUERIES, options);
   const experimentationOpportunities = Object.values(queryResults).flatMap((oppty) => oppty);
   processRageClickOpportunities(experimentationOpportunities);
+
+  // Handle static URLs
+  if (staticUrls && staticUrls.length > 0) {
+    log.info(`Processing ${staticUrls.length} static URLs for experimentation opportunities`);
+
+    // Map real RUM data for the specific static URLs, if none, generate mock data
+    const realDataOpportunities = experimentationOpportunities.filter((oppty) => staticUrls.includes(oppty.page));
+    const urlsWithRealData = realDataOpportunities.map(oppty => oppty.page);
+    const urlsWithoutData = staticUrls.filter(url => !urlsWithRealData.includes(url));
+    const mockOpportunities = generateMockOpportunitiesForMissingUrls(urlsWithoutData);
+    const finalExperimentationOpportunities = [...realDataOpportunities, ...mockOpportunities];
+
+    if (realDataOpportunities.length > 0) {
+      log.info(`Using real RUM data for ${realDataOpportunities.length} URLs: ${urlsWithRealData.join(', ')}`);
+    }
+    if (mockOpportunities.length > 0) {
+      log.info(`Using mock data for ${mockOpportunities.length} URLs: ${urlsWithoutData.join(', ')}`);
+    }
+    
+    return {
+      auditResult: {
+        experimentationOpportunities: finalExperimentationOpportunities,
+      },
+      fullAuditRef: auditUrl,
+    };
+  }
+  
   log.info(`Found ${experimentationOpportunities.length} experimentation opportunites for ${auditUrl}`);
 
   return {
--- a/src/experimentation-opportunities/handler.js
+++ b/src/experimentation-opportunities/handler.js
@@ -129,9 +129,11 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context, stat
     log.info(`Processing ${staticUrls.length} static URLs for experimentation opportunities`);
 
     // Map real RUM data for the specific static URLs, if none, generate mock data
-    const realDataOpportunities = experimentationOpportunities.filter((oppty) => staticUrls.includes(oppty.page));
-    const urlsWithRealData = realDataOpportunities.map(oppty => oppty.page);
-    const urlsWithoutData = staticUrls.filter(url => !urlsWithRealData.includes(url));
+    const realDataOpportunities = experimentationOpportunities.filter(
+      (oppty) => staticUrls.includes(oppty.page),
+    );
+    const urlsWithRealData = realDataOpportunities.map((oppty) => oppty.page);
+    const urlsWithoutData = staticUrls.filter((url) => !urlsWithRealData.includes(url));
     const mockOpportunities = generateMockOpportunitiesForMissingUrls(urlsWithoutData);
     const finalExperimentationOpportunities = [...realDataOpportunities, ...mockOpportunities];
 
@@ -141,7 +143,7 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context, stat
     if (mockOpportunities.length > 0) {
       log.info(`Using mock data for ${mockOpportunities.length} URLs: ${urlsWithoutData.join(', ')}`);
     }
-    
+
     return {
       auditResult: {
         experimentationOpportunities: finalExperimentationOpportunities,
@@ -149,7 +151,7 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context, stat
       fullAuditRef: auditUrl,
     };
   }
-  
+
   log.info(`Found ${experimentationOpportunities.length} experimentation opportunites for ${auditUrl}`);
 
   return {
@@ -162,7 +164,8 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context, stat
 
 export async function runAuditAndScrapeStep(context) {
   const { site, finalUrl } = context;
-  const result = await experimentOpportunitiesAuditRunner(finalUrl, context);
+  const staticUrls = context.auditContext?.additionalAuditData?.staticUrls;
+  const result = await experimentOpportunitiesAuditRunner(finalUrl, context, staticUrls);
 
   return {
     auditResult: result.auditResult,
--- a/test/audits/experimentation-opportunities.test.js
+++ b/test/audits/experimentation-opportunities.test.js
@@ -351,4 +351,92 @@ describe('Experimentation Opportunities Tests', () => {
     await generateOpportunityAndSuggestions(context);
     expect(context.sqs.sendMessage).to.not.have.been.called;
   });
+
+  describe('Static URLs Functionality', () => {
+    it('should use static URLs for audit when provided', async () => {
+      context.auditContext = {
+        additionalAuditData: {
+          staticUrls: ['https://abc.com/page1', 'https://abc.com/page2'],
+        },
+      };
+
+      const auditData = await runAuditAndScrapeStep(context);
+
+      expect(context.rumApiClient.queryMulti).to.have.been.calledOnce;
+      expect(auditData.auditResult.experimentationOpportunities).to.have.length(2);
+      expect(auditData.auditResult.experimentationOpportunities[0].page).to.equal('https://abc.com/page1');
+      expect(auditData.auditResult.experimentationOpportunities[1].page).to.equal('https://abc.com/page2');
+      expect(auditData.auditResult.experimentationOpportunities[0]).to.have.property('pageViews');
+      expect(auditData.auditResult.experimentationOpportunities[0]).to.have.property('trackedPageKPIValue');
+    });
+
+    it('should combine real RUM data with mock data for static URLs', async () => {
+      const mockOpportunities = {
+        'high-organic-low-ctr': [
+          {
+            page: 'https://abc.com/page1',
+            pageViews: 5000,
+            trackedPageKPIValue: 0.02,
+            trackedKPISiteAverage: 0.035,
+            organicTraffic: 1500,
+            type: 'high-organic-low-ctr',
+          },
+        ],
+      };
+
+      context.rumApiClient.queryMulti = sinon.stub().resolves(mockOpportunities);
+      context.auditContext = {
+        additionalAuditData: {
+          staticUrls: ['https://abc.com/page1', 'https://abc.com/page2'],
+        },
+      };
+
+      const auditData = await runAuditAndScrapeStep(context);
+      expect(auditData.auditResult.experimentationOpportunities).to.have.length(2);
+
+      const realOpportunity = auditData.auditResult.experimentationOpportunities.find(
+        (op) => op.page === 'https://abc.com/page1',
+      );
+      expect(realOpportunity.pageViews).to.equal(5000);
+
+      const mockOpportunity = auditData.auditResult.experimentationOpportunities.find(
+        (op) => op.page === 'https://abc.com/page2',
+      );
+      expect(mockOpportunity.pageViews).to.be.within(1000, 5000);
+    });
+
+    it('should work normally when no static URLs provided', async () => {
+      const auditData = await runAuditAndScrapeStep(context);
+      expect(context.rumApiClient.queryMulti).to.have.been.calledOnce;
+      expect(auditData.auditResult.experimentationOpportunities).to.exist;
+    });
+
+    it('should handle empty static URLs array', async () => {
+      context.auditContext = {
+        additionalAuditData: {
+          staticUrls: [],
+        },
+      };
+
+      expect(context.rumApiClient.queryMulti).to.have.been.calledOnce;
+    });
+
+    it('should generate mock data when no RUM data exists for static URLs', async () => {
+      context.rumApiClient.queryMulti = sinon.stub().resolves({});
+      context.auditContext = {
+        additionalAuditData: {
+          staticUrls: ['https://abc.com/new-page1', 'https://abc.com/new-page2'],
+        },
+      };
+
+      const auditData = await runAuditAndScrapeStep(context);
+      expect(auditData.auditResult.experimentationOpportunities).to.have.length(2);
+      auditData.auditResult.experimentationOpportunities.forEach((op) => {
+        expect(op.type).to.equal('high-organic-low-ctr');
+        expect(op.pageViews).to.be.within(1000, 5000);
+        expect(op.trackedPageKPIValue).to.be.within(0.015, 0.045);
+        expect(op.trackedKPISiteAverage).to.equal(0.035);
+      });
+    });
+  });
 });
--- a/test/audits/experimentation-opportunities.test.js
+++ b/test/audits/experimentation-opportunities.test.js
@@ -418,6 +418,7 @@ describe('Experimentation Opportunities Tests', () => {
         },
       };
 
+      await runAuditAndScrapeStep(context);
       expect(context.rumApiClient.queryMulti).to.have.been.calledOnce;
     });
 


src/experimentation-opportunities/handler.js
```diff
@@ -95,7 +111,7 @@ function getHighOrganicLowCtrOpportunityUrls(experimentationOpportunities) {
  * @returns
  */
 
-export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
+export async function experimentOpportunitiesAuditRunner(auditUrl, context, staticUrls = null) {
```
rpapani, line 114: how about custom_urls?

src/experimentation-opportunities/handler.js
```diff
@@ -87,6 +87,22 @@ function getHighOrganicLowCtrOpportunityUrls(experimentationOpportunities) {
   )?.map((oppty) => oppty.page);
 }
 
+/**
+ * Generates mock experimentation opportunities for URLs that don't have real RUM data
+ * @param {string[]} urlsWithoutData - Array of URLs missing real analytics
+ * @returns {Array} Array of mock experimentation opportunities
+ */
+function generateMockOpportunitiesForMissingUrls(urlsWithoutData) {
+  return urlsWithoutData.map((url) => ({
```
rpapani, line 96: I think we should never mock data for any audits in the product code. If the data isn't there, just return what we've

rpapani (changes requested):
please remove the mocking

src/experimentation-opportunities/handler.js
```diff
@@ -74,7 +74,7 @@ export async function generateOpportunityAndSuggestions(context) {
     },
   }));
 
-  for (const message of messages) {
+  for (const message of messages || []) {
```
jeddiec, line 77: Should we add the fallback `|| []` to `const messages = ... ?? []` directly?

src/experimentation-opportunities/handler.js
```diff
@@ -107,6 +125,36 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
   const queryResults = await rumAPIClient.queryMulti(OPPTY_QUERIES, options);
   const experimentationOpportunities = Object.values(queryResults).flatMap((oppty) => oppty);
   processRageClickOpportunities(experimentationOpportunities);
+
+  if (customUrls && customUrls.length > 0) {
+    log.info(`Processing ${customUrls.length} custom URLs for experimentation opportunities`);
+
+    const highOrganicLowCtrOpportunities = experimentationOpportunities.filter(
+      (oppty) => oppty.type === HIGH_ORGANIC_LOW_CTR_OPPTY_TYPE,
+    );
```
jeddiec, line 134: We can use `reduce` here:

```js
const [opptiesWithRumData, opptiesWithoutRumData] = highOrganicLowCtrOpportunities.reduce(
  (accumulator, currentValue) => {
    if (customUrls.includes(currentValue.page)) {
      accumulator[0].push(currentValue);
    } else {
      accumulator[1].push(currentValue);
    }
    return accumulator;
  },
  [[], []]
);
```

src/experimentation-opportunities/handler.js
```diff
@@ -87,15 +87,33 @@ function getHighOrganicLowCtrOpportunityUrls(experimentationOpportunities) {
   )?.map((oppty) => oppty.page);
 }
 
+/**
+ * Creates empty experimentation opportunities for URLs that don't have RUM data
+ * @param {string[]} urlsWithoutData - Array of URLs missing real analytics
+ * @returns {Array} Array of empty experimentation opportunities
+ */
+function createEmptyOpportunitiesForMissingUrls(urlsWithoutRUMData) {
+  return urlsWithoutRUMData.map((url) => ({
```
jeddiec, line 96: Is there a reason why we need this mapping? What do we see originally when an opportunity has no RUM data?

src/utils/url-utils.js
```diff
@@ -40,3 +40,39 @@ export function getCountryCodeFromLang(lang, defaultCountry = 'us') {
...
+
+  // Join all data and split by commas to handle comma-separated URLs
+  const allUrls = additionalData
+    .join(',')
+    .split(',')
+    .map((url) => url.trim())
+    .filter((url) => url.length > 0)
+    .map((url) => {
+      // Normalize relative paths to full URLs for consistency with RUM data
+      if (!url.startsWith('http://') && !url.startsWith('https://')) {
```
jeddiec, line 63: Can you help give an example why we might need the `map` here?

test/audits/experimentation-opportunities.test.js
```diff
@@ -351,4 +353,566 @@ describe('Experimentation Opportunities Tests', () => {
...
+
+      expect(context.rumApiClient.queryMulti).to.have.been.calledOnce;
+      expect(auditData.auditResult.experimentationOpportunities).to.have.length(2);
+      expect(auditData.auditResult.experimentationOpportunities[0].page).to.equal('https://abc.com/page1');
+      expect(auditData.auditResult.experimentationOpportunities[1].page).to.equal('https://abc.com/page2');
+      expect(auditData.auditResult.experimentationOpportunities[0]).to.have.property('pageViews');
+      expect(auditData.auditResult.experimentationOpportunities[0]).to.have.property('trackedPageKPIValue');
+    });
+
+    it('should combine real RUM data with empty data for custom URLs', async () => {
```
jeddiec, line 373: Can we rename it to `it('should handle the mixed URLs with and without RUM data'`. Would this help explain the purpose of the test more clear?

test/audits/experimentation-opportunities.test.js
```diff
@@ -351,4 +353,566 @@ describe('Experimentation Opportunities Tests', () => {
...
+      expect(auditData.auditResult.experimentationOpportunities).to.exist;
+    });
+
+    it('should handle empty custom URLs array', async () => {
+      context.auditContext = {
+        additionalData: [],
+      };
+
+      await runAuditAndScrapeStep(context);
+      expect(context.rumApiClient.queryMulti).to.have.been.calledOnce;
```
jeddiec, line 420: I think we need to check `expect(auditData.auditResult.experimentationOpportunities).to.exist;` here.
`rumApiClient.queryMulti` is called early in the function and we want to ensure the final result is returned.

test/audits/experimentation-opportunities.test.js
```diff
@@ -351,4 +353,566 @@ describe('Experimentation Opportunities Tests', () => {
...
+    });
+
+    it('should handle null additionalData', async () => {
+      context.auditContext = {
+        additionalData: null,
+      };
+
+      await runAuditAndScrapeStep(context);
+      expect(context.rumApiClient.queryMulti).to.have.been.calledOnce;
```
jeddiec, line 429: I think we need to check `expect(auditData.auditResult.experimentationOpportunities).to.exist;` here.
`rumApiClient.queryMulti` is called early in the function and we want to ensure the final result is returned.

test/audits/experimentation-opportunities.test.js
```diff
@@ -351,4 +353,566 @@ describe('Experimentation Opportunities Tests', () => {
...
+    });
+
+    it('should handle undefined auditContext', async () => {
+      context.auditContext = undefined;
+
+      await runAuditAndScrapeStep(context);
+      expect(context.rumApiClient.queryMulti).to.have.been.calledOnce;
```
jeddiec, line 436: I think we need to check `expect(auditData.auditResult.experimentationOpportunities).to.exist;` here.
`rumApiClient.queryMulti` is called early in the function and we want to ensure the final result is returned.

test/audits/experimentation-opportunities.test.js
```diff
@@ -351,4 +353,566 @@ describe('Experimentation Opportunities Tests', () => {
...
+        additionalData: ['/relative-page1', 'relative-page2'],
+      };
+
+      const auditData = await runAuditAndScrapeStep(context);
+      expect(auditData.auditResult.experimentationOpportunities).to.have.length(2);
+      expect(auditData.auditResult.experimentationOpportunities[0].page).to.equal('https://abc.com/relative-page1');
+      expect(auditData.auditResult.experimentationOpportunities[1].page).to.equal('https://abc.com/relative-page2');
+    });
+
+    it('should filter out non-high-organic-low-ctr opportunities when using custom URLs', async () => {
```
jeddiec, line 482: I think this test is for `getHighOrganicLowCtrOpportunityUrls()`?

test/audits/experimentation-opportunities.test.js
```diff
@@ -351,4 +353,566 @@ describe('Experimentation Opportunities Tests', () => {
...
+      };
+
+      context.rumApiClient.queryMulti = sinon.stub().resolves(mockOpportunities);
+
+      const result = await experimentOpportunitiesAuditRunner('https://abc.com', context);
+
+      expect(result.auditResult.experimentationOpportunities[0]).to.have.property('opportunityImpact', 50);
+    });
+
+    it('should handle empty metrics array for rage click', async () => {
```
jeddiec, line 572: Looks like we are testing other type of opportunities? Should we create another test file for these tests?

src/experimentation-opportunities/handler.js
FentPams: The reason I am putting here for some reasons:
1. It looks cleaner putting here rather than after long chain operation
2. Being defensive on iteration happens, if messages is undefined, will throw TypeError.
3. And this way we maintain what messages was without force it to be array. So it is easy to debug, when we see messages is undefined, we will know the chain fails.

src/experimentation-opportunities/handler.js
FentPams: Nice code!! While we are handling the case:
 the custom urls: `['page1', 'page2']`, and RUM has `[{page: 'page1, ...'}, {page: 'other-page', ...}]`
what we need is
`realDataOpportunities = [{page: 'page1', ...}]`  => this goes to final result with all metrics
`urlsWithoutRUMData = ['page2']` => this will go to creating empty opportunities
Then we combine them sending to result

src/experimentation-opportunities/handler.js
FentPams: Good question. My understanding, if there is no rum data, the url will not appear in the results. If user requires page1, page2, page3 RUM API returns only data for page 1 if page 2 doesn't have any. And we need map to fill the page2 and page3 without empty rum data.

test/audits/experimentation-opportunities.test.js
FentPams: Good catch! added

test/audits/experimentation-opportunities.test.js
FentPams: experimentation opportunities will return different type, including rage click, hotlctr, high inorganic high bounce rate, we need just find the opportunities for hotlctr.

test/audits/experimentation-opportunities.test.js
FentPams: similar purpose, to only focus with hotlctr opportunities

test/audits/experimentation-opportunities.test.js
```diff
@@ -351,4 +353,566 @@ describe('Experimentation Opportunities Tests', () => {
...
+      const [, message] = context.sqs.sendMessage.firstCall.args;
+
+      expect(message.time).to.be.a('string');
+      const messageTime = new Date(message.time).getTime();
+      expect(messageTime).to.be.at.least(beforeTime);
+      expect(messageTime).to.be.at.most(afterTime);
+    });
+  });
+
+  describe('parseCustomUrls Function', () => {
```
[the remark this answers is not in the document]
FentPams: yep, will move to right place.

src/experimentation-opportunities/handler.js
jeddiec: Should we throw error or return early here if the `messages` is not defined?

```js
if (!message) {
    log.warning("...")
    // throw error or early return
}

for (const message of messages) { ... }
```

src/experimentation-opportunities/handler.js
```diff
@@ -107,6 +125,36 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
   const queryResults = await rumAPIClient.queryMulti(OPPTY_QUERIES, options);
   const experimentationOpportunities = Object.values(queryResults).flatMap((oppty) => oppty);
   processRageClickOpportunities(experimentationOpportunities);
+
+  if (customUrls && customUrls.length > 0) {
+    log.info(`Processing ${customUrls.length} custom URLs for experimentation opportunities`);
+
+    const highOrganicLowCtrOpportunities = experimentationOpportunities.filter(
```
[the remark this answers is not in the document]
jeddiec: Thanks for the explanation, I did not notice it returns `.page` in that function.
In that case, would it make sense we modify the original function so it can be reused?

We then don't need to duplicate the same logic and the function `getHighOrganicLowCtrOpportunity` also can be more reusable. Customization can come after the function call base on the needs.

```js
const highOrganicLowCtrOpportunities = getHighOrganicLowCtrOpportunity(experimentationOpportunities);
```

```diff
- function getHighOrganicLowCtrOpportunityUrls(experimentationOpportunities) {
+ function getHighOrganicLowCtrOpportunity(experimentationOpportunities) {
  return experimentationOpportunities?.filter(
    (oppty) => oppty.type === HIGH_ORGANIC_LOW_CTR_OPPTY_TYPE,
-  )?.map((oppty) => oppty.page);
+  )?.map((oppty) => oppty);
}

// line 224
- const urls = getHighOrganicLowCtrOpportunityUrls(auditResult.experimentationOpportunities);
+ const urls = getHighOrganicLowCtrOpportunity(auditResult.experimentationOpportunities).map((oppty) => oppty.page);

// line 181
- urls: getHighOrganicLowCtrOpportunityUrls(
+ urls: getHighOrganicLowCtrOpportunity(
      result.auditResult?.experimentationOpportunities,
    ).map((oppty) => ({ oppty.page })),
```

test/audits/experimentation-opportunities.test.js
jeddiec: I am seeing `getHighOrganicLowCtrOpportunityUrls` is being used inside `runAuditAndScrapeStep()`. Would it make sense that we can simplify this test for `getHighOrganicLowCtrOpportunityUrls`? We are testing the filter of the function.

src/experimentation-opportunities/handler.js
jeddiec: Sorry the previous example I have was not quite accurate. Please see the following one.
Would this provide more context of our use case?

```js
// Convert to Set to have O(1) performance finding elements
const customUrlSet = new Set(customUrls);

// In one iteration
const opptiesWithRumData = highOrganicLowCtrOpportunities.reduce(
  (accumulator, currentValue) => {
    if (customUrlSet.has(currentValue.page)) {

      // Collect oppties with original Rum data
      accumulator.push(currentValue);
      
      // delete them from `customUrlSet`
      customUrlSet.delete(currentValue.page);
    }

    return accumulator;
  },
  []
);

// The remain URLS in customUrlSet should be mapped to oopty data format
const customUrlsWithoutRumData = [...customUrlSet];
const opptiesWithoutRumData = createEmptyOpportunitiesForMissingUrls(customUrlsWithoutRumData);

const finalExperimentationOpportunities = [...opptiesWithRumData, ...opptiesWithoutRumData];

```

test/audits/experimentation-opportunities.test.js
```diff
@@ -351,4 +353,566 @@ describe('Experimentation Opportunities Tests', () => {
...
+    });
+
+    it('should handle undefined auditContext', async () => {
+      context.auditContext = undefined;
+
+      await runAuditAndScrapeStep(context);
+      expect(context.rumApiClient.queryMulti).to.have.been.calledOnce;
+    });
+
+    it('should generate empty data when no RUM data exists for custom URLs', async () => {
```
jeddiec, line 439: Can we rename it to `it('should create opportunities with no RUM data for custom URLs'`?

test/audits/experimentation-opportunities.test.js
```diff
@@ -351,4 +353,566 @@ describe('Experimentation Opportunities Tests', () => {
...
+      context.auditContext = {
+        additionalData: ['https://abc.com/page1'],
+      };
+
+      const auditData = await runAuditAndScrapeStep(context);
+      expect(auditData.auditResult.experimentationOpportunities).to.have.length(1);
+      expect(auditData.auditResult.experimentationOpportunities[0].type).to.equal('high-organic-low-ctr');
+    });
+
+    it('should handle whitespace and empty strings in additionalData', async () => {
```
jeddiec, line 512: This is testing `parseCustomUrls()` and we should group it in `describe("parseCustomUrls()")` test set.

test/audits/experimentation-opportunities.test.js
```diff
@@ -351,4 +353,566 @@ describe('Experimentation Opportunities Tests', () => {
...
+            type: 'high-organic-low-ctr',
+            pageViews: 1000,
+          },
+        ],
+      };
+
+      context.rumApiClient.queryMulti = sinon.stub().resolves(mockOpportunities);
+      const customUrls = ['https://abc.com/page1', 'https://abc.com/page2'];
+
+      const result = await experimentOpportunitiesAuditRunner('https://abc.com', context, customUrls);
```
jeddiec, line 655: Yes! This is the test to test the logic of `experimentOpportunitiesAuditRunner()`.

test/audits/experimentation-opportunities.test.js
```diff
@@ -351,4 +353,566 @@ describe('Experimentation Opportunities Tests', () => {
...
+
+    it('should return correct structure without custom URLs', async () => {
+      const result = await experimentOpportunitiesAuditRunner('https://abc.com', context);
+
+      expect(result).to.have.property('auditResult');
+      expect(result).to.have.property('fullAuditRef', 'https://abc.com');
+      expect(result.auditResult).to.have.property('experimentationOpportunities');
+    });
+
+    it('should log correct messages for custom URLs with mixed data', async () => {
```
jeddiec, line 641: Yes! This is the testing the logic of experimentOpportunitiesAuditRunner().

src/experimentation-opportunities/handler.js
```diff
@@ -119,17 +182,20 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
 
 export async function runAuditAndScrapeStep(context) {
   const { site, finalUrl } = context;
-  const result = await experimentOpportunitiesAuditRunner(finalUrl, context);
+  const additionalData = context.auditContext?.additionalData;
```
jeddiec, line 185: Give that `runAuditAndScrapeStep()` is a general entry point for different types of oppty, I think we should move the `additionalData` and `customUrls` to `experimentOpportunitiesAuditRunner()`;

```diff
export async function runAuditAndScrapeStep(context) {
  const { site, finalUrl } = context;
-   const additionalData = context.auditContext?.additionalData;
-   const customUrls = parseCustomUrls(additionalData, finalUrl);

-  const result = await experimentOpportunitiesAuditRunner(finalUrl, context, customUrls);
  const result = await experimentOpportunitiesAuditRunner(finalUrl, context);

##

- export async function experimentOpportunitiesAuditRunner(auditUrl, context, customUrls = null) {
+ export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
  const { log } = context;

  const rumAPIClient = RUMAPIClient.createFrom(context);
  ...
  processRageClickOpportunities(experimentationOpportunities);

+   const additionalData = context.auditContext?.additionalData;
-   const customUrls = parseCustomUrls(additionalData, finalUrl);
+   const customUrls = additionalData ? parseCustomUrls(additionalData, auditUrl) : null;

 if (customUrls && customUrls.length > 0) { ...
```

src/experimentation-opportunities/handler.js
```diff
@@ -74,17 +74,41 @@ export async function generateOpportunityAndSuggestions(context) {
     },
   }));
 
+  if (!messages) {
+    log.info('No experimentation opportunities found or audit result is undefined.');
```
rpapani, line 78: please add the site.getId() and audit.getId() to the log, so that we can relate this log

src/experimentation-opportunities/handler.js
```diff
@@ -107,6 +130,50 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
   const queryResults = await rumAPIClient.queryMulti(OPPTY_QUERIES, options);
   const experimentationOpportunities = Object.values(queryResults).flatMap((oppty) => oppty);
   processRageClickOpportunities(experimentationOpportunities);
+
+  // Handle custom URLs if provided via audit context
+  const additionalData = context.auditContext?.additionalData;
+  const customUrls = additionalData ? parseCustomUrls(additionalData, auditUrl) : null;
+
+  if (customUrls && customUrls.length > 0) {
```
rpapani, line 138: `isNonEmptyArray(customUrls)` from shared-utils?

src/utils/url-utils.js
```diff
@@ -40,3 +40,39 @@ export function getCountryCodeFromLang(lang, defaultCountry = 'us') {
...
+ * @returns {Array|null} Array of normalized URLs or null
+ */
+export function parseCustomUrls(additionalData, domain = null) {
+  if (!additionalData || !Array.isArray(additionalData) || additionalData.length === 0) {
+    return null;
+  }
+
+  // Join all data and split by commas to handle comma-separated URLs
+  const allUrls = additionalData
+    .join(',')
```
rpapani, line 57: - your data (input) is comma separated urls, so we don't need that to be an array, please update the checks as well.
- not sure if we need this heavy processing, simply split, trim would do the trick, right?
- we can expect the user to provide the fully qualified url, i.e, including the domain etc. rather than you constructing it.
- instead of filtering, just use Set(urls), that should get rid of the duplicates

src/experimentation-opportunities/handler.js
```diff
@@ -107,6 +130,50 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
...
+          accumulator.push(currentValue);
+          customUrlSet.delete(currentValue.page);
+        }
+        return accumulator;
+      },
+      [],
+    );
+
+    const customUrlsWithoutRumData = [...customUrlSet];
+    const opptiesWithoutRumData = mapToOpportunities(customUrlsWithoutRumData);
```
rpapani, line 158: please note that the mapping wouldn't help, as `runAuditAndScrapeStep` will only read the urls from the results and send it for scraping, all the mapping would be dropped, so no mapping is required. When there're custom urls, we don't need to run the `experimentOpportunitiesAuditRunner`, just extract the urls and send those urls to the next step. This logic you can do directly in `runAuditAndScrapeStep`

rpapani (changes requested):
please check the comments

src/experimentation-opportunities/handler.js
```diff
@@ -107,6 +112,54 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
   const queryResults = await rumAPIClient.queryMulti(OPPTY_QUERIES, options);
   const experimentationOpportunities = Object.values(queryResults).flatMap((oppty) => oppty);
   processRageClickOpportunities(experimentationOpportunities);
+
+  // Handle custom URLs if provided via audit context
+  const { data } = context;
+  const customUrls = data ? parseCustomUrls(data) : null;
```
rpapani, line 118: may be you can use `[]` instead of null

src/experimentation-opportunities/handler.js
```diff
@@ -107,6 +112,54 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
...
+      log.info(`No RUM data found for ${opptiesWithoutRumData.length} URLs: ${customUrlsWithoutRumData.join(', ')} - using URLs without performance metrics`);
+    }
+
+    return {
+      auditResult: {
+        experimentationOpportunities: finalExperimentationOpportunities,
+      },
+      fullAuditRef: auditUrl,
+    };
+  }
```
rpapani, line 161: can you please move the [customUrls logic](https://github.com/adobe/spacecat-audit-worker/pull/1086/files#diff-3139033b5464618c7105c52ca4ae0bf72dbf69c4054818213703460230e6348aR121-R153) to a new method `processCustomUrls` and update like
```
if (isNonEmptyArray(customUrls)) {
customUrlOpportunities = processCustomUrls(customUrls, experimentationOpportunities);
return {
      auditResult: {
        experimentationOpportunities: customUrlOpportunities,
      },
      fullAuditRef: auditUrl,
    };
}
```
that'd make it more readable.

rpapani (approved):
@FentPams please incorporate the final feedback. Please validate in dev before merging. Thank you for your contribution!!

src/experimentation-opportunities/handler.js
```diff
@@ -107,6 +112,54 @@ export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
...
+
+  if (isNonEmptyArray(customUrls)) {
+    log.info(`Processing ${customUrls.length} custom URLs for experimentation opportunities`);
+
+    const highOrganicLowCtrOpportunities = getHighOrganicLowCtrOpportunity(
+      experimentationOpportunities,
+    );
+
+    const customUrlSet = new Set(customUrls);
+    const opptiesWithRumData = highOrganicLowCtrOpportunities.reduce(
```
rpapani, line 128: @FentPams upon further thinking, `highOrganicLowCtrOpportunities` will have only top 10 opportunities returned from [rum api client](https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-rum-api-client/src/functions/opportunities/high-organic-low-ctr.js#L142), so there's a high chance that you won't have any data for these custom urls in those 10. So, for custom urls case, you can do
```
const highOrganicLowCtrOpportunities = await rumAPIClient.query(HIGH_ORGANIC_LOW_CTR_OPPTY_TYPE, {...options, maxOpportunities:1000});
//.. The above will get you only high-organic-low-ctr opportunities max of 1000,
//..rest of the logic works fine from here.
```
I've introduced `maxOpportunities` option in https://github.com/adobe/spacecat-shared/pull/930

PR #1098, opened 2025-08-08
feat: keep the a11y individual opportunity if exists and update the suggestions, mark the ones not found with fixed

iuliag:
Ticket: https://jira.corp.adobe.com/browse/SITES-32889
Wiki: https://wiki.corp.adobe.com/display/AEMSites/Project+SpaceCat+-+Highlight+manual+fixes

It implements iteration 1, keeping the changes minimal and customizing the syncSuggestions only for a11y audits for now.

Collateral change from non-a11y files/tests:
- log as input parameter is not necessary, as context contains it.
- also, added debug logging in the syncSuggestions function.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,7 +45,6 @@
         "jsdom": "26.1.0",
         "robots-parser": "3.0.1",
         "text-readability": "1.1.1",
-        "ts-deepmerge": "7.0.3",
         "urijs": "1.19.11"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


src/accessibility/handler.js
```diff
@@ -228,9 +228,10 @@ export async function processAccessibilityOpportunities(context) {
 }
 
 export default new AuditBuilder()
-  .addStep('processImport', processImportStep, AUDIT_STEP_DESTINATIONS.IMPORT_WORKER)
+  // .addStep('processImport', processImportStep, AUDIT_STEP_DESTINATIONS.IMPORT_WORKER)
```
iuliag, line 231: These are just for easier testing, will be reverted before merging.

petcud (approved):

PR #1118, opened 2025-08-14
feat: implement scrapeClient for audit steps

LeChef318:
META-TAGS only
added the new AUDIT_STEP_DESTINATION.SCRAPE_CLIENT
-> instead of sending a message to a SQS-queue, the scrapeClient now handles that
-> results will be fetched by the step-audit and directly added to the stepContext as Map(["url" => "path"])


src/common/step-audit.js
```diff
@@ -72,9 +74,18 @@ export class StepAudit extends BaseAudit {
...
     };
 
-    const queueUrl = destination.getQueueUrl(context);
-    const payload = destination.formatPayload(stepResult, auditContext, context);
-    await sendContinuationMessage({ queueUrl, payload }, context);
+    if (step.destination === AUDIT_STEP_DESTINATIONS.SCRAPE_CLIENT) {
+      const scrapeClient = ScrapeClient.createFrom(context);
+      const payload = destination.formatPayload(stepResult, auditContext, context);
+      log.info(`Creating new scrapeJob with the ScrapeClient. Payload: ${JSON.stringify(payload)}`);
+      const scrapeJob = await scrapeClient.createScrapeJob(payload);
```
rpapani, line 81: @LeChef318 after scrape job is complete, which logic sends the message back to audit worker so that the stepped audit proceeds to next step? 

Is this how scrape client handles the job?
- scrape client might be sending the list of urls (say 200) in a job to a queue/dynamo
- job manager will take each url/set of urls from the job and use content-scraper to scrape the url(s) with a completionQueueUrl, so each url is executed in a separate lambda
- after content-scraper completes, it sends message to completionQueueUrl
- job manager aggregates these messages and updates the job config, for success/failure/reason/etc..
- when all the urls are processed, job manager will send message to audit queue to resume the stepped audit processing

Where can I read more about scrape client execution, to get more clarity on the overall flow.

src/support/sqs.js
```diff
@@ -32,6 +32,7 @@ class SQS {
     const msgCommand = new SendMessageCommand({
       MessageBody: asJSON,
       QueueUrl: queueUrl,
+      MessageGroupId: msgGroupId, // Only needed for FIFO queues
```
rpapani, line 35: are you using this any where for current flow or this is for future?

src/common/step-audit.js
LeChef318: there is no detailed documentation yet but i will add that to the scrape client's README.

here a short overview how it works:
- the scrape client creates a ScrapeJob in dynamo and sends the urls to the scrape-job-manager
- the scrape-job-manager creates for each url an entry in dynamo and sends a messages to the content-scraper (only 1 url in a message)
- after content scraper completes a message is sent to the content-processor
- the content-processor updates the updates the ScrapeUrl (status, path,...) and the ScrapeJob (failed/success count) in dynamo and checks when all urls of a job are done
- when the job is done the content-processor will send the the continuation message to the audit worker. the message contains the audit context and the scrapeJobId

src/common/step-audit.js
```diff
@@ -72,9 +74,18 @@ export class StepAudit extends BaseAudit {
       fullAuditRef: audit.getFullAuditRef(),
     };
 
-    const queueUrl = destination.getQueueUrl(context);
-    const payload = destination.formatPayload(stepResult, auditContext, context);
-    await sendContinuationMessage({ queueUrl, payload }, context);
+    if (step.destination === AUDIT_STEP_DESTINATIONS.SCRAPE_CLIENT) {
+      const scrapeClient = ScrapeClient.createFrom(context);
+      const payload = destination.formatPayload(stepResult, auditContext, context);
```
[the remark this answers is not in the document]
LeChef318: the old allowCache flag doesn't work anymore with the scrape client. we have our own caching logic which is by default 24 hours and is handled in the scrape-job-manager (i'll add that to the README too). let me know if it should be longer

ssilare-adobe (approved):
LGTM 👍

src/metatags/handler.js
```diff
@@ -242,38 +235,16 @@ export async function metatagsAutoDetect(site, pagesSet, context) {
...
-  const topPages = await getTopPagesForSiteId(dataAccess, siteId, context, log);
-  const includedURLs = await site?.getConfig()?.getIncludedURLs('meta-tags') || [];
-
-  // Transform URLs into scrape.json paths and combine them into a Set
-  const topPagePaths = topPages.map((page) => getScrapeJsonPath(page.url, siteId));
-  const includedUrlPaths = includedURLs.map((url) => getScrapeJsonPath(url, siteId));
-  const totalPagesSet = new Set([...topPagePaths, ...includedUrlPaths]);
-
-  log.info(`Received topPages: ${topPagePaths.length}, includedURLs: ${includedUrlPaths.length}, totalPages to process after removing duplicates: ${totalPagesSet.size}`);
-
```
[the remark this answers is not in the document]
ssilare-adobe: Can you clear that do you mean that whether the top pages are just sent while `meta-tags` audit has been triggered? Hence, the removal?

dzehnder:
I had to partially revert this PR as it broke the preflight metatags audit and we have a customer and leadership demo today: https://github.com/adobe/spacecat-audit-worker/pull/1196

PR #1134, opened 2025-08-18
fix(page-intent): store total number of missing page intents

cwjwisse:
- add custom persister to only persist the number of missing page intents
- previously, the default persister was used, which would cause issues when persisting too large audit results


ekremney (approved):

cwjwisse:
- add custom persister to only persist the number of missing page intents
- previously, the default persister was used, which would cause issues when persisting too large audit results

PR #1145, opened 2025-08-20
fix: base_url_fix removing URI from baseURL

abhishekgarg18:

## Related Issues

Thanks for contributing!

## Issue
When processing meta tag suggestions for URLs like `https://kpmg.com/ch/en.html`, the system is incorrectly appending path segments, resulting in malformed URLs:

- Current: `https://kpmg.com/ch/en.html/ch/en/about.html`
- Expected: `https://kpmg.com/ch/en/about.html`

The issue occurs because we're directly concatenating the endpoint to the base URL without proper path handling, causing path segments to be duplicated.

## Solution
Added `useHostnameOnly` configuration in site's delivery config to control URL handling:
- When `true`: Strips URL to just protocol + hostname (e.g., `https://kpmg.com`)
- When `false` (default): Preserves the full path, only removing trailing slashes

This allows sites to configure how their base URLs should be handled, which is particularly important for sites with complex URL structures like KPMG where we need to prevent path segment duplication.

## Implementation
- Added `useHostnameOnly` flag in delivery config
- Updated `getBaseUrl` to respect this configuration
- Added proper error handling and logging
- Default behavior remains unchanged (false) for backward compatibility

## Testing
Added test cases for:
- URLs with port numbers
- URLs with query parameters
- Error handling scenarios
- Default behavior verification


src/metatags/opportunity-utils.js
```diff
@@ -11,7 +11,13 @@
  */
 
 export function removeTrailingSlash(url) {
```
absarasw, line 13: Can we create a new function as getBaseUrl and use that? This might be used somewhere where we expect to only to remove the trailing slash.

abhishekgarg18:
@absarasw @dipratap please review.

dipratap:
@abhishekgarg18 can you please add a brief PR description of the change?

src/metatags/handler.js
```diff
@@ -47,6 +47,9 @@ export async function opportunityAndSuggestions(finalUrl, auditData, context) {
   );
   const { log } = context;
   const { detectedTags } = auditData.auditResult;
+  const site = opportunity.getSite?.();
+  const deliveryConfig = site !== null ? site?.getDeliveryConfig() : null;
```
dipratap, line 52: Can you please share the purpose of delivery config?

dipratap, line 52:
```suggestion
const useHostnameOnly = opportunity.getSite?.()
  ?.getDeliveryConfig()
  ?.useHostnameOnly ?? false;
```

src/metatags/opportunity-utils.js
```diff
@@ -14,6 +14,19 @@ export function removeTrailingSlash(url) {
   return url.endsWith('/') ? url.slice(0, -1) : url;
 }
 
+export function getBaseUrl(url, useHostnameOnly = true) {
```
dipratap, line 17: useHostnameOnly default should be false. If the urls structure change then all the previous suggestions, detected before this change, will have different urls than the new ones - which will lead to duplicate suggestions

dipratap: Before that, I would like to understand why are we doing this change, as it can break things.

dipratap (changes requested):

dipratap:
There is a overrideBaseUrl param that is currently used to overwrite the site baseurl params, we should check if the final Url param, which we are not using in this case, uses that overrideBaseUrl config, and if yes what should be the behaviour.

src/metatags/opportunity-utils.js
abhishekgarg18: changed default value to `false`

abhishekgarg18:
@dipratap I have added a description of PR.

Currently, we are using finalUrl,
``` url: getBaseUrl(auditData.auditResult.finalUrl, useHostnameOnly) + endpoint,```
i added 
```"fetchConfig": {
            "overrideBaseURL": "https://kpmg.com"
        }```
        
  it still didn't work.

src/metatags/handler.js
abhishekgarg18: I tried using config, but I am not able to add new value in this "config", so reverted to use deliveryConfig
Cc : @absarasw @dipratap

absarasw (approved):

alinarublea (approved):

PR #1149, opened 2025-08-22
fix: check if bucket exist

akshaymagapu:
Check if bucket exist in s3 for cdn logs report and then only create report.

### DIFF
--- a/src/cdn-logs-report/handler.js
+++ b/src/cdn-logs-report/handler.js
@@ -11,14 +11,30 @@
  */
 import { createFrom } from '@adobe/spacecat-helix-content-sdk';
 import { AWSAthenaClient } from '@adobe/spacecat-shared-athena-client';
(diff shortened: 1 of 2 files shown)


src/cdn-logs-report/handler.js
```diff
@@ -11,15 +11,31 @@
  */
 import { createFrom } from '@adobe/spacecat-helix-content-sdk';
 import { AWSAthenaClient } from '@adobe/spacecat-shared-athena-client';
+import { HeadBucketCommand } from '@aws-sdk/client-s3';
```
ctinp, line 14: The HeadBucketCommand should be provided by the same wrapper that the context uses to provide access to `s3Client` . Not a big deal, but this seems to be the standard across the codebase

akshaymagapu: will fix here: https://github.com/adobe/spacecat-audit-worker/pull/1147

PR #1155, opened 2025-08-22
fix: proper country lang pattern

akshaymagapu:
proper country lang pattern

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.157.8](https://github.com/adobe/spacecat-audit-worker/compare/v1.157.7...v1.157.8) (2025-08-26)
+
+
+### Bug Fixes
+
+* base_url_fix removing URI from baseURL ([#1145](https://github.com/adobe/spacecat-audit-worker/issues/1145)) ([7f079c5](https://github.com/adobe/spacecat-audit-worker/commit/7f079c5954aedd65964fbed38e9fe5f6c3e2c526))
+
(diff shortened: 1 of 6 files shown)


src/cdn-logs-report/constants/country-patterns.js
```diff
@@ -23,8 +23,8 @@ export const DEFAULT_COUNTRY_PATTERNS = [
   // Matches countries/regions prefix: /countries/us/, /regions/fr/, https://example.com/country/de/
   { name: 'countries_prefix', regex: '^(?:/|(?:https?:\\/\\/|\\/\\/)?[^/]+/)(?:countries?|regions?)/([a-z]{2})(?:/|$)' },
 
-  // Matches language/country format: /en/us/, /fr/fr/, https://example.com/de/de/page
-  { name: 'lang_country', regex: '^(?:/|(?:https?:\\/\\/|\\/\\/)?[^/]+/)[a-z]{2}/([a-z]{2})(?:/|$)' },
+  // Matches country/language format: /in/en/, /us/en/, https://example.com/de/en/page
+  { name: 'country_lang', regex: '^(?:/|(?:https?:\\/\\/|\\/\\/)?[^/]+/)([a-z]{2})/[a-z]{2}(?:/|$)' },
```
ctinp, line 27: The regex looks fine but do we check at some point that the two-letter codes extracted by this query are actual **country** codes, and not just any letters?

akshaymagapu: yeah we validate in processing, if not a valid country code we fallback to global.

https://github.com/adobe/spacecat-audit-worker/blob/main/src/cdn-logs-report/utils/report-utils.js#L64

PR #1158, opened 2025-08-24
fix: add support for content api urls for high organic low ctr auto fix

rpapani:
- Updates the url to content api url when sending Message to Mystique when conten api is enabled for AEM CS delivery type sites


terinmez:
ok, sounds good. We need to resolve pageRef -> pageId in Mystique because it has access to the AEM CS via it's client_id + secret.

PR #1191, opened 2025-08-29
feat: s3 restructure to use single bucket for CDN logs

akshaymagapu:
S3 - restructure to use single bucket for CDN logs


ekremney (approved):
**Important**: please adjust referral traffic related queries to this format as well.

PR #1220, opened 2025-09-03
feat: migrate experimentation opportunities to ScrapeClient

LeChef318:
- add ScrapeClient as destination
- refactor organicKeywordStep to use scrapeResultPaths directly
- update unit tests

### DIFF
--- a/src/experimentation-opportunities/handler.js
+++ b/src/experimentation-opportunities/handler.js
@@ -225,10 +225,8 @@ async function toggleImport(site, importType, enable, log) {
   }
 }
 
-async function getLangFromScrape(s3Client, bucketName, s3BucketPrefix, pathname, log) {
+async function getLangFromScrape(s3Client, bucketName, key, log) {
   // remove the trailing slash from the pathname
-  const pathnameWithoutTrailingSlash = pathname.replace(/\/$/, '');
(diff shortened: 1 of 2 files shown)


LeChef318:
> @LeChef318 looks like you're trying to use the scrapeResultPaths for importing organic keywords, I don't think we need to change the existing logic. Existing logic will read from the auditResult and find its scrape location to find the geo and use that to import organic keywords. The only change required is to update the step destination from Content Scraper to scrape client.

When changing to the scrape client destination it won't be able to find the scrapes anymore since they no longer are stored under the prefix `scrapes/{siteId}`. In `scrapeResultPaths` we have all the S3keys available for the URLs that were submitted for scraping by the audit in the previous step.

rpapani:
@LeChef318 
>When changing to the scrape client destination it won't be able to find the scrapes anymore since they no longer are stored under the prefix scrapes/{siteId}.

Is there a reason we're changing the path of scrapes? In today's scrape of celestyal.com/gb site, I see the scrapes are in the same format like earlier, see below image
image

> In scrapeResultPaths we have all the S3keys available for the URLs that were submitted for scraping by the audit in the previous step.

Does scrapeResultPaths send the unscraped url paths as well (due to the caching)?

PR #1227, opened 2025-09-04
fix(internal-links): update destination to ScrapeClient

LeChef318:
- update scrape destination to SCRAPE_CLIENT
- check if urls to scrape are empty


nitinja:
- Looks like AUDIT_STEP_DESTINATIONS.SCRAPE_CLIENT is a drop-in replacement for AUDIT_STEP_DESTINATIONS.CONTENT_SCRAPER  and does start scraping necessary pages. 
- CONTENT_SCRAPER  fetched specified pages and puts in s3 bucket, which this audit accesses later for sending to LLM for suggestion generation.
- Audit does fetch scraped page data for suggestion generations here: (this suggestions-generator.js is used by internal audit handler), but it does not scrape pages on its own.
https://github.com/adobe/spacecat-audit-worker/blob/d73078f26d38dd3ba0cafeadb33cbb8b022b4831/src/internal-links/suggestions-generator.js#L30

nitinja (approved):

PR #1236, opened 2025-09-05
fix: showing ai agents blocked if all user agents are blocked

hannessolo:
Previously, we showed ai agents blocked even if we block *. Now only shown if * is allowed.

This was previously implemented but somehow got lost.


hannessolo:
> This PR will trigger no release when merged.

Why not?

PR #1248, opened 2025-09-10
feat: add security vulnerability audit

kronnox:

## Related Issues

Thanks for contributing!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,6 +22,7 @@
         "@adobe/spacecat-shared-google-client": "1.4.49",
         "@adobe/spacecat-shared-gpt-client": "1.6.5",
         "@adobe/spacecat-shared-http-utils": "1.17.6",
+        "@adobe/spacecat-shared-ims-client": "1.8.9",
         "@adobe/spacecat-shared-rum-api-client": "2.38.0",
         "@adobe/spacecat-shared-rum-api-client-v1": "npm:@adobe/spacecat-shared-rum-api-client@1.8.4",
         "@adobe/spacecat-shared-scrape-client": "2.1.5",
@@ -961,7 +962,6 @@
       "resolved": "https://registry.npmjs.org/@adobe/helix-universal/-/helix-universal-5.2.3.tgz",
       "integrity": "sha512-13lpLUAyPvitjNh6eVZKjh8+kE+0LGHElyIaV0X7wOqUn371+BR9LrfS5seOpcIDAFNpEbL4T68rgFa1/aQrbA==",
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "@adobe/fetch": "4.2.3",
         "aws4": "1.13.2"
@@ -3027,7 +3027,6 @@
       "resolved": "https://registry.npmjs.org/@aws-sdk/client-sts/-/client-sts-3.726.1.tgz",
       "integrity": "sha512-qh9Q9Vu1hrM/wMBOBIaskwnE4GTFaZu26Q6WHwyWNfj7J8a40vBxpW16c2vYXHLBtwRKM1be8uRLkmDwghpiNw==",
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "@aws-crypto/sha256-browser": "5.2.0",
         "@aws-crypto/sha256-js": "5.2.0",
@@ -3612,7 +3611,6 @@
       "resolved": "https://registry.npmjs.org/aws-xray-sdk-core/-/aws-xray-sdk-core-3.10.2.tgz",
       "integrity": "sha512-hAFEB+Stqm4FoQmIuyw5AzGVJh3BSfvLjK7IK4YYRXXLt1Oq9KS6pv2samYgRTTTXsxhmVpDjiYF3Xo/gfXIXA==",
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "@aws-sdk/types": "^3.4.1",
         "@smithy/service-error-classification": "^2.0.4",
@@ -5474,7 +5472,6 @@
       "resolved": "https://registry.npmjs.org/@aws-sdk/client-dynamodb/-/client-dynamodb-3.859.0.tgz",
       "integrity": "sha512-Bt840uICsGcn7IFewif8ARCF0CxtdTx9DX/LfUGRI+SVZcqyeEccmH2JJRRzThtEzKTXr+rCN6yaNB3c4RQY2g==",
       "license": "Apache-2.0",
-      "peer": true,
       "dependencies": {
         "@aws-crypto/sha256-browser": "5.2.0",
         "@aws-crypto/sha256-js": "5.2.0",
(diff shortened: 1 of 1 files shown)


src/common/opportunity.js
```diff
@@ -71,7 +71,8 @@ export async function convertToOpportunity(auditUrl, auditData, context, createO
       opportunity.setAuditId(auditData.id);
       if (auditType === Audit.AUDIT_TYPES.CWV
           || auditType === Audit.AUDIT_TYPES.META_TAGS
-          || auditType === Audit.AUDIT_TYPES.SECURITY_CSP) {
+          || auditType === Audit.AUDIT_TYPES.SECURITY_CSP
+          || auditType === 'security-vulnerabilities') { // TODO: Audit.AUDIT_TYPES.SECURITY_VULNERABILITIES
```
bisswanger, line 75: this needs a PR + release + dependency bump on https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-data-access/src/models/audit/audit.model.js

bisswanger: not sure if we can trigger it via Slackbot etc without adding it to the model - never tested

src/vulnerabilities/vulnerability-report.d.ts
```diff
@@ -0,0 +1,68 @@
+/**
```
bisswanger, line 1: potentially also move to spacecat-shared to that the UI may reuse the same type definitions

src/vulnerabilities/suggestion-data-mapper.js
```diff
@@ -0,0 +1,65 @@
...
+  const {
+    name, version, recommendedVersion, vulnerabilities,
+  } = vulnerability;
+
+  // Handle null/undefined vulnerabilities
+  const safeVulnerabilities = vulnerabilities || [];
+
+  return {
+    opportunityId: opportunity.getId(),
+    type: 'CONTENT_UPDATE',
```
bisswanger, line 50: I would rather say this should be type CODE_CHANGE?

src/vulnerabilities/handler.js
```diff
@@ -0,0 +1,238 @@
...
+    buildKey,
+    mapNewSuggestion: (entry) => mapVulnerabilityToSuggestion(opportunity, entry),
+    log,
+  });
+
+  return { status: 'complete' };
+};
+
+export default new AuditBuilder()
+  .withUrlResolver(wwwUrlResolver)
```
bisswanger, line 235: do we really need the URL resolver?
This one fetches RUM-related information but should not be needed for the VULN audit?

src/vulnerabilities/handler.js
```diff
@@ -0,0 +1,238 @@
...
+ * @param {Object} site - The site object
+ * @returns {Object} The audit data unchanged (opportunities created as side effect).
+ */
+export const opportunityAndSuggestionsStep = async (auditUrl, auditData, context, site) => {
+  const { log, dataAccess } = context;
+  const { Configuration, Suggestion } = dataAccess;
+
+  // Check whether the audit is enabled for the site
+  const configuration = await Configuration.findLatest();
+  if (!configuration.isHandlerEnabledForSite('security-vulnerabilities', site)) {
```
bisswanger, line 138: hmm, as this function is only triggered during an audit of this type, not sure we need an explicit check here?

src/vulnerabilities/handler.js
```diff
@@ -0,0 +1,238 @@
...
+    log.info(`[${AUDIT_TYPE}] [Site: ${site.getId()}] audit is disabled for site`);
+    return { status: 'complete' };
+  }
+
+  // This opportunity is only relevant for aem_cs delivery-type at the moment
+  if (site.getDeliveryType() !== 'aem_cs') {
```
bisswanger, line 144: hm, I would rather move this part to the audit step - i.e. if not CS there is no need in even downloading the report?

src/vulnerabilities/handler.js
```diff
@@ -0,0 +1,238 @@
...
+ * @param {object} context - The context object of the audit
+ * @param {object} site - The site object containing delivery configuration and details.
+ * @return {Promise<VulnerabilityReport>} A promise that resolves to the vulnerability report data.
+ */
+export async function fetchVulnerabilityReport(baseURL, context, site) {
+  const { log, env, dataAccess } = context;
+
+  // Retrieve site details
+  const imsOrg = await getImsOrgId(site, dataAccess, log);
+  const { programId, environmentId } = site.getDeliveryConfig();
```
bisswanger, line 44: add a check to not continue if no IMS org was retrieved? => function returns null in that case

src/vulnerabilities/handler.js
bisswanger: ok, it's implicitly already there by checking for the program + environment id

src/vulnerabilities/handler.js
```diff
@@ -0,0 +1,238 @@
...
+  if (!programId || !environmentId) {
+    throw new Error('Invalid delivery config for AEM_CS');
+  }
+
+  // Get service access-token
+  const imsContext = {
+    log, env,
+  };
+  const imsClient = ImsClient.createFrom(imsContext);
```
bisswanger, line 53: move into a try ... catch to have an explicit error like "failed to retrieve IMS token"?

src/vulnerabilities/handler.js
```diff
@@ -0,0 +1,238 @@
...
+
+  // Fetch vulnerability report
+  try {
+    const headers = {
+      Authorization: `Bearer ${token.access_token}`,
+      'x-api-key': env.IMS_CLIENT_ID,
+      'x-gw-ims-org-id': imsOrg,
+    };
+    const resp = await fetch(
+      `https://aem-trustcenter-dev.adobe.io/api/reports/${programId}/${environmentId}/vulnerabilities`,
```
bisswanger, line 64: can we move the Trustcenter URL into some environment variable / config parameter?

src/vulnerabilities/handler.js
```diff
@@ -0,0 +1,238 @@
...
+  const opportunity = await convertToOpportunity(
+    auditUrl,
+    { siteId: auditData.siteId, id: auditData.auditId },
+    context,
+    createOpportunityData,
+    AUDIT_TYPE,
+    createOpportunityProps(auditData.auditResult.vulnerabilityReport),
+  );
+
+  if (!configuration.isHandlerEnabledForSite('security-vulnerabilities-auto-suggest', site)) {
```
bisswanger, line 202: the convertToOpportunity only creates an opportunity but without any libraries listed

If the flag is disabled, there will then be an empty opportunity with no suggestions?
Not sure we need a flag here - or if we have it, it should rather just set "empty" on the recommendedVersion received from the report?

src/common/opportunity.js
kronnox: PR opened https://github.com/adobe/spacecat-shared/pull/953

src/vulnerabilities/handler.js
kronnox: Pretty sure the default is `default` here and not null. Eventually we should have a proper validation, but for now it works regardless of what's in there. The Starfish API doesn't seem to care right now

PR #1256, opened 2025-09-12
feat(prflight): migrate to scrape client

LeChef318:
- add SCRAPE_CLIENT destination for scraping
- fetch scrape result paths and attach it to the context
- add scrape client to accessibility check
- updated unit tests

Note:
this PR also includes changes from PR [https://github.com/adobe/spacecat-audit-worker/pull/1216](https://github.com/adobe/spacecat-audit-worker/pull/1216) that were rolled back due issues with preflight


dzehnder (commented):
Looks good! How long does it take with the new scrape client to execute the audits? Can we do a small performance analysis before we merge as this is critical for the success of preflight?

PR #1266, opened 2025-09-15
fix: updating response handling from mystique to spacecat

ymadobe:
- Removed matching of UAs from spacecat to mystique, suggestions are now updated if the URL matches.

### DIFF
--- a/src/llm-error-pages/guidance-handler.js
+++ b/src/llm-error-pages/guidance-handler.js
@@ -59,19 +59,30 @@ export default async function handler(message, context) {
     const sheet = workbook.worksheets[0] || workbook.addWorksheet('data');
 
     const baseUrl = site.getBaseURL?.() || 'https://example.com';
+    const SPREADSHEET_COLUMNS = [
+      'Agent Type',
+      'User Agent',
+      'Number of Hits',
+      'Avg TTFB (ms)',
+      'Country Code',
+      'URL',
+      'Product',
+      'Category',
+      'Suggested URLs',
+      'AI Rationale',
+      'Confidence score',
+    ];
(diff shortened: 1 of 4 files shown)


src/llm-error-pages/guidance-handler.js
```diff
@@ -75,30 +85,38 @@ export default async function handler(message, context) {
...
+
       if (urlCell) {
         const pathOnlyUrl = toPathOnly(urlCell, baseUrl);
+        log.info(`Row ${i}: Checking URL="${pathOnlyUrl}", UserAgent="${userAgentCell}"`);
 
         // Look up the URL in brokenUrls and update if found
         const brokenUrlData = brokenUrlsMap.get(pathOnlyUrl);
-        if (brokenUrlData && brokenUrlData.userAgents.includes(userAgentCell)) {
+        if (brokenUrlData) {
+          log.info(`Found match for URL: ${pathOnlyUrl}, Mystique userAgents: "${brokenUrlData.userAgents}", Excel userAgent: "${userAgentCell}"`);
```
duynguyen, line 102: debug logs

src/llm-error-pages/guidance-handler.js
```diff
@@ -59,14 +59,35 @@ export default async function handler(message, context) {
     const sheet = workbook.worksheets[0] || workbook.addWorksheet('data');
 
     const baseUrl = site.getBaseURL?.() || 'https://example.com';
+    const SPREADSHEET_COLUMNS = [
```
duynguyen, line 62: given this is shared between handler and guidance-handler, can we move it to utils and used in both audits?

duynguyen (approved):

PR #1280, opened 2025-09-18
feat: new audit type - prerender

anuj-adobe:

## Related Issues

Thanks for contributing!


src/prerender/handler.js
```diff
@@ -0,0 +1,353 @@
...
+
+    const { SiteTopPage } = dataAccess;
+    const topPages = await SiteTopPage.allBySiteIdAndSourceAndGeo(siteId, 'ahrefs', 'global');
+
+    topPages.forEach((page) => {
+      trafficMap.set(page.getUrl(), page.getTraffic());
+    });
+
+    // Try to get URLs from the audit context first
+    if (scrapeResultPaths?.size > 0) {
```
dipratap, line 282: We are using fallbacks which is fine, however, I think not getting the URLs in scrapeResultPaths is not an expected behaviour. Also, we are setting needsPrerender to false in such cases where the scraped content is not there which isn't ideal. If we can somehow flag these anomalies - maybe starting with adding error log for these and monitoring them, so that we know when something isn’t working as expected

dipratap (changes requested):

anuj-adobe:
I've updated suggestion type to CONFIG_UPDATE which is released in 2.64.4 version https://github.com/adobe/spacecat-shared/commit/33d569c87879fbdffb6c39bbb3c3cdfe880179e5

The build should work with auto-dep update in this repo.

src/prerender/handler.js
```diff
@@ -0,0 +1,358 @@
...
+export async function importTopPages(context) {
+  const { site, finalUrl } = context;
+
+  const s3BucketPath = `scrapes/${site.getId()}/`;
+  return {
+    type: 'top-pages',
+    siteId: site.getId(),
+    auditResult: { status: 'preparing', finalUrl },
+    fullAuditRef: s3BucketPath,
+    allowCache: false,
```
dipratap, line 148: Lets remove allowCache: false as it will import top-pages on each run.

dipratap (approved):

PR #1423, opened 2025-10-16
feat: read rum threshold values from dlivery config if available

abhishekgarg18:

## Related Issues

Thanks for contributing!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.216.2](https://github.com/adobe/spacecat-audit-worker/compare/v1.216.1...v1.216.2) (2025-10-24)
+
+
+### Bug Fixes
+
+* improve fastly onboarding ([#1452](https://github.com/adobe/spacecat-audit-worker/issues/1452)) ([c2e8fcb](https://github.com/adobe/spacecat-audit-worker/commit/c2e8fcb84853094b1352090044d9eae5f38e217f))
+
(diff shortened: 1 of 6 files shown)


src/cwv/handler.js
```diff
@@ -25,19 +25,22 @@ const INTERVAL = 7; // days
 const auditType = Audit.AUDIT_TYPES.CWV;
 
 export async function CWVRunner(auditUrl, context, site) {
+  const deliveryConfig = site?.getDeliveryConfig() || {};
+  const dailyThreshold = deliveryConfig.dailyThreshold || DAILY_THRESHOLD;
+  const interval = deliveryConfig.interval || INTERVAL; // days
```
vdua: This is needed only for testing on our demo sites where the traffic is not enough. The thresholds were decided based on that fact that for sites with low traffic, the revenue impact will not be enough and hence we do not consider low traffic pages.

Unless we have an agreement regarding the same, we should not merge this PR.

On a side note, there are multiple opportunities, each with their own threshold/intervals. If we go ahead with this PR we should 
* use opportunity specific thresholds (for instance `deliveryConfig.cwv.dailyThreshold`) and 
* ensure `INTERVAL` cannot be set to any arbitrary large value as that can impact fetching data from RUM.

abhishekgarg18: @vdua Thanks for the review! I’ve incorporated your feedback in the latest commit.
imho, for testing any opportunity end-to-end, I think it’s important to keep this configurable.

naydav (approved):

PR #1442, opened 2025-10-21
feat: add FAQs audit

duynguyen:
Add FAQs audit:
* Handler takes the top prompts from brand presence data, group them by URL and topic, and send to Mystique for AI suggestion generation
* Guidance handler receives the message back from mystique, save the suggestions to spacecat as generic oppty

Example output: https://test.d2ikwb7s634epv.amplifyapp.com/bulk.com/opportunities/generic/476f0e28-17bc-48dc-9562-cfc122efa880


src/faqs/guidance-handler.js
```diff
@@ -0,0 +1,146 @@
...
+      opportunity.setAuditId(auditId);
+      opportunity.setData({
+        ...opportunity.getData(),
+        ...entity.data,
+      });
+      opportunity.setGuidance(wrappedGuidance);
+      opportunity.setUpdatedBy('system');
+      opportunity = await opportunity.save();
+      log.info(`Updated existing FAQ opportunity: ${opportunity.getId()}`);
+    }
```
hannessolo: Not opposed to this, but curious if you have a specific reason for not using `convertToOpportunity`?

src/faqs/handler.js
```diff
@@ -0,0 +1,223 @@
...
+    const sharepointClient = await createLLMOSharepointClient(context);
+    const outputLocation = getOutputLocation
+      ? getOutputLocation(site)
+      : `${site.getConfig().getLlmoDataFolder()}/brand-presence`;
+    const brandPresenceFilename = `brandpresence-all-${periodIdentifier}.xlsx`;
+
+    // Read brand presence spreadsheet
+    const topPrompts = await readBrandPresenceSpreadsheet(
+      brandPresenceFilename,
+      outputLocation,
```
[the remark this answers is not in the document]
duynguyen: agree 👍  but `outputLocation` was already used in the function called by this, and also from the audit context (`getOutputLocation`).
hence naming can be improved, but probably decoupled from this change for consistency...

src/faqs/guidance-handler.js
duynguyen: `convertToOpportunity` only works for updating oppties of specific types (like `structured-data`)
as this audit produces a generic oppty, we're using a `subType` field in order to differentiate other oppties of the same kind (e.g. summarization)
there are custom logics here than the default in `convertToOpportunity` to look up the audit by sub type and update.

PR #1486, opened 2025-10-30
trimming metatags from scraper before processing

miakobchuk:
Fix for https://jira.corp.adobe.com/browse/SITES-36652

### DIFF
--- a/src/utils/seo-utils.js
+++ b/src/utils/seo-utils.js
@@ -44,10 +44,11 @@ const issueRankings = {
  * Trims whitespace from string or array of strings.
  * Used for cleaning scraped meta tag content (title, description, h1).
  * @param {string|string[]|null|undefined} value - Value to trim
- * @returns {string|string[]|null} - Trimmed value(s)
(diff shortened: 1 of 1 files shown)


src/metatags/handler.js
```diff
@@ -108,9 +108,9 @@ export async function fetchAndProcessPageObject(s3Client, bucketName, url, key,
   // handling for homepage
   return {
     [pageUrl]: {
-      title: object.scrapeResult.tags.title,
-      description: object.scrapeResult.tags.description,
-      h1: object.scrapeResult.tags.h1 || [],
+      title: trimTagValue(object.scrapeResult.tags.title),
+      description: trimTagValue(object.scrapeResult.tags.description),
+      h1: trimTagValue(object.scrapeResult.tags.h1) || [],
```
naydav: Looks like fix should be
`h1: (object.scrapeResult.tags.h1 || []).map(h => trimTagValue(h)),`

because
object.scrapeResult.tags.h1 is `list`, but it isn't a `string`

naydav (changes requested):

naydav (approved):

PR #1495, opened 2025-11-01
fix: check if retry after header present

akshaymagapu:
check if retry after header present from helix admin api


ekremney (approved):

ekremney:
@akshaymagapu let's go with https://github.com/adobe/spacecat-audit-worker/pull/1496

PR #1506, opened 2025-11-04
LLMO-1023: integrate logWrapper for trace ID logging

HollywoodTonight:
- Add logWrapper to wrapper chain after sqsEventAdapter
- Replace context.log with context.contextualLog for enhanced logging
- Logs will now include AWS X-Ray trace IDs automatically

## Related Issues

Thanks for contributing!


iuliag (changes requested):
I know I'm late to review this, but the implementation needs to be revisited:
The idea was to include ootb the traceId in the logs without requiring to change the log statements throughout the codebase.
Because `contextualLog` needs to be used instead of `log` for having the trace id included, it means we need to do this replacement everywhere, including in the handlers implementation, not just here in the common before and after running the handler logic.

PR #1508, opened 2025-11-04
feat: add specific opportunity for faq

duynguyen:

## Related Issues

Thanks for contributing!


hannessolo (approved):
Seems good to me, but could you explain why we're doing `analyzeScrapeData` on the guidance handler, and not in Mystique, since Mystique probably already looks at the scrape anyway.

Should we long-term do it in a unified way between the 2 opportunities or is there a reason for it to be different?

PR #1531, opened 2025-11-06
fix: canonical self referenced urls

oiatsiuk:
Summary:
Optimize canonical audit by skipping redundant accessibility checks for self-referenced canonical URLs.

### DIFF
--- a/test/audits/canonical.test.js
+++ b/test/audits/canonical.test.js
@@ -884,8 +884,8 @@ describe('Canonical URL Tests', () => {
       const pageURL = 'https://example.com/page1';
       const html = `<html lang="en"><head><link rel="canonical" href="${pageURL}"><title>test</title></head><body></body></html>`;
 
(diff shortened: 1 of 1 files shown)


src/canonical/handler.js
```diff
@@ -540,8 +540,28 @@ export async function canonicalAuditRunner(baseURL, context, site) {
...
+        const isSelfReferenced = selfRefCheck?.success === true;
+
+        // if self-referenced - skip accessibility
+        if (isSelfReferenced) {
+          checks.push({
+            check: CANONICAL_CHECKS.CANONICAL_URL_STATUS_OK.check,
+            success: true,
+          });
+          checks.push({
+            check: CANONICAL_CHECKS.CANONICAL_URL_NO_REDIRECT.check,
```
mikeweis: It seems that this line should be part of the previous `push` (after line 552).  Unless I am missing something? 🤔

PR #1535, opened 2025-11-07
feat: use new aggregated bucket for reports

akshaymagapu:
use new aggregated bucket for reports


src/cdn-logs-report/constants/report-configs.js
```diff
@@ -9,21 +9,15 @@
...
-
+export function getConfigs(bucket, customerDomain, siteId) {
   return [
     {
       name: 'agentic',
       createTableSql: 'create-aggregated-table',
-      aggregatedLocation: `s3://${s3path}/aggregated/`,
-      tableName: `aggregated_logs_${customerDomain}`,
+      aggregatedLocation: `s3://${bucket}/aggregated/${siteId}/`,
+      tableName: `aggregated_logs_${customerDomain}_consolidated`,
```
ctinp: curious why we need the `_consolidated` suffix. Are we disambiguating against something else?

ctinp (approved):

PR #1598, opened 2025-11-18
fix: prerender audit adding agentic urls

anuj-adobe:
Code change -

1. Using athena query to find top agentic URLs in last 4 weeks with fallback to sheet as athena query on dev only works for wknd.site.
2. Not saving traffic value in suggestion entity, traffic value is calculated on UI as done in other opportunity as well
3. Rank is set to 0 for all URLs and sorting is done on UI based on agentic traffic value calculated

Code refactoring -
1. Created a `utils` folder
2. moved `html-comparator-util.js` under `utils` folder with name html-comparator.js
4. Created a constant.js file to use same value at multiple places in prerender
5. Created a shared.js to wrap external audit methods getting used in prerender audit. Can later extract those out to common utility.

## Related Issues

Thanks for contributing!


src/cdn-logs-report/sql/agentic-hits-for-urls.sql
```diff
@@ -0,0 +1,8 @@
+SELECT
+  url,
+  SUM(number_of_hits) AS number_of_hits
+FROM {{databaseName}}.{{tableName}}
+{{whereClause}}
+AND url IN ({{urlList}})
+GROUP BY url;
```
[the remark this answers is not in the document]
anuj-adobe: removed this file.

src/prerender/handler.js
```diff
@@ -12,17 +12,179 @@
...
+    const recentWeeks = Array.isSame ? periods.weeks : periods.weeks;
+    const windowWeeks = recentWeeks.slice(0, 4);
+    const startBoundary = windowWeeks.length > 0
+      ? windowWeeks[windowWeeks.length - 1].startDate
+      : periods.weeks[periods.weeks.length - 1].startDate;
+    const endBoundary = windowWeeks.length > 0
+      ? windowWeeks[0].endDate
+      : periods.weeks[0].endDate;
+    const fourWeekPeriods = { weeks: [{ startDate: startBoundary, endDate: endBoundary }] };
+    const athenaClient = AWSAthenaClient.fromContext(context, s3Config.getAthenaTempLocation());
```
dipratap: separate discussion: any purging we can enable on this s3 result location since we will be saving results on it.

src/prerender/handler.js
```diff
@@ -12,17 +12,179 @@
...
+    // Build query using agentic report query builder (date-range + LLM UA + site filters)
+    const query = await weeklyBreakdownQueries.createAgenticReportQuery({
```
dipratap: since we fetching for 4 weeks, any idea what the data size would be?

src/prerender/handler.js
```diff
@@ -12,17 +12,179 @@
...
+ * @param {number} limit
+ * @returns {Promise<Array<{url:string}>>}
+ */
+async function getTopAgenticUrlsFromAthena(site, context, limit = TOP_AGENTIC_URLS_LIMIT) {
+  const { log } = context;
+  try {
+    const s3Config = await getS3Config(site, context);
+    const periods = generateReportingPeriods();
+    // Build a single contiguous date window covering the last up to 4 weeks
+    const recentWeeks = Array.isSame ? periods.weeks : periods.weeks;
```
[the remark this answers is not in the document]
anuj-adobe: vibe coding side-effect, removed.

src/prerender/handler.js
anuj-adobe: It's good point. I think auto-purge could be applied to buckets storing athena query after 7 days. But that would be applicable to all buckets storing the athena query result and we could discuss this outside the PR scope?

dipratap (approved):

PR #1668, opened 2025-11-26
readability detect improvements w/ cheerio

dicagno:

## Related Issues

Thanks for contributing!


hannessolo (approved):
Same review as before - given time constraints this is fine but we should look at a way to treat text beyond individual elements, what if there's a nested li in a p, as an example - or maybe it makes sense to optimize a `ul` with several `li` at the same time, vs each `li` separately

PR #1675, opened 2025-11-26
feat: introduce selectors in the payload of preflight for highlighting

dzehnder:
## Description
Surface precise DOM locations for every preflight issue so the UI can highlight the offending elements directly. Adds a shared selector utility and threads those selectors through each audit response plus fixtures/tests.

## Summary
- Introduced `getDomElementSelector()`/`toElementTargets()` helper for consistent minimal CSS selectors.
- Added element selector payloads to all preflight checks which are visible (DOM/body-size/lorem/H1, links, headings, readability, accessibility) so each opportunity points to its exact node.
- Updated affected fixtures and expectations to include the new selector fields.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [1.307.1](https://github.com/adobe/spacecat-audit-worker/compare/v1.307.0...v1.307.1) (2026-02-10)
+
+
+### Bug Fixes
+
+* Make code patch disabled by default for a11y fixes ([#1964](https://github.com/adobe/spacecat-audit-worker/issues/1964)) ([638f450](https://github.com/adobe/spacecat-audit-worker/commit/638f4503912e143ab4dec6d6274a2838cae59641))
+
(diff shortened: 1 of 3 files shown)


radhikagpt1208:
Changes LGTM, but I see a potential issue with the links check-
Broken link issues only propagate `elements`, so CS/UE targeting might get dropped because:
1. `toElementTargets()` intentionally returns either `{ elements: [...] }` (standard CSS) or `{ selector: { elements: [...] } }` (CS/UE)
2. `runLinksChecks()` correctly includes whatever it gets from `toElementTargets` 
3. But, later links check does `selectorsByLink.set(..., link.elements)` and destructures `{ ..., elements }` for broken links, so when the broken link object contains `selector` (CS/UE) the final broken-links issues end up with no info.

However, this will work fine for bad-links because it spreads `...toElementTargets(selector)`

To fix this, for broken internal/external links, we can propagate the entire target payload (elements and/or selector) instead of only elements output, or pass through both properties from `runLinksChecks()` results.

@dzehnder can you also please verify this?

PR #1684, opened 2025-11-27
fix: Lambda deployment failures caused by legacy RUM client

solaris007:
# Fix Lambda deployment failures caused by legacy RUM client

**Problem**

After updating to `@adobe/helix-deploy@13.2.0`, all newly deployed Lambdas began failing at runtime before our handler code could execute. AWS CloudWatch logs showed:

```
Error: Cannot find module '../../package.json'
...
TypeError: main2 is not a function
```

**Root Cause**

We were shipping both modern and legacy versions of the RUM API client:
- `@adobe/spacecat-shared-rum-api-client@2.39.0` (current, ESM)
- `@adobe/spacecat-shared-rum-api-client-v1` (alias to `@1.8.4`, legacy CommonJS)

The v1 client bundles an old copy of `@adobe/fetch` that uses `createRequire(import.meta.url)` to load `../../package.json` at runtime. This worked in previous helix-deploy versions where node_modules structure was preserved, but the new bundler flattens everything into a single `dist/spacecat-services/unpacked/index.js` file. When the Lambda tries to execute, that relative path no longer exists, causing initialization to fail before `main` is ever assigned—hence "main2 is not a function."

Comparing with `spacecat-api-service` (which deploys successfully) confirmed it only depends on the modern v2 client and doesn't encounter this issue.

**Solution**

- Removed the `@adobe/spacecat-shared-rum-api-client-v1` dependency alias
- Removed 3 audits using the old client: `notfound`, `conversions` and `experimentation`
- Regenerated lockfile 

The modern client uses proper ESM imports that survive the bundling process, eliminating the runtime `require()` failures.

**Testing**

- ✅ `npm run build` completes successfully
- ✅ `npm run start:unpacked` dev server starts and responds to requests

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,29 +82,80 @@ After logging into KLAM, you’ll receive the following credentials required to 
 
 ### Steps to use `nodemon`
 
-#### 1. Create an `.env` File
+#### 1. Configure Local Environment Variables
 
-Create a `.env` file in the root directory with the following environment variables, which are required for all audits.  
-Add any additional environment variables specific to the audit you're working on.
+Both development scripts (`npm start` and `npm run start:unpacked`) require environment variables to be set. The **`start:unpacked`** script uses [dotenv](https://github.com/motdotla/dotenv) to automatically load them from a `.env` file in the project root. For **`npm start`**, you need to manually export the variables in your shell or use a tool to load `.env` before running the command.
 
-```
+The `.env` file should contain:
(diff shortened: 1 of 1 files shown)


davidaurelio (approved):
lgtm, I will try the dev deployment

akshaymagapu (approved):

PR #2309, opened 2026-04-06
fix: prompt generation as part of RCV audit flow

dipratap:
## Summary

Wires RCV prompt generation into the prerender audit pipeline. The default audit behaviour
is completely unchanged — prompts are generated only when `generatePrompts=true` is explicitly
passed (opt-in). Also plumbs LLMO site config (categories/topics/regions) through to Mystique
so generated prompts can be correctly classified.

### What changed

**`src/prerender/handler.js` — sending the Mystique request**
- `sendPrerenderGuidanceRequestToMystique` accepts a new `generatePrompts` parameter
  (defaults to `false`) and includes it in the SQS payload
- Each suggestion now carries `hasPrompts: true` when it already has stored prompts, so
  Mystique can skip re-generation
- Reads LLMO site config (`categories`, `topics`, region codes) from
  `site.getConfig().getLlmoConfig()` and sends them as `llmoCategories`, `llmoTopics`,
  `llmoRegions`
- `handleAiOnlyMode` parses `generatePrompts` from the SQS `data` field using the same
  safe JSON parsing pattern already used for `opportunityId` and `scrapeJobId`

**`src/prerender/guidance-handler.js` — receiving Mystique's response**
- Destructures `prompts` from each incoming suggestion alongside `url`, `aiSummary`, `valuable`
- Persists prompts with a preserve-or-update strategy:
  `prompts: hasNewPrompts ? prompts : (currentData.prompts ?? [])`
  New prompts win; existing prompts are kept when Mystique returns an empty array
- Adds `suggestionsWithPrompts` and `totalPromptCount` to the quality-metrics log line


src/prerender/handler.js
```diff
@@ -793,6 +808,20 @@ export async function importTopPages(context) {
...
+
+  // Extract generatePrompts so it can be forwarded to downstream steps via auditContext.
+  // context.data is only populated on the SQS message that triggers Step 1 (the original
+  // Slack command payload) — it is NOT automatically forwarded to Steps 2 and 3.
+  let generatePromptsFlag = false;
+  try {
+    const parsedData = typeof data === 'string' ? JSON.parse(data) : data;
+    generatePromptsFlag = !!parsedData?.generatePrompts;
+  } catch {
+    // Non-JSON data — graceful degradation, flag stays false
```
nit23uec: pls do log atleast a warn in catch - otherwise debugging may be difficult.

src/prerender/handler.js
```diff
@@ -803,9 +832,14 @@ export async function importTopPages(context) {
       ? {
         auditContext: {
           urls: auditContext.urls,
+          generatePrompts: generatePromptsFlag,
         },
       }
-      : {}),
+      : {
+        auditContext: {
```
nit23uec: lets make sure to preserve the existing context and then append new props - less error prone in future when new props are added in context.

src/prerender/handler.js
```diff
@@ -855,6 +889,8 @@ export async function submitForScraping(context) {
         pageLoadTimeout: 20000,
         storagePrefix: AUDIT_TYPE,
       },
+      // Forward generatePrompts so it survives in auditContext for Step 3
+      auditContext: { generatePrompts: !!auditContext?.generatePrompts },
```
nit23uec: lets make sure to preserve the existing context and then append new props - less error prone in future when new props are added in context.

src/prerender/handler.js
```diff
@@ -950,6 +986,8 @@ export async function submitForScraping(context) {
       pageLoadTimeout: 20000,
       storagePrefix: AUDIT_TYPE,
     },
+    // Forward generatePrompts so it survives in auditContext for Step 3
+    auditContext: { generatePrompts: !!auditContext?.generatePrompts },
```
nit23uec: ditto

src/prerender/handler.js
```diff
@@ -793,6 +808,20 @@ export async function importTopPages(context) {
     return handleAiOnlyMode(context);
   }
 
+  // DEBUG: log context.data to confirm what the Slack command delivers at Step 1
+  log.info(`${LOG_PREFIX} [DEBUG] Step 1 context.data type=${typeof data} value=${JSON.stringify(data)} baseUrl=${site.getBaseURL()}`);
```
nit23uec: avoid logging the complete data in logs info level to optimize the volume.

src/prerender/handler.js
```diff
@@ -1535,6 +1573,22 @@ export async function processContentAndGenerateOpportunities(context) {
...
+  // Read generatePrompts from auditContext (forwarded from Step 1 via the step return value)
+  // or fall back to context.data (populated only on the SQS message that triggers each step,
+  // so context.data is empty on Steps 2 and 3 for normal Slack-triggered runs).
+  const generatePromptsFromAuditContext = !!auditContext?.generatePrompts;
+  let generatePromptsFromData = false;
+  try {
+    const parsedData = typeof data === 'string' ? JSON.parse(data) : data;
+    generatePromptsFromData = !!parsedData?.generatePrompts;
+  } catch {
+    // Non-JSON data — graceful degradation, flag stays false
```
nit23uec: does it makes sense to avoid this code duplication?

src/prerender/guidance-handler.js
```diff
@@ -170,10 +172,18 @@ export default async function handler(message, context) {
...
+        suggestionsWithPrompts += 1;
+        totalPromptCount += prompts.length;
+      }
+
       const updatedData = {
         ...currentData,
         // Use new summary if valid; otherwise preserve existing (don't overwrite with empty)
         aiSummary: hasValidAiSummary ? aiSummary : (currentData.aiSummary ?? ''),
+        // Use new prompts if provided; otherwise preserve existing
+        prompts: hasNewPrompts ? prompts : (currentData.prompts ?? []),
```
nit23uec, line 189: how do we take into account the page updates? if content in the page updated to influence the existing prompts, how do the prompts get updated?

src/prerender/guidance-handler.js
```diff
@@ -170,10 +172,18 @@ export default async function handler(message, context) {
         }
       }
 
+      const hasNewPrompts = Array.isArray(prompts) && prompts.length > 0;
```
nit23uec, line 178: we should have some metadata in each prompt that marks it as aiGenerated.

nit23uec (dismissed):
inline comments.

src/prerender/guidance-handler.js
Danialkafeel: Yes it is already attached in mystique when the prompts are generated. 
2 fields are there: `origin: ai` & `source: audit`
https://git.corp.adobe.com/experience-platform/mystique/blob/fb7a2eacb58cd7de011499daee67b538f2c43c72/app/agents/crews/prerender/types.py#L25-L27

src/prerender/handler.js
Danialkafeel: Done

src/prerender/handler.js
Danialkafeel: > lets make sure to preserve the existing context and then append new props - less error prone in future when new props are added in context.

Done at all places except for the step 1 as with this change we will be introducing new data passed to further steps which was not the case before. 
https://github.com/adobe/spacecat-audit-worker/pull/2309#pullrequestreview-4425562933

src/prerender/handler.js
Danialkafeel: Done

src/prerender/handler.js
Danialkafeel: Removed, it was not needed here as in the first audit step we are adding this flag in the audit context & used from there afterwards

src/prerender/guidance-handler.js
Danialkafeel: Valid point, will discuss with @dipratap further on this

anuj-adobe (commented):
## PR Review: Prompt Generation Pipeline (SQS → Mystique → DB)

**Overall: LGTM with a few notes.** The core flow is correct, the regular audit path has no functional regression, and the new logic is properly guarded behind `generatePrompts=false` by default.

---

### ✅ What works well

**KISS/DRY/YAGNI**
- `generatePrompts` is a boolean flag — no new abstractions, no new classes, no new modules.
- `hasPrompts` on each suggestion candidate is a one-liner that avoids a full DB reload in Mystique.
- `siteRegion` forwarding is a single `site.getRegion() ?? ''` call — appropriately minimal.

**Logging**
- All new log calls carry the `Prerender -` prefix ✓
- `siteId=` and `baseUrl=` present in the existing `prerender_ai_summary_metrics` log where `suggestionsWithPrompts` and `totalPromptCount` are appended ✓
- `log.warn` added for JSON parse failures in both `handleAiOnlyMode` and `importTopPages` — makes misconfigured SQS messages debuggable ✓
- `generatePrompts=` logged in the ai-only mode info line ✓

**Opt-in / Slack-only**
- `generatePrompts` defaults to `false` in all three code paths (`handleAiOnlyMode`, `importTopPages`, `processContentAndGenerateOpportunities`). Only a Slack command with `generatePrompts: true` in the payload triggers prompt generation. Regular scheduled runs are unaffected ✓

**Preserve-or-update logic**
- `prompts: hasNewPrompts ? prompts : (currentData.prompts ?? [])` correctly avoids clobbering existing prompts when Mystique returns an empty list or omits the field ✓

---

### ⚠️ Issues to address

**1. Test fixtures use wrong DRS prompt shape (misleading, not a runtime bug)**

All test fixtures in `guidance-handler.test.js` and `ai-only-mode.test.js` use `{ question, category, topic, region }`, e.g.:
```js
prompts: [{ question: 'What is prerendering?', category: 'SEO', topic: 'Performance', region: '' }]
```

The required DRS format is:
```js
{ id: '', origin: 'ai', prompt: '...', source: 'audit', regions: ['US'], type: 'Non-Branded', topic: '', category: '', intent: 'Commercial Investigation' }
```

The production runtime is fine because prompts are stored as-is from Mystique, which generates the correct shape. But a future engineer reading these tests will think the fixture represents a valid stored prompt — it doesn't. Consider updating at least one fixture to use the real DRS shape so the test doubles match what Mystique actually produces.

**2. `importTopPages` now always spreads `auditContext` (behavior change for scheduled runs)**

Before this PR, the return value only included `auditContext` when `auditContext.urls` was non-empty:
```js
// Before
...(Array.isArray(auditContext?.urls) && auditContext.urls.length > 0 ? { auditContext: { urls } } : {})
```

After:
```js
// After — auditContext always present
auditContext: { ...auditContext, ...(urls present ? { urls } : {}), generatePrompts: false }
```

This is correct for forwarding `generatePrompts`, but it also means fields like `slackContext`, `mode`, `domainBlocked`, etc. from the original trigger message are now unconditionally spread into step 2's `auditContext` for ALL runs, where previously they were absent in scheduled (non-Slack, non-CSV) runs. In practice this is benign since step 2 reads these fields defensively (`auditContext?.slackContext`, etc.), but worth a comment in the code explaining WHY the unconditional spread is safe.

**3. Duplicate `generatePrompts` parse logic (minor DRY concern)**

The same `JSON.parse + !!parsedData?.generatePrompts` pattern appears in both `handleAiOnlyMode` and `importTopPages`. Since they're both in the same file and serve the same step-1 dispatching logic, a small shared helper would reduce duplication:
```js
function parseGeneratePrompts(data, log) {
  try {
    const parsed = typeof data === 'string' ? JSON.parse(data) : data;
    return !!parsed?.generatePrompts;
  } catch (e) {
    log.warn(`${LOG_PREFIX} Failed to parse context.data for generatePrompts, defaulting to false: ${e.message}`);
    return false;
  }
}
```
Not blocking, but worth tidying.

**4. `submitForScraping` has `auditContext` forwarding in two places**

```js
// Slack path (line ~947)
auditContext: { ...auditContext, generatePrompts: !!auditContext?.generatePrompts }

// Normal path (line ~1056)  
auditContext: { ...auditContext, generatePrompts: !!auditContext?.generatePrompts }
```

Identical line, two locations. If we ever add another field to forward, we'll need to remember to update both. Could be refactored to build the forwarded context once before the branch. Not blocking.

---

### No regression in regular flow ✓

The regular scheduled audit path:
1. `importTopPages` → `generatePromptsFlag=false` (from JSON parse of absent/empty `data`) → forwarded as `auditContext.generatePrompts=false`
2. `submitForScraping` → forwards `generatePrompts=false`
3. `processContentAndGenerateOpportunities` → `generatePrompts=false` → `sendPrerenderGuidanceRequestToMystique(..., false)`
4. SQS payload to Mystique has `generatePrompts: false` — Mystique skips prompt generation
5. Mystique response may omit `prompts` field → `guidance-handler` preserves `currentData.prompts ?? []`

All four steps behave identically to pre-PR for scheduled runs.

---

**Summary**: Items 1 and 2 are the most important — fix fixture format for at least one test case, and add a comment explaining the unconditional spread. Items 3 and 4 are nice-to-have cleanups.

MysticatBot (dismissed):
Hey @dipratap,

## Strengths

- **Opt-in gate is well-implemented** (`src/prerender/handler.js:583`): `generatePrompts = false` default means the entire prompt-generation path is inert unless explicitly activated. The regular scheduled audit path remains unchanged.
- **Preserve-or-update pattern** (`src/prerender/guidance-handler.js:189`): `hasNewPrompts ? prompts : (currentData.prompts ?? [])` mirrors the existing `aiSummary` preservation logic and avoids data loss when Mystique returns empty results. Clean and consistent.
- **`hasPrompts` per-suggestion** (`src/prerender/handler.js:681`): Letting Mystique skip re-generation for suggestions that already have prompts is the right boundary decision - avoids redundant computation without leaking prompt-generation internals into the audit worker.
- **Metrics added to existing log line** (`src/prerender/guidance-handler.js:214-215`): `suggestionsWithPrompts` and `totalPromptCount` appended to the existing quality-metrics log provides operational visibility without new infrastructure.
- **Thorough test coverage for the preserve-or-update matrix** (`test/audits/prerender/guidance-handler.test.js:1039-1248`): All four states (new prompts, empty array, field omitted, neither side has prompts) are explicitly covered. Edge cases like string coercion (`'true'`) and null region fallback are tested.
- **Refactored function signature** (`src/prerender/handler.js:577-588`): Breaking the long single-line signature into multi-line format and removing the eslint-disable comment is a clean improvement.

## Issues

### Important (Should Fix)

**1. `importTopPages` unconditionally spreads full `auditContext` into the step return value - step-boundary coupling risk**

`src/prerender/handler.js` (the `importTopPages` return block, around line 872-886 of the diff):

Before this PR, `importTopPages` only set `auditContext` when `auditContext.urls` was non-empty, producing a minimal typed shape. Now it spreads the full incoming `auditContext`:

```js
auditContext: {
  ...auditContext,
  ...(Array.isArray(auditContext?.urls) && auditContext.urls.length > 0
    ? { urls: auditContext.urls }
    : {}),
  generatePrompts: generatePromptsFlag,
},
```

For Slack-triggered runs, `auditContext` may contain `slackContext`, `mode`, `domainBlocked`, and other fields from the original trigger. Previously these were NOT forwarded to Step 2. Now they always are. While downstream reads defensively today, this turns the step boundary from "Step 1 outputs a minimal typed shape" into "Step 1 passes through an opaque bag." Any future upstream addition silently leaks across the step boundary, making it impossible to reason about inter-step data flow without reading the full trigger schema.

**How to fix:** Explicitly pick only the fields downstream steps need:
```js
auditContext: {
  ...(Array.isArray(auditContext?.urls) && auditContext.urls.length > 0
    ? { urls: auditContext.urls }
    : {}),
  generatePrompts: generatePromptsFlag,
},
```

The same consideration applies to the two `submitForScraping` sites (`src/prerender/handler.js` lines ~947 and ~1056 of the diff) where `{ ...auditContext, generatePrompts: ... }` passes through the full bag. Consider forwarding only the keys Step 3 actually reads.

**2. No test verifying the SQS payload contains `generatePrompts: true` in the normal audit path (Step 3)**

`src/prerender/handler.js:1783` passes `generatePrompts` to `sendPrerenderGuidanceRequestToMystique` from `processContentAndGenerateOpportunities`. The ai-only-mode tests verify the SQS payload for that path, but the normal Step 3 path (with `auditRunCandidates`) has no equivalent assertion. The new `handler.test.js` tests for `processContentAndGenerateOpportunities` only assert `result.status === 'complete'` without verifying the parsed value was actually forwarded to the SQS message.

**How to fix:** Add a test that stubs `sqs.sendMessage`, runs `processContentAndGenerateOpportunities` with `auditContext.generatePrompts = true` and valid scrape results, and asserts the SQS message includes `generatePrompts: true`. Even asserting on the existing `log.info` call matching `generatePrompts=true` would add verification depth beyond "didn't throw."

### Minor (Nice to Have)

**3. Duplicate JSON parse logic between `handleAiOnlyMode` and `importTopPages`**

`src/prerender/handler.js` (~line 745 and ~line 875): The identical try/catch/warn/`!!parsedData?.generatePrompts` block appears in both functions. The author removed the duplication from `processContentAndGenerateOpportunities` per nit23uec's feedback, but this pair remains. A small shared helper would prevent them from drifting independently:

```js
function parseGeneratePrompts(data, log) {
  try {
    const parsed = typeof data === 'string' ? JSON.parse(data) : data;
    return !!parsed?.generatePrompts;
  } catch (e) {
    log.warn(`${LOG_PREFIX} Failed to parse generatePrompts, defaulting to false: ${e.message}`);
    return false;
  }
}
```

**4. Unvalidated prompts array stored directly from Mystique response**

`src/prerender/guidance-handler.js:189`: The `prompts` array is persisted without a length cap. If Mystique returns an unexpectedly large array, it is stored verbatim. A basic guard (`prompts.slice(0, MAX_PROMPTS)`) would add defense-in-depth.

## Recommendations

- The JSDoc for `sendPrerenderGuidanceRequestToMystique` (`src/prerender/handler.js:580`) has the `@param {boolean} [generatePrompts]` description line placed between the continuation of `preBuiltCandidates` and `@returns`, which makes it read as if the "Each entry is..." description belongs to `generatePrompts`. Consider reordering so each param's multi-line description is self-contained.
- The open question from nit23uec about prompt staleness (how prompts get updated when page content changes) is worth tracking as a follow-up. The current design correctly preserves prompts when `generatePrompts=false`, but there is no mechanism to detect or clear stale prompts when `aiSummary` changes significantly.

## Assessment

Ready to merge? With fixes

The core logic is sound, the opt-in gate works correctly, and test coverage is strong for the new prompt-handling paths. The main concern is the unconditional `...auditContext` spread at step boundaries (finding 1), which introduces implicit coupling between pipeline steps. The test gap for the normal-mode path (finding 2) leaves an assertion hole in the most common production flow. Addressing these two items would make the change solid.

## Next Steps

1. Tighten the `auditContext` spread in `importTopPages` and `submitForScraping` to forward only needed fields.
2. Add an assertion on the SQS payload in the normal Step 3 path to prove `generatePrompts` is forwarded end-to-end.
3. Minor items are optional but would improve maintainability.


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 38s | <b>Cost:</b> $3.52 | <b>Commit:</b> <code>4e84fa8e04bcbf74c997dab66b87c9421487adae</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/prerender/guidance-handler.js
Danialkafeel: Created JIRA to track for further: [LLMO-5312](https://jira.corp.adobe.com/browse/LLMO-5312)

MysticatBot (commented):
Hey @dipratap,

## Strengths

- **Previously flagged SQS payload test gap now closed** - The new `should forward generatePrompts:true to Mystique SQS message in normal audit path` test (`test/audits/prerender/handler.test.js`) with esmock wiring verifies the flag survives the full Step 1 -> Step 2 -> Step 3 chain. This was the main Important finding from the prior review.
- **Code duplication in `processContentAndGenerateOpportunities` resolved** - Step 3 now reads `generatePrompts` directly from `auditContext` (`src/prerender/handler.js:1641`) instead of re-parsing `context.data`. Clean and correct.
- **Reviewer feedback fully incorporated** - warn logging in catch blocks (per nit23uec), DEBUG log removed, existing context preserved in `submitForScraping` spreads. All prior reviewer requests addressed.
- **Preserve-or-update logic** (`src/prerender/guidance-handler.js:189`) mirrors the existing `aiSummary` pattern exactly. The `hasNewPrompts ? prompts : (currentData.prompts ?? [])` idiom prevents data loss on empty Mystique responses without diverging from established patterns.
- **`hasPrompts` per-suggestion boundary decision** (`src/prerender/handler.js:681-682`) - passing a signal to Mystique rather than implementing skip-logic locally keeps the audit worker's responsibility boundary clean.
- **Opt-in gate is architecturally sound** - `generatePrompts = false` as a default parameter means the entire new capability is inert unless explicitly activated. The existing pipeline semantics are unchanged for all current callers.
- **Thorough preserve-or-update test matrix** (`test/audits/prerender/guidance-handler.test.js:1037-1248`) covers all four permutations: new prompts present, empty array from Mystique, field omitted, and neither side has prompts.

## Issues

### Minor (Nice to Have)

**1. Remaining duplicate parse logic between `handleAiOnlyMode` and `importTopPages`**

`src/prerender/handler.js` (~line 745 and ~line 878): The identical try/catch/warn/`!!parsedData?.generatePrompts` pattern appears in both Step 1 entry points. The duplication in `processContentAndGenerateOpportunities` was correctly cleaned up, but this pair remains. A small helper would prevent them from drifting independently:

```js
function parseGeneratePromptsFromData(data, log) {
  try {
    const parsed = typeof data === 'string' ? JSON.parse(data) : data;
    return !!parsed?.generatePrompts;
  } catch (e) {
    log.warn(`${LOG_PREFIX} Failed to parse context.data for generatePrompts, defaulting to false: ${e.message}`);
    return false;
  }
}
```

**2. Unbounded `prompts` array stored without size cap**

`src/prerender/guidance-handler.js:189`: The `prompts` array from Mystique is stored verbatim. While Mystique is an internal service and the SQS message has a 256KB limit, a defense-in-depth cap (e.g., `prompts.slice(0, 50)`) would prevent edge cases where a buggy Mystique response produces a large array that hits DynamoDB item size limits.

**3. Two `processContentAndGenerateOpportunities` tests assert only `result.status` without behavioral verification**

`test/audits/prerender/handler.test.js` (`should parse generatePrompts:true from JSON string data` and `should read generatePrompts from auditContext when data is absent`): These tests set up `generatePrompts: true` but only assert `result.status === 'complete'`. Since the function returns `'complete'` regardless of the flag value, these prove "didn't throw" rather than "the flag was parsed correctly." The dedicated esmock test already covers the end-to-end SQS assertion, so these add limited incremental value. Consider adding a `log.info` assertion matching `generatePrompts=true` to give them a behavioral check.

**4. `submitForScraping` redundantly re-asserts `generatePrompts` after spreading**

`src/prerender/handler.js` (~line 946 and ~1054):
```js
auditContext: { ...auditContext, generatePrompts: !!auditContext?.generatePrompts },
```

The spread already includes `generatePrompts` (set in `importTopPages`). The trailing re-assertion only coerces to boolean what is already boolean from Step 1. Since nit23uec explicitly requested this pattern for forward-compatibility, this is not blocking - just noting the redundancy.

## Recommendations

- Consider defining an explicit `AuditContext` JSDoc `@typedef` that lists the fields each step is expected to consume and produce. The pipeline is now four steps deep and three of them read/write `auditContext`. An explicit shape makes inter-step data flow reasoning possible without reading every step's implementation.
- The JSDoc for `sendPrerenderGuidanceRequestToMystique` (`src/prerender/handler.js:582`) places `@param {boolean} [generatePrompts]` after the multi-line description of `preBuiltCandidates`, making it read as if the "Each entry is..." continuation belongs to `generatePrompts`. Consider reordering.

## Assessment

The core logic is sound, the opt-in gate works correctly, and the prior review's main findings have been addressed. All five review perspectives agree on merge-readiness. The remaining items are incremental improvements (helper extraction, defense-in-depth cap, test strengthening) - none represent correctness risks or regressions.

Note: CI checks are currently pending - ensure the build passes before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 4s | <b>Cost:</b> $3.40 | <b>Commit:</b> <code>e63d0383aec70a2e08742b359808e0599139bf6a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/prerender/guidance-handler.js
shrotia: Valid point — we'll defer prompt generation to experimentation launch time in future iterations. Since this doesn't impact the current happy path (prompt generation audit can be re-triggered before experimentation kicks off), let's proceed with merging the PR imho

shrotia (approved):

shrotia:
LGTM on the current agreed-upon flow — let's merge and start using for happy-path deployments with manual QA gates. Going forward, we need Mystique-generated prompts to pass quality gates:

Staleness detection — flag & re-generate if the page has changed significantly since the last prompt was generated (per @nit23uec's point)
Confidence scoring — currently getting 4 prompts per URL with no way to rank them; need a score to surface the best candidate out of them
Generation metadata — timestamp (date/time) of when each prompt was generated

PR #2319, opened 2026-04-06
fix: prevent job cancellation on transient database errors during entitlement checks

bhellema:
Preflight jobs were incorrectly cancelled when database connection pool
  timeouts (PGRST003) occurred during entitlement validation. The error
  handling treated all TierClient failures as "site not entitled" and
  returned false, causing AsyncJobRunner to set job status to CANCELLED
  instead of allowing Lambda/SQS to retry.

  Root cause: checkProductCodeEntitlements() caught all errors and
  returned false, including transient infrastructure failures like:
  - PGRST000-003: PostgREST connection/timeout errors (503/504)
  - Network errors: ECONNREFUSED, ETIMEDOUT, ENOTFOUND
  - HTTP server errors: 429, 500, 502, 503, 504

  Solution: Created isTransientTierClientError() classifier to distinguish
  transient errors (should retry) from permanent errors (not entitled).
  Transient errors now throw to trigger Lambda/SQS retry mechanism instead
  of incorrectly cancelling jobs.

  Changes:
  - Added tier-client-error-classifier.js with error classification logic
  - Updated checkProductCodeEntitlements() to rethrow transient errors
  - Updated checkSiteRequiresValidation() to rethrow transient errors
  - Updated isPaidLLMOCustomer() to rethrow transient errors
  - Added comprehensive test coverage for all error scenarios
  - Documented scope analysis confirming fix addresses all vulnerable areas

### DIFF
--- a/docs/plans/fix-database-timeout-entitlement-check.md
+++ b/docs/plans/fix-database-timeout-entitlement-check.md
@@ -1,0 +1,277 @@
+# Fix Database Timeout Issue in Entitlement Checking
+
+## Context
+
+Preflight job `d272041d-f014-4792-a1a1-944e5c0c57ab` was incorrectly **cancelled** on March 31, 2026 due to a database connection pool timeout (`PGRST003`). The entitlement check in `checkProductCodeEntitlements()` caught the transient infrastructure error and returned `false`, causing the system to treat it as "site not entitled" and set the job status to `CANCELLED`.
+
+The problem: **All errors are treated equally** - transient infrastructure failures (database timeouts, network issues) are handled the same as legitimate "site not entitled" scenarios. This causes jobs to be cancelled instead of allowing Lambda/SQS to retry.
+
+### Error Flow from Logs
+1. Database: `PGRST003: Timed out acquiring connection from connection pool`
+2. TierClient throws error → caught at `/src/common/audit-utils.js:38-40`
+3. Returns `false` → `isAuditEnabledForSite()` returns `false`
+4. AsyncJobRunner sets `AsyncJob.Status.CANCELLED` (line 91)
+5. Job skipped with reason: "preflight audits disabled for site"
+
+### Expected Behavior
+- **Transient errors** (database timeouts, network failures) → Throw error, let Lambda retry
+- **Permanent errors** (not entitled, not found) → Return `false`, cancel job gracefully
+
+## Solution Approach
+
+**Strategy**: Distinguish transient from permanent errors and **rethrow transient errors** to trigger Lambda/SQS retry mechanism (no local retry logic).
+
+**Why no local retry?** 
+- Lambda/SQS provides automatic retries with backoff and DLQ
+- Local retry adds complexity and consumes Lambda timeout budget
+- Matches existing pattern in codebase (internal-links lets "SQS retry complete finalization")
+
+## Implementation Plan
+
+### 1. Create Error Classifier Utility
+
+**New file**: `/src/common/tier-client-error-classifier.js`
+
+Create `isTransientTierClientError(error)` function to classify errors:
+
+**Transient patterns** (should retry):
+- Database: `PGRST000`, `PGRST001`, `PGRST002`, `PGRST003`, `connection pool`, `timeout`, `ECONNREFUSED`, `ETIMEDOUT`
+  - PGRST000: Could not connect with database (503)
+  - PGRST001: Could not connect due to internal error (503)
+  - PGRST002: Could not connect when building schema cache (503)
+  - PGRST003: Timed out acquiring connection from connection pool (504)
+- Network: `ENOTFOUND`, `ECONNRESET`, `EAI_AGAIN`, `socket hang up`, `network error`
+- HTTP: 408, 429, 500, 502, 503, 504 (if exposed)
+- Generic: `temporary failure`, `service unavailable`
+
+**Permanent patterns** (skip audit):
+- HTTP: 401, 403, 404
+- PostgREST: `PGRST100-122` (bad request), `PGRST200-204` (not found), `PGRST300` (config error)
+- Business: `not enrolled`, `no entitlement`, `invalid product code`
+
+Implementation:
+```javascript
+export function isTransientTierClientError(error) {
+  const message = error?.message?.toLowerCase() || '';
+  const code = error?.code?.toUpperCase() || '';
+  const statusCode = error?.statusCode || error?.status;
+  
+  // Database errors - PostgREST connection and timeout issues
+  const transientPostgrestCodes = ['PGRST000', 'PGRST001', 'PGRST002', 'PGRST003'];
+  if (transientPostgrestCodes.includes(code)) return true;
+  
+  if (message.includes('connection pool') || message.includes('timed out')) return true;
+  
+  // Network errors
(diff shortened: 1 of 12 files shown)


ekremney (commented):
## Review

### Critical: Classifier won't match real PGRST003 errors

I traced the error propagation end-to-end through the data-access layer and TierClient.

**PostgREST returns 504** → `@supabase/postgrest-js` returns `{ error: { code: 'PGRST003', message: 'Timed out...' }, data: null }` → `base.collection.js#logAndThrowError('Failed to query', pgrstError)` → throws `DataAccessError`

The `DataAccessError` that reaches the catch blocks looks like:

```javascript
{
  name: 'DataAccessError',
  message: 'Failed to query',              // ← generic wrapper
  details: { entityName: 'Entitlement', tableName: 'entitlements' },
  cause: {                                  // ← the real info is HERE
    code: 'PGRST003',
    message: 'Timed out acquiring connection from connection pool',
    details: '...',
    hint: '...'
  }
}
```

TierClient re-throws this unchanged (`tier-client.js:127`).

The classifier checks:
- `error.code` → **undefined** (DataAccessError has no `.code`)
- `error.statusCode` / `error.status` → **undefined**
- `error.message` → **`'Failed to query'`** — does NOT contain `'connection pool'`, `'timed out'`, etc.

**None of the transient checks match.** The classifier returns `false`, and the bug remains unfixed. The PGRST code and meaningful message live at `error.cause.code` and `error.cause.message`, but the classifier never looks there.

The data-access layer itself already has the correct pattern — `#isInvalidInputError` in `base.collection.js` walks the `.cause` chain:

```javascript
#isInvalidInputError(error) {
  let current = error;
  while (current) {
    if (current?.code === '22P02') return true;
    current = current.cause;
  }
  return false;
}
```

The classifier needs to do the same.

---

### Critical: Tests use wrong error shapes

All tests construct errors like:
```javascript
const dbError = new Error('Timed out acquiring connection from connection pool');
dbError.code = 'PGRST003';
```

But real errors are `DataAccessError` where `.code` is undefined and `.message` is `'Failed to query'`. The tests pass because they use fabricated error shapes that don't match production, masking the classifier bug above.

---

### Medium: `checkSiteRequiresValidation` throw is silently caught

In `src/index.js:296-310`:
```javascript
try {
  const site = await Site.findById(siteId);
  if (site) {
    const requiresValidation = await checkSiteRequiresValidation(site, context, type);
    site.requiresValidation = requiresValidation;
    context.site = site;
  }
} catch (e) {
  if (!siteId.startsWith('warmup-site-')) {
    log.warn(`Failed to fetch site ${siteId}: ${e.message}`);
  }
}
```

If `checkSiteRequiresValidation` throws (per the new change), the catch here **swallows it**. No Lambda retry happens. The audit handler still runs, but now `context.site` is never assigned (the throw happens before line 303-304). This is a behavioral change — previously the site was always set on context; now on transient errors it won't be. Downstream impact depends on what handlers do when `context.site` is undefined.

---

### Medium: Outer catch in `checkProductCodeEntitlements` semantics changed

```javascript
} catch (error) {
  // Old: return false (cancel job)
  // New: throw error (retry)
  context.log.error('Transient error in entitlement check, job will retry:', error);
  throw error;
}
```

The comment says *"a transient error was thrown from inner loop"* but any error landing here is now rethrown — including non-transient unexpected errors (e.g., from `Promise.all` itself). The old code returned `false` for all of these (cancel). The new code retries all of them. This is probably safer in practice, but the comment is misleading.

---

### Low: HTTP 500 classified as transient

500 from PostgREST could indicate a bug (bad query, schema mismatch), not just a transient outage. Retrying until DLQ is low risk for simple TierClient queries but worth noting.

---

### Infra context

The PostgREST pool is configured as:
- `PGRST_DB_POOL=100` per ECS task
- `PGRST_DB_POOL_ACQUISITION_TIMEOUT=10` seconds
- 10 prod ECS tasks = 1000 total PostgREST connections
- Auto-scaling is **disabled** because it *"churns connection pools and causes sporadic fetch failed errors"*

The 10-second acquisition timeout with auto-scaling disabled means PGRST003 is a real and recurring scenario under load, making it important to get this fix right.

---

### Summary

| Severity | Finding |
|----------|---------|
| **Critical** | Classifier checks `error.code`/`.message` but real errors are `DataAccessError` with info nested in `.cause` — the fix doesn't fix the bug |
| **Critical** | Tests use fabricated error shapes that don't match production, masking the classifier bug |
| **Medium** | `checkSiteRequiresValidation` throws are caught by caller in `index.js` — no retry, and `context.site` not set |
| **Medium** | Outer catch now rethrows ALL errors, not just transient ones |
| **Low** | HTTP 500 as transient |

The classifier needs to walk the `error.cause` chain to find PGRST codes and messages, and tests should use `DataAccessError` with the real nested error shape.

ekremney (approved):
All previous issues have been addressed. A few minor observations:

### ✅ Fixed: Classifier now walks the `.cause` chain

The `while (current) { ... current = current.cause; }` loop correctly reaches the PostgREST error nested inside `DataAccessError`. The original PGRST003 scenario will now be caught.

### ✅ Fixed: Tests use `DataAccessError` with realistic error shapes

Tests now import `DataAccessError` from `@adobe/spacecat-shared-data-access` and construct errors like:
```js
const pgrstError = { code: 'PGRST003', message: 'Timed out acquiring connection from connection pool' };
const dbError = new DataAccessError('Failed to query', { entityName: 'Entitlement' }, pgrstError);
```
This matches the production error shape.

### ✅ Fixed: `index.js` now rethrows transient errors

Transient errors during site validation are now rethrown to trigger Lambda retry instead of being silently swallowed.

---

### Minor: Two remaining test inconsistencies

**1.** `test/common/index-site-validation.test.js` constructs its transient error as a plain `Error` with `.statusCode = 503` directly on it, rather than a `DataAccessError`:
```js
const transientError = new Error('tier service unavailable');
transientError.statusCode = 503;
```
In production the error would be a `DataAccessError` wrapping a PostgREST error. The test works because `isTransientTierClientError` at level 0 finds `.statusCode = 503`, but it doesn't exercise the cause-chain walking that the fix actually relies on.

**2.** The "throws immediately when transient error occurs even with multiple product codes" test in `audit-utils.test.js` uses `new Error('Connection pool exhausted')` — matches via `message.includes('connection pool')` at level 0. Same issue: works, but exercises message-matching rather than the cause-chain path.

Both tests would be more rigorous wrapped in a `DataAccessError`.

---

### Minor: Undocumented scope expansion in `index.js`

The catch block now applies `isTransientTierClientError` to all errors in the `try` block, which includes `Site.findById(siteId)` — not just `checkSiteRequiresValidation`. Transient DB errors from `Site.findById` were previously swallowed; now they trigger retry too. This is actually the right behavior (if the DB is timing out, the whole Lambda should retry), but the PR description and commit message don't mention it.

---

### Minor: Constant arrays recreated per iteration

Inside the `while` loop, `transientPostgrestCodes`, `transientNetworkCodes`, and `transientStatusCodes` are re-declared on every iteration. Hoisting them outside the function would be a micro-optimization, but not impactful for a function that typically exits on the first or second iteration.

bhellema:
PR is too old, if the issue still is around it should be revisted.

PR #2358, opened 2026-04-09
feat: byocdn-imperva cdn audits

ctinp:
byocdn-imperva CDN support wires Imperva logs into the existing CDN analysis pipeline. Three things make it different from other providers:

- S3 path layout — raw logs land at {pathId}_raw_byocdn-imperva/ (underscores, not slashes). buildCdnPaths emits this format; getBucketInfo runs a second ListObjectsV2 scan with Prefix: "{pathId}_raw_" to discover it alongside the normal slash-layout providers.

- Daily-only processing — Imperva SQL filters by date only (no hourly partition), so it joins Cloudflare/Other in hasDailyPartitioningOnly and is skipped at every hour except 23.

- Flat raw location — files land directly at the prefix with no date/hour subdirs, so rawDataPath = paths.rawLocation with nothing appended.

Everything else (SQL execution, aggregation, Athena table management) flows through the existing pipeline unchanged via CDN_TYPES.IMPERVA → sql/imperva/.

### DIFF
--- a/src/cdn-analysis/sql/imperva/insert-aggregated-referral.sql
+++ b/src/cdn-analysis/sql/imperva/insert-aggregated-referral.sql
@@ -1,0 +1,125 @@
+INSERT INTO {{database}}.{{aggregatedTable}}
+WITH hosts AS (
+  -- first, identify the hosts from the cdn logs so that self-referrals can be filtered out later on
+  SELECT DISTINCT s_computername AS host
+  FROM {{database}}.{{rawTable}}
+  WHERE date = '{{year}}-{{month}}-{{day}}'
+),
+
+base AS (
+  -- normalize key fields and construct a full URL from path + query
+  SELECT
+    CASE
+      WHEN cs_uri_query IS NOT NULL AND cs_uri_query <> ''
+        THEN CONCAT(cs_uri, '?', cs_uri_query)
+      ELSE cs_uri
+    END AS url,
+    s_computername AS host,
+    cs_referrer AS referer_raw,
+    cs_user_agent AS user_agent
+  FROM {{database}}.{{rawTable}}
+  WHERE date = '{{year}}-{{month}}-{{day}}'
+),
+
+referrals_raw AS (
+  SELECT
+    url,
+    host,
+    try(url_extract_host(referer_raw)) AS referrer,
+    url_extract_parameter(url, 'utm_source') AS utm_source,
+    url_extract_parameter(url, 'utm_medium') AS utm_medium,
+
+    -- only the tracking_param, not the value (no PII)
+    -- for the tracking_param list, please refer to https://github.com/adobe/helix-rum-enhancer/blob/main/plugins/martech.js#L13-L22
+    -- list is limited to have feature parity with rum-enhancer
+    CASE
+      WHEN REGEXP_LIKE(
+        url,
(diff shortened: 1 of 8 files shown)


src/cdn-analysis/sql/imperva/insert-aggregated.sql
```diff
@@ -0,0 +1,44 @@
...
+
+WHERE 
+  date = '{{year}}-{{month}}-{{day}}'
+
+  -- match known LLM-related user-agents
+  AND REGEXP_LIKE(cs_user_agent, '(?i)(ChatGPT|GPTBot|OAI-SearchBot|Perplexity|Claude|Anthropic|Gemini|Copilot|MistralAI-User|Google-NotebookLM|GoogleAgent|Google-Extended|Googlebot|bingbot|Amzn-User|^Google$)')
+
+  -- only count HTML/PDF/Markdown responses, plus .md paths, robots.txt and sitemaps
+  AND (
+    REGEXP_LIKE(lower(url_extract_path(cs_uri)), '(?i)((\.html|\.pdf|\.md)(\?.*)?)$')
```
akshaymagapu: Add a negative-match fallback to include URLs that don't match known static extensions, or confirm this is intentional.

like in frontdoor or other csse

src/cdn-analysis/sql/imperva/insert-aggregated-referral.sql
```diff
@@ -0,0 +1,125 @@
...
+  url_extract_path(url) as url,
+  host,
+  referrer,
+  utm_source,
+  utm_medium,
+  tracking_param,
+  device,
+  date,
+  cdn_provider,
+  COALESCE(host, '') as x_forwarded_host,
```
akshaymagapu, line 115: for both agentic and referral, could we do same empty or use host in both places?

src/cdn-analysis/sql/imperva/insert-aggregated.sql
akshaymagapu: `  AND (
      NOT REGEXP_LIKE(url_extract_path(cs_uri), '(?i)\.(css|js|png|jpg|jpeg|gif|webp|php|svg|ico|woff|woff2|otf|ttf|eot|mp4|mp3|avi|mov|zip|tar|gz|json|xml|txt)(\?.*)?$')
      OR REGEXP_LIKE(url_extract_path(cs_uri), '(?i)(\.htm|\.pdf|robots\.txt|sitemap)')
  )`

ctinp: Good catch, I was really wondering how to improve the robustness of the content-type detection since the response headers are not available in Imperva (or at least I did not see any so far)

src/cdn-analysis/sql/imperva/insert-aggregated.sql
```diff
@@ -8,7 +8,7 @@ SELECT
   0.0 AS time_to_first_byte,
   COUNT(*) AS count,
   '{{serviceProvider}}' AS cdn_provider,
-  '' AS x_forwarded_host,
+  COALESCE(s_computername, '') as x_forwarded_host,
```
akshaymagapu, line 11: here 
s_computername or host ?

akshaymagapu (approved):

PR #2363, opened 2026-04-10
fix(cdn-logs-report): add dispatch log correlation context

akshaymagapu:
## Summary
- add `batchId` to the daily agentic export dispatch log so it matches the outbound analytics `correlationId`
- include `site.getBaseURL()` in the same log line for easier log filtering and correlation
- assert the updated log output in the focused agentic daily export test

## Testing
- `npx mocha --require test/setup-env.js test/audits/cdn-logs-report/agentic-daily-export.test.js`


davidaurelio (approved):
lgtm

PR #2343, opened 2026-04-08
feat: enhance the Sitemap audit to proportionately probe across the s…

mikeweis:
…itemap.xml files

 * net result: better results at a faster pace

Please ensure your pull request adheres to the following guidelines:
- [ ] make sure to link the related issues in this description
- [ ] when merging / squashing, make sure the fixed issue references are visible in the commits, for easy compilation of release notes
- [ ] If data sources for any opportunity has been updated/added, please update the [wiki](https://wiki.corp.adobe.com/display/AEMSites/Data+Sources+for+Opportunities) for same opportunity.

## Related Issues

Thanks for contributing!

PR #2364, opened 2026-04-10
feat: enhance the Sitemap audit to be able to slow down its probing rate

mikeweis:
* if needed, dynamically slow down the rate of probing the page URLs
 * if slowed down, dynamically downsize the number of pages probed

## Related Issues

Thanks for contributing!


mikeweis:
Related links:
 * ticket: https://jira.corp.adobe.com/browse/SITES-39177
 * PR with proportioned sampling: https://github.com/adobe/spacecat-audit-worker/pull/2343

src/sitemap/common.js
```diff
@@ -64,6 +77,22 @@ function delay(ms) {
...
 }
 
+/**
+ * After {@link applyPageUrlProbeSampling}, further cap how many page URLs we probe per sitemap
+ * while in slow mode: floor(length * SLOW_MAX / FAST_MAX), at least 1 when non-empty.
+ *
+ * @param {string[]} urls
+ * @returns {string[]}
+ */
+export function slicePageUrlsForSlowProbeSampling(urls) {
```
jsanchez07, line 87: If urls is null or undefined, this returns it as-is. Any caller that passes the result straight into filterValidUrls would hit Cannot read property 'length' of null at common.js:93. In practice it's protected by the outer if                
  (urlsFromSampling?.length) guard in handler.js, but the function's contract is inconsistent — should return [] for all falsy-length inputs.

src/sitemap/common.js
```diff
@@ -544,7 +585,10 @@ export async function getSitemapUrls(inputUrl, log) {
       `${protocol}://${domain}/sitemap.xml`,
       `${protocol}://${domain}/sitemap_index.xml`,
     ];
-    sitemapUrls = await filterValidUrls(commonSitemapUrls, log, SITEMAP_GET_TIMEOUT_MS);
+    sitemapUrls = await filterValidUrls(commonSitemapUrls, log, SITEMAP_GET_TIMEOUT_MS, {
+      pageUrlBatchSize: SITEMAP_XML_BATCH_SIZE,
```
jsanchez07, line 590: SITEMAP_XML_BATCH_SIZE and SITEMAP_XML_BATCH_DELAY_MS are meant for fetching sitemap XML files, not for validating the 2–3 common sitemap URLs (/sitemap.xml, /sitemap_index.xml). This is functionally harmless (2–3 URLs never hit the batch  
  boundary), but semantically wrong — it would be cleaner to pass null here and let the fast defaults apply, or use dedicated constants.

jsanchez07 (changes requested):

src/sitemap/common.js
mikeweis: 👍  Excellent point!  Will fix.

src/sitemap/common.js
mikeweis: - the 2 files being checked really are (potential) sitemap.xml files, hence why we use the "sitemap" constants
- nonetheless, I added a comment on why we are using `filterValidUrls` since that function is typically used on the page URLs (as opposed to any sitemap.xml files)

jsanchez07 (approved):

PR #2366, opened 2026-04-10
feat: filter moneyPages URLs when enableMoneyPageUrls flag is false

nitinja:
## Summary

- Reads `enableMoneyPageUrls` from site config in `getAuditTargetUrls()`
- When the flag is explicitly `false`, moneyPages-sourced entries are excluded from the returned URL list
- When the flag is `true` or absent (the default), all sources are included unchanged
- Also replaces a raw `JSON.stringify` with the existing `safeStringify` utility for the top-pages log line

## Notes

- The `=== false` check correctly handles three states: `undefined` (absent → enabled), `true` (enabled), `false` (disabled)
- Depends on `isMoneyPageUrlsEnabled()` being available on the site config object (added in spacecat-shared-data-access)

## Test plan

- [x] Returns all URLs when `enableMoneyPageUrls` is `true`
- [x] Returns all URLs when flag is absent (default enabled)
- [x] Filters out moneyPages URLs when `enableMoneyPageUrls` is `false`
- [x] Always returns manual-source URLs regardless of flag value


slitviachenko (commented):
Suggestion on the filtering approach: consider flipping from **exclusion-based** (filter out `moneyPages` when disabled) to **inclusion-based** (explicitly opt-in each source). Right now:

```javascript
const filtered = config?.isMoneyPageUrlsEnabled?.() === false
  ? entries.filter(({ source }) => source !== MONEY_PAGES_SOURCE)
  : entries;
```

This means any new source added in data-access (e.g. `informational` pages in the future) would be **included by default** in audit scope, even before the audit worker explicitly handles it.

An inclusion-based approach would be safer:

```javascript
const MANUAL_SOURCE = 'manual';

const isSourceIncluded = (source, config) => {
  if (source === MANUAL_SOURCE) return true;
  if (source === MONEY_PAGES_SOURCE) return config?.isMoneyPageUrlsEnabled?.() !== false;
  return false;
};

const filtered = entries.filter(({ source }) => isSourceIncluded(source, config));
```

Benefits:
- **Deny-by-default** for non-manual sources — unknown sources silently drop rather than leaking into audit scope. This is a desirable state: it forces a coordinated change in the audit worker whenever a new source is introduced in data-access, rather than letting unknown data slip through unintentionally.
- **Self-documenting** — `isSourceIncluded` acts as a manifest of "what sources feed into audits and under what conditions"
- **Easy to extend** — when a new source like `informational` comes along, you add one line + its config accessor

nitinja:
> Suggestion on the filtering approach: consider flipping from **exclusion-based** (filter out `moneyPages` when disabled) to **inclusion-based** (explicitly opt-in each source). Right now:
> 
> ```js
> const filtered = config?.isMoneyPageUrlsEnabled?.() === false
>   ? entries.filter(({ source }) => source !== MONEY_PAGES_SOURCE)
>   : entries;
> ```
> 
> This means any new source added in data-access (e.g. `informational` pages in the future) would be **included by default** in audit scope, even before the audit worker explicitly handles it.
> 
> An inclusion-based approach would be safer:
> 
> ```js
> const MANUAL_SOURCE = 'manual';
> 
> const isSourceIncluded = (source, config) => {
>   if (source === MANUAL_SOURCE) return true;
>   if (source === MONEY_PAGES_SOURCE) return config?.isMoneyPageUrlsEnabled?.() !== false;
>   return false;
> };
> 
> const filtered = entries.filter(({ source }) => isSourceIncluded(source, config));
> ```
> 
> Benefits:
> 
> * **Deny-by-default** for non-manual sources — unknown sources silently drop rather than leaking into audit scope. This is a desirable state: it forces a coordinated change in the audit worker whenever a new source is introduced in data-access, rather than letting unknown data slip through unintentionally.
> * **Self-documenting** — `isSourceIncluded` acts as a manifest of "what sources feed into audits and under what conditions"
> * **Easy to extend** — when a new source like `informational` comes along, you add one line + its config accessor

Implemented, thanks

slitviachenko (approved):

PR #2369, opened 2026-04-11
fix(prerender): keep valuable flag in sync with aiSummary in guidance handler

anuj-adobe:
## Problem

In `guidance-handler.js`, when updating suggestion data from Mystique AI responses:

- `aiSummary` is conditionally updated — only when the new value is valid (not null, undefined, or "Not available"). Otherwise the existing value is preserved.
- `valuable` was **always** overwritten with the new value, regardless of whether `aiSummary` was valid.

This meant: if a suggestion had `valuable: false` and a valid `aiSummary` from a previous run, and the next Mystique response had no valid summary (`aiSummary: null`), the `valuable` flag would be overwritten with `true` (the default) while `aiSummary` was correctly preserved. The two fields fell out of sync.

## Fix

```js
// Before
valuable: isValuable,  // Always overwrites

// After
valuable: hasValidAiSummary ? isValuable : (currentData.valuable ?? true),
```

Now `valuable` follows the same pattern as `aiSummary` — only updates when the new AI response has a valid summary.

## Test plan

- [x] New test: preserves existing `valuable: false` when `aiSummary` is null
- [x] Existing tests pass (29 passing, 0 failing)
- [x] 100% coverage on guidance-handler.js

### DIFF
--- a/src/prerender/guidance-handler.js
+++ b/src/prerender/guidance-handler.js
@@ -175,8 +175,8 @@ export default async function handler(message, context) {
         ...currentData,
         // Use new summary if valid; otherwise preserve existing (don't overwrite with empty)
         aiSummary: hasValidAiSummary ? aiSummary : (currentData.aiSummary ?? ''),
-        // Default to true if not provided, but respect explicit boolean from Mystique
-        valuable: isValuable,
+        // Keep valuable in sync with aiSummary — only update when new AI response is valid
+        valuable: hasValidAiSummary ? isValuable : (currentData.valuable ?? true),
       };
 
       warnOnInvalidSuggestionData(updatedData, opportunity.getType(), log);
--- a/test/audits/prerender/guidance-handler.test.js
+++ b/test/audits/prerender/guidance-handler.test.js
@@ -754,6 +754,47 @@ describe('Prerender Guidance Handler (Presigned URL)', () => {
       );
     });
 
+    it('should preserve existing valuable flag when aiSummary is invalid', async () => {
+      // Suggestion already has valuable=false and a valid aiSummary from a previous run
+      mockSuggestions[0].getData.returns({
+        url: 'https://example.com/page1',
+        isDomainWide: false,
+        aiSummary: 'Previous valid summary',
+        valuable: false,
+      });
+
+      mockFetchSuccess({
+        opportunityId: 'opportunity-123',
+        suggestions: [
+          {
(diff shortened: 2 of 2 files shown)


ssilare-adobe (approved):
The fix is correct and well-reasoned. Traced through all input combinations — the core logic is sound and all targeted bug cases are properly handled.

One pre-existing edge case worth a follow-up (not a blocker): when Mystique returns a valid `aiSummary` but omits `valuable` entirely, `isValuable` defaults to `true` and overwrites an existing `false`. The fix doesn't make this worse, but a more defensive variant could preserve the existing value in that scenario too:

```js
valuable: hasValidAiSummary
  ? (typeof valuable === 'boolean' ? valuable : (currentData.valuable ?? true))
  : (currentData.valuable ?? true)
```

Approving — the stated bug is fixed, tests are comprehensive, and coverage is at 100%.

ssilare-adobe (approved):
The fix is correct and well-reasoned. Traced through all input combinations — the core logic is sound and all targeted bug cases are properly handled.

One pre-existing edge case worth a follow-up (not a blocker): when Mystique returns a valid `aiSummary` but omits `valuable` entirely, `isValuable` defaults to `true` and overwrites an existing `false`. The fix does not make this worse, but a more defensive variant could preserve the existing value in that scenario too:

```js
valuable: hasValidAiSummary
  ? (typeof valuable === 'boolean' ? valuable : (currentData.valuable ?? true))
  : (currentData.valuable ?? true)
```

Approving — the stated bug is fixed, tests are comprehensive, and coverage is at 100%.

ssilare-adobe (approved):
The fix is correct and well-reasoned. Traced through all input combinations — the core logic is sound and all targeted bug cases are properly handled.

One pre-existing edge case worth a follow-up (not a blocker): when Mystique returns a valid aiSummary but omits valuable entirely, isValuable defaults to true and overwrites an existing false. The fix does not make this worse, but a more defensive variant could preserve the existing value in that scenario too.

Approving — the stated bug is fixed, tests are comprehensive, and coverage is at 100%.

ssilare-adobe (approved):
The fix is correct and well-reasoned. Traced through all input combinations — the core logic is sound and all targeted bug cases are properly handled. Approving — the stated bug is fixed, tests are comprehensive, and coverage is at 100%.

ssilare-adobe (approved):
Approving. Fix is correct and well-tested.

PR #2370, opened 2026-04-11
fix(prerender): increase Mystique batch size to 320 and include suggestionId in guidance message

ssilare-adobe:
## Summary

- **Batch size**: Increase `MYSTIQUE_BATCH_SIZE` from 100 to 320 in `src/prerender/utils/constants.js` to align with the intended throughput for the prerender guidance flow.
- **Missing `suggestionId`**: The normal-run path in `processOpportunityAndSuggestions` was building `auditRunCandidates` without `suggestionId`. Mystique requires this field to correlate AI summaries back to the correct suggestion. Fixed by fetching saved suggestions after `syncSuggestions` and mapping each URL to its DB-assigned `suggestionId` before sending the message.

## Root Cause

In the normal audit flow, `auditRunCandidates` were constructed directly from raw `preRenderSuggestions` (plain result objects), which have no DB identity. The ai-only mode already included `suggestionId` correctly. This inconsistency meant that any normal-run prerender guidance message to Mystique lacked the field required to update the right suggestion record on the way back.

## Test Plan

- [x] Updated `ai-only-mode.test.js`: updated test to assert `getSuggestions` is called twice and that `auditRunCandidates[0].suggestionId` is populated from saved suggestions
- [x] Added new test in `handler.test.js`: `should include suggestionId in auditRunCandidates from saved suggestions`
- [x] Updated 7 existing tests that used mock suggestion objects without `getId()` — those objects are now returned by the second `getSuggestions` call and require the method
- [x] All 292 prerender tests pass

### DIFF
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -1163,14 +1163,7 @@ export async function processOpportunityAndSuggestions(
   // suggestionId is required by Mystique to correlate AI summaries back to the right suggestion.
   const savedSuggestions = await opportunity.getSuggestions();
   const urlToSuggestionId = new Map(
-    savedSuggestions
-      .filter((s) => {
-        const status = s.getStatus();
-        return status !== Suggestion.STATUSES.OUTDATED
-          && status !== Suggestion.STATUSES.SKIPPED
-          && status !== Suggestion.STATUSES.FIXED;
-      })
-      .map((s) => [s.getData()?.url, s.getId()]),
+    savedSuggestions.map((s) => [s.getData()?.url, s.getId()]),
   );
 
   // Build Mystique candidates directly from the URL list processed in this audit run.
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -487,8 +487,8 @@ async function fetchLatestScrapeJobId(siteId, context) {
  * @param {Object} auditData - Audit data used to build the message
  * @param {Object} opportunity - The prerender opportunity entity
  * @param {Object} context - Processing context
- * @param {Array|null} [preBuiltCandidates] - Pre-built candidate objects for normal audit runs
- *   (avoids a DB round-trip). Each entry is { url, originalHtmlMarkdownKey, markdownDiffKey }.
+ * @param {Array|null} [preBuiltCandidates] - Pre-built candidate objects for normal audit runs.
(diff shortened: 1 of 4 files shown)


src/prerender/handler.js
```diff
@@ -1159,12 +1159,26 @@ export async function processOpportunityAndSuggestions(
...
+  const urlToSuggestionId = new Map(
+    savedSuggestions
+      .filter((s) => {
+        const status = s.getStatus();
+        return status !== Suggestion.STATUSES.OUTDATED
+          && status !== Suggestion.STATUSES.SKIPPED
+          && status !== Suggestion.STATUSES.FIXED;
+      })
+      .map((s) => [s.getData()?.url, s.getId()]),
+  );
```
anuj-adobe, line 1165: does this mean you are fetching all suggestions from the db and sending to mystique?
We need to send only suggestions created in current audit..

anuj-adobe (commented):
I think we need to remove the batching logic for normal flow, in normal flow we want to send suggestions created in that run. 
I don't see any change here https://github.com/adobe/spacecat-audit-worker/blob/9c0209fc8d09ef718d7930ba135c7818995578a6/src/prerender/handler.js#L600

anuj-adobe (approved):

src/prerender/handler.js
```diff
@@ -1159,12 +1159,26 @@ export async function processOpportunityAndSuggestions(
...
+  // Fetch saved suggestions to map URL → suggestionId for Mystique payload.
+  // suggestionId is required by Mystique to correlate AI summaries back to the right suggestion.
+  const savedSuggestions = await opportunity.getSuggestions();
+  const urlToSuggestionId = new Map(
+    savedSuggestions
+      .filter((s) => {
+        const status = s.getStatus();
+        return status !== Suggestion.STATUSES.OUTDATED
+          && status !== Suggestion.STATUSES.SKIPPED
+          && status !== Suggestion.STATUSES.FIXED;
```
anuj-adobe: do you think we need this status check here, we are doing this in sendPrerenderGuidanceRequestToMystique(), would be easier to maintain if we apply status check at one place, wdyt?
outdated check is fine, rest checks could be handled in the method itself.

PR #2373, opened 2026-04-13
fix(prerender): fix suggestionId lookup, baseUrl in payload, and coveredByDomainWide guards

anuj-adobe:
## Problem

PR #2370 added `suggestionId` to the Mystique payload, but production data showed **8 out of 13** suggestions missing it for micron.com.

### Root cause: URL mismatch between audit runs

The `urlToSuggestionId` map used exact string match. The DB held both `https://www.micron.com/...` and `https://micron.com/...` variants from different audit runs — so lookups silently missed and `suggestionId` was `undefined`.

### Why a missing `suggestionId` crashed Mystique

Mystique parses SQS messages via Pydantic union types. With `suggestionId` absent:

1. `PrerenderSuggestionMessageData` requires `suggestion_id: str` → **validation fails**
2. Pydantic falls through to the next union member
3. `MetatagsOpportunityData` has all fields defaulted → **silently matches the payload**
4. Code calls `.suggestions` on `MetatagsOpportunityData` → **AttributeError crash**

### Missing `baseUrl` in SQS message

The SQS message to Mystique omitted `url` (baseUrl). `opportunity_service.py` fell back to the hardcoded `https://www.adobe.com/` for every site.

---

### 1. Send URL directly as `suggestionId`

`guidance-handler.js` matches Mystique responses back to DB suggestions **by URL**, not by `suggestionId` — so the URL is a valid correlation token and always available without a DB lookup.

This removes the post-sync `getSuggestions()` call and the `urlToSuggestionId` map entirely.

**No gap between synced and sent URLs:**
`syncSuggestions` receives `newData = preRenderSuggestions` and creates/updates a DB suggestion for every URL in the current audit run. `auditRunCandidates` is built from the same `preRenderSuggestions` array, so every synced URL — new or existing — maps 1:1 to a Mystique candidate. The only intentional exclusion is the domain-wide aggregate suggestion, which Mystique does not process.

### 2. Include `url: baseUrl` in SQS message

Adds `url: baseUrl` to the `guidance:prerender` SQS message so Mystique receives the correct site URL.

### 3. Exclude `edgeDeployed` suggestions from `coveredByDomainWide` marking

`markDeployedUrlSuggestionsAsCovered` was setting `coveredByDomainWide` on suggestions that already had `edgeDeployed` set. These are individually-deployed URL suggestions and should not be overridden by the domain-wide suggestion.

### 4. Diagnostic warning for non-`NEW` + `edgeDeployed` coexistence

Added a warning that fires when any suggestion has `edgeDeployed` set but `status !== NEW` — a combination that should never occur since status should not change after a URL is deployed at the CDN edge. Makes regressions visible in Splunk.

---

## Companion Mystique changes

Branch `fix/prerender-suggestion-id-optional`:

- Makes `suggestion_id` optional in `PrerenderSuggestionMessageData` so a missing field no longer fails the entire batch
- Fixes `save_result_to_s3` signature — `opportunity_service.py` calls it without `suggestions`; param made optional with fallback to `opportunity.data.suggestions`
- Extends SQS visibility timeout to 6 hours to prevent message redelivery on large batches

---

## Test plan

- [x] 224 prerender handler tests passing
- [ ] Deploy and verify Mystique no longer crashes on prerender payloads
- [ ] Verify site logs show correct `base_url` instead of `https://www.adobe.com/`
- [ ] Confirm `nonNewEdgeDeployedCount` warning never fires in production


ssilare-adobe (commented):
REMOVED

ssilare-adobe (commented):
## Code Review

### Overview

Small, well-described fix. Addresses an immediate Mystique crash caused by 8/13 suggestions missing `suggestionId` due to www/non-www URL mismatch between audit runs. Also adds the missing `url: baseUrl` field in the Mystique SQS message.

---

### Fix 1: `url: baseUrl` in SQS message

Placement is correct (top-level alongside `siteId`, `auditId`). Clean fix. However, **no test asserts `message.url === baseUrl`** at the top level of the SQS payload — the existing `ai-only-mode.test.js` assertions only check `message.data.suggestions[*].url`.

---

### Fix 2: Filter OUTDATED from `urlToSuggestionId` map

Good defensive fix — consistent with the ai-only mode path. The added null-check on `s.getData()?.url` is a nice touch to prevent `undefined` keys in the Map.

---

### Fix 3: Fallback to `s.url` when `suggestionId` is undefined

The intent is sound (prevent Mystique Pydantic silent type coercion), but a few concerns:

- **Semantics**: Sending a URL string as `suggestionId` is a workaround, not a correct value. The companion Mystique PR makes `suggestion_id` optional (`str | None`), so the cleaner fix once that lands would be to omit the field or send `null`. Until the www/non-www root cause (#1 in Follow-up items) is fixed, the fallback silently hides the mismatch — the warning log is the only signal.
- **Deployment order dependency**: If the Mystique companion PR is not deployed first, messages with URL-valued `suggestionId` will still crash via a different error path. Worth calling this out explicitly as a deploy prerequisite.

---

### Test Coverage

The diff only adds `warn: sandbox.stub()` to prevent errors — it is not an assertion. The following new code paths appear untested:

1. **`missingSuggestionIdCount > 0` warning log path** — the case where a suggestion URL does not match the saved map (the primary fix).
2. **`message.url === baseUrl`** assertion in the SQS payload (Fix 1).
3. **OUTDATED suggestion excluded from `urlToSuggestionId`** within `processOpportunityAndSuggestions` specifically (~line 1165). OUTDATED filtering in `sendPrerenderGuidanceRequestToMystique` has coverage in `ai-only-mode.test.js`, but this is a separate code path.

Given the 100% line/branch coverage requirement, worth confirming CI is enforcing coverage on this function.

---

### Suggested blocking items before merge

1. Add a test asserting `log.warn` is called with the expected message when a suggestion URL does not match the saved map.
2. Add a test asserting `message.url` equals the site base URL in the SQS payload.
3. Document the Mystique companion PR as a **deploy prerequisite** (deploy Mystique first, then this).

anuj-adobe (commented):
Addressed the three items in the review:

1. **Test for `log.warn` when suggestion URL misses the map** — Added `should warn and use URL as fallback when suggestion URL does not match saved suggestion (www/non-www mismatch)` in `handler.test.js`. Confirms that when a saved suggestion URL uses www (e.g. `https://www.example.com/page1`) but the current audit produces the non-www variant (`https://example.com/page1`), the exact-match lookup fails, the URL is used as `suggestionId` fallback, and `log.warn` is called with the expected message.

2. **Test that OUTDATED suggestions are excluded from the map** — Added `should exclude OUTDATED suggestions from the urlToSuggestionId map` in `handler.test.js`. Confirms that even when an OUTDATED suggestion has a URL matching the current audit run, it is filtered out of the `urlToSuggestionId` Map (consistent with the ai-only path), the URL fallback fires, and a warning is logged.

3. **Test for `message.url === baseUrl` in SQS payload** — Added `should include the site base URL as top-level url in the SQS message` in `ai-only-mode.test.js`. Confirms that the `url` field added by #2373 equals the site's base URL (`https://example.com`) so Mystique's `opportunity_service.py` receives the correct domain instead of defaulting to `https://www.adobe.com/`.

**Mystique companion PR deploy note**: The companion Mystique PR (branch `fix/prerender-suggestion-id-optional`) must be deployed before or alongside this PR. It makes `suggestion_id` optional in Pydantic to prevent a missing field from failing the entire batch, and extends SQS visibility to 3 hours to prevent redelivery on large batches.

ssilare-adobe (approved):
All three blocking items from my previous review have been addressed. Two minor nits:

1. The PR description is missing two changes included in the diff: the `edgeDeployed` guard added to `markDeployedUrlSuggestionsAsCovered` (prevents re-marking already edge-deployed suggestions as `coveredByDomainWide`) and the SKIPPED+edgeDeployed diagnostic warning in `processOpportunityAndSuggestions`. Both are well-tested and correct — just worth documenting.

2. The three new standalone tests build context/auditData from scratch instead of using `MockContextBuilder`. Non-blocking, but inconsistent with project test patterns.

Good fix overall. Deploy Mystique companion PR first as noted.

ssilare-adobe (approved):
All three blocking items from my previous review have been addressed. Two minor nits:

1. The PR description is missing two changes included in the diff: the `edgeDeployed` guard added to `markDeployedUrlSuggestionsAsCovered` (prevents re-marking already edge-deployed suggestions as `coveredByDomainWide`) and the SKIPPED+edgeDeployed diagnostic warning in `processOpportunityAndSuggestions`. Both are well-tested and correct — just worth documenting.

2. The three new standalone tests build context/auditData from scratch instead of using `MockContextBuilder`. Non-blocking, but inconsistent with project test patterns.

Good fix overall. Deploy Mystique companion PR for visiblity timeout issue.

src/prerender/handler.js
```diff
@@ -1161,15 +1163,32 @@ export async function processOpportunityAndSuggestions(
...
+      .filter((s) => s.getData()?.url && s.getStatus() !== Suggestion.STATUSES.OUTDATED)
+      .map((s) => [s.getData().url, s.getId()]),
   );
 
+  // Diagnostic: SKIPPED + edgeDeployed should never coexist — log a warning if found.
+  const skippedEdgeDeployedCount = savedSuggestions.filter(
+    (s) => s.getStatus() === Suggestion.STATUSES.SKIPPED && s.getData()?.edgeDeployed,
+  ).length;
+  if (skippedEdgeDeployedCount > 0) {
+    log.warn(`${LOG_PREFIX} Unexpected SKIPPED suggestions with edgeDeployed set. baseUrl=${auditUrl}, siteId=${auditData.siteId}, skippedEdgeDeployedCount=${skippedEdgeDeployedCount}`);
```
nit23uec: pls log this warning for any non-NEW suggestions also that have `edgeDeployed` set - this should be logged at the audit start.

nit23uec (commented):
## PR Review

---

## Strengths

**1. Precise root cause chain.** The PR correctly traces the failure all the way through: www/non-www URL mismatch → `undefined` `suggestionId` → Pydantic union type silent cast (`PrerenderSuggestionMessageData` → `MetatagsOpportunityData`) → `AttributeError` crash. That's non-trivial to diagnose from a production SQS payload.

**2. Multi-layer defense instead of a single point fix.** The PR attacks four independent failure modes in one shot: OUTDATED filter, URL fallback, `url` in SQS payload, `edgeDeployed` guard. Each protects a distinct invariant.

**3. URL fallback is more sound than it first appears.** The guidance handler (`guidance-handler.js:127-133`) looks up suggestions by `url` from Mystique's response — **not** by `suggestionId`. So whether Mystique receives a real UUID or a URL string, the guidance handler's ability to apply AI summaries is unchanged. The fallback's only real job is preventing Pydantic from silently casting the message type, which it does correctly.

**4. `coveredByDomainWide` guard is logically correct.** Individually-deployed suggestions (`edgeDeployed` set) have their own CDN-level deployment. Marking them `coveredByDomainWide` conflates two distinct deployment pathways and complicates rollback semantics. The fix is architecturally cleaner.

**5. Good observability design.** `missingSuggestionIdCount` and `skippedEdgeDeployedCount` give Splunk watchers concrete signals. The log line includes both `savedSuggestions` and `candidates` counts, making the ratio debuggable.

---

## Critical (Must Fix)

### C1 — `missingSuggestionIdCount` can over-count when `getS3Path` throws

**Location:** `handler.js` — the `auditRunCandidates` reduce block

```js
try {
  const suggestionId = urlToSuggestionId.get(s.url);
  if (!suggestionId) {
    missingSuggestionIdCount += 1;   // ← incremented here
  }
  acc.push({
    suggestionId: suggestionId || s.url,
    ...
    originalHtmlMarkdownKey: getS3Path(s.url, ...),   // ← can throw
    markdownDiffKey: getS3Path(s.url, ...),
  });
} catch {
  // skip malformed URLs
}
```

If `getS3Path` throws for a malformed URL that also had no matching `suggestionId`, the counter increments but the candidate is never pushed. The warning then logs:
```
2 suggestions missing suggestionId (using URL as fallback). candidates=0
```
That's a lie — the "fallback" never actually reached Mystique. Move the increment to after `acc.push` succeeds, or restructure so the count reflects candidates actually sent.

---

## Major (Should Fix)

### M1 — The OUTDATED test proves the wrong scenario

**Location:** `handler.test.js` — `'should exclude OUTDATED suggestions from the urlToSuggestionId map'`

The test creates a single OUTDATED suggestion for `https://example.com/page1` and mocks `syncSuggestions` away entirely. The test asserts the URL fallback fires and a warning is logged.

The problem: in the real code path, `syncSuggestions` runs **before** `getSuggestions()`, and when an OUTDATED suggestion is a regression (same URL now needs prerender again), `syncSuggestions`'s `defaultMergeStatusFunction` transitions it back to NEW **in-place** (same UUID, new status). If `getSuggestions()` returns fresh DB data, the entity is NEW and its UUID is correctly in the map — no fallback, no warning.

The test as written validates the stale-cache edge case (OUTDATED entity visible after sync), but misses the primary intent: **confirm that a legitimately OUTDATED suggestion from a prior cycle (not a regression) is excluded**. Add a test where the OUTDATED suggestion's URL does NOT appear in the current audit's `preRenderSuggestions` — that's the actual scenario being defended.

### M2 — The `url: baseUrl` field has no assertion in the normal-flow path

**Location:** `handler.test.js` — no test checks `url` in the SQS payload when called from `processContentAndGenerateOpportunities`

The ai-only mode test (`ai-only-mode.test.js`) covers the `url` field. But the normal flow invocation of `sendPrerenderGuidanceRequestToMystique` from `processContentAndGenerateOpportunities` has no explicit assertion that the message includes `url`. Given the project's 100% coverage requirement, the line is covered by execution, but the behavioral contract (correct `url` in the payload) is untested for the main flow. Add an assertion in at least one of the existing `processContentAndGenerateOpportunities` integration-style tests.

### M3 — OUTDATED is the only excluded status; REJECTED suggestions pollute the map

**Location:** `handler.js:1166`

```js
.filter((s) => s.getData()?.url && s.getStatus() !== Suggestion.STATUSES.OUTDATED)
```

`REJECTED` suggestions pass this filter and land in the `urlToSuggestionId` map. If the same URL appears in a subsequent audit as still needing prerender (user disagrees or it regressed), the REJECTED suggestion's UUID is sent to Mystique, which generates an AI summary for it. The guidance handler (`guidance-handler.js:114-117`) doesn't filter REJECTED — only OUTDATED — so Mystique's response would update `aiSummary` on a user-rejected suggestion, effectively undoing the user's decision signal on the data.

Consider widening the filter to exclude terminal statuses: `OUTDATED`, `REJECTED`, `FIXED` (if a FIXED suggestion for a URL now needs prerender again, `syncSuggestions` has already transitioned it to NEW before `getSuggestions()` is called, so including FIXED in the exclusion is safe and avoids stale-ID risk).

---

## Minor (Nice to Fix)

### m1 — Double `getData()` call in the `markDeployedUrlSuggestionsAsCovered` filter

```js
? newSuggestions.filter((s) => deployedAtEdgeUrls.has(s.getData()?.url)
  && !s.getData()?.edgeDeployed)
```

Two `getData()` calls per suggestion. Extract to a local variable:
```js
? newSuggestions.filter((s) => {
    const data = s.getData();
    return deployedAtEdgeUrls.has(data?.url) && !data?.edgeDeployed;
  })
```

### m2 — `skippedEdgeDeployedCount` warning fires on every audit run once the state exists

If even one suggestion is in `SKIPPED + edgeDeployed` state, the warning fires on **every** subsequent audit run for that site until the state is resolved. In production this would spam Splunk with recurring warnings for the same root cause. Consider enriching the message with a suggestion ID so the same event isn't treated as a new alarm on each run.

### m3 — Test data uses hardcoded status strings instead of imported constants

New tests construct `dataAccess.Suggestion.STATUSES` inline and `OUTDATED` is conspicuously absent from the stub even though it's what the test exercises. If a code path calls `Suggestion.STATUSES.OUTDATED` against this stub, it silently returns `undefined`. Use the real `Suggestion.STATUSES` from `@adobe/spacecat-shared-data-access` in tests.

### m4 — Missing code comment explaining guidance handler URL-based matching

The `suggestionId || s.url` fallback comment says "Mystique requires non-empty string" but omits a critical context: **the guidance handler matches by `url`, not by `suggestionId`**, so the fallback doesn't break AI summary delivery for URLs that are properly matched on the Mystique→handler return path. Add this note so future maintainers don't remove the fallback thinking it's semantically wrong.

---

## New Regression Risk: None

Going through each change:
- `coveredByDomainWide` guard — only restricts what gets set; no behavioral regression for non-`edgeDeployed` suggestions
- `url: baseUrl` in SQS — additive field; Mystique's hardcoded fallback is overridden with the correct value
- OUTDATED filter — conservative narrowing; AI summary outcome is the same either way
- URL fallback — prevents crashes without changing guidance handler's lookup logic
- Diagnostic warning — log-only, no behavior change

The PR does not introduce regressions into the existing audit, sync, or guidance-handler flows.

anuj-adobe:
## Summary of changes addressing review feedback

### C1 — `missingSuggestionIdCount` counter moved after `acc.push()`
The counter was incremented before `getS3Path` was called. If `getS3Path` threw for a malformed URL that also had no matching `suggestionId`, the count would be inflated — the log would claim N suggestions used the URL fallback when some of those never actually reached Mystique. Moved the increment to after `acc.push()` succeeds so the count only reflects candidates sent.

### m1 — `getData()` extracted to local variable in `markDeployedUrlSuggestionsAsCovered`
Removed the redundant double `s.getData()` call in the filter. Extracted to `const data = s.getData()` once per suggestion.

### m3 — `OUTDATED` added to `Suggestion.STATUSES` stub in tests
The test mock for `context.dataAccess.Suggestion.STATUSES` was missing `OUTDATED`. While the source code uses the module-level `Suggestion` import (not the mock) for this constant, the mock was incomplete and misleading. Added `OUTDATED: 'OUTDATED'` to both affected test stubs.

### m4 — Comment added explaining guidance handler URL-based matching
Added a comment near the `suggestionId || s.url` fallback explaining that the guidance handler (`guidance-handler.js`) matches Mystique's response back to suggestions **by URL, not by `suggestionId`**. This means the URL fallback does not break AI summary delivery — future maintainers won't remove it thinking it's semantically wrong.

### m2 (revised) — `edgeDeployed` diagnostic broadened and moved to audit start
- **Broadened from `=== SKIPPED` to `!== NEW`**: any suggestion whose status was changed after edge deployment is suspicious, not just SKIPPED ones.
- **Renamed** to `detectWrongEdgeDeployedStatus()` with all calculation self-contained (Opportunity lookup + `getSuggestions` + count inside the function).
- **Moved to start of `processContentAndGenerateOpportunities`** (before `syncSuggestions`): previously the check was inside `processOpportunityAndSuggestions`, which is skipped entirely when no URLs need prerender. Now it runs unconditionally on every audit, catching pre-existing bad state even on zero-prerender runs.
- The post-sync duplicate call was removed — `syncSuggestions` cannot create non-NEW+edgeDeployed combinations, so one check at audit start is sufficient.

nit23uec (approved):

nit23uec (approved):

PR #2374, opened 2026-04-13
feat:  replace Excel/SharePoint output with DB Opportunities

ymadobe:
Replaces legacy Excel/SharePoint persistence with DB-backed Opportunities
and Suggestions. One Opportunity per status code bucket (404, 403, 5xx),
one Suggestion per unique error URL.

- Adds groupErrorsByUrl and parsePeriodIdentifier to utils.js
- Adds opportunity-data-mapper.js with per-bucket title/description/guidance
- Syncs suggestions weekly via syncSuggestions; preserves Mystique-enriched
  fields (suggestedUrls, aiRationale, confidenceScore) on re-merge
- Marks suggestions OUTDATED after 4 weeks of absence


src/llm-error-pages/handler.js
```diff
@@ -185,59 +203,119 @@ export async function runAuditAndSendToMystique(context) {
...
-            return;
-          }
+        // ── Opportunity + Suggestion sync ──────────────────────────────────────
+        // One Opportunity per status code bucket. One Suggestion per unique URL.
+        const opportunityMap = {};
+
+        for (const { code, auditType, suggestionType } of STATUS_BUCKETS) {
+          const rawErrors = categorizedResults[code] || [];
+
+          if (rawErrors.length > 0) {
```
anagarwa, line 220: @ymadobe correct me If I have missed something but The retention logic only runs inside if (rawErrors.length > 0). That means if a bucket stops producing errors entirely, we skip the bucket and never revisit
  its existing suggestions. In that case, old 403 / 5xx suggestions can stay NEW indefinitely instead of aging out after the 4-week retention window.

  I think the stale-suggestion cleanup needs to run even when rawErrors is empty, using the existing opportunity/suggestions for that bucket. Otherwise “not
  seen in the last N weeks” only works for buckets that still have at least one current result.

src/llm-error-pages/handler.js
```diff
@@ -280,7 +358,7 @@ export async function runAuditAndSendToMystique(context) {
                   }))
                   .filter((link) => link.urlFrom.length > 0),
                 alternativeUrls,
-                opportunityId: `llm-404-${periodIdentifier}`,
+                opportunityId: opportunityMap[404]?.getId(),
```
anagarwa, line 374: @ymadobe are we saving the response from mystique specially aiRationale, suggestedUrls and confidenceScore ?

PR #2382, opened 2026-04-14
feat: replace Coralogix site detection poller with async job handler …

papineni05:
## Summary

Replaces the hourly Coralogix polling loop with an async job handler. The DaaS team now ingests Fastly CDN logs and calls \`POST /sites/detect/jobs\` (in spacecat-api-service) which enqueues an SQS message processed by this handler.

**Worker flow:**
1. Receives \`{ type: 'site-detection', jobId }\` from SQS
2. Loads \`AsyncJob\` and extracts \`{ domain, hlxVersion }\` from job metadata
3. Validates domain (IP check, ignored subdomain/domain filters)
4. Checks Site/SiteCandidate tables → marks job \`COMPLETED\` with \`action: duplicate\` if found
5. Verifies Helix/EDS site via DOM structure check
6. Extracts \`hlxConfig\` (RSO parse + admin API if \`HLX_ADMIN_TOKEN\` available)
7. Creates \`SiteCandidate\` record
8. Sends Slack discovery notification (best-effort)
9. Marks job \`COMPLETED\` with \`action: created\`

---

### \`POST /sites/detect/jobs\` _(in spacecat-api-service — triggers this worker)_

**Request**
\`\`\`json
{
  "domain": "www.example.com",
  "hlxVersion": 5
}
\`\`\`

| Field | Type | Required | Description |
|-------|------|----------|-------------|
| \`domain\` | string | Yes | Hostname to detect (e.g. \`www.example.com\`) |
| \`hlxVersion\` | integer | No | Helix version detected from CDN logs |

**Response — 202 Accepted**
\`\`\`json
{
  "jobId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "status": "IN_PROGRESS",
  "createdAt": "2025-01-01T00:00:00.000Z",
  "pollUrl": "https://spacecat.experiencecloud.live/api/v1/sites/detect/jobs/3fa85f64-5717-4562-b3fc-2c963f66afa6"
}
\`\`\`

**Response Statuses**

| Status | Condition |
|--------|-----------|
| \`202 Accepted\` | Job created, poll \`pollUrl\` for result |
| \`400 Bad Request\` | Missing/invalid \`domain\`, or \`hlxVersion\` is not an integer |
| \`500 Internal Server Error\` | Unexpected failure |

---

### \`GET /sites/detect/jobs/:jobId\` _(in spacecat-api-service — polls this worker's result)_

**Request**

No request body. \`jobId\` is a UUID path parameter.

**Response — 200 OK**
\`\`\`json
{
  "jobId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "status": "COMPLETED",
  "createdAt": "2025-01-01T00:00:00.000Z",
  "updatedAt": "2025-01-01T00:01:00.000Z",
  "result": {
    "action": "created",
    "domain": "www.example.com",
    "baseURL": "https://www.example.com",
    "reason": null
  },
  "error": null
}
\`\`\`

**Possible \`status\` values**

| Status | Meaning |
|--------|---------|
| \`IN_PROGRESS\` | Worker is processing the domain |
| \`COMPLETED\` | Evaluation complete — check \`result.action\` for the outcome |
| \`FAILED\` | Unexpected exception; safe to retry. Check \`error.code\` and \`error.message\` |

**Possible \`result.action\` values** (only present when \`status: COMPLETED\`)

| Action | Meaning |
|--------|---------|
| \`created\` | SiteCandidate successfully created |
| \`duplicate\` | Site or SiteCandidate already exists |
| \`rejected\` | Domain failed validation (IP address, ignored subdomain, not a Helix site) |

**Response Statuses**

| Status | Condition |
|--------|-----------|
| \`200 OK\` | Job found, returns current state |
| \`400 Bad Request\` | \`jobId\` is not a valid UUID |
| \`404 Not Found\` | No job found for given \`jobId\` |
| \`500 Internal Server Error\` | Unexpected DB error |

---

## Infra Changes Needed (not in this PR)

- Add \`HLX_ADMIN_TOKEN\` to audit-worker Lambda secrets
- Add \`SLACK_SITE_DISCOVERY_CHANNEL_INTERNAL\` to audit-worker Lambda env
- Remove \`CORALOGIX_API_KEY\` and \`SITE_DETECTION_WEBHOOK\` from audit-worker env

## Test Plan

- [ ] SQS message with valid new EDS domain → SiteCandidate created, job \`COMPLETED\`, \`action: created\`
- [ ] SQS message with domain already in Site table → job \`COMPLETED\`, \`action: duplicate\`
- [ ] SQS message with domain already in SiteCandidate table → job \`COMPLETED\`, \`action: duplicate\`
- [ ] SQS message with IP address domain → job \`COMPLETED\`, \`action: rejected\`
- [ ] SQS message with ignored subdomain token → job \`COMPLETED\`, \`action: rejected\`
- [ ] SQS message with non-Helix domain → job \`COMPLETED\`, \`action: rejected\`
- [ ] Unexpected exception → job \`FAILED\`, \`error.code: EXCEPTION\`
- [ ] Missing \`jobId\` in SQS message → logged error, no crash

## Related
- Jira: [SITES-42916](https://jira.corp.adobe.com/browse/SITES-42916)
- Companion PR (api-service): [adobe/spacecat-api-service#2203](https://github.com/adobe/spacecat-api-service/pull/2203)


solaris007 (changes requested):
Hey @papineni05,

This review covers spacecat-audit-worker#2382 together with its companion spacecat-api-service#2203. Findings are split by repo where they are fixable; cross-cutting concerns are called out explicitly.

## Strengths

- Replacing `urijs` with native `URL` (src/site-detection/handler.js:14-28) reduces a third-party dep and gives better malformed-input resilience (exercised by the `'  not a domain  '` test).
- Apex and www-only rejection (src/site-detection/handler.js:76-94) surfaces a real class of false positives that the old poller silently onboarded.
- Job-status check at src/site-detection/handler.js:420-422 short-circuits SQS redelivery for non-IN_PROGRESS jobs - clean idempotency guard with an observable log.warn.
- `parseHlxRSO` (src/site-detection/handler.js:175-184) restricts `owner`/`site` to `[\w-]+`. `HLX_ADMIN_TOKEN` only ever reaches the hardcoded `admin.hlx.page` host - no URL injection primitive.
- Slack notification at src/site-detection/handler.js:509-520 is best-effort and cannot fail the job.
- Env overrides for ignored subdomains/domains with graceful fallback on invalid regex at src/site-detection/handler.js:386-397.
- Step-by-step happy path at src/site-detection/handler.js:449-527 makes terminal state transitions auditable in one read.
- Test coverage is proportional and covers non-happy paths: SQS redelivery short-circuit, save failure, admin API 404/500/throw, Slack failure, missing payload.

## Issues

### Important (Should Fix)

1. **FAILED status conflates rejected/duplicate with exceptions** (src/site-detection/handler.js:452-490)
   - The worker marks `rejected` (invalid domain), `duplicate` (already known), and runtime exceptions all as `AsyncJob.Status.FAILED`. `rejected` and `duplicate` are valid terminal outcomes of evaluation - not failures. DaaS (consumer) cannot distinguish "your job crashed, retry is appropriate" from "we correctly decided not to onboard this domain" from `status` alone. Monitoring that watches FAILED async-job counts will fire on expected outcomes.
   - Fix: mark `created`/`rejected`/`duplicate` as `COMPLETED` (result.action carries the verdict). Reserve `FAILED` for exceptions only (the catch at line 534). This matches how preflight modelled it: COMPLETED with a result payload, FAILED only on exception.

2. **SSRF feedback channel via `isHelixSite` fetch** (src/site-detection/handler.js:150-167)
   - `isHelixSite(baseURL)` fetches a URL derived from the caller's `domain`. Direct-IP SSRF is blocked by the regex, but a hostname that DNS-resolves to a private address (e.g. `internal-api.corp.example.com` -> 10.x.x.x) is not. The primitive is amplified by:
     - `reason: 'DOM is not in Helix format. Status: ${resp.status}'` (line 160) leaks the HTTP status of the internal host.
     - `reason: 'Cannot fetch the site: ${e.message}'` (line 166) leaks DNS/TCP error details (ECONNREFUSED on 10.0.0.5:443, ENOTFOUND on internal.corp).
     - No `redirect: 'manual'` - a public host can 302 to a private one.
   - The api-service GET endpoint returns `result.reason` to the caller, so any holder of a `site:write` S2S key gets a generic internal-host enumeration oracle.
   - Fix: resolve hostname and reject RFC1918, loopback, link-local, AWS metadata (169.254.169.254) ranges before the fetch. Pass `redirect: 'manual'` and re-validate after each redirect. Genericize `reason` on unreachable/non-200 to `"Site did not serve a Helix-format DOM"` without status codes or error messages.

3. **Missing-domain path's `job.save()` is outside the outer try/catch** (src/site-detection/handler.js:432-438)
   - The "missing domain" early-exit calls `await job.save()` at line 437, before the outer `try { ... }` at line 449. If save throws (DB transient), the exception escapes AuditBuilder, SQS redelivers, the next attempt hits the same invalid-payload path and re-throws forever. Poison pill.
   - Fix: wrap the save in try/catch + `log.error`, or move the invalid-payload handling inside the outer try. Add a test for the save-failure branch on the missing-domain path.

4. **AuditBuilder wrapper is all no-ops for this handler** (src/site-detection/handler.js:560-564)
   - The handler no longer operates on a Site, has no siteId in the message, and the runner owns its own AsyncJob lifecycle. Every AuditBuilder hook (`withUrlResolver`, `withPersister`, `withMessageSender`) is no-op'd. A future engineer has to learn the AuditBuilder abstraction, discover it is all no-op'd, then track down where the actual dispatch happens.
   - Fix: register the runner directly in src/index.js HANDLERS map (`'site-detection': siteDetectionRunner`) and drop the builder wrapping.

### Minor (Nice to Have)

1. **Dead code: redundant `hlxConfig` initializer** (src/site-detection/handler.js:494-495)
   - `let hlxConfig = { hlxVersion: hlxVersion ?? null, rso: {} };` is immediately overwritten by the next line's `await extractHlxConfig(...)`. Use `const hlxConfig = await extractHlxConfig(...)`.

2. **Rejection `reason` is always generic** (src/site-detection/handler.js:454, 466, 477, 488)
   - The specific reason (ignored subdomain, one-character subdomain, IP, port, apex, www-only) is only in the handler log, not in `result.reason`. DaaS polls the API and gets `"Domain failed validation rules"` for every rejection. Return the specific reason from `isValidCandidate` and propagate into `result.reason`.

3. **Subdomain extraction does not handle multi-level TLDs** (src/site-detection/handler.js:80)
   - `url.hostname.split('.').slice(0, -2).join('.')` treats `.co.uk`, `.com.br`, `.com.au` as subdomains - `site.company.co.uk` yields `site.company` as the "subdomain". The `ignoredDomains` regex catches most problematic cases today; add a comment at the declaration acknowledging the limitation.

4. **`error.stack` stored in `job.error.details`** (src/site-detection/handler.js:546)
   - The api-service controller strips `details` in the GET response (verified), so this does not leak to HTTP callers today. But the job row is readable by anyone with DB access. If a future controller serializes `error` without filtering, the stack leaks. Consider truncating or gating on an env flag.

5. **Missing test for `aem.live` + v4 RSO combination** (test/audits/site-detection.test.js)
   - The branch `(hlxVersion ?? 0) >= 5 || rso.tld === 'aem.live'` is covered separately (hlx.live + v5; aem.live + null), but the specific aem.live + v4 combination that forces the admin API call despite low version is untested.

6. **Job-not-IN_PROGRESS test only covers COMPLETED** (test/audits/site-detection.test.js:751-757)
   - FAILED and CANCELLED are also terminal states that can arrive via SQS redelivery. Parameterize or add tests so the short-circuit and warn log are confirmed for every non-IN_PROGRESS state.

## Recommendations

- After fixing Important item 2, add SSRF test coverage: domains that DNS-resolve to 127.0.0.1, 10.0.0.1, 169.254.169.254, ::1, and a public host that 302-redirects to a private IP.
- Introduce helpers for terminal job transitions: `finalizeJobAsCompleted(job, result)` / `finalizeJobAsFailed(job, error)` that setEndedAt internally. The current handler sets endedAt at seven call sites - easy to forget in the next path added.
- Normalize `auditResult` shape across return paths. Today it varies: `{action, domain}`, `{action, domain, baseURL}`, `{error}`, `{skipped: true}`. A union type simplifies downstream consumers.
- Consider a dedicated `SITE_DETECTION_QUEUE_URL` (today the code sends to `AUDIT_JOBS_QUEUE_URL`). Clearer cost attribution, independent DLQ, independent rate limits. Follow-up PR.
- Worth an ADR or short doc note on the apex + www-only rejection rule (handler.js:82-94). Future operators will wonder why a legitimate-looking `adobe.com` got dropped.

- Infra changes: add `HLX_ADMIN_TOKEN`, `SLACK_SITE_DISCOVERY_CHANNEL_INTERNAL`; remove `CORALOGIX_API_KEY`, `SITE_DETECTION_WEBHOOK`. Called out in PR body.
- `hlxVersion` as a top-level field in POST /sites/detect couples the endpoint to the CDN-log detection source. Future sources (GSC, manual, partner feeds) will have their own metadata. Consider a source-agnostic `hints` object or source-keyed envelope before DaaS wires up.
- Formalize `result.action` as a shared type in spacecat-shared-data-access. The OpenAPI already declares the enum; lift into code so both the worker and api-service agree on values.

## Assessment

Ready to merge? With fixes.

Reasoning: The handler rewrite is structurally sound and test coverage is substantial. The 4 Important items are proportional fixes: 1 and 2 shape the observable public contract and the SSRF surface and should be resolved before DaaS calls this in production; 3 is a small poison-pill fix; 4 is cleanup that gets harder the longer it sits.

## Next Steps

1. Address the 4 Important items first.
2. Review the 6 Minor items.
3. Recommendations and out-of-scope items can be separate PRs.

papineni05:
@solaris007 addressed the review. Summary of changes in [55c51b93](https://github.com/adobe/spacecat-audit-worker/commit/55c51b93):

**Important**
1. **FAILED vs COMPLETED semantics** — `rejected` and `duplicate` are now `AsyncJob.Status.COMPLETED` with `result.action` carrying the verdict. `FAILED` is reserved for unexpected exceptions only. Monitoring on FAILED counts no longer fires on expected outcomes. Extracted `finalizeCompleted` / `finalizeFailed` helpers so `endedAt` and state transitions live in one place (addresses the "seven call sites" recommendation too).
2. **SSRF guard on `isHelixSite`** —
   - `dns.lookup(hostname, { all: true })` before fetch. Any resolved address in RFC1918 (10/8, 172.16/12, 192.168/16), loopback (127/8), link-local + AWS metadata (169.254/16), ULA (fc/fd), CGNAT (100.64/10), multicast (≥224), IPv6 loopback (::1, ::), IPv6 link-local (fe80:), or v4-mapped IPv6 pointing at a private v4 → rejected without calling fetch.
   - `fetch(url, { redirect: 'manual' })`. A 30x response is treated as non-Helix (no chasing into unvalidated hosts).
   - Only a generic `"Site did not serve a Helix-format DOM"` reaches the caller. Status codes, DNS errors, and resolved IPs stay in logs.
   - 17 new SSRF tests including each private range, DNS failure, empty DNS, mixed public+private, non-string address, and public IPv6 happy path.
3. **Missing-domain `job.save()`** — wrapped in try/catch via `finalizeFailed`. A DB transient can no longer escape the outer `try` and poison-pill SQS. New test asserts the runner returns normally when save throws on this path.
4. **AuditBuilder wrapper dropped** — default-export is now `siteDetectionRunner` directly. The dispatcher at [src/index.js:316](https://github.com/adobe/spacecat-audit-worker/blob/main/src/index.js#L316) already handles both `.run()` builder outputs and plain functions, so HANDLERS registration stayed the same.

**Minor**
- Dead `let hlxConfig = { ... }` initializer removed.
- Rejection reasons now carry the specific cause where safe (validation returns `"Domain failed validation rules"`; Helix check returns generic reason by design).
- Existing subdomain-extraction comment on multi-level TLDs can be a follow-up.
- New happy-path-with-`save()`-failure test covers the `finalizeCompleted` swallowed-save branch.
- `aem.live + v4` RSO combination now has a dedicated test.
- Short-circuit test now covers both `COMPLETED` and `FAILED` terminal states via SQS redelivery.

100% coverage on the handler; full lint clean.

**Out of scope**
- Infra changes (`HLX_ADMIN_TOKEN`, `SLACK_SITE_DISCOVERY_CHANNEL_INTERNAL`, Coralogix envs), dedicated `SITE_DETECTION_QUEUE_URL`, `result.action` shared type, ADR for apex/www-only rejection — tracked separately.

Companion api-service changes in [adobe/spacecat-api-service#2203](https://github.com/adobe/spacecat-api-service/pull/2203).

solaris007 (commented):
Hey @papineni05,

Re-review of spacecat-audit-worker after your changes addressing the prior round. Reviewed in conjunction with the companion spacecat-api-service#2203.

## Previously flagged issues now addressed

All four prior Important findings are correctly resolved:

- **FAILED/COMPLETED semantics** (src/site-detection/handler.js:348-372): `finalizeCompleted` and `finalizeFailed` cleanly split "evaluated and rejected/duplicate" (COMPLETED + result.action) from "exception" (FAILED + error). All call sites route through these helpers consistently.
- **SSRF defense** (src/site-detection/handler.js:105-214): comprehensive layered approach - `dns.lookup` with `isPrivateIP` covering RFC1918, loopback, link-local, AWS metadata, ULA, CGNAT, multicast, v4-mapped IPv6; `redirect: 'manual'`; generic `NOT_HELIX_REASON` for callers. 17 SSRF tests verified and thorough.
- **Poison-pill on missing-domain** (src/site-detection/handler.js:444-448): save now routed through `finalizeFailed` with try/catch. Test confirms save failure doesn't escape.
- **AuditBuilder no-ops removed** (src/site-detection/handler.js:547): plain function export.

Additionally: dead hlxConfig initializer removed, aem.live + v4 RSO test added, short-circuit test covers FAILED state, rejection reasons now carry specific cause where safe.

## Issues

### Important (Should Fix)

1. **DNS rebinding window between `resolvesToPublicAddress` and `fetch`** (src/site-detection/handler.js:182-188)
   - The hostname is resolved twice: once explicitly via `dns.lookup` for the SSRF guard, and again implicitly by `fetch` when it opens the socket. An attacker who controls DNS for the submitted domain can serve a public A record with TTL 0 to pass the guard, then flip to a private address so the actual fetch lands on an internal target.
   - Practical risk is bounded: this is blind SSRF since response details never reach the caller (only `NOT_HELIX_REASON`), and IMDSv2 should reject unauthenticated GETs.
   - Fix (this PR or follow-up): pin the validated address through to the connect via a custom Undici dispatcher with `{ connect: { lookup: pinnedLookup } }`, or accept the residual and document the gap with a tracking ticket. If deferring, please add a comment in `isHelixSite` noting the limitation so the next reader doesn't assume the guard is complete.

2. **Non-atomic check-then-create for concurrent same-domain jobs** (src/site-detection/handler.js:481-500)
   - With the api-service early-exit removed, the worker is the sole duplicate-detection authority. But `SiteCandidate.findByBaseURL` followed by `SiteCandidate.create` is not atomic. Two SQS messages for the same domain processed concurrently can both observe "no candidate" and both create, producing duplicate rows and double Slack notifications.
   - DaaS call volume is low, so the window is small, but the design's stated contract is "single decision point."
   - Fix (this PR or follow-up): add a unique index on `site_candidates.base_url` and catch the unique-violation, mapping it to `finalizeCompleted({ action: 'duplicate' })`. If the schema migration is non-trivial, document the race in the handler and file a tracking ticket.

### Minor (Nice to Have)

1. **IPv6 coverage gaps in `isPrivateIP`** (src/site-detection/handler.js:127-139)
   - Link-local check `startsWith('fe80:')` covers `fe80::/16` but RFC 4291 defines link-local as `fe80::/10` (through `febf:`). Theoretical only since kernels produce `fe80::/64` in practice.
   - IPv6 multicast (`ff00::/8`) is not blocked, asymmetric to the IPv4 side which blocks `>= 224`. Theoretical since Lambda drops multicast traffic.
   - Fix: add `lower.startsWith('ff')` for multicast, and widen link-local to `/^fe[89ab][0-9a-f]:/`. Or adopt `ipaddr.js` which handles all reserved ranges canonically.

2. **No test for IPv6 ULA (`fc00::/7`) rejection** (test/audits/site-detection.test.js)
   - `isPrivateIP` correctly rejects ULA (handler.js:131-133) but no test locks it in. Two-line addition mirroring existing IPv6 SSRF tests.

3. **`finalizeCompleted` save failure leaves zombie IN_PROGRESS jobs with no alarmable signal** (src/site-detection/handler.js:348-357)
   - The save error is deliberately swallowed (to prevent SQS redelivery storms), the test codifies this, and the rationale is sound. But `log.error` alone is not a usable on-call signal at scale. A partial DB outage could leave many jobs stuck.
   - Recommendation: emit a distinct log marker (e.g. `event=site-detection.finalize_save_failed`) suitable for a CloudWatch metric/alarm.

## Recommendations

- Decide explicitly whether DNS-rebinding hardening is in scope for this PR or a follow-up. If deferred, add a comment in `isHelixSite` and a tracking ticket.
- Consider `ipaddr.js` as a replacement for the hand-rolled `isPrivateIP`. It handles all reserved ranges, canonical forms, and v4-mapped/v4-compatible addresses.
- Add boundary unit tests for `isPrivateIP`: `172.15.x.x` and `172.32.x.x` (just outside RFC1918), `100.63.x.x` (just outside CGNAT) to make future range edits safer.

## Assessment

Ready to merge? Yes, with the two Important items addressed in this PR or as tracked follow-ups.

Reasoning: All prior blocking findings are correctly and thoroughly addressed. The remaining Important items are defense-in-depth gaps (DNS rebinding is a known class of guard-then-fetch limitation with bounded blast radius; the race condition requires concurrent duplicate submissions at low DaaS volume). Both are defensible as tracked follow-ups if the gaps are documented in code.

solaris007 (approved):
Hey @papineni05,

All five round-2 findings are correctly addressed. The SSRF guard is now symmetric across IPv4 and IPv6, the DNS rebinding limitation is honestly documented with bounded blast radius, and the race-recovery pattern handles concurrent same-domain SQS deliveries without a schema migration.

## Strengths

- Previously flagged DNS rebinding window is now documented at `handler.js:181-187` with clear blast-radius bounds (blind SSRF, NOT_HELIX_REASON-only, IMDSv2). Proper fix path (Undici connect-options pinning) tracked as follow-up
- Previously flagged non-atomic check-then-create is handled at `handler.js:521-540`: try/catch around `SiteCandidate.create`, retry `findByBaseURL`, graceful mapping to COMPLETED/duplicate. Two tests pin both the race-recovery and non-race paths
- Previously flagged IPv6 coverage gaps are closed: `/^fe[89ab][0-9a-f]:/` correctly covers fe80::/10 (verified at boundaries: fe80/fe90/feaf/febf match, fec0 does not), `startsWith('ff')` covers ff00::/8 multicast, ULA test at fc00::1 added
- Previously flagged finalize_save_failed observability is addressed with structured `event=site-detection.finalize_save_failed` marker
- Cross-repo contract is stable: worker writes {action, domain, baseURL, reason}; api-service (just approved) reads as-is. No coupling to the reverted imsOrgId feature

## Issues

### Minor (Nice to Have)

1. **Race-recovery catch doesn't log the original `createErr.message`** (`handler.js:530`) - When `SiteCandidate.create` fails and the retry finds a candidate, the log says "concurrent duplicate detected" but never records why `create` failed. Since the catch is error-class-agnostic (catches any error, not just unique-violations), logging the original error preserves diagnostic signal for distinguishing real races from masked transient failures. One-line fix: include `createErr.message` in the existing log.info line.

2. **`finalizeFailed` save error has no alarmable event marker** (`handler.js:368`) - `finalizeCompleted` save failures emit `event=site-detection.finalize_save_failed` (alarmable), but the symmetric `finalizeFailed` save failure logs a plain error message with no `event=` marker. A CloudWatch metric filter on the existing event will miss FAILED-path zombie jobs. One-line fix: add a matching `event=` marker.

3. **finalize_save_failed test asserts the event marker but not the jobId** (`site-detection.test.js:570-572`) - The regex confirms the marker but not the `job=<id>` segment. If a future refactor drops the jobId, operators lose the field needed to correlate alarms to specific jobs. One-line fix: extend the assertion.

## Recommendations

- Once the unique-index migration on `site_candidates.base_url` lands, narrow the race-recovery catch to the specific unique-violation error shape. Reference this code path in the migration ticket.
- Replace inline "tracked as follow-up" notes with actual issue numbers once filed, so the comments don't decay.

## Assessment

Ready to merge? **Yes.**

All prior Important findings confirmed resolved across three review rounds. The three remaining Minor items are observability polish - none affect correctness or security.

PR #2388, opened 2026-04-15
fix(toc): convert transformRules.value to HAST in mergeDataFunction

anagarwa:
The mergeDataFunction in opportunityAndSuggestions was spreading raw newSuggestion data (valueFormat: 'html', value: array) over existing suggestions, overwriting correctly formatted HAST values. This caused ~970 suggestions across 27 sites to revert from HAST to raw array format on subsequent audit runs.

The HAST conversion via tocArrayToHast() was only applied in mapNewSuggestion (for new suggestions) but not in mergeDataFunction (for existing suggestions being updated). Now both newSuggestion and preserved isEdited values are converted to HAST before merging.

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/toc/handler.js
+++ b/src/toc/handler.js
@@ -489,12 +489,23 @@ export async function opportunityAndSuggestions(auditUrl, auditData, context) {
   );
 
   const mergeDataFunction = (existingSuggestion, newSuggestion) => {
+    const converted = { ...newSuggestion };
+    if (converted.transformRules && Array.isArray(converted.transformRules.value)) {
+      converted.transformRules = {
+        ...converted.transformRules,
+        value: tocArrayToHast(converted.transformRules.value),
+        valueFormat: 'hast',
+      };
+    }
     const mergedSuggestion = {
       ...existingSuggestion,
-      ...newSuggestion,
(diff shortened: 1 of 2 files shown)


duynguyen (approved):
The fix is correct and well-scoped. The bug description matches the code change.

**Observations (non-blocking):**

1. **Misleading test fixture in the new test** — `existingSuggestion.transformRules.valueFormat` is set to `'html'` but the value is already HAST. The `'html'` value is never actually exercised by the assertion. The test passes because `valueFormat` comes from the converted `newSuggestion` path, not from preserving the existing format. Minor, but worth a note for readability.

2. **Missing edge-case test** — no test for `mergeDataFunction` when `newSuggestion.transformRules` is absent entirely. The `if (converted.transformRules && ...)` guard makes it safe, but an explicit test would lock in that behavior.

3. **Subtle safety in the `isEdited` path** — `existingSuggestion.transformRules.value` is accessed without optional chaining, but it's unreachable when `transformRules` is null due to the `transformRules?.value !== undefined` condition guard above. Safe, just non-obvious.

These are all optional improvements. LGTM.

PR #2392, opened 2026-04-15
fix: use pathname when building site keys and urls

ctinp:

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/utils/cdn-utils.js
+++ b/src/utils/cdn-utils.js
@@ -71,12 +71,20 @@ export function mapServiceToCdnProvider(serviceProvider) {
 }
 
 /**
- * Extracts and sanitizes customer domain from site
+ * Extracts and sanitizes a site-specific key from the base URL.
  */
-export function extractCustomerDomain(site) {
-  const { host } = new URL(site.getBaseURL());
+export function extractSiteKeyFromBaseURL(site) {
+  const { host, pathname } = new URL(site.getBaseURL());
   const cleanHost = host.startsWith('www.') ? host.substring(4) : host;
-  return cleanHost.replace(/[^a-zA-Z0-9]/g, '_').toLowerCase();
+  const normalizedHost = cleanHost.replace(/[^a-zA-Z0-9]/g, '_').toLowerCase();
+  const normalizedPath = pathname
+    .split('/')
+    .filter(Boolean)
+    .map((segment) => segment.replace(/[^a-zA-Z0-9]/g, '_').toLowerCase())
+    .filter(Boolean)
+    .join('_');
+
+  return normalizedPath ? `${normalizedHost}_${normalizedPath}` : normalizedHost;
 }
 
 /**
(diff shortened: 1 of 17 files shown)


src/utils/cdn-utils.js
```diff
@@ -71,12 +71,20 @@ export function mapServiceToCdnProvider(serviceProvider) {
...
+  const { host, pathname } = new URL(site.getBaseURL());
   const cleanHost = host.startsWith('www.') ? host.substring(4) : host;
-  return cleanHost.replace(/[^a-zA-Z0-9]/g, '_').toLowerCase();
+  const normalizedHost = cleanHost.replace(/[^a-zA-Z0-9]/g, '_').toLowerCase();
+  const normalizedPath = pathname
+    .split('/')
+    .filter(Boolean)
+    .map((segment) => segment.replace(/[^a-zA-Z0-9]/g, '_').toLowerCase())
+    .filter(Boolean)
+    .join('_');
```
akshaymagapu: **Collision risk:** the `_` used to join segments is the same character produced by sanitization. So `example.com/us-en` (single segment, dash sanitized to `_`) and `example.com/us/en` (two segments joined with `_`) both produce `example_com_us_en` - the same Athena database/table name for two distinct sites.

Fix: collapse consecutive non-alphanumeric chars per segment, then join with `__`:
```js
.map((segment) => segment.replace(/[^a-zA-Z0-9]+/g, '_').toLowerCase())
.filter(Boolean)
.join('__');
```
With the `+` quantifier, no single sanitized segment can contain `__`, so double underscore becomes an unambiguous segment boundary.

src/utils/cdn-utils.js
```diff
@@ -502,9 +510,20 @@ export async function shouldRecreateTable(
...
     const baseURL = site.getBaseURL();
-    const { host } = new URL(baseURL);
+    const { host, pathname } = new URL(baseURL);
     const rootHost = host.replace(/^www\./, '');
-    return `(REGEXP_LIKE(host, '(?i)^(www.)?${rootHost}$') OR REGEXP_LIKE(x_forwarded_host, '(?i)^(www.)?${rootHost}$'))`;
+    const hostFilter = `(REGEXP_LIKE(host, '(?i)^(www.)?${rootHost}$') OR REGEXP_LIKE(x_forwarded_host, '(?i)^(www.)?${rootHost}$'))`;
+    const normalizedPath = pathname
+      .split('/')
+      .filter(Boolean)
+      .join('/');
```
akshaymagapu: The `split('/').filter(Boolean).join('/')` normalization here duplicates the logic in `extractSiteKeyFromBaseURL` (lines 80-84) with a different join character. If the normalization rules change in one place they won't automatically stay in sync with the other. Consider extracting a shared helper, e.g. `normalizePathname(pathname, separator)`, used by both functions.

src/utils/cdn-utils.js
```diff
@@ -71,12 +71,20 @@ export function mapServiceToCdnProvider(serviceProvider) {
...
-  return cleanHost.replace(/[^a-zA-Z0-9]/g, '_').toLowerCase();
+  const normalizedHost = cleanHost.replace(/[^a-zA-Z0-9]/g, '_').toLowerCase();
+  const normalizedPath = pathname
+    .split('/')
+    .filter(Boolean)
+    .map((segment) => segment.replace(/[^a-zA-Z0-9]/g, '_').toLowerCase())
+    .filter(Boolean)
+    .join('_');
+
+  return normalizedPath ? `${normalizedHost}_${normalizedPath}` : normalizedHost;
```
akshaymagapu, line 97: No guard on key length. This value flows into Athena database names (`cdn_logs_${siteKey}`, +14 char prefix) and table names (`aggregated_logs_${siteKey}_consolidated`, +27 char prefix). A deeply nested base URL path can push these toward AWS Glue Catalog's 255-char limit. Consider logging a warning or asserting when `siteKey.length > 100` so long keys are caught during onboarding rather than silently truncated by AWS.

src/utils/url-utils.js
```diff
@@ -197,10 +197,23 @@ export function joinBaseAndPath(baseURL, path) {
...
-  const normalizedPath = path.startsWith('/') ? path : `/${path}`;
+  try {
+    // Prefer URL parsing when possible, but keep string-join fallback for invalid bases.
+    const base = new URL(baseURL);
+    const normalizedBasePath = base.pathname === '/' ? '' : base.pathname.replace(/\/$/, '');
+    const normalizedPath = path.startsWith('/') ? path : `/${path}`;
+    const joinedPath = normalizedBasePath
+      && (normalizedPath === normalizedBasePath || normalizedPath.startsWith(`${normalizedBasePath}/`))
+      ? normalizedPath
+      : `${normalizedBasePath}${normalizedPath}`;
```
akshaymagapu, line 211: The prefix check is case-sensitive. If a site's base URL has an uppercase path (e.g. `https://example.com/US`) but CDN log URLs are lowercased (Fastly preserves case, other providers may normalize), this check fails for `/us/page` and the result doubles the prefix: `https://example.com/US/us/page`.

Either lowercase both sides before comparison:
```js
normalizedPath.toLowerCase().startsWith(`${normalizedBasePath.toLowerCase()}/`)
```
Or document that base URL paths must be lowercase and enforce that on site registration.

src/utils/url-utils.js
```diff
@@ -197,10 +197,23 @@ export function joinBaseAndPath(baseURL, path) {
...
+
+    return `${base.origin}${joinedPath}`;
+  } catch {
+    const normalizedBase = baseURL.endsWith('/') ? baseURL.slice(0, -1) : baseURL;
```
akshaymagapu, line 217: `base.origin` drops any query string or fragment present in `baseURL` - the previous string-concatenation path preserved them. Site base URLs realistically won't carry query strings, so this is likely fine. If intentional, add a comment here so future readers don't flag it as an accidental regression. If not, preserve them with `${base.origin}${joinedPath}${base.search}${base.hash}`.

src/utils/cdn-utils.js
ctinp: This is quite unlikely, I will skip it for now

src/utils/url-utils.js
ctinp: Added a comment

akshaymagapu (approved):

PR #2397, opened 2026-04-17
fix(prerender): key suggestions on pathname to prevent duplicates after domain shift

ssilare-adobe:
## Problem

After the page-citability migration, prerender suggestions are being duplicated in the UI for ~10,657 entries across 100 paying customers.

**Root cause:** Before the migration, suggestions were keyed on the full URL using `site.getBaseURL()` as the base domain. After the migration, `getPreferredBaseUrl()` can resolve to a different domain (e.g. `context.finalUrl` or `overrideBaseURL` from site config). The same page path then produces two different full URL keys:

```
"https://www.adobe.com/test|prerender"  ← old domain key (existing suggestion)
"https://adobe.com/test|prerender"      ← new domain key (duplicate)
```

`syncSuggestions` treats these as distinct entries, creating duplicates instead of updating the existing one.

Additionally, trailing slash variants of the same path (e.g. `/products` vs `/products/`) could also cause duplicates — normalized as part of this fix.

---

## Fix

All URL comparisons across the prerender audit are now normalized to **pathname** via a shared `toPathname` utility. Trailing slashes are stripped from non-root paths (`/products/` → `/products`, but `/` is preserved).

| Location | Change |
|----------|--------|
| `utils/shared.js` — `toPathname` | New shared utility: extracts pathname, strips trailing slash on non-root paths, falls back to raw string on parse failure |
| `handler.js` — `buildKey` in `processOpportunityAndSuggestions` | Keys suggestions on `pathname\|prerender` instead of full URL |
| `handler.js` — `scrapedUrlsSet` in `processOpportunityAndSuggestions` | Pathname-normalizing wrapper so domain-shifted suggestions are still marked OUTDATED |
| `handler.js` — `deployedAtEdgeUrls` | Set built from pathnames so `markDeployedUrlSuggestionsAsCovered` matches correctly |
| `handler.js` — cleanup-path `buildKey` (no-opportunity path) | Aligned to `pathname\|prerender` keying |
| `guidance-handler.js` — `suggestionsByUrl` Map | Keys and lookups normalized to pathname so Mystique AI summaries attach correctly after domain shift |

---

### Going forward (new audits)

The same page path always resolves to the same key regardless of which domain `getPreferredBaseUrl()` returns or whether a trailing slash is present. `syncSuggestions` will always match and update in place — no new duplicates will be created.

### Existing duplicates

This PR only prevents **new** duplicates. The ~10,657 existing duplicate pairs in production were created before this fix and will **not** self-resolve — both copies keep getting updated on every audit run (their `contentGainRatio`, `wordCount`, S3 keys etc. are refreshed, but `edgeDeployed` state differs between the pair and is preserved separately). A one-time cleanup script is required to remove stale domain-variant rows.

---

## Edge cases covered

| Scenario | Before | After |
|----------|--------|-------|
| `www.adobe.com/test` vs `adobe.com/test` | Duplicate created | Same key `/test\|prerender` |
| `adobe.com/products/` vs `adobe.com/products` | Duplicate created | Same key `/products\|prerender` |
| `https://adobe.com` vs `https://adobe.com/` | Same key `/` | Same key `/` (unchanged) |
| Guidance handler Mystique match after domain shift | AI summary lost | Matched by pathname |
| Outdating after domain shift | Suggestion left as NEW forever | Correctly marked OUTDATED |

---

## Next steps — cleanup of existing duplicates

1. **One-time cleanup script** — for each site, find suggestion pairs that share the same pathname key within an opportunity, keep the most recent/active one, mark the other OUTDATED.
2. **Validation** — run against staging first, verify counts match the CSV in the Slack thread before applying to production.
3. **Monitoring** — after the script runs, confirm the duplicate count drops to 0.

The CSV with the per-site breakdown is in the Slack thread: https://cq-dev.slack.com/archives/C09B49M6TV3/p1776145687666189


anuj-adobe:
Could you pls test on dev using csv method for different cases like -
- suggestion has without www and you add with www
- suggestion has both with and without www, then you add with www
....

ssilare-adobe:
Pushed `e33ffe39` addressing two gaps in this PR:

### 1. `scrapedUrlsSet` exact URL matching (medium severity)

`handleOutdatedSuggestions` in `data-access.js` filters by exact URL match:
```js
const suggestionUrl = existing.getData()?.url;
return suggestionUrl && scrapedUrlsSet.has(suggestionUrl);
```

If an existing suggestion was stored with `https://www.example.com/page` but the current audit scraped `https://example.com/page` (via `getPreferredBaseUrl`), the set wouldn't contain the old domain variant — so old-domain suggestions that **should** be marked outdated were silently preserved.

**Fix:** `processOpportunityAndSuggestions` now wraps `scrapedUrlsSet` in a pathname-normalizing adapter that extracts `new URL(url).pathname` from both stored values and `.has()` lookups. This keeps the change entirely within the prerender handler without modifying the shared `data-access.js`.

### 2. Cleanup-path `buildKey` at line ~1714 used raw URL (low severity)

The "no URLs need prerender" path (`newData: []`) used `buildKey: (s) => s.url`, which was inconsistent with the pathname-based key in the main path. While functionally irrelevant today (since `newDataKeys` is empty), it would break if the logic changes.

**Fix:** Aligned to pathname-based keying: `${new URL(s.url).pathname}|prerender`.

### Tests
- Added new test: `should normalize scrapedUrlsSet to pathname so domain-shifted suggestions are correctly outdated` — verifies that both `https://www.example.com/some/page` and `https://example.com/some/page` match the normalized set, and unscraped paths don't.
- Updated two existing tests that asserted `instanceof Set` to use duck-type checks (`has` property is a function).
- All 226 tests passing, lint clean.

ssilare-adobe:
Pushed `3f1a8b58` addressing two more domain-shift vulnerabilities found during second review:

### 3. `deployedAtEdgeUrls` exact URL match (high severity)

**`handler.js:163`** — `markDeployedUrlSuggestionsAsCovered` checks `deployedAtEdgeUrls.has(data?.url)` to mark NEW suggestions as `coveredByDomainWide`. The set was built from `successfulComparisons` URLs (new domain), but existing DB suggestions may carry the old domain — causing the match to fail silently.

**Fix:** Both the set construction (line ~1770) and the `.has()` lookup (line ~163) now normalize to `new URL(url).pathname`.

### 4. `suggestionsByUrl` Map in guidance handler (medium severity)

**`guidance-handler.js:131,155`** — Suggestions are indexed by full URL in a Map, then Mystique responses are looked up by URL. If the domain shifted between audit and Mystique response, the lookup silently fails and AI summaries never get attached to suggestions.

**Fix:** Both the Map key (line 131) and the `.get()` lookup (line 155) now use a `toPathname()` normalizer.

All 255 prerender tests passing, lint clean.

src/prerender/handler.js
```diff
@@ -1150,6 +1166,45 @@ export async function processOpportunityAndSuggestions(
...
+      && !data?.edgeDeployed
+      && scrapedUrlsSet?.has(data?.url);
+  }).length;
+
+  // Case 5 — new data items with no matching existing suggestion → created fresh
+  const syncCase5Created = allSuggestions.filter(
+    (data) => !existingKeySet.has(buildKey(data)),
+  ).length;
+
   await syncSuggestions({
```
shrotia, line 1243: do we need to pass the suggestions list at syncSuggestions ?

src/prerender/utils/shared.js
```diff
@@ -34,6 +34,22 @@ export function generateReportingPeriods(referenceDate = new Date()) {
...
+ * @returns {string} pathname, or the original string on parse failure
+ */
+export function toPathname(url) {
+  try {
+    const { pathname } = new URL(url);
+    return pathname === '/' ? pathname : pathname.replace(/\/$/, '');
+  } catch {
+    return url;
+  }
+}
```
anuj-adobe: shared.js is not the right file for this utility. I would recommend adding it in util.js.
`shared.js` was created as a utility that wraps other audit's methods.

src/prerender/handler.js
```diff
@@ -1150,6 +1166,45 @@ export async function processOpportunityAndSuggestions(
...
+      && !data?.tokowakaDeployed
+      && !data?.edgeDeployed
+      && scrapedUrlsSet?.has(data?.url);
+  }).length;
+
+  // Case 5 — new data items with no matching existing suggestion → created fresh
+  const syncCase5Created = allSuggestions.filter(
+    (data) => !existingKeySet.has(buildKey(data)),
+  ).length;
+
```
anuj-adobe: are we doing all this specifically for adding logs?

src/prerender/handler.js
```diff
@@ -1703,14 +1760,23 @@ export async function processContentAndGenerateOpportunities(context) {
...
+        );
+        const scrapedUrlsForNoOppty = {
+          has: (url) => scrapedPathnames.has(toPathname(url)),
+        };
         await syncSuggestions({
           opportunity: existingOpportunity,
           newData: [],
           context,
-          buildKey: (suggestionData) => suggestionData.url,
+          buildKey: (suggestionData) => `${toPathname(suggestionData.url)}|${AUDIT_TYPE}`,
```
anuj-adobe: what is the significance of adding ${AUDIT_TYPE} here? Is build key stored in db or just used at runtime? If it is only runtime, we will always have same '| Prerender' suffix for all urls and doesn't make any difference

anuj-adobe (commented):
Please check inline comments

ssilare-adobe:
Addressed all review comments in `90f7e6d7`:

1. **`toPathname` moved to `utils.js`** ([@anuj-adobe](https://github.com/anuj-adobe)) — Removed from `shared.js` and added to `utils/utils.js` where general prerender utilities live. Updated imports in `handler.js` and `guidance-handler.js` accordingly.

2. **Pre-fetched suggestions passed to `syncSuggestions`** ([@shrotia](https://github.com/shrotia)) — `existingSuggestionsForSync` is now passed as `existingSuggestions` to `syncSuggestions`, avoiding the redundant DB round-trip inside it. This also directly answers the question on whether the pre-fetch block is just for logs — it now serves a dual purpose: diagnostics + avoiding the double query.

3. **Cleanup-path `buildKey` simplified** ([@anuj-adobe](https://github.com/anuj-adobe)) — Correct that the `|${AUDIT_TYPE}` suffix is redundant since the key is only used at runtime for diffing and never stored. Simplified to `toPathname(suggestionData.url)`.

anuj-adobe:
> 2\. **Pre-fetched suggestions passed to `syncSuggestions`** ([@shrotia](https://github.com/shrotia)) — `existingSuggestionsForSync` is now passed as `existingSuggestions` to `syncSuggestions`, avoiding the redundant DB round-trip inside it. This also directly answers the question on whether the pre-fetch block is just for logs — it now serves a dual purpose: diagnostics + avoiding the double query.

still seeing this fetch -
```
const existingSuggestionsForSync = await opportunity.getSuggestions();
```

And I am not sure why do we need this much code for monitoring. Can we optimise it by avoiding multiple pass through suggestions array and also encapsulate in a separate method.

anuj-adobe:
Would be best to move the diagnostic code for different cases you added in handler.js in unit tests, that would confirm the hypothesis and confirm the logic.
- Add test/audits/prerender/sync-suggestions.test.js (or extend data-access.test.js) with 3–4 tests using real syncSuggestions + prerender buildKey / mergeDataFunction / pathname scrapedUrlsSet.

image

src/prerender/handler.js
```diff
@@ -1703,14 +1722,23 @@ export async function processContentAndGenerateOpportunities(context) {
...
-        scrapedUrlsForNoOppty.add(getDomainWideSuggestionUrl(site.getBaseURL()));
+        // Include domain-wide URL so aggregate suggestion can be marked outdated
+        // when appropriate. Normalize to pathnames so domain shifts don't prevent
+        // outdating.
+        const scrapedPathnames = new Set(
+          [...scrapedUrlsSet].map(toPathname),
+        );
+        scrapedPathnames.add(
+          toPathname(getDomainWideSuggestionUrl(site.getBaseURL())),
+        );
```
anuj-adobe: why do we need to add domainWideSuggestion in syncing? 
The comment above doesn't seems correct to me " // Include domain-wide URL so aggregate suggestion can be marked outdated
        // when appropriate. Normalize to pathnames so domain shifts don't prevent
        // outdating."

I don't think we need to outdate domain wide suggestion

anuj-adobe:
Regarding your latest commit: https://github.com/adobe/spacecat-audit-worker/pull/2397/changes/1044f0ca81214061e773756e7290ddf67bf7b66f

Is it possible to add tests for processOpportunityAndSuggestions() method, to replace the monitoring code you wrote in handler.js and removed in commit https://github.com/adobe/spacecat-audit-worker/pull/2397/commits/13824253e5f4b14a42b517b4a1ef4a49bda57a63 ?

Pls check https://github.com/adobe/spacecat-audit-worker/pull/2397#issuecomment-4479469547

src/prerender/handler.js
ssilare-adobe: You're right — the comment and the added URL were misleading. The domain-wide suggestion is already protected from being outdated by the `isDomainWide: true` guard in `handleOutdatedSuggestions` (which runs before the `scrapedUrlsSet` filter). Adding the URL to `scrapedPathnames` was redundant dead code. Removed it along with the incorrect comment. Fixed in the latest commit.

ssilare-adobe:
Added `test/audits/prerender/sync-suggestions.test.js` with 5 tests that use the **real `syncSuggestions`** function (no mocking of the sync logic) together with the prerender `buildKey`, `mergeDataFunction`, and pathname-normalized `scrapedUrlsSet`:

- **Case 1** — existing suggestion updated in place when pathname key matches new audit data (even with a different domain prefix)
- **Case 2** — existing NEW suggestion marked OUTDATED when its URL was scraped but is absent from new data
- **Case 3** — existing suggestion preserved when its URL was NOT scraped this run
- **Case 5** — new suggestion created when no matching existing suggestion exists
- **pathname normalization** — domain shift (www → no-www) does not create a duplicate suggestion

These replace the diagnostic logging that was removed in the earlier commit.

src/prerender/handler.js
```diff
@@ -1723,7 +1747,7 @@ export async function processContentAndGenerateOpportunities(context) {
     const deployedAtEdgeUrls = new Set(
       successfulComparisons
         .filter((r) => r.isDeployedAtEdge)
-        .map((r) => r.url),
+        .map((r) => toPathname(r.url)),
```
anuj-adobe, line 1835: changing the url here to pathname but `markDeployedUrlSuggestionsAsCovered` checks for url and not pathnames

ssilare-adobe: The reviewer's concern is not correct. `markDeployedUrlSuggestionsAsCovered` at line 164 normalizes the suggestion's stored URL to a pathname **before** the lookup:

```js
return deployedAtEdgeUrls.has(toPathname(data?.url)) && !data?.edgeDeployed;
```

So `deployedAtEdgeUrls` must contain pathnames for the `.has()` call to match. If we kept raw URLs in the set, `toPathname(data?.url)` (e.g. `/page`) would never equal a full URL (e.g. `https://example.com/page`), and no suggestions would ever be marked as covered. The change to store pathnames is intentional and correct.

src/prerender/handler.js
```diff
@@ -1703,14 +1722,19 @@ export async function processContentAndGenerateOpportunities(context) {
...
+        );
+        const scrapedUrlsForNoOppty = {
+          has: (url) => scrapedPathnames.has(toPathname(url)),
+        };
         await syncSuggestions({
           opportunity: existingOpportunity,
           newData: [],
           context,
-          buildKey: (suggestionData) => suggestionData.url,
+          buildKey: (suggestionData) => toPathname(suggestionData.url),
```
anuj-adobe, line 1822: buildKey method still has - `return `${toPathname(data.url)}|${AUDIT_TYPE}`;`
I think these two areas should be in sync?
So, either keep `${AUDIT_TYPE} ` at both places or remove.

anuj-adobe: I wonder why was this not caught in unit tests that we wrote?

ssilare-adobe: Good catch — both `buildKey` functions are now in sync. Removed the `|AUDIT_TYPE` suffix from `processOpportunityAndSuggestions` so both use pathname-only keying:

```js
buildKey: (data) => toPathname(data.url)
```

As for why the unit test didn't catch it: the test had a comment that explicitly said "no audit type suffix needed at runtime", which baked the wrong assumption in. Updated the test to assert the correct pathname-only format so this stays in sync going forward.

src/prerender/handler.js
ssilare-adobe: No @shrotia , that's correct, I have removed it.

anuj-adobe (approved):

PR #2398, opened 2026-04-17
fix: honor configured cdn provider in cdn analysis

akshaymagapu:
## Summary
- read the configured `cdnProvider` from the latest LLMO config
- filter discovered CDN service providers to the configured provider when it is present
- keep existing discovery behavior when `cdnProvider` is empty or unavailable
- add regression coverage for the mixed `byocdn-other` plus configured-provider path so `scanAndTriggerOnly` does not suppress other configured CDN audits

## Testing
- `npx mocha -i -g 'Post-Deploy' --spec=test/audits/cdn-analysis/handler.test.js`
- `npx mocha -i -g 'Post-Deploy' --spec=test/audits/cdn-logs-report/handler.test.js`
- `npx eslint src/cdn-logs-report/handler.js`


test/audits/cdn-analysis/handler.test.js
```diff
@@ -243,6 +303,194 @@ describe('CDN Analysis Handler', () => {
...
+        auditContext,
+      );
+
+      expect(result.auditResult.providers).to.be.an('array').with.length(0);
+      expect(context.log.warn).to.have.been.calledWith(
+        'Configured cdnProvider=byocdn-cloudflare was not found among discovered service providers: fastly for siteId=test-site-id, host=example.com',
+      );
+    });
+
+    it('keeps all discovered service providers when config cdnProvider is empty', async () => {
```
ctinp, line 469: This basically means that if we fail to read the configured cdn provider, then the old behaviour (the buggy one) would still occur, right?

test/audits/cdn-analysis/handler.test.js
```diff
@@ -243,6 +303,194 @@ describe('CDN Analysis Handler', () => {
...
+      expect(result.auditResult.providers).to.be.an('array').with.length(1);
+      expect(result.auditResult.providers[0]).to.include({
+        serviceProvider: 'fastly',
+        cdnType: 'fastly',
+      });
+      expect(context.log.info).to.have.been.calledWith(
+        'Filtered discovered service providers to configured cdnProvider=byocdn-fastly for siteId=test-site-id, host=example.com',
+      );
+    });
+
```
ctinp, line 371: This fix emerged from the fact that byocdn-other "surpressed" the other cdn providers from running cdn audits. I'd suggest adding an unit test to make sure the problem was indeed fixed

 ```
   it('does not trigger scanAndTriggerOnly when byocdn-other is discovered alongside a different configured provider', async () => {
      const mockedHandler = await loadMockedHandler('aem-cs-fastly');
      const auditContext = {
        year: 2025,
        month: 6,
        day: 15,
        hour: 23,
      };

      context.s3Client.send.callsFake(
        createS3MockForServiceProviders(['byocdn-other', 'aem-cs-fastly'], {
          year: '2025',
          month: '06',
          day: '15',
          hour: '23',
        }),
      );

      const result = await mockedHandler.cdnLogsAnalysisRunner(
        'https://example.com',
        context,
        site,
        auditContext,
      );

      expect(result.auditResult).to.not.have.property('scanAndTriggerOnly');
      expect(result.auditResult.providers).to.be.an('array').with.length(1);
      expect(result.auditResult.providers[0]).to.include({
        serviceProvider: 'aem-cs-fastly',
        cdnType: 'fastly',
      });
      expect(context.sqs.sendMessage).to.not.have.been.called;
      expect(context.log.info).to.have.been.calledWith(
        'Filtered discovered service providers to configured cdnProvider=aem-cs-fastly for siteId=test-site-id, host=example.com',
      );
    });
```

akshaymagapu: thanks, added now

test/audits/cdn-analysis/handler.test.js
akshaymagapu: Yes, that’s correct for the read-failure case.
That was intentional to avoid dropping audits on transient config/S3 lookup issues

ctinp (approved):

PR #2412, opened 2026-04-21
feat: improve Sitemap audit suggestions for redirected pages

mikeweis:
* try to use the terminal "final" page
 * otherwise use the "first hop" page

## Related Issues

Thanks for contributing!


mikeweis:
related Jira ticket: https://jira.corp.adobe.com/browse/SITES-40573

src/sitemap/common.js
```diff
@@ -209,78 +238,106 @@ export async function filterValidUrls(
...
       if (response.status === 404) {
         return { type: 'notOk', url, statusCode: response.status };
       }
 
+      // All remaining status codes
       return { type: 'otherStatus', url, statusCode: response.status };
-    } catch {
+    } catch (err) {
       // exception during the fetch (network error, timeout, etc.) is considered a network error
+      /* c8 ignore next 3 */
```
jsanchez07: This branch is straightforward to cover with nock().replyWithError('...'), the same pattern used in other tests in this file (e.g. the                 
  "broken-redirect" test). Given the PR adds thorough log coverage tests for every other branch, it's worth covering this one too rather than ignoring it. 
  The /* c8 ignore */ should be removed and a test added, something like:                                                                                  
  ```
it('calls log.debug on network error during URL probe', async () => {
  nock('https://example.com').head('/network-err').replyWithError('ECONNRESET');
    const log = { debug: sandbox.spy(), error: sandbox.spy() };                                                                                          
     const result = await filterValidUrls(['https://example.com/network-err'], log);                                                                      
    expect(result.networkErrors).to.have.length(1);                                                                                                      
    expect(log.debug).to.have.been.calledWith(sinon.match(/network error/i));                                                                            
  });          
```

File: src/sitemap/common.js [hunk 2]
```diff
@@ -209,78 +238,106 @@ export async function filterValidUrls(
...
+          + `first hop: ${firstHopUrl}, terminal URL: ${terminalUrl}. No urlsSuggested.`,
+        );
         return {
           type: 'notOk',
           url,
-          statusCode: response.status,
-          // Only suggest homepage if it's valid, otherwise provide no suggestion
-          ...(isHomepageValid && { urlsSuggested: homepageUrl }),
+          statusCode: response.status, // redirected
+          urlsSuggested: '', // no reasonable suggestion available
```
jsanchez07, line 330: The tests assert urlsSuggested: undefined (i.e. property absent) for this case, but the code sets it to '' (empty string). These are not the same —    
  deep.equal will pass if the property is missing but fail if it's '' when undefined is expected, and vice versa. One of them needs to match the other. If 
  the intent is "no suggestion", the property should be omitted entirely.

src/sitemap/common.js
mikeweis: 👍  Will fix.

src/sitemap/common.js
mikeweis: 👍  Will update the unit tests.
- I do intend to use an empty string in the main code: it makes the intent very obvious. If I omitted the property, the code would become obscure.  The `urlsSuggested` property is the natural outcome of the Sitemap audit.

jsanchez07 (approved):

mikeweis:
retest

PR #2415, opened 2026-04-22
fix: toc audit update to scrape client

anagarwa:

## Related Issues

Thanks for contributing!


src/toc/handler.js
```diff
@@ -582,7 +580,9 @@ export async function tocPersister(auditData, context) {
 
 export default new AuditBuilder()
   .withUrlResolver(noopUrlResolver)
-  .withRunner(tocAuditRunner)
+  .addStep('import-top-pages', importTopPages, AUDIT_STEP_DESTINATIONS.IMPORT_WORKER)
+  .addStep('submit-for-scraping', submitForScraping, AUDIT_STEP_DESTINATIONS.SCRAPE_CLIENT)
+  .addStep('process-toc-results', processTocResults)
```
duynguyen, line 621: This final step returns the TOC findings, but the suggestion/opportunity generation is still wired as `withPostProcessors` below. For `StepAudit`, those postprocessors run when the initial step is persisted, not after the scrape-client continuation finishes. I think suggestion generation should be part of the final step itself: process the scrape results, build suggestions, and sync the opportunity from that final-step context.

src/toc/handler.js
```diff
@@ -280,63 +280,77 @@ export async function validatePageTocFromScrapeJson(
...
+  if (auditResult.success === false) {
+    log.warn('[TOC] Audit failed in previous step, skipping scraping');
+    throw new Error('Audit failed, skipping scraping');
   }
-  return validatePageTocFromScrapeJson(url, scrapeJsonObject, log, context);
+
+  const { urls } = await getTocInputUrls(context, site);
+  if (urls.length === 0) {
+    log.warn('[TOC] No URLs to submit for scraping');
+    throw new Error('No URLs to submit for scraping');
```
duynguyen: I would avoid throwing for this case. `urls.length === 0` is a valid no-op state, not a transient failure, so retrying the SQS message will not help. Can we complete the audit early with a no-op / no-opportunities result before entering the scrape-client path, or otherwise route this to a final no-op step instead of throwing?

src/toc/handler.js
```diff
@@ -566,6 +481,116 @@ export function slimTocAuditResult(auditResult) {
...
+        }
+      }
+    });
+
+    log.debug(`Successfully completed TOC Audit for site: ${baseURL}. Found ${totalIssuesFound} issues.`);
+
+    auditResult = totalIssuesFound === 0 ? { toc: {} } : { toc: aggregatedResults };
+
+    // Build suggestions from FULL result (transformRules present) before slimming
+    const auditData = {
```
duynguyen, line 576: This final-step audit data is missing the current audit id. `opportunityAndSuggestions` calls `convertToOpportunity`, and `convertToOpportunity` uses `auditData.id` when creating/updating the Opportunity `auditId`. Without `id: audit.getId()` here, TOC opportunities can be linked to `undefined` instead of the audit record.

src/common/step-audit.js
```diff
@@ -86,6 +86,17 @@ export class StepAudit extends BaseAudit {
     };
 
     if (step.destination === AUDIT_STEP_DESTINATIONS.SCRAPE_CLIENT) {
+      if (!stepResult.urls?.length) {
```
duynguyen: I would avoid making this bypass global for every `SCRAPE_CLIENT` step. Several existing scrape-client audits either expect `scrapeResultPaths` in the final step or leave result persistence to audit-specific code; routing all empty-url scrape steps directly to the final step can leave audits in their initial/preparing state or fail in a different place. Can this be TOC-local or guarded by an explicit opt-in flag on the step result?

anagarwa: Hi @duynguyen 
Without a change in `step-audit.js`, removing the bypass would cause a runtime error. When submitForScraping returns urls: [], chainStep unconditionally     
  calls `ScrapeClient.createScrapeJob(payload).` Inside the scrape-job-supervisor, the very first thing it does is:                                            
                                                                                                                                                             
  `baseURL: composeBaseURL(new URL(urls[0]).host)`,      ([Link](https://github.com/adobe/spacecat-shared/blob/main/packages/spacecat-shared-scrape-client/src/clients/scrape-job-supervisor.js#L58))                                                                                                      
                                                                                                                                                             
  With an empty array, urls[0] is undefined, so new URL(undefined) throws `TypeError: Invalid URL`. This exception propagates up through chainStep. 

This also can't be handled purely inside the TOC handler — chainStep owns the routing decision and the step handler can only control what it returns, not  whether ScrapeClient.createScrapeJob gets called.                                                                
 
So IMO opt in approach is better. The bypassOnEmpty: true opt-in flag keeps this safe and backward-compatible — existing audits that don't set the flag continue to flow through `ScrapeClient` as before. Only steps that explicitly signal they want the bypass (like TOC's submitForScraping when there are no pages) will take the alternate route.

src/toc/handler.js
```diff
@@ -280,148 +280,67 @@ export async function validatePageTocFromScrapeJson(
...
-      },
+      urls: [], siteId: site.getId(), processingType: auditType, bypassOnEmpty: true,
     };
-  } catch (error) {
-    log.error(`TOC audit failed: ${error.message}`);
+  }
+
+  if (topPages.length === 0) {
+    log.warn('[TOC] No URLs to submit for scraping, routing to process-toc-results');
     return {
```
duynguyen: Can we follow the `image-alt-text` step model here instead of routing an empty scrape through the final scrape-processing step? In `image-alt-text`, the step that discovers no pages builds a terminal failed/no-op audit result, persists it with `Audit.updateByKeys(...)`, and returns that result; continuation-step errors also persist the failed status before throwing. That would make TOC clearer in both places: no URLs becomes a persisted terminal state immediately, and `process-toc-results` should persist its catch-path error result the same way it persists success/no-scrape results.

src/toc/handler.js
```diff
@@ -566,6 +485,117 @@ export function slimTocAuditResult(auditResult) {
...
+    // Slim AFTER suggestions are persisted — strips transformRules from the audit DB record
+    const slimmedAuditResult = slimTocAuditResult(auditResult);
+
+    await AuditModel.updateByKeys(
+      { auditId: audit.getId() },
+      { auditResult: slimmedAuditResult },
+    );
+
+    return { fullAuditRef: baseURL, auditResult: slimmedAuditResult };
+  } catch (error) {
```
duynguyen, line 595: Same persistence concern on the final-step error path: this is a `StepAudit` continuation, so returning this error-shaped result does not update the existing audit record. Since this function explicitly persists the no-scrape and success results with `AuditModel.updateByKeys(...)`, the catch path should also build `errorResult` and persist it before returning (or persist and rethrow), similar to the `image-alt-text` continuation-step pattern.

duynguyen (approved):
Approving since the latest version follows the existing image-alt-text continuation-step pattern: no-URL and final-step failure states are explicitly persisted, the shared StepAudit change was removed, and CI is green.

PR #2416, opened 2026-04-22
feat: instrument audit-worker to post lifecycle events to Slack

tkotthakota-adobe:
https://jira.corp.adobe.com/browse/SITES-45625

**Summary**

Audit-worker had no Slack instrumentation, so audits triggered from Slack silently failed (most visibly under bot protection). This adds a Slack notification surface in the base audit classes and threads the                                                                                                                                    originating slackContext through to downstream queue messages so the worker can post on the right thread across the full audit lifecycle.                                                                                                                                 
                                                                                                                                                                                                       
Lifecycle events posted to the originating Slack thread:                                                                                                                                             
  - Audit triggered (already wired upstream)                                                                                                                                                           
  - Per-step progress for StepAudit (auto-humanized step name; non-final
    steps only — the final step is covered by "Audit Completed")                                                                                                                                       
  - Audit completed (RunnerAudit success path; StepAudit's last step)                                                                                                                                  
  - Audit failed (any error caught in RunnerAudit / StepAudit / outer                                                                                                                                  
    dispatcher; deduped via slackContext.notifiedAt marker)                                                                                                                                            
  - Bot-protection fully-blocked abort (rich payload: HTTP status +                                                                                                                                    
    blocker breakdown + sample URLs + IP+UA allowlist instructions)                                                                                                                                    
  - Bot-protection partially-blocked warning (audit continues; warning                                                                                                                                 
    posted but Mystique still receives all URLs)                                                                                                                                                       
                                                                                                                                                                                                       
API surface (minimal by design):                                                                                                                                                                     
  - one primitive: say(env, log, slackContext, message) — no-ops cleanly when not triggered from Slack                                                                                                                                                                      
  - one wrapper that carries real logic: sendAuditFailureNotification —
    bot-protection vs generic discriminator + dedup marker                                                                                                                                             
  - pure formatters for each message type                                                                                                                                                              
  - everything else inlines say(...) + formatX(...) at the call site                                                                                                                                   
                    
**Output examples**                                                                                                                                                                                      
                                                            
Clean CWV audit from Slack:                                                                                                                                                                          
   
  ▎ :arrows_counterclockwise: Collect CWV Data And Import Code — done                                                                                                                                  
  ▎ :white_check_mark: Audit Completed                      
                                                                                                                                                                                                       
Partial bot-protection block (e.g. 3 of 10 URLs blocked, audit continues):

  ▎ :warning: Bot Protection — Partial Block · 3/10 URLs blocked — continuing audit for all URLs                                                                                                       
  ▎ :arrows_counterclockwise: Sync Opportunity And Suggestions Step — done
  ▎ :white_check_mark: Audit Completed                                                                                                                                                                 
                                                            
Full bot-protection block:                                                                                                                                                                           
                                                            
  ▎ :rotating_light: :warning: Bot Protection Detected (rich payload with IP+UA allowlist instructions, no further work sent to Mystique)                                                              
   
Mid-step failure:                                                                                                                                                                                    
                                                            
  ▎ :x: Audit Failed — single message (dedup prevents the outer-catch double-fire)


ramboz (commented):
# PR Review: feat: instrument audit-worker to post lifecycle events to Slack

## Summary
This PR adds a Slack notification surface to the base audit classes (`RunnerAudit`, `StepAudit`, `handleAbort`, and the `index.js` dispatcher) so audits triggered from Slack post lifecycle events — triggered/progress/completed/failed plus bot-protection block details — back to the originating thread. The design is clean and minimal: a single `say()` primitive that no-ops when not Slack-triggered, one `sendAuditFailureNotification()` wrapper carrying the real logic, and pure formatters. It's well-tested (390+ passing on the touched specs, 100% coverage gate holds, lint clean) and I verified the no-op and dedup paths behave as documented. It's close to mergeable; my main concern is a readability bug in `humanizeStepName` for the dominant step-naming convention, plus a fragile dedup mechanism.

**Perspectives applied:** Architecture, SRE/Reliability, QA/Testing, Security (brief), Engineering Practices. Skipped LLM/AI Systems (no LLM calls in the diff).

## Strengths
- **Genuinely non-invasive.** `say()` and `sendAuditFailureNotification()` bail out early when `slackContext.channelId`/`threadTs` are absent, so the overwhelmingly common non-Slack audit path is untouched. This is the right design for retrofitting instrumentation into ~100 handlers.
- **Failures never break audits.** `say()` wraps the client call in try/catch and only logs on error — a Slack outage can't fail an audit. Good reliability posture.
- **Centralized, consistent failure reporting.** One entry point used by all four call sites, reusing the existing `formatAllowlistMessage` from shared-utils and `preserveSlackContext` from `audit-utils.js`, and mirroring the task-processor's `say()` for cross-service consistency.
- **Solid, behavior-focused tests.** 487 new lines in `slack-utils.test.js` cover the happy path, error swallowing, both null and undefined `slackContext`, the `abort.details` fallback, and the dedup no-op — test names that double as documentation. The `humanizeStepName` and formatter tests assert on output, not just code paths.
- **The dedup intent is well-reasoned and commented.** The double-fire problem (inner catch + outer `index.js` catch on the re-thrown error) is real, and the marker approach solves it; the comment in `slack-utils.js` explains it clearly.

## Issues

### Should Fix

**S1. `humanizeStepName` mangles kebab-case step names — the dominant convention in this repo** — `src/utils/slack-utils.js:321`

The function only splits on underscores and camelCase boundaries, and only capitalizes the **first** word. But a survey of `addStep(...)` calls shows the majority of real step names are **kebab-case**: `send-to-mystique`, `run-audit-and-generate-suggestions`, `analyze-and-report`, `submit-for-scraping`, etc. Running the actual function against them:

```
send-to-mystique                   -> "Send-to-mystique"
run-audit-and-generate-suggestions -> "Run-audit-and-generate-suggestions"
analyze-and-report                 -> "Analyze-and-report"
```

So the per-step progress messages — the headline feature of this PR — render with raw hyphens and no title-casing for most audits. The docstring's own snake_case example (`'sync_..._step' -> 'Sync Opportunity And Suggestions Step'`) is also unreachable: snake_case produces `"Sync opportunity and suggestions step"` because only the first word is capitalized. The `humanizeStepName` tests only exercise camelCase (`collectCWVDataAndImportCode`), which is why this slipped through.

Concrete fix — handle both separators and title-case every word:
```js
return stepName
  .replace(/[_-]/g, ' ')                       // snake_case AND kebab-case
  .replace(/([a-z])([A-Z])/g, '$1 $2')
  .replace(/([A-Z]+)([A-Z][a-z])/g, '$1 $2')
  .trim()
  .replace(/\b\w/g, (c) => c.toUpperCase());    // capitalize each word
```
`\b\w` is a no-op on already-uppercase acronyms, so `collectCWVDataAndImportCode` still yields `Collect CWV Data And Import Code`. Please add a kebab-case case to the test (e.g. `send-to-mystique` → `Send To Mystique`).
*Perspective: QA/Testing, Product/Consumer Impact*

**S2. Dedup marker mutates a serializable object that gets propagated to downstream queues** — `src/utils/slack-utils.js:409`

`sendAuditFailureNotification` records the dedup marker by mutating the inbound `slackContext`: `slackContext.notifiedAt = new Date().toISOString()`. That object is *propagated data* — `preserveSlackContext` forwards `slackContext` onto step-continuation and result messages sent to SQS. Today this is safe only because every write to `notifiedAt` happens on a failure/abort path that re-throws or returns without chaining, so the marker never reaches a `chainStep`. But it couples correctness to "failures never chain," and the dedup itself relies on the inner and outer catches sharing the *same* `slackContext` reference (true now because `buildRunnerAuditContext` does a shallow spread — but a future deep-clone of `auditContext` would silently re-introduce double alerts).

Both inner audit classes and the `index.js` catch share the same `context` object reference within an invocation. Tracking the marker there (e.g. `context.__slackFailureNotified`) keeps the transient invocation-state off the serializable payload and removes the reference-sharing fragility. Worth at least a code comment if you keep the current approach.
*Perspective: Architecture, SRE/Reliability, AI-Native Maintainability*

### Nice to Have

**N1. `say()` overlaps with the existing `postMessageOptional()` in the same file** — `src/utils/slack-utils.js:26`

`say(env, log, slackContext, message)` and the pre-existing `postMessageOptional(context, channelId, text, options)` both guard on `hasText(channelId) && hasText(threadTs)`, build a `WORKSPACE_INTERNAL` client, and post safely. The new one differs by signature, a curated `env` subset, and `unfurl_links: false`. The stated rationale (mirror task-processor's `say`) is reasonable, but two near-duplicate Slack senders in one file invites drift. Consider implementing `say()` in terms of `postMessageOptional`, or adding a one-line cross-reference comment noting why both exist.
*Perspective: Architecture*

**N2. Scope: unrelated test changes bundled in, unmentioned in the description** — `test/audits/product-metatags.test.js`, `test/audits/canonical.test.js`

The PR removes 106 lines from `product-metatags.test.js` and adds `robots.txt` nocks to two `canonical.test.js` tests. These are harmless — the removed product-metatags cases were redundant (the truthy `scrapeResultPaths?.size` branch is still covered by the "missing site config" test at `product-metatags.test.js:3229`, and `npm test` passes the 100% gate), and the canonical nocks look like flakiness stabilization for the `processScrapedContent` step's robots fetch. But they're invisible churn relative to "instrument audit-worker for Slack." A one-line note in the PR description ("also de-duplicates product-metatags tests / stabilizes canonical robots nock") would save the next reviewer the dig.
*Perspective: Engineering Practices*

**N3. Dedup is unit-tested but not at the integration boundary** — `test/utils/slack-utils.test.js:545`

The dedup test re-uses one `slackContext` object across two direct calls — it proves the marker logic but assumes the reference-sharing that S2 flags. There's no test asserting that a `RunnerAudit`/`StepAudit` failure followed by the `index.js` catch produces exactly one Slack post end-to-end. Adding one would lock in the behavior the whole dedup exists for, and would fail loudly if someone later clones `auditContext`.
*Perspective: QA/Testing*

## Out of Scope, Worth Tracking
- **slackContext propagation to ad-hoc Mystique fan-out is CWV-only.** Every audit's *result* message (via `base-audit.js`) and step continuations (via `preserveSlackContext`) already carry `slackContext`, but only `cwv/auto-suggest.js` was updated for per-suggestion Mystique messages. Other handlers that build their own Mystique SQS payloads (image-alt-text, forms, internal-links, …) won't thread it through. Likely intentional CWV-first scoping — worth a tracking ticket so the lifecycle coverage is known to be partial.

## Verdict
**Ready to merge:** With fixes
**Reasoning:** Sound, well-tested design that's safe for the non-Slack path; fix the `humanizeStepName` kebab-case bug (S1) since it degrades the feature's primary output for most audits, and ideally relocate the dedup marker off the propagated object (S2) before merge.

*For a deeper dedicated security pass you could run Adobe's security-foundations review, but I don't think it's warranted here — the only outbound data is audit error text and blocked URLs posted to an internal Slack workspace, within the same trust domain as whoever triggered the audit, and no secrets are emitted (the env tokens are passed to the client, never into message text).*

---
*Review method: fetched the PR, read the full changed files at the PR head, and verified in an isolated worktree — ran the full suite (`npm test`, exit 0, 100% coverage gate holds), the touched specs directly (390 passing), `eslint` (clean), and exercised `humanizeStepName` against real step names. Generated with [Claude Code](https://claude.com/claude-code).*

ramboz (approved):

solaris007 (commented):
Hey @tkotthakota-adobe,

**Verdict: Comment** - sound, well-tested instrumentation that's correctly a no-op on the non-Slack path; one real-but-cosmetic defect (already flagged by @ramboz) plus cleanup and scope notes, none blocking a merge a maintainer has already approved.
_Changes: adds Slack lifecycle instrumentation to the base audit classes - a `say()` no-op-when-not-Slack primitive, a `sendAuditFailureNotification()` wrapper, and pure message formatters - so Slack-triggered audits post per-step / completion / failure and bot-protection events back to the originating thread (14 files)._
_Scope note: the diff also deletes 106 lines from `product-metatags.test.js` and adds `robots.txt` nocks in `canonical.test.js`, unrelated to Slack instrumentation and unmentioned in the description (verified harmless - the product-metatags branch stays covered by the test near `test/audits/product-metatags.test.js:3233`, and the canonical nocks are flakiness stabilization)._

<details>
<summary>Non-blocking (10): the kebab-case bug, cleanups, and test-quality notes</summary>

- suggestion: `humanizeStepName` renders kebab-case step names with raw hyphens and capitalizes only the first word - and kebab-case is the dominant convention (measured: 53 of 82 `addStep` names in `src/` are kebab-case; `send-to-mystique` -> "Send-to-mystique", `run-audit-and-generate-suggestions` -> "Run-audit-and-generate-suggestions"). The headline per-step messages look broken for most audits. Reinforces @ramboz's S1, still unaddressed at head - fix with `.replace(/-/g, ' ')` plus per-word capitalization (`/\b\w/g`) and add a kebab-case test - `src/utils/slack-utils.js:326`.
- suggestion: the new `say()` overlaps the pre-existing `postMessageOptional()` in the same file (both guard channelId+threadTs, default WORKSPACE_INTERNAL, catch-and-log) - implement one over the other or add a cross-reference so they don't drift (reinforces @ramboz's N1) - `src/utils/slack-utils.js:324`.
- suggestion: the dedup marker mutates the inbound `slackContext` (`slackContext.notifiedAt = ...`), which `preserveSlackContext` propagates onto downstream SQS payloads - transient invocation state on a wire object, safe today only because failure paths don't chain; prefer an invocation-scoped flag on `context` (reinforces @ramboz's S2) - `src/utils/slack-utils.js:635`.
- suggestion: best-effort progress `say()` calls are `await`ed inline on the step-chaining critical path, so a slow/rate-limited Slack response can stall the audit and extend Lambda billed duration (Slack-triggered path only) - consider fire-and-forget with `.catch(log...)` - `src/common/step-audit.js:223`, `src/common/runner-audit.js:88`.
- nit: the `let siteUrl = siteId; try { siteUrl = site.getBaseURL(); } catch {}` block is copy-pasted in both base classes - extract a shared `resolveSiteUrl(site, siteId)` - `src/common/runner-audit.js`, `src/common/step-audit.js`.
- nit: `handleAbort`'s new `site.getBaseURL()` call is unguarded while the rest of the PR wraps the same call in try/catch - `src/common/bot-detection.js:75`.
- nit: `slackContext` threading is hand-rolled (`...(slackContext && {...})`) where the existing `preserveSlackContext` helper already does it - `src/common/base-audit.js`, `src/cwv/auto-suggest.js`.
- nit: emoji style is mixed - raw Unicode in the formatters vs Slack shortcodes in the scaffolding; pick one (shortcodes render most reliably) - `src/utils/slack-utils.js`.
- nit: the `sendAuditFailureNotification` docstring claims it makes "every audit/opportunity" surface failures, but only RunnerAudit/StepAudit routes are wired - soften to match reality - `src/utils/slack-utils.js:580`.
- suggestion: the new `index.test.js` "uses site base URL..." test asserts only `status === 500` (same as the test above it), doesn't stub `slack-utils` or set a `slackContext`, so it verifies nothing about the base-URL path it names; and dedup is only unit-tested by reusing one object - add an end-to-end test through the `index.js` re-throw asserting exactly one post (reinforces @ramboz's N3) - `test/index.test.js:88`.
</details>

- **Strategic (the main one):** audit-worker is on the path to the Mystique migration - investing ~1000 lines of new lifecycle instrumentation here is worth a conscious "yes, on purpose" call rather than a default. If audit-worker stays the execution engine near-term, the operator visibility is genuinely useful; if not, this is effort that may be rebuilt or removed post-migration. Good idea on the merits - just worth confirming it's being invested in the right place.
- **Lifecycle visibility is CWV-only:** only `cwv/auto-suggest.js` threads `slackContext` onto its Mystique messages; ~15 other handlers build their own Mystique payloads and won't, so the "all audits" framing over-promises. Scope the docs to CWV or lift the threading into the shared message-construction seam.
- **Three-way formatter duplication:** `say()` + the bot-protection formatters are now a third verbatim copy alongside `spacecat-task-processor` and `spacecat-api-service`. Promote to a shared package (next to `formatAllowlistMessage`) or add a canonical-copy reference + tracked follow-up before the copies drift.
- **CLAUDE.md docs follow-ups:** no spec in `docs/specs/` and no ADR in `docs/decisions/` for this cross-cutting feature / the dedup-marker decision, and the README `auditContext` shape omits the new `slackContext` field. A retroactive follow-up is acceptable per the rule's "specs reflect implemented state."
- **Producer-side trust (security read otherwise clean):** `slackContext` is settable by the SQS producer, but this PR is more constrained than the existing `detect-cdn` path (hardcodes WORKSPACE_INTERNAL), so no new exposure; any per-tenant validation belongs producer-side. Tokens never enter message text, and `error.message` posted to the internal workspace is the same trust level as the existing logs.

tkotthakota-adobe:
Hi @ramboz Thank you for the review. Fixed.

S1 — humanizeStepName kebab-case bug. Fixed: separator regex now handles [_-], title-case applied to every word via \b\w. Acronym preservation still works (collectCWVDataAndImportCode → Collect CWV Data And Import Code). Added tests for kebab-case (send-to-mystique → Send To Mystique, run-audit-and-generate-suggestions) and snake_case (sync_opportunity_and_suggestions_step → Sync Opportunity And Suggestions Step). Docstring updated with all three input styles.                                

S2 — dedup marker off the propagated payload. Fixed: marker moved from slackContext.notifiedAt to context.__slackFailureNotifiedAt. The context object is the Lambda invocation handle shared by every catch in one invocation; never serialized; never propagated to SQS. Removes coupling to reference-sharing of slackContext and the "failures-never-chain" assumption. Comment in sendAuditFailureNotification explains the choice. Existing dedup test updated to use a  cloned slackContext on the second call — proves dedup works across structural copies (the exact future-refactor risk S2 flagged).

Nice-to-Have                                                                                                                                                 
  
N1 — say vs postMessageOptional overlap. Added an explicit cross-reference comment at the top of say documenting why both exist and when to use which.  Light-touch fix per the reviewer's suggested "or a one-line cross-reference comment."
                                                                                                                                                               
N3 — dedup tested only at unit boundary. The updated test now uses a cloned slackContext between calls, which is what S2 was guarding against. This is the integration-equivalent for the dedup invariant. A true end-to-end RunnerAudit/StepAudit→outer-dispatcher test would be more invasive (requires esmock  + slack-client stub orchestration); the cloned-payload test catches the same regression with much less plumbing.                                               
                                                         
To address in the PR description (not code)

N2 — unrelated test changes. Adding a one-liner to the PR description acknowledging:                                                                         
  - test/audits/product-metatags.test.js — removed redundant coverage cases (the scrapeResultPaths?.size truthy branch is still covered by the "missing site config" test at line 3229; 100% gate still passes)                                                                                                           
  - test/audits/canonical.test.js — added robots.txt nocks to stabilize flaky processScrapedContent step tests
                                                                                                                                                               
Out of scope (tracking ticket)                                                                                                                               
                                                                                                                                                               
slackContext Mystique fan-out is CWV-only. Other handlers that build their own per-suggestion Mystique SQS payloads (image-alt-text, forms, internal-links,  etc.) don't thread slackContext through. Audit-result message and step continuations already do (via base-audit.js + preserveSlackContext), but ad-hoc fan-out handlers won't. Worth a follow-up ticket — same pattern as the auto-suggest.js one-liner addition per handler.

tkotthakota-adobe:
> @/tmp/review-body-2416.md

Hi @solaris007 I am not seeing review comments on this PR. Looking forward to get them.

tathagat2241:
Overview
Adds a Slack notification surface so audits triggered from Slack post lifecycle events (triggered / per-step progress / completed / failed / bot-protection blocked) back to the originating thread. New src/utils/slack-utils.js exposes one primitive (say), one logic-carrying wrapper (sendAuditFailureNotification), and pure formatters; the audit base classes (runner-audit, step-audit), bot-detection.handleAbort, the outer index.js dispatcher, and cwv/auto-suggest thread slackContext through to downstream SQS messages. I reviewed it through the lens of (a) does slackContext propagate intact, (b) is the failure-notification dedup actually correct, and (c) does the happy path stay zero-cost for non-Slack audits — so I traced the call sites and the shared context reference, not just the diff.

Blockers
None. The dedup design is sound: index.js:336 passes the same context object into handler.run(); StepAudit/RunnerAudit set context.slackFailureNotifiedAt before re-throwing, and the outer catch's sendAuditFailureNotification hits that guard → exactly one alert. Verified end-to-end, and there's a unit test for it (slack-utils.test.js:545).
Should-fix
src/common/step-audit.js:159 — the abort path calls handleAbort(..., { ...context, auditContext }), a throwaway shallow copy. sendAuditFailureNotification then sets the slackFailureNotifiedAt dedup marker on that copy, not the shared context. It's harmless today (handleAbort returns a response and never re-throws, so nothing double-fires), but it's a latent trap: if handleAbort ever re-throws or a second notify is added, dedup silently breaks. Prefer mutating/ passing the real context (e.g. set context.auditContext = auditContext and pass context).
src/common/runner-audit.js:88 & src/common/step-audit.js:266 — await sendAuditFailureNotification(...) runs unguarded inside the catch. say() swallows its own errors, but the formatters and formatAllowlistMessage(botIps) execute before say and aren't wrapped — a formatter/exception there would replace the original audit error with a Slack-formatting error, masking the root cause in logs and the re-thrown cause. Wrap the notify call in try/catch (at the call site or inside sendAuditFailureNotification).
Nits
src/common/step-audit.js:265 (and the PR description) say failures are "deduped via slackContext.notifiedAt", but the real marker is context.slackFailureNotifiedAt. The slack-utils.js:408 doc is correct and even explains why it's on context not slackContext — just align the stale comment + PR text.
Per-step say() is awaited inside each step, so a Slack-triggered audit now serializes a Slack POST per non-final step. Intended and fine — worth noting only that a slow Slack API adds to audit latency on that path. Non-Slack audits are unaffected because say() early-returns before any network call (good).
Strengths
say()'s early no-op when channelId/threadTs are absent means zero added cost for non-Slack-triggered audits — the right default, and well-tested.
Genuinely thorough docs: the say vs postMessageOptional rationale block and the dedup-marker-placement comment answer the exact "why" questions a reviewer would raise.
Formatters kept compatible with the task-processor's equivalents for consistent output across services; humanizeStepName correctly preserves acronym runs (CWVData → CWV Data).
Tests
Strong unit coverage on the new module — test/utils/slack-utils.test.js (659 lines) exercises every formatter, say's no-op/error-swallow paths, the bot-protection rich/partial messages, and crucially the double-fire dedup (line 545). Caveat I could verify directly: the wiring tests in runner-audit.test.js/step-audit.test.js stub sendAuditFailureNotification or use auditContext: {} to short-circuit it, so the success-path completion say(...) integration is asserted more lightly than the util itself. Worth one wiring test that asserts the completion/step messages actually fire on the happy path.

Verdict
Approve with minor notes. Solid, well-documented, well-tested feature with a correct dedup design; the two should-fix items are robustness hardening (error masking + fragile marker placement on the abort path), neither merge-blocking.

solaris007 (commented):
@tathagat2241 i have fixed the missing review

tkotthakota-adobe:
@solaris007 @tathagat2241 addressed review comments. Thank you for the review

src/utils/slack-utils.js
```diff
@@ -11,7 +11,70 @@
...
+ * @param {string} message - The message text to send
+ * @returns {Promise<void>}
+ */
+export async function say(context, slackContext, message) {
+  const { log } = context;
+  // Temporary diagnostic — fires before any early-return so it proves whether
+  // this code path is reached at all on a Slack-triggered audit. Remove once
+  // we confirm Slack messages are landing.
+  if (log?.info) {
+    log.info('[slack-debug] say() called', {
```
ramboz, line 49: S1 (Should Fix): This `[slack-debug] say() called` log fires before the `channelId`/`threadTs` early-return, so it runs on every audit that reaches a `say(...)` call site — including all non-Slack audits, since the completion `say(...)` is on the success path of both RunnerAudit and StepAudit. That's fleet-wide INFO noise that the zero-cost non-Slack design was meant to avoid, and the comment says to remove it once Slack posting is confirmed. Please drop it (or gate behind a debug level) before merge. (No secret leak — booleans only.)

ramboz (commented):
## Summary

Re-review since my approval at `e142cb8`. The head has moved and the author has worked through the prior review round. I re-verified the full diff fresh-eyes rather than trusting the changelog.

Everything I flagged before is genuinely fixed: `humanizeStepName` now handles kebab-case, snake_case, and camelCase and title-cases every word while preserving acronym runs (S1); the dedup marker moved off the serializable `slackContext` onto the in-memory `context.slackFailureNotifiedAt` (S2); the `say` vs `postMessageOptional` rationale comment is in place (N1); and the stale "deduped via slackContext.notifiedAt" comment now correctly names `context.slackFailureNotifiedAt`. The two should-fix items other humans raised are also resolved: `sendAuditFailureNotification` wraps its formatters in try/catch so a formatter throw can't mask the original audit error, and the `handleAbort` abort path now passes the live `context` (with `context.auditContext = auditContext`) instead of a throwaway shallow copy, so the dedup marker lands on the shared object.

One new issue slipped in with the reworked `say()`: a self-described temporary debug log. Everything else is in good shape.

**Perspectives applied:** Logging/Observability, Architecture, SRE/Reliability, QA/Testing, Security (brief).

## Resolution status (prior findings)

- **S1 — `humanizeStepName` kebab-case** — Fixed. New impl + kebab/snake tests.
- **S2 — dedup marker on propagated payload** — Fixed. Moved to `context.slackFailureNotifiedAt`; commented; dedup test now uses a cloned `slackContext`.
- **N1 — `say`/`postMessageOptional` overlap** — Fixed. Cross-reference comment added.
- **N2 — unrelated test churn undocumented** — Addressed in PR discussion.
- **N3 — dedup only unit-tested** — Partially addressed (cloned-payload test). A true end-to-end wiring test is still absent; already noted by other reviewers, not re-raising.
- **Human (tathagat2241) — formatter error masking** — Fixed (try/catch in `sendAuditFailureNotification`).
- **Human (tathagat2241) — `handleAbort` throwaway shallow copy** — Fixed (live `context` passed).
- **Human (solaris007) nits** — awaited `say()` latency, copy-pasted `resolveSiteUrl`, unguarded `getBaseURL` in `handleAbort`, mixed emoji style, docstring over-promise, weak `index.test.js` — remain as-is; these were human-flagged nits, so I'm not re-raising them.

## Issues

### Should Fix

**S1. Leftover temporary debug log in `say()` fires on every audit** — `src/utils/slack-utils.js:49`

The reworked `say()` opens with a diagnostic that is emitted *before* the non-Slack early-return:

```js
if (log?.info) {
  log.info('[slack-debug] say() called', { hasChannelId, hasThreadTs, hasWorkspaceToken, hasOpsChannel });
}
```

Because it runs before the `channelId`/`threadTs` guard, it fires on **every** audit that reaches a `say(...)` call site — and the completion `say(...)` is on the success path of both `RunnerAudit` and `StepAudit`, so effectively every audit in the fleet now logs `[slack-debug] say() called` at INFO, not just Slack-triggered ones. That is exactly the log noise the "zero-cost non-Slack path" design was built to avoid. The comment itself says "Remove once we confirm Slack messages are landing," which confirms it's a merge-blocker-by-intent leftover rather than permanent instrumentation. No secret is leaked (booleans only), so this is noise/hygiene, not security. Please drop it (or gate it behind a debug level) before merge.
*Perspective: Logging/Observability*

## Verdict

**Ready to merge:** With one small fix — remove the temporary `[slack-debug]` log (S1). Every prior finding from me and the other reviewers is resolved; the design is sound and the non-Slack path is a clean no-op once the diagnostic is gone. Not blocking the merge over a stray log line, hence a comment rather than a change request.

*This is an automated review.*

PR #2417, opened 2026-04-22
fix: Readability - Exclude Invalid content

noruiz:

## Related Issues
https://jira.corp.adobe.com/browse/LLMO-4147

Thanks for contributing!


src/readability/opportunities/guidance-handler.js
```diff
@@ -195,6 +232,20 @@ export default async function handler(message, context) {
...
+          shouldExclude: true,
+          exclusionReason: newData.exclusionReason,
+          suggestionStatus: 'excluded',
+          mystiqueProcessingCompleted: new Date().toISOString(),
+        };
+        delete merged.category;
+        delete merged.seoImpact;
+        delete merged.transformRules;
+        delete merged.improvedText;
+        return enrichExcludedReadabilitySuggestionData(merged);
```
baranovskyalexandr, line 247: enrichExcludedReadabilitySuggestionData overwrites url with data.pageUrl. If the existing DB suggestion was previously enriched (url was set, pageUrl was deleted), then merged.pageUrl is undefined and url gets silently overwritten with undefined. Verify that pageUrl always survives in existingData.

src/readability/shared/analysis-utils.js
```diff
@@ -354,15 +355,32 @@ export async function analyzeTextReadability(
  * @param {Cheerio} $ - The Cheerio object to search for text elements.
  * @returns {Element[]} Array of meaningful text elements for readability analysis and enhancement.
  */
+const NAV_CLASS_PATTERNS = ['nav', 'menu', 'breadcrumb', 'filter', 'pagination', 'sidebar', 'review', 'rating', 'promo', 'banner', 'cookie'];
```
baranovskyalexandr: 'review' and 'rating' match via substring — a <p class="product-review-text"> or <p class="customer-rating-summary"> would be silently filtered out as navigation chrome. These are legitimate editorial paragraphs. Consider restricting to word-boundary matches or removing the most aggressive patterns.

baranovskyalexandr (approved):

PR #2422, opened 2026-04-23
Fix: skip reconcile disappeared suggestions for TBYB sites

absarasw:
## Problem
TBYB (Try Before You Buy) sites, identified by the `summit-plg` handler being
enabled, have access to a limited view of suggestions. Running "Reconcile
Disappeared Suggestions" for these sites risks marking suggestions as FIXED and
creating fix entities for issues the user has never seen, which is misleading.

## Change
In `syncSuggestionsWithPublishDetection`, skip Step 1 (Reconcile Disappeared
Suggestions) when the site has the `summit-plg` handler enabled. The check
is only performed when the reconcile callbacks are provided, so there is no
overhead for callers that don't use publish detection. If the configuration
lookup fails, a warning is logged and reconcile proceeds normally as a safe
fallback.

## Tests
- TBYB site → reconcile step is skipped
- Non-TBYB site → reconcile step runs as before
- `Configuration.findLatest` throws → logs warning, reconcile still runs

## Related Issues

Thanks for contributing!

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "start": "nodemon",
-    "test": "true",
+    "test": "c8 mocha -i -g 'Post-Deploy' --spec=test/**/*.test.js",
     "test:spec": "c8 mocha -i -g 'Post-Deploy'",
     "test:bundle": "HELIX_TEST_BUNDLE_NAME=dist/spacecat-services/audit-worker@$(npm pkg get version | tr -d \\\")-bundle.mjs mocha --spec=test/index.test.js",
     "test-postdeploy": "mocha -g 'Post-Deploy' --spec=test/**/*.test.js",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "scripts": {
     "start": "nodemon",
-    "test": "c8 mocha -i -g 'Post-Deploy' --spec=test/**/*.test.js",
+    "test": "true",
     "test:spec": "c8 mocha -i -g 'Post-Deploy'",
     "test:bundle": "HELIX_TEST_BUNDLE_NAME=dist/spacecat-services/audit-worker@$(npm pkg get version | tr -d \\\")-bundle.mjs mocha --spec=test/index.test.js",
     "test-postdeploy": "mocha -g 'Post-Deploy' --spec=test/**/*.test.js",
--- a/src/backlinks/handler.js
+++ b/src/backlinks/handler.js
@@ -294,9 +294,11 @@ export const generateSuggestionData = async (context) => {
     || !Array.isArray(auditResult.brokenBacklinks)
     || auditResult.brokenBacklinks.length === 0) {
     log.info(`No broken backlinks found for ${site.getId()}, skipping opportunity creation`);
+    /*
     return {
       status: 'complete',
     };
+    */
   }
 
   const kpiDeltas = await calculateKpiMetrics(audit, context, site);
--- a/src/backlinks/handler.js
+++ b/src/backlinks/handler.js
@@ -294,11 +294,9 @@ export const generateSuggestionData = async (context) => {
     || !Array.isArray(auditResult.brokenBacklinks)
     || auditResult.brokenBacklinks.length === 0) {
     log.info(`No broken backlinks found for ${site.getId()}, skipping opportunity creation`);
-    /*
     return {
       status: 'complete',
     };
-    */
   }
 
   const kpiDeltas = await calculateKpiMetrics(audit, context, site);
--- a/src/image-alt-text/handler.js
+++ b/src/image-alt-text/handler.js
@@ -18,6 +18,7 @@ import {
   SCRAPE_MAX_AGE_HOURS, SCRAPE_PAGE_LOAD_TIMEOUT, ALT_TEXT_PROCESSING_ERROR_TAG,
 } from './constants.js';
 import { getTopPageUrls } from './url-utils.js';
+import { SUMMIT_PLG_HANDLER } from '../utils/data-access.js';
 
 const AUDIT_TYPE = AuditModel.AUDIT_TYPES.ALT_TEXT;
 const { AUDIT_STEP_DESTINATIONS } = AuditModel;
@@ -49,7 +50,7 @@ async function getTopPagesLimit(site, context) {
   const onDemand = rawOnDemand === true || rawOnDemand === 'true';
   const { Configuration } = dataAccess;
   const configuration = await Configuration.findLatest();
-  const summitPlgFromConfig = configuration.isHandlerEnabledForSite('summit-plg', site);
+  const summitPlgFromConfig = configuration.isHandlerEnabledForSite(SUMMIT_PLG_HANDLER, site);
   const isSummitPlgEnabled = !onDemand && summitPlgFromConfig;
   const pageLimit = isSummitPlgEnabled ? SUMMIT_PLG_PAGE_LIMIT : DEFAULT_PAGE_LIMIT;
   log.debug(`[${AUDIT_TYPE}]: Page limit set to ${pageLimit} (summit-plg active: ${isSummitPlgEnabled}, onDemand: ${onDemand})`);
--- a/src/utils/data-access.js
+++ b/src/utils/data-access.js
@@ -19,6 +19,8 @@ import { limitConcurrencyAllSettled } from '../support/utils.js';
 // Max concurrent HTTP calls to prevent Lambda timeout (15 min)
 const MAX_CONCURRENT_CHECKS = 5;
 
+export const SUMMIT_PLG_HANDLER = 'summit-plg';
+
 /**
  * Opportunity types that only make changes in the author environment (no publish state).
  * For these types:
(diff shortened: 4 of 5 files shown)


sandsinh:
**Suggestion: use entitlements instead of `Configuration` handler check**

The TBYB detection here goes through `Configuration.findLatest()` + `isHandlerEnabledForSite('summit-plg', site)`, but the canonical source of truth for TBYB/PLG status is the entitlement tier. The codebase already has an established pattern for this in `src/utils/site-validation.js`:

```js
import { TierClient } from '@adobe/spacecat-shared-tier-client';
import { Entitlement } from '@adobe/spacecat-shared-data-access';

const tierClient = await TierClient.createForSite(context, site, Entitlement.PRODUCT_CODES.ASO);
const { entitlement } = await tierClient.checkValidEntitlement();
isTbybSite = entitlement?.getTier?.() === Entitlement.TIERS.PLG;
```

This avoids the magic string `'summit-plg'`, stays consistent with how the rest of the worker gates behavior, and uses the intended data model (`SiteEnrollments` + `Entitlements`) rather than the general-purpose config bag.

More broadly, it would be worth tracking a follow-up to remove the `summit-plg` handler from `Configuration` entirely and migrate all TBYB detection across the codebase to the entitlement-driven approach. The handler-in-config pattern is an infra signal that should be superseded by the entitlement tier now that that model is in place.

sandsinh (approved):

tarunsinghdev:
Hey @absarasw,                                                                                           
                                                                                                           
  Strengths                                                                                                
                                                                                                           
  - Fail-open default is the right call (data-access.js:802). If site is missing, Configuration is missing,
   or findLatest() throws, isTbybSite stays false and reconcile proceeds normally. The existing behavior is
   the safe baseline.                                                                                      
  - Defense-in-depth guard clauses: checks site, then Configuration, then wraps in try/catch. Three layers
  of null-safety before the business logic.                                                                
  - Tight change scope - two files, one behavior change, no unrelated modifications.
  - Good test coverage (data-access.test.js:3026-3140). Three tests hit the three meaningful states: TBYB  
  skip, non-TBYB proceed, and findLatest failure fallback. Each verifies downstream effects, not just the  
  flag value.                                                                                              
  - No overhead for callers that don't use publish detection - the TBYB check only runs inside the existing
   callback guard.                                                                                         
   
  Issues                                                                                                   
                                                                                                         
  Important (Should Fix)                                                                                   
   
  1. Duplicate Configuration.findLatest() call - same data, fetched twice per audit run                    
                                                                                                         
  src/utils/data-access.js:807 (new) and src/utils/data-access.js:~440 (existing in syncSuggestions)       
                                                                                                         
  Both call Configuration.findLatest() + isHandlerEnabledForSite('summit-plg', site). Since                
  syncSuggestionsWithPublishDetection calls syncSuggestions at Step 3, the same configuration is fetched 
  twice per invocation. Beyond the redundant async call, the two implementations diverge in error handling:
   the new code wraps in try/catch with a graceful fallback, while the existing code at line ~440 does a 
  bare await Configuration.findLatest() with no try/catch (it will throw and abort the entire sync on DB
  errors). If one approach is correct, the other is wrong.

  Fix: Extract a shared helper like isSummitPlgSite(context) that encapsulates the lookup, handles errors  
  consistently, and is used in both places. Alternatively, compute the flag once at the top of
  syncSuggestionsWithPublishDetection and pass it through to syncSuggestions.                              
                                                                                                         
  2. Naming inconsistency: isTbybSite vs isSummitPlg                                                       
   
  src/utils/data-access.js:802 (new) vs src/utils/data-access.js:~437 (existing)                           
                                                                                                         
  Both check configuration.isHandlerEnabledForSite('summit-plg', site). Two different names for the same   
  boolean in the same file makes the code harder to reason about. Use isSummitPlg consistently (the
  established name), or rename both in a follow-up.                                                        
                                                                                                         
  3. Unguarded context.dataAccess destructuring

  src/utils/data-access.js:804                                                                             
   
  The code does const { Configuration } = context.dataAccess without a null check on context.dataAccess.   
  But later in the same function (~line 828), the code uses context.dataAccess?.Suggestion with optional 
  chaining, acknowledging that dataAccess can be undefined. If dataAccess is null/undefined, the           
  destructuring throws a TypeError outside the try/catch, crashing the entire                            
  syncSuggestionsWithPublishDetection call.

  Fix: const Configuration = context.dataAccess?.Configuration;                                            
   
  4. Should Step 2 (publish fix entities) also be skipped for TBYB sites?                                  
                                                                                                         
  The PR description says TBYB sites "have access to a limited view of suggestions" and that creating fix  
  entities "for issues the user has never seen" is misleading. But Step 2 (publishDeployedFixEntities)
  still runs and can transition fix entities to PUBLISHED status. If a fix entity was created before this  
  PR deployed, it could still be published for a TBYB site. If Step 2 is intentionally left running, a code
   comment explaining why would prevent future confusion.

  Minor (Nice to Have)

  5. Magic string 'summit-plg' used in multiple locations

  src/utils/data-access.js:808 - The handler name 'summit-plg' now appears in at least two places in this  
  file and across the codebase. Consider extracting it as a named constant to prevent typo-induced bugs.
                                                                                                           
  6. Missing test for context.site being undefined                                                       

  test/utils/data-access.test.js - The guard at line 803 (if (site)) is untested. The default test setup   
  always includes context.site. An explicit test would confirm the guard works and prevent regression.
                                                                                                           
  7. Missing test for context.dataAccess being null when site is present                                 

  test/utils/data-access.test.js - Related to issue 3. No test covers the case where context.site exists   
  but context.dataAccess is null/undefined.
                                                                                                           
  8. Log message could include internal error details                                                      
   
  src/utils/data-access.js:810 - log.warn(\Failed to check TBYB status: ${e.message}`)` could surface      
  connection strings or credentials from DB errors. Consider logging the error object directly so the    
  logging framework can sanitize.                                                                          
                                                                                                         
  Recommendations

  1. Extract the TBYB check into a shared helper (isSummitPlgSite(context)) that handles the               
  Configuration.findLatest() call, error handling, and null safety in one place. This eliminates the
  duplication and ensures consistent behavior.                                                             
  2. Consider resolving the TBYB flag once during audit initialization and attaching it to the context   
  object, so every downstream consumer can check it without redundant configuration lookups.               
  3. Consider log.info instead of log.debug for the TBYB skip (data-access.js:816). The existing
  syncSuggestions code uses log.info for the similar PLG bypass at line ~443. Skipping reconcile for an    
  entire site category is an operational event worth surfacing.                                          
                                                                                                           
  Assessment                                                                                             

  Ready to merge? No - with fixes.

  Reasoning: The behavioral change is correct and well-tested. The fail-open fallback is the right design. 
  However, the duplicate Configuration.findLatest() call creates redundant data layer round-trips and
  divergent error handling that will confuse future maintainers. The unguarded dataAccess destructuring is 
  a potential crash. Address issues 1-3 before merging; issue 4 deserves an explicit answer (code or     
  comment). Minor items and recommendations can be follow-ups.

  Next Steps

  1. Address the duplicate Configuration lookup (issue 1) and naming inconsistency (issue 2) - ideally by  
  extracting a shared helper.
  2. Fix the unguarded context.dataAccess destructuring (issue 3) - one-line change.                       
  3. Clarify whether Step 2 should also skip for TBYB sites (issue 4) - code comment or code change.       
  4. Minor items are optional.

absarasw:
> **Suggestion: use entitlements instead of `Configuration` handler check**
> 
> The TBYB detection here goes through `Configuration.findLatest()` + `isHandlerEnabledForSite('summit-plg', site)`, but the canonical source of truth for TBYB/PLG status is the entitlement tier. The codebase already has an established pattern for this in `src/utils/site-validation.js`:
> 
> ```js
> import { TierClient } from '@adobe/spacecat-shared-tier-client';
> import { Entitlement } from '@adobe/spacecat-shared-data-access';
> 
> const tierClient = await TierClient.createForSite(context, site, Entitlement.PRODUCT_CODES.ASO);
> const { entitlement } = await tierClient.checkValidEntitlement();
> isTbybSite = entitlement?.getTier?.() === Entitlement.TIERS.PLG;
> ```
> 
> This avoids the magic string `'summit-plg'`, stays consistent with how the rest of the worker gates behavior, and uses the intended data model (`SiteEnrollments` + `Entitlements`) rather than the general-purpose config bag.
> 
> More broadly, it would be worth tracking a follow-up to remove the `summit-plg` handler from `Configuration` entirely and migrate all TBYB detection across the codebase to the entitlement-driven approach. The handler-in-config pattern is an infra signal that should be superseded by the entitlement tier now that that model is in place.

@sandsinh The reason why I used this configuration rather than the entitlement+enrolment check is because a lot of sites are in internal orgs and they are part of TBYB onboarding and we are running audits on them even before moving them to customer's org.

absarasw:
> Should Step 2 (publish fix entities) also be skipped for TBYB sites?
The PR description says TBYB sites "have access to a limited view of suggestions" and that creating fix
entities "for issues the user has never seen" is misleading. But Step 2 (publishDeployedFixEntities)
still runs and can transition fix entities to PUBLISHED status. If a fix entity was created before this
PR deployed, it could still be published for a TBYB site. If Step 2 is intentionally left running, a code
comment explaining why would prevent future confusion.

@tarunsinghdev , We don't want to skip this step if the customers have used auto-optimization or mark as deployed on their own.

tarunsinghdev (approved):

PR #2424, opened 2026-04-24
fix: skip chrome links for sites using ARIA landmarks or AEM XFs (SITES-43574)

GeezerPelletier:
## Summary

- The preflight links audit skips anchors inside page chrome so repeated header/footer links don't get checked on every page. The previous guard used `closest('header') || closest('footer')`, which only matches semantic HTML5 tags — so sites expressing chrome with other markup (AEM Experience Fragments, ARIA landmarks) bypassed the skip and had every chrome link link-checked as if it were content.
- On the OneWalmart intranet (`one.walmart.com`), the page renders no semantic `<header>` or `<footer>` at all — the chrome is a pair of AEM XF wrappers (`.cmp-experiencefragment--header` / `--footer`) around `<div class="header">` / `<div class="footer">`. Social links, SSO gateways (`pfedprod.wal-mart.com`), the OWA email link, etc. were all being fetched server-side, and several of those return 4xx/5xx to non-browser UAs.
- Broaden the skip to cover three shapes: semantic tags, ARIA landmarks (`role="banner"` / `role="contentinfo"`), and AEM XF header/footer classes. Shape list is kept explicit in `CHROME_ANCESTOR_SELECTOR` with comments explaining each.

Fixes [SITES-43574](https://jira.corp.adobe.com/browse/SITES-43574).

Related prior work:
- #2207 (v1.388.3) — stopped 400/401 from being classified as broken; still the right call, but didn't address pages where chrome links were being checked in the first place.
- #1007 — introduced the original semantic-tag skip.

## Test plan

- [x] Existing `skips links inside header and footer elements` test still passes
- [x] New test: `skips links inside ARIA banner and contentinfo landmarks`
- [x] New test: `skips links inside AEM experience-fragment header/footer wrappers` — uses Walmart-representative markup, asserts only the main-body link is fetched
- [x] `npm run test:spec -- test/preflight/links-checks.test.js` — 23 passing, 100% coverage on `src/preflight/links-checks.js`
- [x] `npx eslint src/preflight/links-checks.js test/preflight/links-checks.test.js` — clean
- [ ] Post-merge: re-run preflight against a one.walmart.com page and confirm chrome links no longer appear in the Links section

### DIFF
--- a/src/preflight/links-checks.js
+++ b/src/preflight/links-checks.js
@@ -32,6 +32,11 @@ const HEAD_FALLBACK_STATUSES = new Set([400, 401, 403, 405, 429, 451]);
  *   - AEM Experience Fragment wrappers: `cmp-experiencefragment--header` /
  *     `cmp-experiencefragment--footer`, emitted by AEM XFs on many enterprise sites
  *     (including OneWalmart) that don't render semantic header/footer tags at all.
+ *
+ * Intentionally excluded: `<nav>` and `role="navigation"`. Navigation links point at real
+ * in-site destinations that we want to verify, so treating them as chrome would hide
+ * genuine broken links. Keep this list conservative — ambiguous wrappers like `.nav`,
+ * `.menu`, `.sidebar`, or generic `class*="header"` risk false-skipping content links.
  */
 const CHROME_ANCESTOR_SELECTOR = [
   'header',
@@ -40,6 +45,7 @@ const CHROME_ANCESTOR_SELECTOR = [
   '[role="contentinfo"]',
   '.cmp-experiencefragment--header',
   '.cmp-experiencefragment--footer',
+  // Do not add broader class-matchers here without evidence — see docblock above.
 ].join(', ');
 
 /**
--- a/src/preflight/links-checks.js
+++ b/src/preflight/links-checks.js
@@ -22,6 +22,27 @@ import { DEFAULT_USER_AGENT } from '../internal-links/helpers.js';
 const HEAD_FALLBACK_STATUSES = new Set([400, 401, 403, 405, 429, 451]);
 
 /**
+ * Ancestor selectors identifying page chrome (site header / footer). Anchors inside these
+ * are skipped because they are typically repeated across every page — checking them per
+ * page multiplies traffic to the same URLs and produces noisy, non-actionable findings.
+ *
+ * Covers three shapes seen in practice:
+ *   - Semantic HTML5 tags: `<header>`, `<footer>`
+ *   - ARIA landmarks: `role="banner"` (site header), `role="contentinfo"` (site footer)
+ *   - AEM Experience Fragment wrappers: `cmp-experiencefragment--header` /
+ *     `cmp-experiencefragment--footer`, emitted by AEM XFs on many enterprise sites
+ *     (including OneWalmart) that don't render semantic header/footer tags at all.
+ */
+const CHROME_ANCESTOR_SELECTOR = [
+  'header',
+  'footer',
+  '[role="banner"]',
+  '[role="contentinfo"]',
+  '.cmp-experiencefragment--header',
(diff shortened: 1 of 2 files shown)


src/preflight/links-checks.js
```diff
@@ -21,6 +21,27 @@ import { DEFAULT_USER_AGENT } from '../internal-links/helpers.js';
...
+ * are skipped because they are typically repeated across every page — checking them per
+ * page multiplies traffic to the same URLs and produces noisy, non-actionable findings.
+ *
+ * Covers three shapes seen in practice:
+ *   - Semantic HTML5 tags: `<header>`, `<footer>`
+ *   - ARIA landmarks: `role="banner"` (site header), `role="contentinfo"` (site footer)
+ *   - AEM Experience Fragment wrappers: `cmp-experiencefragment--header` /
+ *     `cmp-experiencefragment--footer`, emitted by AEM XFs on many enterprise sites
+ *     (including OneWalmart) that don't render semantic header/footer tags at all.
+ */
```
clotton, line 40: Worth noting here (or in the constant name/doc) that `<nav>` / `role="navigation"` are intentionally excluded — nav links are valid content targets worth checking, unlike header/footer chrome. Without that callout, a future reader might add `nav` here assuming it was just overlooked.

src/preflight/links-checks.js
```diff
@@ -21,6 +21,27 @@ import { DEFAULT_USER_AGENT } from '../internal-links/helpers.js';
...
+const CHROME_ANCESTOR_SELECTOR = [
+  'header',
+  'footer',
+  '[role="banner"]',
+  '[role="contentinfo"]',
+  '.cmp-experiencefragment--header',
+  '.cmp-experiencefragment--footer',
+].join(', ');
```
clotton, line 49: Nit: consider a one-liner comment above the closing bracket noting that this list is intentionally conservative (unambiguous header/footer chrome only), so reviewers know not to add `.nav`, `.menu`, `.sidebar`, etc. as the list grows over time.

test/preflight/links-checks.test.js
```diff
@@ -392,4 +392,46 @@ describe('preflight/links-checks - runLinksChecks', () => {
...
+    );
+
+    expect(fetchStub.callCount).to.equal(0);
+    expect(result.auditResult.brokenExternalLinks).to.have.lengthOf(0);
+  });
+
+  it('skips links inside AEM experience-fragment header/footer wrappers', async () => {
+    fetchStub.resolves(makeResponse(404));
+
+    await runLinksChecks(
```
clotton: Minor style inconsistency: the first new test captures the return value in `const result`, but this test uses bare `await` and then checks `fetchStub` directly. Both work, but picking one style across the describe block would be cleaner.

clotton (approved):
Clean, well-motivated fix. The `CHROME_ANCESTOR_SELECTOR` constant is nicely documented and makes future additions obvious. The AEM XF test is particularly solid — asserting on `fetchStub.firstCall.args[0]` proves the right URL is fetched, not just that no broken links were reported. A few nits inline, nothing blocking.

GeezerPelletier:
Thanks for the review — addressed all three nits in fe3db010:

1. **L35 (docblock)** — added an "Intentionally excluded" paragraph calling out `<nav>` / `role="navigation"` and the reasoning (nav links are real destinations we want to verify).
2. **L43 (selector list)** — added a conservative-by-design note both in the docblock and as an inline `// Do not add broader class-matchers here without evidence` comment above the closing bracket.
3. **Test L415** — the XF test now captures `const result` and asserts on the result shape (`brokenExternalLinks.lengthOf(1)` + `urlTo` check), matching the style of the ARIA and semantic-tags tests. Strong `fetchStub.firstCall.args[0]` assertion is retained.

Tests still 23/23 passing, `links-checks.js` at 100% coverage, lint clean.

PR #2425, opened 2026-04-24
fix(readability): skip nav elements and high link-density divs in preflight

clotton:
## Summary

- Removes `header`, `footer`, `nav`, and `[role="navigation"]` elements before readability analysis
- Adds a link-density filter that skips elements where ≥50% of collapsed text is inside `<a>` tags
- Fixes false positives where AEM vertical navigation components (using plain `<div>` wrappers) were flagged as poor readability

## Context

AEM pages can render navigation menus entirely in `<div>` elements with no semantic nav wrapper. Concatenated link labels score very poorly on Flesch Reading Ease (-53 to +28 in the reported case) despite being navigation, not readable prose. The AEM `dynamic-vertical-navigation-component` renders the same nav twice (desktop + mobile), causing each nav section to be double-counted — accounting for all 21 reported false positives.

The link-density filter catches this pattern generically: the affected elements have ~97% anchor text by character count.

## Test plan

- [ ] `should not flag navigation links inside <nav> elements as readability issues`
- [ ] `should not flag elements with role="navigation" as readability issues`
- [ ] `should still flag poor readability in body content when nav is present`
- [ ] `should not flag AEM custom nav divs with high link density (SITES-43577)`
- [ ] Deploy to dev and re-run preflight on the originally reported page to confirm 0 readability issues

Fixes SITES-43577.

### DIFF
--- a/src/readability/preflight/handler.js
+++ b/src/readability/preflight/handler.js
@@ -25,6 +25,7 @@ import {
   TARGET_READABILITY_SCORE,
   MIN_TEXT_LENGTH,
   MAX_CHARACTERS_DISPLAY,
+  MAX_LINK_DENSITY_RATIO,
 } from '../shared/constants.js';
 import { getDomElementSelector, toElementTargets } from '../../preflight/utils/dom-selector.js';
 
@@ -178,10 +179,10 @@ export default async function readability(context, auditContext) {
 
     const $ = cheerioLoad(rawBody);
 
-    // Remove structural/navigation elements before analysis — they produce false positives
-    // because their concatenated link text scores poorly on Flesch despite being valid nav labels.
-    // The opportunity handler's getMeaningfulElementsForReadability() already does this;
-    // keeping parity here fixes SITES-43577 (Walmart nav links flagged as readability issues).
+    // Remove structural/navigation elements before analysis — their concatenated link text
+    // scores very poorly on Flesch despite not being readable prose (SITES-43577).
+    // Note: the shared getMeaningfulElementsForReadability() only removes header/footer/style/
+    // script/noscript; nav removal and the link-density filter below are preflight-only for now.
     $('header, footer, nav, [role="navigation"]').remove();
 
     // Get all paragraph, div, and list item elements
@@ -284,8 +285,8 @@ export default async function readability(context, auditContext) {
         // very poorly on Flesch despite not being readable prose. SITES-43577.
         const $el = $(element);
         const collapsed = $el.text().replace(/\s+/g, ' ').trim();
-        const anchorText = $el.find('a').map((_, a) => $(a).text().replace(/\s+/g, ' ').trim()).get().join('');
-        return (anchorText.length / collapsed.length) < 0.5;
+        const anchorText = $el.find('a').map((_, a) => $(a).text().replace(/\s+/g, ' ').trim()).get().join(' ');
+        return (anchorText.length / collapsed.length) < MAX_LINK_DENSITY_RATIO;
       });
 
     // Process filtered elements
--- a/src/readability/preflight/handler.js
+++ b/src/readability/preflight/handler.js
@@ -178,6 +178,12 @@ export default async function readability(context, auditContext) {
 
     const $ = cheerioLoad(rawBody);
 
+    // Remove structural/navigation elements before analysis — they produce false positives
+    // because their concatenated link text scores poorly on Flesch despite being valid nav labels.
+    // The opportunity handler's getMeaningfulElementsForReadability() already does this;
+    // keeping parity here fixes SITES-43577 (Walmart nav links flagged as readability issues).
+    $('header, footer, nav, [role="navigation"]').remove();
+
     // Get all paragraph, div, and list item elements
     const textElements = $('p, div, li').toArray();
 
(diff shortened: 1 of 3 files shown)


GeezerPelletier:
## Code Review

### Overview
Fixes SITES-43577 by filtering out navigation content before readability scoring. Two mechanisms are added to `src/readability/preflight/handler.js`:

1. **Semantic removal** — strips `<header>`, `<footer>`, `<nav>`, and `[role="navigation"]` from the parsed DOM before element collection.
2. **Link-density filter** — after existing block-children and `MIN_TEXT_LENGTH` filters, drops any candidate whose concatenated anchor text is ≥50% of its total text (targets AEM vertical-nav components that render as plain `<div>`s).

Four new tests cover both mechanisms, including a regression test that confirms body content is still flagged when a nav is present.

### What looks good
- The link-density approach is the right call: `nav` removal alone wouldn't catch the AEM custom-component case described in the ticket, and blindly removing all high-link-density divs would over-filter.
- The "still flag poor readability in body content when nav is present" test is exactly the right negative guard — prevents future regressions where the filter starts swallowing legitimate issues.
- The AEM test case uses a realistic DOM shape (a `<div>` wrapper, a `<span>` inline-only child so block-children filter passes, then many anchors) — it genuinely exercises the new link-density path, not just the `<nav>` removal.
- Comments cite the ticket and the concrete site, which will be useful when someone revisits this in a year.

### Issues / suggestions

**1. Inaccurate parity claim — the opportunity handler still has the bug.**
The inline comment on the new `.remove()` call says:
> The opportunity handler's getMeaningfulElementsForReadability() already does this; keeping parity here fixes SITES-43577

But [`getMeaningfulElementsForReadability()` in `shared/analysis-utils.js`](https://github.com/adobe/spacecat-audit-worker/blob/main/src/readability/shared/analysis-utils.js#L148-L154) only removes `header, footer, style, script, noscript` — it does **not** remove `nav`, `[role="navigation"]`, or apply any link-density filter. So after this PR, preflight does *more* filtering than the opportunity handler, not the same. The opportunity path is still susceptible to the same false-positive on `<nav>` and AEM vertical-nav. Either update `getMeaningfulElementsForReadability()` in the same PR (preferred — that's the actual fix for parity) or correct the comment.

**2. Magic number `0.5` should be a named constant.**
Put it in [`src/readability/shared/constants.js`](https://github.com/adobe/spacecat-audit-worker/blob/main/src/readability/shared/constants.js) alongside `MIN_TEXT_LENGTH`, e.g.:
```js
// Elements whose anchor text exceeds this fraction of total text are treated
// as navigation (not prose) and skipped during readability analysis.
export const MAX_LINK_DENSITY_RATIO = 0.5;
```
That file already holds the tunable knobs for this audit; keeping `0.5` inline means the next person tuning thresholds won't find it by grepping constants.

**3. Aggressive removal of `<header>` / `<footer>`.**
HTML5 allows `<header>` and `<footer>` inside `<article>`/`<section>` for titles/subtitles, bylines, or copy blocks. Stripping *all* of them wholesale will silently hide legitimate prose on pages that follow HTML5 semantics. This mirrors what `getMeaningfulElementsForReadability()` already does, so it's a preexisting pattern — but worth calling out, and possibly narrowing to `body > header, body > footer` in a follow-up to avoid false negatives on article-scoped headers/footers.

**4. `anchorText` join loses separators.**
```js
const anchorText = $el.find('a').map((_, a) => $(a).text().replace(/\s+/g, ' ').trim()).get().join('');
```
`join('')` with no separator means `anchorText.length` is the sum of stripped anchor text; meanwhile `collapsed.length` includes whitespace/separators between anchors (the `|` pipes in the test fixtures, for example). So the ratio slightly *understates* link density. Fine for a 0.5 threshold, but `join(' ')` would be more faithful and is a one-line change.

**5. Duplicate text traversal.**
The preceding filter calls `$(element).text()?.trim()` and the new one calls `$el.text().replace(...).trim()` again. Each `.text()` walks the subtree. For a page with many candidate divs this is minor overhead, but you could thread the already-computed text through instead of recomputing. Ignorable if perf isn't a concern here.

**6. No unit test for the link-density boundary.**
The four new tests all exercise near-100% anchor density. A test at ~40% link density that *still gets flagged* would pin the threshold and prevent someone from accidentally bumping `0.5` → `0.2` in the future without test failure.

### Correctness & safety
- No security or performance risk introduced.
- Filter ordering (block-children → min-length → link-density) is sensible — cheapest filters first, heaviest last.
- No risk of `NaN` from divide-by-zero reaching the score: `MIN_TEXT_LENGTH ≥ 150` runs first, so `collapsed.length` is guaranteed non-zero. Even if it were zero, `NaN < 0.5` is `false`, which correctly excludes the element.

### Recommendation
**Approve with comments** — the fix is correct and well-tested for the reported case. I'd ask for (1) the comment/parity correction and ideally the matching fix in `getMeaningfulElementsForReadability()`, and (2) extracting `0.5` to a named constant before merge. The `<header>`/`<footer>` scoping and boundary test are nice-to-haves for a follow-up.

GeezerPelletier (approved):
With comments

clotton:
Thanks for the thorough review! Addressed the required items and one of the nice-to-haves in the follow-up commit:

**Fixed:**
1. **Inaccurate parity comment** — corrected to clarify that `getMeaningfulElementsForReadability()` only removes `header/footer/style/script/noscript`; nav removal and the link-density filter are preflight-only for now. (Fixing the shared handler is a good follow-up but out of scope for this bug fix.)
2. **Magic number `0.5`** — extracted to `MAX_LINK_DENSITY_RATIO` in `constants.js` alongside `MIN_TEXT_LENGTH` and `MAX_CHARACTERS_DISPLAY`.
3. **`join('')` → `join(' ')`** — more faithful density calculation.

**Deferred (nice-to-haves):**
- `body > header / body > footer` scoping — valid point, leaving for a follow-up to keep this fix focused.
- Boundary test at ~40% link density — also a good follow-up; the threshold is now a named constant so at least it's grep-able.
- Duplicate text traversal — minor perf concern, deferring.

PR #2429, opened 2026-04-27
fix(llmo-4010): preserve edge-deployed suggestions in audit sync flow

anagarwa:
## Summary

- **Root cause (nit23uec review on spacecat-api-service#2172):** The audit sync flow was not preserving state of already-edge-deployed suggestions, including incorrectly overwriting the `shouldOptimize` flag. The fix was wrongly placed in `patchSuggestion` (API layer); the correct layer is the audit handlers.
- **headings handler:** Added `edgeDeployed` guard at the top of `mergeDataFunction` — returns existing data unchanged when `edgeDeployed` is truthy, so re-audit runs cannot overwrite deployed suggestion data.
- **faqs guidance handler:** Added a `mergeDataFunction` to the `syncSuggestions` call with the same guard — prevents `shouldOptimize` (and other state) being overwritten for deployed suggestions when the audit re-analyses the page.
- Note: PR #2183 already handled the domain-wide case in the prerender audit; this covers individual URL suggestions in headings and faqs.

## Test plan
- [ ] `mergeDataFunction` — edgeDeployed=true returns existing suggestion unchanged (headings handler test)
- [ ] `mergeDataFunction` — edgeDeployed=false performs normal merge (headings handler tests already covered)
- [ ] FAQs guidance handler — `mergeDataFunction` passed to `syncSuggestions` preserves edge-deployed suggestion data including `shouldOptimize`
- [ ] FAQs guidance handler — normal (non-deployed) suggestions still receive new audit data

Jira: LLMO-4010


anagarwa:
## How `edgeDeployed` and `shouldOptimize` are handled

### Background

When a suggestion is deployed to the edge CDN, `suggestion.data.edgeDeployed` is set to `true`. This means the optimization is **live at the CDN layer** — for example, an FAQ section is being injected into the page HTML on every request by Fastly/Cloudflare, even though the origin page itself is unchanged.

### The bug this PR fixes

On every subsequent audit run, `syncSuggestions` calls `mergeDataFunction(existingData, newData)` to update stored suggestions with fresh audit results. Before this fix, the merge looked like:

```js
// Old behaviour — always overwrites existing with new audit data
const mergedSuggestion = { ...existingSuggestion, ...newSuggestion };
```

**Example scenario (FAQs handler):**

```
Audit run 1:
  origin page has no FAQ heading → shouldOptimize = true
  Suggestion created: { shouldOptimize: true, selector: 'main' }
  Suggestion deployed to edge → edgeDeployed = true set on the record
  Edge CDN now injects FAQ section on every request

Audit run 2:
  Scraper hits origin page (fix is at edge, not in CMS — origin unchanged)
  Customer has since added their own FAQ heading to origin CMS
  analyzeScrapeData() → shouldOptimize = false (FAQ heading detected)

  Old merge: { ...existingData, ...newData }
  Result: { edgeDeployed: true, shouldOptimize: false }  ← contradictory state
  UI now shows a deployed suggestion as 'does not need optimization'
```

### The fix

```js
// New behaviour — skip merge entirely if suggestion is already edge-deployed
const mergeDataFunction = (existingSuggestion, newSuggestion) => {
  if (existingSuggestion.edgeDeployed) {
    return { ...existingSuggestion };  // return as-is, audit result ignored
  }
  // ... normal merge for non-deployed suggestions
};
```

**Same example with the fix:**

```
Audit run 2 (with fix):
  analyzeScrapeData() → shouldOptimize = false
  mergeDataFunction called: existingSuggestion.edgeDeployed = true → early return
  Result: { edgeDeployed: true, shouldOptimize: true }  ← deployed state preserved
  Edge CDN continues injecting FAQ section correctly
```

### Scope

- **headings/handler.js** — `mergeDataFunction` guards `edgeDeployed` before merging heading fix suggestions
- **toc/handler.js** — same guard for TOC suggestions
- **faqs/guidance-handler.js** — `mergeDataFunction` added (was missing entirely); guards `shouldOptimize` and other state for FAQ suggestions
- **summarization** — no change needed; has no `syncSuggestions` call (sends to Mystique via SQS instead)
- PR #2183 already handled the domain-wide case in the prerender audit; this covers individual URL suggestions

dipratap:
Changes look fine to me, please fix linting.

duynguyen (approved):
Technical changes look good. The edgeDeployed guards are applied at the audit handler merge layer and preserve deployed suggestion data while leaving normal merge behavior intact.

PR #2434, opened 2026-04-28
fix(preflight): SITES-43720 full-page link coverage and GET-confirm broken status

GeezerPelletier:
## Summary

Two related changes to the preflight links audit, both motivated by SITES-43720:

- **Stop skipping anchors inside page chrome** (semantic `<header>` / `<footer>`, ARIA `role="banner"` / `role="contentinfo"`, and AEM Experience Fragment `cmp-experiencefragment--header` / `--footer` wrappers). Preflight audits a single page on demand and is expected to report every anchor on that page; the chrome-skip pattern was inherited from the bulk-audit world where chrome links repeat across hundreds of pages and per-page reporting produces N copies of the same finding. That motivation does not apply to single-page preflight. Mystique already made this same change in [SITES-42820](https://jira.corp.adobe.com/browse/SITES-42820); this aligns SAW with that decision.
- **Treat HEAD as a fast-path optimization and GET as the source of truth for "is this link broken."** Previously a HEAD response of 404 / 410 / 5xx was reported as broken without any GET retry. Real servers (misconfigured Apache origins, SSO endpoints) commonly return broken-looking statuses to HEAD on routes that respond 200 to GET, producing false positives. We now GET-confirm any HEAD result that would otherwise be flagged.

## Validation

Tested against a real one.walmart.com fuel-merchandise page (~520 anchors):

| Link | Before | After | Notes |
|---|---|---|---|
| `sparkshop.com` (footer) | reported 404 | not reported | False positive: HEAD 404, GET 200 (Apache misconfig). Caught by GET-confirm. |
| `pfedprod ... startSSO.ping?...` (footer) | reported 500 | not reported | Transient HEAD 5xx; GET 200. Caught by GET-confirm. |
| `linkedin.com/in/dougmcmillon/` (footer) | reported 404 | reported 404 | True positive — former exec, slug genuinely dead. |
| AEM `/me/health/helpr-video.html` (header) | reported 404 | reported 404 | True positive — server returns 404 (not 401/302) for an SSO-required page; surfacing this nudges the page owner to fix the misconfigured SSO interstitial. |

Net: 4 reports → 2 reports, with both remaining reports being genuinely actionable.

## Updates since opening

In response to review feedback (commit `5bc2118b`):

- **`Range: bytes=0-0` on GET retry only.** Cooperating servers reply with `206 Partial Content` + 1 byte instead of streaming the full body; servers that ignore the header fall back to whatever they would have returned anyway. `206` and `416 Range Not Satisfiable` both fall outside `isBrokenStatus`, so they are correctly treated as healthy. HEAD requests are unchanged. This minimizes bandwidth in the false-positive correction case (HEAD broken / GET healthy), where the server would otherwise stream a full HTML page.
- **Explicit `HEAD` / `GET` method assertions** on the 410, 500, and HEAD-500-GET-200 tests for parity with the 404 test, plus a new test that confirms `Range: bytes=0-0` is sent only on the GET retry, never on HEAD.

## Impact on related tickets

- Reverses the chrome-skip direction taken in [SITES-43574](https://jira.corp.adobe.com/browse/SITES-43574). That fix aligned preflight with the historical ASO behaviour; we now believe that alignment was the wrong direction for single-page audits.
- Test pages [SITES-43687](https://jira.corp.adobe.com/browse/SITES-43687) and [SITES-43694](https://jira.corp.adobe.com/browse/SITES-43694) need their pass criteria inverted (chrome links should now surface, not be skipped). Heads-up comments left on both tickets.

## Test plan

- [x] `npm run test:spec -- test/preflight/**/*.test.js test/audits/preflight*.test.js` — 339 passing locally (was 338 before the review-feedback test was added).
- [x] 100% coverage on `src/preflight/links-checks.js` and `src/preflight/links.js` per repo policy.
- [x] End-to-end probe against the real fuel.html surfaces only the two true-positive broken links.
- [x] SITES-43574 Walmart 400 / 401 footer noise stays suppressed under the new logic — those statuses still go through GET-confirm and remain excluded by `isBrokenStatus`.
- [ ] CI green.


src/preflight/links-checks.js
```diff
@@ -94,33 +67,34 @@ async function checkLinkStatus(href, pageUrl, context, options = {
...
   } catch (err) {
     log.warn(`[preflight-audit] HEAD request failed (${err.message}), retrying with GET: ${href}`);
   }
 
-  // GET fallback — reached when HEAD was inconclusive (fallback status or network error)
-  fetchOptions.method = 'GET';
+  // GET fallback — HEAD was inconclusive (broken status, fallback status, or network error).
+  const getOptions = { method: 'GET', decode: false, headers };
   try {
-    const res = await fetch(href, fetchOptions);
```
clotton, line 123: **Worth discussing: GET response body pulled for every broken-looking HEAD.**

Previously GET was only invoked for `HEAD_FALLBACK_STATUSES` — statuses where HEAD was known to be unreliable. Now it fires for *all* non-healthy HEAD responses, including genuine 404s and 5xxs.

A page with 10 broken-looking links will now issue 10 GET requests pulling full response bodies just to confirm the status code. For enterprise sites with links to large destinations (video pages, SPAs, SSO endpoints) this could add meaningful latency and bandwidth per audit.

Worth considering: **abort the GET after reading the status line**, or send a `Range: bytes=0-0` request header. Most servers honour it and return the status with a minimal body; servers that ignore it fall back to the full body, same as today.

Not a blocker if the performance profile is acceptable in practice, but worth a deliberate call before this ships broadly.

clotton (commented):
Clean, well-motivated changes. The Walmart validation data makes the case for GET-confirm convincingly — 2 false positives eliminated, 2 genuine ones preserved. The chrome-skip reversal is correct; the PR honestly acknowledges that SITES-43574 aligned SAW with Mystique behaviour and this intentionally un-aligns them, with good reasoning.

**Approve with one concern worth discussing** (inline below).

test/preflight/links-checks.test.js
```diff
@@ -105,9 +126,24 @@ describe('preflight/links-checks - runLinksChecks', () => {
 
     expect(result.auditResult.brokenExternalLinks).to.have.lengthOf(1);
     expect(result.auditResult.brokenExternalLinks[0].status).to.equal(410);
+    expect(fetchStub.callCount).to.equal(2);
+  });
+
```
clotton, line 133: **Nit: missing method assertion for consistency.**

The 404 test above asserts both `firstCall.method === 'HEAD'` and `secondCall.method === 'GET'` explicitly. This test only asserts `callCount === 2`. Worth adding the same method checks here for consistency — makes the intent of the test explicit and catches a future regression where both calls somehow use the same method.

clotton (approved):

src/preflight/links-checks.js
GeezerPelletier: Thanks — addressed in 5bc2118b. Added `Range: bytes=0-0` on the GET retry only (HEAD remains body-less by definition). Cooperating servers will respond 206 + 1 byte instead of streaming the full response; servers that ignore the header fall back to whatever they would have returned anyway. 206 and 416 (Range Not Satisfiable) both fall outside `isBrokenStatus`, so they are correctly treated as healthy.

Honest note on the trade-off: this only helps the false-positive correction case (HEAD broken / GET healthy). For genuinely-broken links, the server still returns its full error page regardless of the Range header — but those bodies are typically small. Aborting the GET stream after status would cut bytes in that case too, but the half-closed-connection complexity wasn’t worth it here. Happy to revisit if profiling later shows it matters.

test/preflight/links-checks.test.js
GeezerPelletier: Done in 5bc2118b — added `firstCall.method === "HEAD"` / `secondCall.method === "GET"` to the 410 / 500 / HEAD-500-GET-200 tests, plus a new test that confirms `Range: bytes=0-0` is sent only on the GET retry.

GeezerPelletier:
@clotton — thanks for the review. Both points addressed in 5bc2118b:

**1. GET-confirm body cost.** Added `Range: bytes=0-0` on the GET retry only (HEAD remains body-less by definition). Cooperating servers reply with `206` + 1 byte instead of streaming the full body; servers that ignore it fall back to whatever they would have returned anyway. `206` and `416` both fall outside `isBrokenStatus`, so they are correctly treated as healthy.

Honest trade-off note: this only helps the false-positive correction case (HEAD broken / GET healthy → server otherwise streams a full HTML page). For genuinely-broken links the server still returns its full error page regardless of `Range` — but those bodies are typically small. Aborting the GET stream after status would cut bytes in that case too, but the half-closed-connection complexity wasn't worth it here. Happy to revisit if profiling later shows it matters.

**2. Method assertion parity.** Added `firstCall.method === "HEAD"` / `secondCall.method === "GET"` to the 410 / 500 / HEAD-500-GET-200 tests, plus a new test that asserts `Range: bytes=0-0` is sent on the GET retry only.

PR description updated with an "Updates since opening" section. 339 preflight tests passing locally; 100% coverage on `links-checks.js` and `links.js`. Ready for another look when you have a minute.

PR #2437, opened 2026-04-29
feat(paid-keyword-optimizer): multi-keyword cluster support

danieljchuser:
## Summary

Rewrite the paid-keyword-optimizer guidance handler and opportunity mapper for multi-keyword intent clustering:

- **Failure envelope handling (PR 0):** Add `status === "failed"` guard at the top of the guidance handler — Mystique failures are now properly routed instead of silently dropped
- **Handler rewrite:** New gate on `body.clusterResults` (replaces `issueSeverity` check). Creates 1 opportunity per URL with N suggestions (one per cluster)
- **Opportunity mapper:** Includes `portfolioMetrics`, `hasConflictingHeadlineRecommendations` (computed locally from clusterResults), `langfuseTraceId`/`langfuseTraceUrl` from `body.observability`
- **Suggestion mapper:** Per-cluster mapper with `recommendation.type` lift to `recommendationType`. Cluster data nested under `data.cluster.*`
- **Composite ranking:** Mismatched clusters by spend desc, aligned by traffic desc, failed last
- **Replace-on-re-audit:** Prior NEW/IN_PROGRESS opportunities on same (siteId, url, type) marked IGNORED before creating new
- **WSIS scoring:** Aggregate `sum(kw.cpc * kw.traffic)` across all keywords (was single-keyword CPC)
- **Empty keywords filter:** Pages without `keywords[]` excluded from Mystique enqueue
- **Null CPC preservation:** `clusterCpc ?? null` and `clusterMisalignedSpend ?? null` (not `|| 0`)
- **Dead code removal:** `inferRecommendationType()`, `isLowSeverityGuidanceBody()`, `reconcileOpportunities()`, `MAX_OPPORTUNITIES_PER_TYPE`

Part of the multi-keyword intent cluster feature for ad-intent-mismatch. Design spec: `mystique/docs/plans/ad-intent-mismatch/2026-04-09-keyword-cluster-design.md` (v6.3).

## Test plan

- [x] 1 opportunity per URL from clusterResults + portfolioMetrics
- [x] 1 suggestion per cluster (including well-aligned and failed)
- [x] `hasConflictingHeadlineRecommendations` true when >=2 modify_heading clusters
- [x] `langfuseTraceId` and `langfuseTraceUrl` populated from observability
- [x] Composite rank: mismatched by spend desc, aligned by traffic desc, failed last
- [x] Replace-on-re-audit: prior NEW→IGNORED, IN_PROGRESS→IGNORED, RESOLVED untouched
- [x] No body or missing clusterResults → no opportunity
- [x] Status "failed" → no opportunity, structured log emitted
- [x] WSIS uses aggregate keywords spend
- [x] Empty keywords filter excludes pages
- [x] Old-format opportunity in DB doesn't crash
- [x] Null CPC preserved (not coerced to 0)
- [x] All tests passing, 100% coverage


LeChef318 (changes requested):
Hey @danieljchuser,

## Strengths

- The dead-code retirement is clean: `inferRecommendationType`, `isLowSeverityGuidanceBody`, `reconcileOpportunities`, and `MAX_OPPORTUNITIES_PER_TYPE` are gone from `src/` and `test/` after this PR. Net source is meaningfully simpler.
- `src/paid-keyword-optimizer/guidance-handler.js:50-60` Hoisting the failure-envelope guard above any DB work is correct ordering. The log captures `error.type`, `langfuseTraceId`, and IDs but deliberately omits `error.message`, which is the right restraint for LLM-sourced error strings.
- `src/paid-keyword-optimizer/guidance-opportunity-mapper.js:188` `clusterCpc ?? null` and `clusterMisalignedSpend ?? null` preserve the unknown-vs-zero distinction. Easy to get wrong, you got it right.
- One opportunity per URL with N suggestions is the right cardinality. The previous shape would have fragmented the consumer UX.
- `assignClusterRanks` is a tiered sort with explicit predicates (`isMismatched`/`isAligned`/`isFailed`) and immutable returns. Tests cover the three tiers, single-cluster, empty input, and null/undefined fallback paths thoroughly.
- DDB writes use the parameterized object form, no injection surface introduced. No new endpoints, no auth code, no dependency or lockfile changes.
- Test structure is behavior-oriented (failure envelope, replace-on-re-audit, observability logging) rather than asserting internal mock invocation order. WSIS edge cases (null cpc, null traffic, alignment clamp at both ends, empty keywords) are well covered.

## Issues

### Important (Should Fix)

1. **Cookie-consent spec doc is unrelated to this PR.** `docs/specs/2026-04-15-cookie-consent-seo-importer-fix.md` (+195 lines) is an implementation plan for `src/paid-cookie-consent/handler.js` (the `importAhrefPaidStep` enable-import fix). It has nothing to do with the multi-keyword cluster work. The file's content references different code paths, a different production incident, and a different goal. Drop it from this PR and submit it on its own branch.

2. **Handler crashes on missing `data`.** `src/paid-keyword-optimizer/guidance-handler.js:49` destructures `const { auditId, siteId, data } = message;`. Line 51 then guards with `data?.status === 'failed'`, which explicitly admits `data` can be undefined. But line 62 does `const { guidance } = data;` which throws `TypeError: Cannot destructure property 'guidance' of 'data' as it is undefined`. Because this is an SQS handler, an unhandled throw becomes a poison pill that retries to DLQ. Either treat missing `data` as a hard error and return `notFound()`/`ok()` before line 62, or use `data?.guidance`.

3. **Empty `clusterResults: []` creates a zombie opportunity.** `src/paid-keyword-optimizer/guidance-handler.js:82` gates with `if (!guidanceBody || !guidanceBody.clusterResults)` - but an empty array is truthy, so the handler proceeds and creates one opportunity with `totalClusters: 0`, `misalignedClusters: 0`, description "0 of 0 clusters show alignment gaps", and zero suggestions. Either tighten the gate to `clusterResults?.length > 0`, or intentionally support empty-cluster opportunities and add a test asserting that contract.

4. **Mass-IGNORE bug when `url` is undefined.** `src/paid-keyword-optimizer/guidance-handler.js:108` filters `oppty.getData()?.url === url`. If `data?.url` is missing on the message, `url` is `undefined`, and any prior opportunity whose stored `data.url` is also `undefined` matches - `undefined === undefined` is true. That can flip unrelated `ad-intent-mismatch` opportunities for the site to IGNORED. Guard with `url && oppty.getData()?.url === url`.

5. **Replace-on-re-audit is not failure-atomic and races on duplicate delivery.** `src/paid-keyword-optimizer/guidance-handler.js:99-139` runs four DB ops without a transaction: `allBySiteIdAndStatus` x 2 -> `saveMany(IGNORED)` -> `Opportunity.create` -> `Promise.all(map(Suggestion.create))`. Two failure modes: (a) if `Opportunity.create` throws after `saveMany` succeeds, the URL is left with zero NEW opportunities until the next audit; (b) two concurrent guidance messages for the same `(siteId, url)` (SQS at-least-once delivery is real here) both read the same prior set, both flip them to IGNORED, both create new - leaving duplicates. Smallest-diff fix: create the new opportunity first, then flip priors to IGNORED in a second pass. Alternative: try/catch that re-promotes priors on failure.

6. **Silent-degradation pattern on Mystique failure envelope.** When `data.status === 'failed'`, the handler logs and `ok()`s. There is no metric tag (e.g., `mystique_url_failed`), no audit-result annotation, no opportunity-level marker. From the outside, the URL just disappears. If Mystique starts failing 30% of URLs, no signal surfaces. At minimum, emit a structured metric and track failed-URL counts on the audit result so alerting can pick them up.

7. **`recommendation` without `recommendation.type` produces a malformed suggestion.** In `mapClusterToSuggestion` (`src/paid-keyword-optimizer/guidance-opportunity-mapper.js`), when `recommendation` is present but `recommendation.type` is missing, the destructure leaves `cleanedRecommendation = {}` (an empty object, truthy), so `data.recommendation` is set to `{}` and `recommendationType` is omitted. Either guard `Object.keys(rest).length > 0` before including `recommendation`, or add a test that locks in the current behavior so the contract is intentional.

### Minor (Nice to Have)

8. **Sort uses `|| 0` while data layer preserves `null`.** `src/paid-keyword-optimizer/guidance-opportunity-mapper.js:53,57,61` sort with `(b.clusterMisalignedSpend || 0) - (a.clusterMisalignedSpend || 0)`, treating `null` and `0` identically; `mapClusterToSuggestion` preserves `null` in the suggestion data. Pick a stance: coalesce to 0 everywhere, or sort nulls last explicitly.

9. **No deterministic tie-breaker in ranking.** `assignClusterRanks` ties resolve to input order. Modern V8 is stable so it is deterministic for a given input, but the input order across re-audits is not. Cheap fix: secondary key on `clusterId` or the cluster's representative keyword.

10. **`Promise.all(map(Suggestion.create))` masquerades as bulk-create.** `src/paid-keyword-optimizer/guidance-handler.js:135` comment claims "Bulk-create suggestions" but issues N independent round-trips. Partial failure leaves the opportunity with a partial suggestion set and rejects the handler, triggering SQS retry which then re-IGNORE's a partial opportunity. Use `Suggestion.bulkCreate`/`saveMany` if available, or update the comment and add `Promise.allSettled` with explicit handling.

11. **The "old-format opportunity" test does not actually validate backward compatibility.** It asserts no crash when `getData()` returns extra legacy fields (`gapAnalysis`, `cpc`, `sumTraffic`), but the handler never reads those fields - only `getType()` and `getData()?.url`. The test passes trivially regardless of the old shape. Either add a test that exercises an old shape that lacks `url` (which collides with finding 4), or drop the test since it gives false confidence.

12. **`paidLog.createdSuggestion` called N times with no per-suggestion ID.** `src/paid-keyword-optimizer/guidance-handler.js:137-139` discards `Promise.all` results, so the per-suggestion log line cannot include `Suggestion.getId()`. Either capture the results and include the id, or log once with a count.

13. **Unredacted `...observability` spread into INFO log.** `src/paid-keyword-optimizer/guidance-handler.js:90` does `log.info({ ...observability, ... })`. Today `observability` only carries `langfuseTraceId`/`Url`, but there is no schema. If Mystique adds prompt fragments or partial model output to that field, it ships verbatim to Coralogix. Allow-list the known fields rather than spread.

14. **Failure-envelope test coverage is limited to one shape.** Only `data.status === 'failed'` with a fully populated `error` object is tested. Missing: `error` undefined (works but locks in `trace_id: undefined` shape), and `failed` + `clusterResults` co-present (asserts the failure path takes precedence).

## Recommendations

- Extract the replace-on-re-audit block into a named helper, e.g. `markPriorOpportunitiesIgnored(Opportunity, siteId, url)`. Localizes finding 5 and makes the happy path readable.
- The `formatCurrency` `c8 ignore next 3` annotation on the `num == null` branch is small enough to test directly. A two-line test removes the ignore comment.
- Add a top-of-file comment in `guidance-handler.js` linking the Mystique design doc (`mystique/docs/plans/ad-intent-mismatch/2026-04-09-keyword-cluster-design.md`). Future maintainers will not find the spec via grep.
- Validate `audit.getSiteId() === message.siteId` defensively. The handler trusts `siteId` and `auditId` from the SQS message independently. The queue is internal-only so exploitability is gated, but a one-line assertion is cheap defense in depth.
- Confirm in the PR description that the WSIS `priorityScore > 0.01` threshold was re-validated against the new aggregate scale, not just inherited from the per-keyword formula. Even one site's before/after counts would put this to bed.
- Discuss with the Mystique team whether `hasConflictingHeadlineRecommendations` should travel in the message envelope rather than be recomputed here. Computing it locally couples this repo to Mystique's recommendation-type taxonomy (`'modify_heading'` is now a magic string in two repos). At minimum, lift it to a named constant with a comment pointing at the spec.

The Mystique-side spec change and the cross-service contract (`clusterResults`, `portfolioMetrics`, `body.observability` shape) are now de-facto frozen by both sides shipping to it. A schema version field on `body`, or a JSON schema in `spacecat-shared` validated by both repos, would make the next contract change cheaper. Not fixable in this PR.

## Assessment

**Ready to merge?** No, with fixes.

**Reasoning:** Findings 1, 2, 3, and 4 are concrete, reachable bugs or change-discipline failures with small targeted fixes. Findings 5 and 6 are real production-failure modes worth addressing before this becomes the canonical handler shape. The architectural direction is sound and the simplification is real, but these need a follow-up commit before merge.

## Next Steps

1. Drop `docs/specs/2026-04-15-cookie-consent-seo-importer-fix.md` from this branch (item 1).
2. Fix the four concrete handler bugs (items 2-5): missing-`data` crash, empty-`clusterResults` zombie, mass-IGNORE on undefined url, non-atomic replace-on-re-audit.
3. Add a metric and audit-result annotation for Mystique failure envelopes (item 6).
4. Decide on the malformed-recommendation behavior and add a test (item 7).
5. Minor items and recommendations are batchable.

danieljchuser:
Thanks for the thorough review @LeChef318! Here's the response to each finding:

## Fixed (in 439f7a11)

**#2 — Handler crashes on missing `data`**: Fixed. Added `!data` guard at the top of the handler that returns `ok()` with a warning log, before any destructuring.

**#3 — Empty `clusterResults: []` creates zombie opportunity**: Fixed. Tightened the gate to `clusterResults.length === 0`. Note that Mystique won't produce this (it gates on `has_recommendations` before building the body), but defense-in-depth is warranted.

**#4 — Mass-IGNORE on undefined url**: Fixed. Added `url &&` guard to the replace-on-re-audit filter. When url is undefined, no prior opportunities are matched.

All three fixes have dedicated tests (4 new test cases). 100% coverage maintained.

## Dismissed

**#1 — Cookie-consent spec unrelated to PR**: These files came from `main` via the merge commit `bc200026`. They are not part of this PR's changes — the diff shows them because our branch diverged before they landed on main. Not actionable here.

## Acknowledged — Pre-existing patterns, not introduced by this PR

**#5 — Replace-on-re-audit race condition**: Valid concern. However, this is the same non-transactional pattern used by all other guidance handlers in the repo (youtube-analysis, page-type, etc.). Fixing it properly requires a data-access layer change (transaction support). Tracked, but out of scope for this PR.

**#6 — Silent degradation on failure envelope**: We log with structured fields (`trace_id`, `audit_id`, `site_id`, `url`, `error_type`) which are queryable in Coralogix. Adding a dedicated metric is a good idea but better done as a cross-cutting concern across all guidance handlers, not just this one.

**#10 — `Promise.all(map(Suggestion.create))` not bulk**: Same pattern as other handlers. `Suggestion.saveMany` could replace this, but that's a refactor across all guidance handlers. Comment is misleading — updated to say "Create suggestions" instead of "Bulk-create".

## Won't fix (by design)

**#7 — Recommendation without `recommendation.type`**: This cannot happen. Mystique's `Recommendation` Pydantic model has `type: Literal["modify_heading", "audit_required"]` — it's a required field enforced by the model. A recommendation without type would fail Pydantic validation before it reaches the SQS message.

**#8 — Sort uses `|| 0` while data layer preserves `null`**: Intentional. Sorting needs numeric values (null breaks comparison), while storage preserves the unknown-vs-zero distinction. These serve different purposes.

**#9 — No deterministic tie-breaker**: Correct observation, but ties within a tier are extremely rare (requires identical spend or traffic values). The input order from Mystique is deterministic for a given run. Low-value change.

**#11 — Old-format opportunity test**: The test validates that the replace-on-re-audit logic doesn't crash on old DB records with a different `data` shape. This is a real scenario during the transition period after deploy.

**#12 — `createdSuggestion` log without per-suggestion ID**: Valid observation. Minor improvement, can be done in a follow-up.

**#13 — Unredacted `...observability` spread**: The observability object shape is controlled by our code in Mystique (`_build_guidance_body`), not by LLM output. It only contains `langfuseTraceId` and `langfuseTraceUrl`. Adding an allow-list is over-engineering for a field we fully control.

**#14 — Limited failure-envelope test coverage**: The one test covers the production-relevant shape. Testing `error: undefined` and `failed + clusterResults co-present` are theoretical edge cases — Mystique never produces either shape.

docs/specs/2026-04-15-cookie-consent-seo-importer-fix.md
```diff
@@ -0,0 +1,195 @@
+# Cookie Consent SEO Importer Fix — Implementation Plan
```
LeChef318, line 1: The verification trail does not match the dismissal rationale: this file was added in commit `ffb0c35e` on this feature branch (2026-04-27, alongside the null-CPC fix), two days before the merge from main. `main` does not contain this file (contents API returns 404; commits-by-path returns empty). Most likely an accidental bundle with the null-CPC commit. Please `git rm` it on this branch and ship it as its own PR with the matching `importAhrefPaidStep` code change.

src/paid-keyword-optimizer/guidance-handler.js
```diff
@@ -132,58 +83,70 @@ export default async function handler(message, context) {
...
   const entity = mapToKeywordOptimizerOpportunity(siteId, audit, message);
   const opportunity = await Opportunity.create(entity);
   paidLog.createdOpportunity(siteId, url, opportunity.getId());
 
-  const suggestionData = mapToKeywordOptimizerSuggestion(
+  // Assign composite ranks to clusters
+  const rankedClusters = assignClusterRanks(clusterResults);
+
+  // Create 1 suggestion per cluster
+  const suggestions = rankedClusters.map((cluster) => mapClusterToSuggestion(
```
LeChef318, line 134: Stale comment: `// Bulk-create suggestions`. The earlier response said this was updated to "Create suggestions" but the change is not in the tree at the current head. `Promise.all(map(Suggestion.create))` is N independent inserts, not a bulk write, so the comment is misleading. Trivial one-line fix.

src/paid-keyword-optimizer/guidance-handler.js
```diff
@@ -113,9 +46,27 @@ export default async function handler(message, context) {
   const { log, dataAccess } = context;
   const { Audit, Opportunity, Suggestion } = dataAccess;
   const { auditId, siteId, data } = message;
+
+  if (!data) {
```
LeChef318, line 50: Consider adding correlation context here. `auditId` and `siteId` are destructured at line 48 and the failure-envelope log at lines 54-60 already includes them as structured fields. Without them, a wave of missing-data warnings is untraceable in Coralogix. Suggest: `log.warn({ audit_id: auditId, site_id: siteId }, [ad-intent-mismatch] Received message with no data, skipping);`

LeChef318 (commented):
Hey @danieljchuser,

Thanks for the quick turnaround on the hardening commit and the detailed response to the prior review. Re-review based on commit `439f7a1` plus the two merges from main.

## Strengths

- The hardening commit (`439f7a1`) is the right shape: small, focused, three guards that map 1:1 onto the previously flagged failure modes, each with a dedicated test.
  - Missing-`data` guard: `src/paid-keyword-optimizer/guidance-handler.js:49-52`
  - Empty `clusterResults` gate: `src/paid-keyword-optimizer/guidance-handler.js:87`
  - Undefined-url filter: `src/paid-keyword-optimizer/guidance-handler.js:113`
- Simplifying `data?.status === 'failed'` to `data.status === 'failed'` at line 53 after the early-return is the right move. The optional chaining was load-bearing in the prior version; it is dead weight now and the cleanup signals "this invariant is established above".
- The undefined-url test is well constructed: it seeds two existing opportunities (one matching the url, one with `url: undefined`) and asserts neither is touched, which prevents the "matching pre-existing url is also undefined" false-positive from regressing.
- Defensive guards on the consumer side reduce blast radius from upstream Mystique contract drift, which is the right posture given Mystique deploys independently.

Previously flagged issues now addressed: handler-crashes-on-missing-`data`, empty-clusterResults-zombie, mass-IGNORE-on-undefined-url.

## Issues

### Important (Should Fix)

**Cookie-consent spec doc still bundled in this PR.** `docs/specs/2026-04-15-cookie-consent-seo-importer-fix.md` (+195 lines) is unrelated to multi-keyword cluster work. Re-raising the prior concern because the dismissal rationale does not line up with git history.

The pushback said the file came from `main` via merge commit `bc20002628`. What I see:

- The file was added in commit `ffb0c35e` on this feature branch (the "preserve null CPC in suggestion mapper" commit, 2026-04-27 15:15Z), two days before the merge from main (`bc20002628`, 2026-04-29 05:38Z). At `ffb0c35e` the file is `status: "added"` against parent `dd438ff9`.
- `main` does not contain this file: `gh api repos/adobe/spacecat-audit-worker/contents/docs/specs/2026-04-15-cookie-consent-seo-importer-fix.md` returns 404.
- No commit on main has ever touched this path: `gh api repos/.../commits?path=docs/specs/2026-04-15-cookie-consent-seo-importer-fix.md` returns empty.

Most likely it was bundled accidentally with the null-CPC commit. How to fix: `git rm docs/specs/2026-04-15-cookie-consent-seo-importer-fix.md` on this branch and ship it on its own PR with the matching `importAhrefPaidStep` code change. Worth doing for change discipline so the commit graph for ad-intent-mismatch stays focused.

### Minor (Nice to Have)

**Stale `// Bulk-create suggestions` comment at `src/paid-keyword-optimizer/guidance-handler.js:134`.** The earlier response indicated the comment was updated to "Create suggestions" but the change is not in the tree at the current head (`4a88757`). Trivial fix while you are in the file. The semantics matter because `Promise.all(map(Suggestion.create))` is N independent inserts, not a bulk write, and the distinction matters for partial-failure reasoning.

**Missing correlation context on the new `!data` warn log at `src/paid-keyword-optimizer/guidance-handler.js:50`.** `auditId` and `siteId` are already destructured at line 48 and the failure-envelope log at lines 54-60 already includes them as structured fields. The new warn line is just a string. If this fires under load, a wave of "no data, skipping" lines in Coralogix is untraceable to the originating site or audit. Suggest:

```js
log.warn({ audit_id: auditId, site_id: siteId }, '[ad-intent-mismatch] Received message with no data, skipping');
```

**Failure-envelope test covers only the fully populated `error` shape.** The handler defends each field with optional chaining (`data?.error?.langfuseTraceId`, `data?.error?.type`), but no test fails if someone later removes the optional chaining and an envelope arrives as `{ status: 'failed' }` with no `error` object. One additional test case covering that shape would lock the contract in. Not urgent; raising as nice-to-have.

## Recommendations

- The defensive-guards-per-consumer pattern this PR adds is a symptom worth tracking at the platform level. Mystique enforces the contract on the producer side via Pydantic; consumers each carry their own null-handling. A schema artifact at the boundary (JSON Schema, or shared types generated from the Pydantic models) would let consumers drop most of these guards and centralize the contract. Track against `spacecat-shared`, not this PR.

- The replace-on-re-audit atomicity concern and the silent-degradation-on-failure-envelope concern from the prior review are legitimate platform-wide patterns and the cross-cutting-fix argument is correct in scope. Worth tracking issues against the data-access layer and the guidance-handler observability template respectively, not blocking this PR.

## Assessment

Ready to merge? With fixes.

Reasoning: The handler hardening is solid, well-tested, and fully resolves three of the four prior Important findings. The remaining blocker is the cookie-consent spec doc, which is unrelated to this PR. Plus three small cleanups noted above.

## Next Steps

1. Drop the cookie-consent spec doc from this branch.
2. Update the stale "Bulk-create suggestions" comment.
3. Optional: structured fields on the new warn log; failure-envelope test for the `error: undefined` shape.

LeChef318 (approved):

PR #2441, opened 2026-04-29
feat: sync competitors, brand aliases, ignored prompts  | LLMO-4477

calvarezg:
## Summary

Extends the `llmo-config-db-sync` audit to sync the remaining entities from S3 config to Postgres needed for the LLMO-4403 PG-read rollout.

**New entities synced:**
- **Brand aliases** (`brand_aliases` table) — flattens `brands.aliases[].aliases[]` entries, normalises regions to uppercase arrays, reconciles additions/updates/deletions against existing rows
- **Competitors** (`competitors` table) — maps `competitors.competitors[]`, handles `region` as string or array, logs a warning when multiple URLs are present (DB has a single `url` column, only `urls[0]` is persisted)
- **Ignored prompts** — extends the existing prompt sync to include `ignored.prompts` entries, stored in the `prompts` table with `status='ignored'`, `topic_id=null`, `category_id=null`; maps the singular `region` field (not the array `regions` used elsewhere)
- **`topic_categories` junction** — upserts desired (topic, category) pairs and batches orphan deletions grouped by `topic_id`; scoped to the org by filtering on `topic_id IN [orgTopicUuids]` (table has no `org_id` column)

**Not synced (intentional):**
- **`topic_prompts` junction** — auto-populated by the `prompts_sync_junction_tables` DB trigger on `prompts` (LLMO-4288). Because the worker uses `ON CONFLICT DO UPDATE` upserts, every prompt write fires the trigger and keeps the junction in sync. Writing it from the worker would be redundant and risk drift. ADR 001 records the rationale.

## Refactor

The original monolithic `handler.js` (~500 lines) is decomposed into focused modules:

| File | Responsibility |
|---|---|
| `constants.js` | All magic values (UUIDs, batch sizes, compare field lists) |
| `diff.js` | `diffRows`, `changedFields`, `normalizeForCompare`, `logDiffSummary` |
| `fetch.js` | `fetchPromptsBatched` (paginated), `fetchExistingState` |
| `sync-categories.js` | Category upsert + lookup update |
| `sync-topics.js` | Topic upsert, `ensureDeletedRefEntities` |
| `sync-prompts.js` | `collectPrompts` (all statuses), `upsertInBatches`, `syncPrompts` |
| `sync-brand-aliases.js` | Brand alias reconcile |
| `sync-competitors.js` | Competitor reconcile |
| `sync-junctions.js` | `syncTopicCategories` |
| `handler.js` | Thin orchestrator (under `c8 ignore`) |

### Brand scoping for `topics` and `prompts` (also closes LLMO-4470)

Both tables carry a `brand_id` column, but the previous fetches filtered only by `organization_id` while writes used a hard-coded brand. This created two bugs:

1. **Topics:** `syncTopicCategories` would have classified other brands' junction rows as orphans and deleted them. New topic rows were also written without `brand_id`. Production data confirmed adobe.com had topic rows from 15 distinct brands sharing the same org.
2. **Prompts (LLMO-4470):** the dedup key `(text, topic_id)` plus a compare-field list that omits `brand_id` meant a config prompt matching a different brand's row was silently classified `unchanged`. The 2026-04-15 adobe.com sync reported 35 175 prompts as unchanged for this reason — only 5 365 of an expected ~34 000 prompts existed under the Adobe brand.

Both fetches now apply `.eq('brand_id', brandId)`, and topic upserts now include `brand_id` in every row written. ADR 001 has the full rationale.

> ⚠️ **Deployment ordering note (LLMO-4470):** the original ticket warns that ghost-brand cleanup must ship before this fix is enabled in production for adobe.com, otherwise the first run will create duplicate prompts (one per brand). This PR is safe to merge today; the operational sequence is to validate against the dev brand first and only roll out to prod brands after the related ghost-brand cleanup work has landed.

### Other behaviour notes

- All sync functions accept a `dryRun` flag — logs diffs without writing to DB.
- `syncTopicCategories` early-exits when the org has no topics in the lookup.
- `fetchExistingState` and `fetchPromptsBatched` now **throw on DB error** instead of logging and continuing with partial data — this matches every other fetch in the worker. (Pre-existing silent-failure behaviour from PR #2210, fixed in this refactor.)

### Behaviour change: array comparison is now order-insensitive

`normalizeForCompare` sorts array fields before comparison:

```js
// before (handler.js, PR #2210)
if (field === 'regions' && Array.isArray(value)) {
  return JSON.stringify(value.map((r) => r.toUpperCase()));
}

// after (diff.js)
if (Array.isArray(value)) {
  const items = field === 'regions' ? value.map((r) => r.toUpperCase()) : value;
  return JSON.stringify([...items].sort());
}
```

This is intentional — order-only differences should not trigger spurious updates — but it is a silent behaviour change for the existing prompt sync. **On the first run after deploy, expect a one-time wave of "unchanged" reclassifications** for any prompt rows whose stored `regions` array was a different ordering than the S3 config. No data is rewritten; only the diff classification changes.

## ADR

`docs/decisions/001-llmo-config-pg-sync-scope.md` (updated in this PR):
- Records which config keys are S3-authoritative, PG-authoritative, trigger-authoritative, or not synced
- Documents per-table upsert conflict keys and the "INSERT label vs. actual UPDATE" caveat in diff logs
- Documents the brand-scoping fix for topics and prompts (closes LLMO-4470) and the pre-existing 256-row cross-brand prompt-topic anomaly
- Documents the `urls[]` → `url` mismatch for competitors
- Documents the `created_by` ← `entry.createdBy` audit-field handling
- Documents that `topic_prompts` is maintained by a DB trigger, not the worker

## Test plan

- [ ] 113 unit tests across 8 test files, all passing
- [ ] 100% line, branch, and statement coverage on all `src/llmo-config-db-sync/*.js` modules (`handler.js` excluded via `c8 ignore`)
- [ ] `npm run lint` clean
- [ ] Dry-run on dev brand/site to confirm diff log output matches expectations
- [ ] After merge: full sync on dev, verify (a) `brand_aliases`, `competitors`, `ignored` prompts populate; (b) `topic_categories` orphan deletions group correctly; (c) `topic_prompts` is populated by the trigger after prompt upserts (no rows written from the worker)

## Related Issues
[LLMO-4477](https://jira.corp.adobe.com/browse/LLMO-4477)
[LLMO-4470](https://jira.corp.adobe.com/browse/LLMO-4470)


rainer-friederich (commented):
## Local multi-agent review (draft PR)

Posted at the author's request while the PR is still in draft. Issues with confidence score > 50, sorted by score descending. Score scale: 100 certain · 75 highly confident · 50 verified-but-low-impact.

Reviewers: 5 standard checks (CLAUDE.md, bug, history, prior PRs, code comments) + 6 topic experts (database, architecture × 2, testing, documentation, nodejs) + 1 domain-expert intent check.

### FIX (score ≥ 75)

| Score | Category | Description | File |
|---|---|---|---|
| 100 | bug / intent / code-comment | `const dryRun = true;` hardcoded with comment "Revert before merge" — overrides caller flag, no writes if merged, defeats stated PR purpose (DRS PG migration) | `src/llmo-config-db-sync/handler.js:706-707` |
| 95 | CLAUDE.md | No spec in `/docs/specs/` for significant new sync surface; CLAUDE.md: "Significant new functionality must be developed spec-first" | `src/llmo-config-db-sync/sync-*.js` |
| 88 | historical regression | `fetchPromptsBatched` changed from soft fail (`log.error + break`, partial result) to hard `throw` — PR 2210 used soft-fail by intent | `src/llmo-config-db-sync/fetch.js:335-337` |
| 88 | architecture | PostgREST as bulk write API is long-term liability (manual URL chunking, per-row delete loops, no transactions); use `wrpc_sync_<entity>` RPCs in mysticat-data-service | `src/llmo-config-db-sync/sync-*.js` |
| 88 | intent / docs | Per-run upsert overwrites `updated_by` from S3 every run; combined with NULL `created_by` policy, any non-S3 writer (UI/autofix) edits silently revert next sync | ADR §"Audit fields", `sync-*.js` |
| 85 | testing | `handler.js` excluded from coverage (`/* c8 ignore start */`) + dryRun hardcoded; no integration test for orchestrator wiring; sync-* signature changes silently bypass tests | `src/llmo-config-db-sync/handler.js:401` |
| 85 | docs | No operator runbook for `data_source_pg=true` flip: no verification, flip, or rollback procedure documented | `docs/decisions/001-llmo-config-pg-sync-scope.md:149-159` |
| 82 | bug | `normalizeForCompare` now sorts ALL arrays, not just `regions`; widens change-detection semantics for competitor `aliases` (string[]) — order-only changes silently treated as unchanged | `src/llmo-config-db-sync/diff.js:222-228` |
| 82 | architecture | Shared mutable lookup maps (`categoryLookup`, `topicLookup`, `topicNameLookup`) passed by reference and mutated across modules — opaque temporal coupling; needs a `SyncContext` object | `src/llmo-config-db-sync/handler.js` |
| 80 | architecture | `ensureDeletedRefEntities` lives in `sync-topics.js` but writes BOTH `categories` and `topics` tables — breaks single-entity-per-module cohesion | `src/llmo-config-db-sync/sync-topics.js:1593-1659` |
| 80 | nodejs | Errors wrap PostgREST `error.message` but drop `error.code/details/hint`; no `Error(msg, {cause})` — debug context lost in logs | `src/llmo-config-db-sync/sync-*.js`, `fetch.js` |
| 80 | docs | ADR Status "Accepted" / Date 2026-04-29 incompatible with `dryRun=true` "Revert before merge" override — implementation cannot deliver Consequences | `docs/decisions/001-llmo-config-pg-sync-scope.md:9` |
| 79 | bug | `syncCategories` does not update `categoryLookup` in dryRun → downstream `syncTopicCategories` silently drops rows with undefined category UUIDs; misleading dryRun report | `src/llmo-config-db-sync/sync-categories.js:1104` |
| 78 | bug | `syncTopicCategories` reads stale `topicLookup` when `syncTopics` skips upsert (dryRun) — `buildTopicCategoryRows` silently drops rows | `src/llmo-config-db-sync/sync-junctions.js:1290` |
| 78 | architecture | No transactional boundary across the 6 sync calls; PostgREST has no multi-statement transaction; mid-run failure → DRS reads half-applied state | `src/llmo-config-db-sync/handler.js` |
| 75 | intent | 256 pre-existing rows where `prompts.brand_id != topics.brand_id`: ADR notes backfill required before `data_source_pg=true` for multi-brand orgs, but PR doesn't make this a hard release-gate | ADR §"Cross-brand mismatches" lines 140-145 |
| 75 | intent / arch | Hardcoded `PROD_BRAND_ID` / `DEV_BRAND_ID` / `ALLOWED_SITE_IDS` ship despite ADR flagging blocker; allow-list-fail mode is data-shaped if extra site is added | `src/llmo-config-db-sync/constants.js:187-197` |
| 75 | database / intent | Competitors persist only `urls[0]` with `log.warn` — silent data loss for multi-URL configs; no follow-up ticket cited in ADR | `src/llmo-config-db-sync/sync-competitors.js:1141-1157` |
| 75 | database | No delete/tombstone for prompts or categories removed from S3; aliases/competitors/junctions have it. Removed prompt stays `status='active'` forever → DRS returns stale | `src/llmo-config-db-sync/sync-prompts.js`, `sync-categories.js` |
| 75 | database | No advisory lock around read-existing → diff → write window; overlapping runs (re-deliveries, manual replay) race; aliases/competitors/junctions deletes can flap | `src/llmo-config-db-sync/handler.js` |
| 75 | database | `existingPrompts` fetched per-org but lookup key omits `brand_id`; in multi-brand orgs same-text prompts cross-collide (topics fetch was tightened, prompts not) | `src/llmo-config-db-sync/fetch.js:329-333` |
| 75 | architecture | 5 entity sync modules repeat fetch → build → diff → log → upsert pattern; no generic `syncEntity({...})` helper; every new entity = new bespoke module | `src/llmo-config-db-sync/sync-*.js` |
| 75 | docs | No JSDoc on exported sync functions; side effects (Map mutation, throws on PostgREST error) undocumented | `src/llmo-config-db-sync/*.js` |
| 75 | nodejs | No client-side timeout on any PostgREST call; hung dependency blocks until Lambda timeout | `src/llmo-config-db-sync/fetch.js`, `sync-*.js` |

### CONSIDER (score 50–74)

| Score | Category | Description | File |
|---|---|---|---|
| 72 | prior PR feedback | Verbose `[DIFF]` `log.info` (per-row JSON, field freq) retained and extended to 4 more entities; PR 2210 review explicitly said "log info should be removed" | `src/llmo-config-db-sync/diff.js:267-302` |
| 72 | architecture | Sync function signatures take 6–10 positional params with inconsistent order (e.g. `syncTopics` org → brand vs `syncPrompts` brand → org) | `src/llmo-config-db-sync/handler.js`, `sync-*.js` |
| 72 | intent / docs | Brand aliases drop `category` and `aliasMode` fields silently; `aliasMode` likely controls matching semantics → DRS PG path differs from S3 path | ADR §"Brand aliases" lines 60-62, `sync-brand-aliases.js` |
| 70 | testing | No PostgREST 207 / partial-success mocking; tests resolve `{error: null}` or `{error}` only; code treats `data \|\| []` as ground truth → silent UUID drops in lookups | `test/audits/llmo-config-db-sync/sync-categories.test.js:2244-2295` |
| 65 | nodejs | Serial `for + await` pagination/upsert (`upsertInBatches`, `fetchPromptsBatched`); no bounded concurrency for thousands of rows | `src/llmo-config-db-sync/sync-prompts.js:1477-1492`, `fetch.js:323-347` |
| 65 | architecture | Cross-domain inconsistency: bypasses `spacecat-shared-data-access` used by adjacent audit handlers; new workers will copy whichever pattern they see first | `src/llmo-config-db-sync/handler.js` |
| 65 | intent | Ignored prompts written with `topic_id: null, category_id: null` regardless of source prompt's scope; semantics (per-org / brand / topic) undocumented | `src/llmo-config-db-sync/sync-prompts.js:1452-1472` |
| 60 | testing | `promptLookupKey` `\0` separator collision risk: text with embedded `\0` could collide with adjacent fields; untested | `src/llmo-config-db-sync/sync-prompts.js`, `test/...sync-prompts.test.js:2728` |
| 60 | intent | PR title hides material domain changes (`topic_categories` junction, `brand_id` scoping, deleted-ref backfill); ADR documents but title doesn't | ADR line 46 |
| 55 | testing | No malformed S3 config tests (`topics: "string"`, `aliases: {}`); `Object.entries` / `forEach` will throw at runtime | `test/audits/llmo-config-db-sync/sync-topics.test.js` |

### Top blockers before merge

1. Revert `dryRun = true` hardcode in `handler.js`. Every other claim about behavior is moot until this is gone.
2. Add `/docs/specs/` entry per CLAUDE.md spec-first rule.
3. Make the 256-row backfill and dynamic brand resolution hard release-gates on `data_source_pg=true`, not deferred items.
4. Wire `Error.cause` and log every `updated_by` overwrite at debug level so non-S3 writer reverts are detectable.
5. Fix dryRun lookup mutation gap, brand_id scope on prompts fetch, and array-sort semantics in `normalizeForCompare`. These break diff correctness silently.

PR not mergeable as-is. Issue 1 alone blocks. Recommend addressing 16 (256-row gate) and 17 (hardcoded brand IDs) before flipping out of draft so reviewers don't re-litigate scope.

PR #2442, opened 2026-04-29
fix(drs-prompt-generation): drop non-alpha-2 region values before writeConfig (SITES-43238)

dzehnder:
## Summary

Step 1 of the [SITES-43238](https://jira.corp.adobe.com/browse/SITES-43238) fix plan. Stops the bleeding by filtering invalid region values at the DRS writer **before** they reach `writeConfig` and corrupt the LLMO config on S3.

### Problem

Upstream DRS occasionally emits region values that don't match the schema-validated alpha-2 format (e.g. `"en-us"`, `"global"`, `"united-states"`). Because `writeConfig` in `@adobe/spacecat-shared-utils` doesn't validate against the `llmoConfig` Zod schema before persisting, these values were silently written to S3 and then broke every subsequent schema-validated read with HTTP 400 (e.g. `GET /api/v1/sites/:id/llmo/config`).

Confirmed broken in production for site `78d59744-e06c-4d14-a77a-9490c1464116` and at risk across the entire DRS-onboarded tenant population.

### Fix

In `src/drs-prompt-generation/drs-config-writer.js`:

- Add a `normalizeRegion(raw)` helper enforcing the same shape the Zod schema requires (`/^[a-z]{2}$/` after lowercasing, non-string rejected).
- Pre-process the incoming `drsPrompts` array: keep prompts, normalize valid regions, strip invalid ones.
- Emit a single aggregated `log.warn` per write — e.g. `{"en-us":23,"global":5}` — so dropped values stay diagnosable without log volume scaling with prompt count.
- Drop the now-redundant downstream `.toLowerCase()` calls.

Prompts themselves are kept (the schema permits empty `regions: []`); only invalid region values are stripped.

### Scope (and what's NOT in this PR)

This is step 1 of 3:

1. **This PR** — filter at the writer (audit-worker). Stops new corruption.
2. *Next PR* — fail-closed validation in `writeConfig` itself (spacecat-shared-utils). Requires step 1 to soak >=24h in prod first, plus an audit of all `writeConfig` callers.
3. *Next PR* — one-shot repair of already-corrupted S3 configs. Requires a read-only sweep first to size the population.

### Test plan

- [x] New tests in `test/drs-prompt-generation/drs-config-writer.test.js`:
  - drops \`en-us\`, \`global\`, mixed valid+invalid; aggregated WARN includes \`"en-us":2\`, \`"global":1\`
  - does not warn when all regions are valid (also covers uppercase normalization, e.g. \`"DE"\` -> \`"de"\`)
  - drops non-string region values (e.g. \`123\`) gracefully
- [x] \`npm test\` — 7579 passing, 100% coverage on \`src/**/*.js\`
- [x] \`npm run lint\` — clean
- [ ] Soak in prod >=24h before merging step 2

### DIFF
--- a/src/drs-prompt-generation/drs-config-writer.js
+++ b/src/drs-prompt-generation/drs-config-writer.js
@@ -13,10 +13,27 @@
 import { randomUUID } from 'crypto';
 import { llmoConfig as sharedLlmoConfig } from '@adobe/spacecat-shared-utils';
 
+// Region values must match the schema in @adobe/spacecat-shared-utils
+// (ISO-3166 alpha-2). Upstream DRS occasionally emits non-conformant values
+// like "en-us" or "global" — those would corrupt the LLMO config and break
+// every subsequent schema-validated read. See SITES-43238.
+const REGION_REGEX = /^[a-z]{2}$/;
+
+function normalizeRegion(raw) {
+  if (typeof raw !== 'string') {
+    return null;
+  }
+  const lc = raw.toLowerCase();
+  return REGION_REGEX.test(lc) ? lc : null;
+}
+
 /**
  * Groups DRS prompts by category and topic, then writes them into
  * the LLMO config as aiTopics so they appear in the UI and are used
  * by downstream brand-presence analysis flows.
+ *
+ * Non-alpha-2 region values are dropped (with an aggregated WARN log)
+ * before grouping so they cannot reach the schema-validated S3 config.
  *
  * @param {object} params
  * @param {Array<object>} params.drsPrompts - Raw prompts from DRS
@@ -36,6 +53,25 @@ export default async function writeDrsPromptsToLlmoConfig({
   log,
   configClient = sharedLlmoConfig,
 }) {
+  const droppedRegions = new Map();
+  const normalizedPrompts = drsPrompts.map((p) => {
+    if (!p.region) {
+      return p;
+    }
+    const valid = normalizeRegion(p.region);
+    if (valid) {
+      return { ...p, region: valid };
+    }
+    droppedRegions.set(p.region, (droppedRegions.get(p.region) || 0) + 1);
+    const rest = { ...p };
+    delete rest.region;
+    return rest;
+  });
+  if (droppedRegions.size > 0) {
+    const summary = Object.fromEntries(droppedRegions);
+    log.warn(`Dropped non-alpha-2 region values from DRS prompts for site ${siteId}: ${JSON.stringify(summary)}`);
+  }
+
   const { config } = await configClient.readConfig(siteId, s3Client, { s3Bucket });
 
   if (!config.aiTopics) {
@@ -54,7 +90,7 @@ export default async function writeDrsPromptsToLlmoConfig({
   // Group prompts by topic (DRS topic field) within each category
   // DRS prompt shape: { prompt, region, category, topic, base_url }
   const grouped = {};
-  for (const p of drsPrompts) {
+  for (const p of normalizedPrompts) {
     const catName = p.category || 'general';
     const topicName = p.topic || 'general';
     const key = `${catName}|||${topicName}`;
@@ -88,7 +124,7 @@ export default async function writeDrsPromptsToLlmoConfig({
         if (!categoryRegions.has(categoryId)) {
           categoryRegions.set(categoryId, new Set());
         }
-        categoryRegions.get(categoryId).add(p.region.toLowerCase());
+        categoryRegions.get(categoryId).add(p.region);
       }
     }
 
@@ -101,7 +137,7 @@ export default async function writeDrsPromptsToLlmoConfig({
           promptMap[text] = { regions: new Set(), type: p.type || '' };
         }
         if (p.region) {
-          promptMap[text].regions.add(p.region.toLowerCase());
+          promptMap[text].regions.add(p.region);
         }
       }
     }
--- a/src/drs-prompt-generation/drs-config-writer.js
+++ b/src/drs-prompt-generation/drs-config-writer.js
@@ -101,12 +101,36 @@ export default async function writeDrsPromptsToLlmoConfig({
     grouped[key].prompts.push(p);
   }
 
+  // Determine which categories have at least one prompt with a valid region
+  // across all their topics. New categories without a valid region cannot be
+  // safely persisted because the schema requires `region`.
+  const categoriesWithValidRegion = new Set();
+  for (const p of normalizedPrompts) {
+    if (p.region) {
+      categoriesWithValidRegion.add((p.category || 'general').toLowerCase());
+    }
+  }
+
+  // Drop any group whose category is new AND has no valid region to assign,
+  // so we never create a category that violates the schema.
+  const droppedCategories = new Set();
+  const groupsToProcess = Object.values(grouped).filter((group) => {
+    const catKey = group.category.toLowerCase();
+    const isNew = !categoryNameToId[catKey];
+    if (isNew && !categoriesWithValidRegion.has(catKey)) {
+      droppedCategories.add(group.category);
+      return false;
+    }
+    return true;
+  });
+
   const now = new Date().toISOString();
   const categoryRegions = new Map();
 
-  for (const { category: catName, topic: topicName, prompts } of Object.values(grouped)) {
-    // Find or create category
-    let categoryId = categoryNameToId[catName.toLowerCase()];
+  for (const { category: catName, topic: topicName, prompts } of groupsToProcess) {
+    const catKey = catName.toLowerCase();
+    let categoryId = categoryNameToId[catKey];
+
     if (!categoryId) {
       categoryId = randomUUID();
       config.categories[categoryId] = {
@@ -115,7 +139,7 @@ export default async function writeDrsPromptsToLlmoConfig({
         updatedBy: 'drs',
         updatedAt: now,
       };
-      categoryNameToId[catName.toLowerCase()] = categoryId;
+      categoryNameToId[catKey] = categoryId;
     }
 
     // Track regions for this category
@@ -167,6 +191,10 @@ export default async function writeDrsPromptsToLlmoConfig({
   for (const [catId, regions] of categoryRegions) {
     const arr = [...regions];
     config.categories[catId].region = arr.length === 1 ? arr[0] : arr;
+  }
+
+  if (droppedCategories.size > 0) {
+    log.warn(`Skipped DRS categories with no valid region for site ${siteId}: ${JSON.stringify([...droppedCategories])}`);
   }
 
   const { version } = await configClient.writeConfig(siteId, config, s3Client, { s3Bucket });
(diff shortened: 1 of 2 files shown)


src/drs-prompt-generation/drs-config-writer.js
```diff
@@ -88,7 +124,7 @@ export default async function writeDrsPromptsToLlmoConfig({
         if (!categoryRegions.has(categoryId)) {
           categoryRegions.set(categoryId, new Set());
         }
-        categoryRegions.get(categoryId).add(p.region.toLowerCase());
+        categoryRegions.get(categoryId).add(p.region);
```
solaris007, line 151: When a brand-new category has ALL prompts with invalid regions, this `.add()` never executes for that category. `categoryRegions` gets no entry, so the final loop never sets `config.categories[catId].region`. Since the Zod schema declares `region` as required, this still produces a config that fails the next read. Consider skipping category creation when no valid region survives normalization.

solaris007 (changes requested):
Hey @dzehnder,

Clean, well-scoped stop-the-bleed fix. The 3-step plan is clear, the PR description is unusually disciplined about what's in and out of scope, and the aggregated WARN with value->count diagnostics gives ops exactly the data needed to drive Step 2 prioritization and Step 3 repair sizing.

## Strengths

- `normalizeRegion` is clean: explicit `typeof` guard, anchored fixed-length regex (not ReDoS-susceptible), single normalization point upstream that eliminates the now-redundant `.toLowerCase()` calls downstream (`drs-config-writer.js:127`, `drs-config-writer.js:140`)
- Aggregated `Map` + `JSON.stringify` is the right primitive - one WARN per write with value->count breakdown keeps log volume flat while preserving diagnostics. JSON encoding closes off log-forging through the raw values. Counting keyed on the raw input preserves diagnostic fidelity (e.g. `"en-us"` vs `"EN-US"` are distinct)
- Schema regex alignment confirmed: `@adobe/spacecat-shared-utils` uses `z.string().length(2).regex(/^[a-z][a-z]$/i)`, and this PR's `/^[a-z]{2}$/` post-lowercase is functionally equivalent
- The comment block above `REGION_REGEX` (lines 16-19) explains the "why" and references SITES-43238 - the next person debugging an LLMO config corruption will find the breadcrumb
- Drop region, keep prompt - preserves user-facing content while shedding the schema-toxic field. Spread + delete correctly avoids mutating the caller's objects
- Tests cover both directions of the warn signal (fires with counts when dirty, silent when clean) plus non-string and uppercase normalization
- Zero new dependencies - supply chain footprint unchanged

## Issues

### Important (Should Fix)

**New category with only invalid regions is written without the required `region` field** - `drs-config-writer.js`

When a brand-new category is created and ALL prompts in that category had invalid regions (e.g. every prompt has `region: "en-us"`), `categoryRegions` never gets an entry for that category ID. The final loop that sets `config.categories[catId].region` skips it entirely. The Zod schema declares `region` as required (`z.union([region, z.array(region)])`), so this produces a config that fails the next schema-validated read - the same class of corruption this PR aims to prevent.

How to fix: skip creating the category when no valid region survives normalization. Add a test for "category where all prompts have only invalid regions is not persisted."

### Minor (Nice to Have)

1. **Aggregated WARN message has no upper bound on size** (`drs-config-writer.js:68`) - If DRS emits a degenerate payload with many unique invalid strings, the single `log.warn` line could become very large or get truncated. Consider capping the number of distinct entries reported (top-N by count plus an "others: M" tail).

2. **Whitespace-padded regions not tested** (`drs-config-writer.test.js`) - `'us '` and `' us '` hit the regex-fail path and get dropped, which is the intended behavior. Pinning this with one assertion prevents future drift if someone later adds `.trim()` inside `normalizeRegion`.

3. **Prompts with `regions: []` persisted when all regions for a prompt text were invalid** - The schema allows empty arrays, so this passes validation, but the operational meaning of "an AI-generated prompt with zero regions" is ambiguous for downstream brand-presence flows. Consider dropping such prompts when handling the empty-category case above.

## Recommendations

- **Lock down the falsy-region contract in a test.** The current code preserves prompts where `region` is `null`, `undefined`, or `''` via `if (!p.region) return p;`. None of the new tests assert this explicitly. A one-line assertion prevents a future maintainer from tightening that check and silently dropping prompts the schema permits.
- **Decide explicitly on `.trim()`**: today the implicit policy is "no trim, padded values are dropped." Either decision is defensible - just make it deliberate.
- **CloudWatch/Coralogix metric for dropped regions**: a counter would make "is DRS regressing this week?" trivially queryable without log grep. Good candidate for Step 2 or a follow-up.
- **When Step 2 lands, decide explicitly whether to keep this writer-side filter**: defense-in-depth is legitimate, but normalizers across producers can diverge from the canonical schema over time. A short note in the Step 2 PR would prevent that drift.
- **"global" is likely meaningful intent** from the DRS side (not a typo). The schema has no representation for "all regions" - worth a sibling ticket so the DRS team has a recourse.

## Assessment

Ready to merge? **With fixes.**

The Important finding is the same class of bug this PR claims to fix - a write that produces a config the next read can't parse. Fixing it is a few lines and one test. Everything else is minor/recommendation-level, and the architectural placement (writer-side stop-gap with explicit Step 2 follow-up in the correct repo) is sound.

## Next Steps

1. Address the category-without-region edge case (Important) - skip category creation when no valid regions survive normalization.
2. Review the minor items at your discretion - none are blocking.

dzehnder:
Thanks for the thorough review @solaris007 — pushed [4c88ae1](https://github.com/adobe/spacecat-audit-worker/commit/4c88ae13) addressing the Important finding plus the two minor tests that were quick to land.

## Addressed

**Important — new category without valid region.** You were right that this was the same class of bug the PR aims to prevent. Note this also existed before my PR for prompts with no region at all, since the schema treats both \`region: missing\` and \`region: dropped\` identically. Approach:

- Pre-compute \`categoriesWithValidRegion\` across all prompts (not just within a single group, since one category can span multiple topics).
- Filter \`grouped\` before the main loop: groups whose category is new AND has no valid region anywhere are dropped, tracked in \`droppedCategories\`, and emitted as a second WARN at write time.
- Existing categories are unaffected — they keep their stored region, and prompts (even with empty \`regions: []\`) get appended.

This did surface a side-effect: the existing test \`handles prompts without region\` was asserting the buggy behavior (category created without \`region\`). I rewrote it as the falsy-region contract test you asked for, and added an anchor prompt to \`defaults category and topic to \"general\"\` so it keeps testing what its name says.

**Tests added/updated:**
- \`skips new categories where every prompt has an invalid region\` (the Important fix)
- \`keeps adding prompts to an existing category when new prompts have only invalid regions\` (covers the merge case — existing region preserved, prompts persisted with empty \`regions: []\`)
- \`preserves prompts with falsy regions (null, undefined, empty string) without WARN\` (your recommendation 1)
- \`drops whitespace-padded region values (no implicit trim)\` (your minor 2 — pins the no-trim policy as a deliberate decision)

\`npm test\` 7582 passing, 100% coverage on \`drs-config-writer.js\`, lint clean.

## Deferred (with rationale)

**Aggregated WARN size cap (minor 1).** Worth doing if log volume becomes an issue, but the realistic bad-value cardinality from DRS is small (handful of language-region tags, a few free-form strings). I'd rather see real log volume from prod first before adding the top-N+others machinery — happy to add a follow-up ticket if you'd prefer it tracked explicitly.

**Empty \`regions: []\` prompts persisted (minor 3).** This is a product question more than a correctness one — the schema permits it, and \"AI prompt with no region\" might still be useful for non-region-scoped flows. Want to leave that decision to whoever owns the brand-presence pipeline rather than encode it silently here.

**CloudWatch/Coralogix counter metric.** Agreed this is the right shape for \"is DRS regressing this week?\" — will pick this up either as part of step 2 or as a small follow-up. Filed mentally; happy to formalize a sub-task on SITES-43238 if useful.

**Sibling ticket for \"global\" intent.** Good catch — \"global\" almost certainly is meaningful intent on the DRS side and our schema has no representation for it. Will file a separate ticket against the DRS team rather than treat it as a defect.

**Step 2 (fail-closed \`writeConfig\`) — keep this writer-side filter?** I'll address explicitly in the step 2 PR description so the decision is visible. Current lean: keep it as defense-in-depth, with the writer-side filter being lossless (drops bad regions but keeps prompts where possible) and \`writeConfig\` validation being the hard backstop.

Ready for re-review when you have a moment.

solaris007 (commented):
Hey @dzehnder,

Solid follow-up. The Important finding from the first review is correctly resolved, the deferred items have substantive rationale, and the new tests lock down the contracts we flagged.

## Strengths

- Previously flagged "New category with only invalid regions written without required `region` field" is resolved (`drs-config-writer.js:104-124`). The `categoriesWithValidRegion` precompute plus `groupsToProcess` filter ensures new categories cannot reach `config.categories` without at least one valid alpha-2 region. Test `skips new categories where every prompt has an invalid region` verifies both `categories === {}` and `aiTopics === {}`, preventing orphaned topic references.
- Previously flagged "whitespace-padded regions not tested" is resolved. New test pins the "no implicit trim" policy.
- Previously flagged "falsy-region contract" recommendation is resolved. Rewritten test covers null, undefined, '' plus a valid anchor, asserting no WARN fires.
- Existing-category invariant correctly preserved: only new categories are dropped; existing categories continue to receive prompts with empty `regions: []` without losing their prior `region` value. Test `keeps adding prompts to an existing category when new prompts have only invalid regions` verifies this.
- Two WARN messages provide clean, separable operational signals: one for upstream data quality (non-conformant region values), one for downstream impact (categories that could not be persisted).
- Author's pushback on prior Minor 1 (WARN size cap) holds - bad-value cardinality from DRS is bounded by upstream emission patterns, not adversarial input. Agreed to defer.
- Author's pushback on prior Minor 3 (empty `regions: []` persisted) holds - schema permits it, product semantics belong to the brand-presence pipeline owner.

## Issues

### Minor (Nice to Have)

**Missing test: new category where ALL prompts have falsy regions (no valid anchor)** - The new logic introduces a behavioral asymmetry between new and existing categories when all prompts have falsy regions (null/undefined/''):
- Existing category: prompts persisted with `regions: []`, no WARN
- New category: category and prompts dropped, category-WARN fires

This exact path is untested. The `preserves prompts with falsy regions` test has a Q4 anchor with valid `'us'` that keeps the category alive, and `skips new categories where every prompt has an invalid region` uses invalid strings, not falsy values. The all-falsy-no-anchor path takes a distinct route: `categoriesWithValidRegion` stays empty, `droppedRegions` stays empty (no region WARN), only the category-WARN fires. A ~5 line test would pin this contract and prevent silent regression if a future refactor changes the filter behavior.

## Recommendations

- Consider a brief comment near the `groupsToProcess` filter noting the asymmetry: new categories require an anchor because the schema requires `region` on creation, while existing categories already satisfy the schema.
- (Pre-existing, not this PR's scope) `categoryNameToId = {}` inherits from `Object.prototype`; if DRS ever emits `category: "constructor"` or `"__proto__"`, the property lookup would return inherited values. Swapping to `Object.create(null)` or a `Map` closes this. Worth considering if this writer gets another pass.

## Assessment

Ready to merge? **Yes.**

The Important finding from the first review is correctly resolved by construction. The author's pushback on deferred items is substantive and well-reasoned. The one minor gap (all-falsy-new-category test) is a contract-pinning improvement, not a correctness issue.

solaris007 (approved):

PR #2445, opened 2026-04-29
fix(no-cta-above-the-fold): filter on absolute thresholds instead of bounce share

alexandraa-kupreeva:
## Summary

The no-cta-above-the-fold audit was missing a lot of pages it should have flagged. The query required each candidate path to absorb at least 10% of its channel's total bounces — a bar that scales with channel size. On sites that run paid traffic across many pages, no single page could clear it, so the audit returned almost nothing actionable. On sites with very low paid traffic the same filter let through paths backed by only a few sampled bounces — not enough signal to trust.

This PR replaces the share-based filter with two absolute thresholds that don't scale with site size, so the audit detects bounce-prone paid pages consistently regardless of how big or small the site is.

 ## Changes
 
  - **Drop the 10% bounce-share filter** — the main reason high-traffic paid pages on busy sites were being hidden.
  - **Add a 50% absolute bounce-rate floor** — surfaced paths must bounce above their channel baseline AND above 50% in absolute terms.
  - **Add a `bounces >= 25` filter** — the bounce-rate signal must be built on enough sampled events to be statistically reliable.
  - **Rank candidates by projected traffic lost** (pageviews × bounce_rate) instead of bounce-share percentage, so the report prioritizes pages by absolute traffic impact.
  - Drop the `bounce_share_pct` column from the output (no downstream consumers).

### DIFF
--- a/src/no-cta-above-the-fold/queries.js
+++ b/src/no-cta-above-the-fold/queries.js
@@ -53,42 +53,33 @@ source_stats AS (
     FROM mobile_paid
     GROUP BY trf_channel
 ),
-highest_pageviews AS (
+candidates AS (
     SELECT
-        path,
-        trf_channel,
-        pageviews,
-        row_count,
-        bounces,
-        CAST(bounces AS DOUBLE) / NULLIF(row_count, 0) AS bounce_rate
-    FROM mobile_paid
-    WHERE pageviews >= ${pageViewThreshold}
-),
-top_bounces AS (
-    SELECT
-        h.path,
-        h.trf_channel,
-        h.pageviews,
-        h.row_count,
-        h.bounce_rate,
+        p.path,
+        p.trf_channel,
+        p.pageviews,
+        p.row_count,
+        p.bounces,
+        CAST(p.bounces AS DOUBLE) / NULLIF(p.row_count, 0) AS bounce_rate,
         ss.channel_bounce_rate,
-        h.bounces,
         ss.channel_bounces,
-        CAST(h.bounces AS DOUBLE) / NULLIF(ss.channel_bounces, 0) AS bounce_share,
-        CAST(h.bounces AS DOUBLE) / NULLIF(ss.channel_bounces, 0) * 100 AS bounce_share_pct
-    FROM highest_pageviews h
-    JOIN source_stats ss ON h.trf_channel = ss.trf_channel
-    WHERE h.bounce_rate >= ss.channel_bounce_rate
-      AND CAST(h.bounces AS DOUBLE) / NULLIF(ss.channel_bounces, 0) >= 0.10
+        CAST(p.pageviews AS DOUBLE) * CAST(p.bounces AS DOUBLE) / NULLIF(p.row_count, 0) AS projected_traffic_lost,
+        CAST(p.pageviews AS DOUBLE) * CAST(p.bounces AS DOUBLE) / NULLIF(p.row_count, 0) * ${ESTIMATED_CPC} AS projected_traffic_value
+    FROM mobile_paid p
+    JOIN source_stats ss
+      ON p.trf_channel = ss.trf_channel
+    WHERE p.pageviews >= ${pageViewThreshold}
+      AND p.bounces >= 25
+      AND CAST(p.bounces AS DOUBLE) / NULLIF(p.row_count, 0) >= GREATEST(ss.channel_bounce_rate, 0.50)
 ),
 deduped AS (
     SELECT
         *,
         ROW_NUMBER() OVER (
             PARTITION BY path
-            ORDER BY bounce_share DESC, pageviews DESC
+            ORDER BY projected_traffic_lost DESC, pageviews DESC, bounces DESC
         ) AS path_rank
-    FROM top_bounces
+    FROM candidates
 )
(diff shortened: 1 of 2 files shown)


src/no-cta-above-the-fold/queries.js
```diff
@@ -53,42 +53,33 @@ source_stats AS (
...
+    FROM mobile_paid p
+    JOIN source_stats ss
+      ON p.trf_channel = ss.trf_channel
+    WHERE p.pageviews >= ${pageViewThreshold}
+      AND p.bounces >= 25
+      AND CAST(p.bounces AS DOUBLE) / NULLIF(p.row_count, 0) >= GREATEST(ss.channel_bounce_rate, 0.50)
 ),
 deduped AS (
     SELECT
         *,
```
danieljchuser, line 88: These two thresholds (25 bounces, 0.50 bounce-rate floor) are core tuning parameters. Consider extracting them to named JS constants (like `ESTIMATED_CPC`) so they are self-documenting and easy to find when tuning is needed.

danieljchuser (changes requested):
Hey @alexandraa-kupreeva,

## Strengths

- Clean simplification at `queries.js:56-82`: merging the two-stage CTE chain (`highest_pageviews` -> `top_bounces`) into a single `candidates` CTE removes intermediate state and makes the query easier to follow.
- Moving `projected_traffic_lost` and `projected_traffic_value` computation into the CTE (line 72-73) instead of recomputing in the final SELECT is a good DRY improvement - the values are now reusable in the dedupe ORDER BY without expression duplication.
- `GREATEST(ss.channel_bounce_rate, 0.50)` at line 78 is an elegant way to enforce both "above channel baseline" and "above 50% absolute" in a single predicate. Nicely done.
- The PR description is excellent - clearly explains the problem (share-based filter scales with channel size), the fix (absolute thresholds), and the tradeoffs. This is how a fix PR should be documented.
- Removal of `bounce_share_pct` verified safe - no downstream consumers in the opportunity mapper, handler, or guidance handler reference this column.

## Issues

### Important (Should Fix)

**`queries.js:77-78` - Tuning thresholds hardcoded as SQL literals without named constants.** The values `25` (minimum bounces) and `0.50` (absolute bounce-rate floor) are core tuning parameters that define what the audit considers "bounce-prone". They're embedded directly in the SQL string with no documentation in code. By contrast, `pageViewThreshold` is env-driven and `ESTIMATED_CPC` is a named JS constant imported from `guidance-opportunity-mapper.js`. These new thresholds follow neither pattern, so a future maintainer won't know the rationale for "25" without finding this PR.

**Why it matters:** When someone needs to tune the audit's sensitivity, they'll have to grep through SQL template strings to find the magic numbers. Named constants are self-documenting.

**How to fix:** Extract to named JS constants alongside `ESTIMATED_CPC`, e.g.:
```js
const MIN_BOUNCES = 25;
const ABSOLUTE_BOUNCE_RATE_FLOOR = 0.50;
```
Then interpolate in the SQL: `AND p.bounces >= ${MIN_BOUNCES}` and `>= GREATEST(ss.channel_bounce_rate, ${ABSOLUTE_BOUNCE_RATE_FLOOR})`.

### Minor (Nice to Have)

**`queries.js:84` - Dedupe sort key change subtly affects channel attribution.** The old dedupe picked the row with highest `bounce_share` when a path appeared in multiple channels; the new dedupe picks by `projected_traffic_lost`. This means the same path could now be reported under a different channel than before. Not a bug - it aligns with the PR's intent to rank by traffic impact - but it's a non-obvious behavioral shift that would benefit from a brief inline comment explaining the tie-breaking intent.

## Recommendations

- **Test assertions for new constructs.** The existing tests (`queries.test.js`) only assert string shape and parameter substitution. They wouldn't catch a regression if the `25`, `0.50`, or `GREATEST(...)` were accidentally removed. Adding targeted assertions like `expect(query).to.include('bounces >= 25')` and `expect(query).to.include('GREATEST')` would catch accidental regressions cheaply.
- **Consider a result row safety cap.** The peer audit `paid-cookie-consent` accepts an optional `limit` parameter. This audit has no result cap, and each row triggers a Mystique LLM call via SQS. After changing the filter criteria, row counts could shift unpredictably for sites with many paid pages. An optional `LIMIT` (or logging the candidate count) would provide a safety net.

## Assessment

Ready to merge? With fixes

The core logic is sound and well-motivated - replacing a scale-dependent share filter with absolute thresholds is the right call. The math equivalence for `projected_traffic_lost` checks out, the column removal is verified safe, and CI is green. The one Important item (extracting named constants) is a small change that improves maintainability and consistency with existing patterns in this codebase.

## Next Steps

1. Extract the two hardcoded thresholds (25, 0.50) to named JS constants.
2. Optionally, add targeted test assertions and consider a brief comment on the dedupe tie-breaker.

danieljchuser (approved):
All findings addressed - nice work!

- Named constants (`MIN_BOUNCES`, `ABSOLUTE_BOUNCE_RATE_FLOOR`) with clear comments - exactly what was needed.
- Dedupe tie-breaker comment clarifies the channel-attribution intent.
- Three new targeted test assertions cover the thresholds and ordering.

LGTM.

PR #2449, opened 2026-04-30
fix(agentic): queue weekly DB export refreshes by date

akshaymagapu:
## Summary
- Preserve the default `cdn-logs-report` behavior: normal/date-based runs still execute one daily Agentic DB export.
- For weekly DB refresh triggers, queue one date-based `cdn-logs-report` message per completed UTC day instead of running 7 exports inline.
- Use `refreshAgenticDailyExport` as the explicit weekly DB refresh signal for LLMO category changes and BYOCDN/CDN config refresh paths.
- Keep `categoriesUpdated` scoped to category/pattern/report refresh behavior, not as its own DB export trigger.

## Use Cases Covered
- Normal scheduled report run: SharePoint/report flow + yesterday's DB export.
- Date-based report run: single-day DB export for the provided date.
- Category-change refresh: weekly report refresh, then queued daily DB exports for that week.
- BYOCDN/CDN config refresh: weekly refresh expands into completed daily DB export jobs.
- Current-week refresh: only completed UTC days are queued; future days are skipped.

## Jira
- LLMO-4654
- LLMO-4655

### DIFF
--- a/src/cdn-analysis/handler.js
+++ b/src/cdn-analysis/handler.js
@@ -212,7 +212,11 @@ async function triggerSubAudits(context, site, detectedDays) {
     await sqs.sendMessage(auditQueue, {
       type: 'cdn-logs-report',
       siteId,
-      auditContext: { weekOffset },
+      auditContext: {
+        weekOffset,
+        refreshAgenticDailyExport: true,
+        triggeredBy: SERVICE_PROVIDER_TYPES.BYOCDN_OTHER,
+      },
     }, null, 900);
     log.info(`Triggered cdn-logs-report for siteId=${siteId} weekOffset=${weekOffset}`);
   }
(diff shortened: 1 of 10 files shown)


src/cdn-analysis/handler.js
```diff
@@ -212,7 +212,11 @@ async function triggerSubAudits(context, site, detectedDays) {
     await sqs.sendMessage(auditQueue, {
       type: 'cdn-logs-report',
       siteId,
-      auditContext: { weekOffset },
+      auditContext: {
+        weekOffset,
+        refreshAgenticDailyExport: true,
+        triggeredBy: SERVICE_PROVIDER_TYPES.BYOCDN_OTHER,
+      },
     }, null, 900);
```
ctinp, line 220: Not a blocket but keepi in mind that the refreshed weekly report is queued a fixed 900 seconds after scheduling the per-day cdn-logs-analysis sub-audits, but there is no dependency check that those reprocessing jobs have actually finished. If the analysis fanout runs longer than 15 minutes, the weekly report and the newly added Agentic refresh flow can still execute against incomplete aggregates, producing another stale refresh.

ctinp (approved):

src/cdn-analysis/handler.js
akshaymagapu: yeah, will think in a followup

PR #2450, opened 2026-04-30
feat(llmo-referral-traffic-daily): add daily LLM referral traffic audit

ekremney:
## Summary

- Adds a new `llmo-referral-traffic-daily` two-step `StepAudit`
- Step 1 (`trigger-traffic-analysis-daily-import`) sends a message to import-worker to fetch one day of RUM data; defaults to yesterday UTC, overridable via `auditContext.date` (YYYY-MM-DD)
- Step 2 (`run-referral-traffic-daily`) reads the stored parquet from S3 via `hyparquet`, transforms it into a 12-column CSV matching the optel column contract, uploads to `rum-metrics-compact/llmo-daily-csvs/`, and dispatches a `batch.completed` analytics SQS event for projector ingestion
- 100% line/branch/statement coverage

## Test plan

- [ ] `npm run test:spec -- test/audits/llmo-referral-traffic-daily/handler.test.js` — 22 tests, all green, 100% coverage
- [ ] Verify step 1 dispatches to `AUDIT_STEP_DESTINATIONS.IMPORT_WORKER`
- [ ] Verify step 2 reads parquet from `rum-metrics-compact/data-daily/`, writes CSV to `rum-metrics-compact/llmo-daily-csvs/`, and sends correct SQS analytics message with `pipeline_id: referral_traffic_optel`
- [ ] Companion import-worker PR: adobe/spacecat-import-worker feat/traffic-analysis-daily-import

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -78,6 +78,7 @@ import missingAltTextGuidance from './image-alt-text/guidance-missing-alt-text-h
 import readabilityOpportunities from './readability/opportunities/handler.js';
 import unifiedReadabilityGuidance from './readability/shared/unified-guidance-handler.js';
 import llmoReferralTraffic from './llmo-referral-traffic/handler.js';
+import llmoReferralTrafficDaily from './llmo-referral-traffic-daily/handler.js';
 import llmErrorPages from './llm-error-pages/handler.js';
 import llmErrorPagesGuidance from './llm-error-pages/guidance-handler.js';
 import paidTrafficAnalysis from './paid-traffic-analysis/handler.js';
@@ -184,6 +185,7 @@ const HANDLERS = {
   'detect:form-details': detectFormDetails,
   'page-intent': pageIntent,
   'llmo-referral-traffic': llmoReferralTraffic,
+  'llmo-referral-traffic-daily': llmoReferralTrafficDaily,
   'llm-error-pages': llmErrorPages,
   'guidance:llm-error-pages': llmErrorPagesGuidance,
   'optimization-report-callback': optimizationReportCallback,
--- a/src/llmo-referral-traffic-daily/handler.js
+++ b/src/llmo-referral-traffic-daily/handler.js
@@ -10,8 +10,8 @@
  * governing permissions and limitations under the License.
  */
 
-import { GetObjectCommand, PutObjectCommand } from '@aws-sdk/client-s3';
-import { v4 as uuidv4 } from 'uuid';
+import { createHash } from 'crypto';
+import { DeleteObjectCommand, GetObjectCommand, PutObjectCommand } from '@aws-sdk/client-s3';
 import { parquetReadObjects } from 'hyparquet';
 import { Audit } from '@adobe/spacecat-shared-data-access';
 import { AuditBuilder } from '../common/audit-builder.js';
@@ -47,6 +47,12 @@ function consentToBool(raw) {
   return ['hidden', 'suppressed', 'accept'].includes(String(raw).toLowerCase());
 }
 
+function validateDate(date) {
+  if (!/^\d{4}-\d{2}-\d{2}$/.test(date)) {
+    throw new Error(`Invalid date format: ${date}`);
+  }
+}
+
 const CSV_COLUMNS = [
   'traffic_date', 'host', 'url_path', 'trf_platform', 'device', 'region',
   'pageviews', 'consent', 'trf_type', 'trf_channel', 'bounced', 'updated_by',
@@ -54,7 +60,7 @@ const CSV_COLUMNS = [
 
 function escapeCsvValue(value) {
   const normalized = String(value ?? '');
-  if (/["\n,]/.test(normalized)) {
+  if (/["\r\n,]/.test(normalized)) {
     return `"${normalized.replace(/"/g, '""')}"`;
   }
   return normalized;
@@ -77,10 +83,12 @@ function buildCsvRows(records, host) {
       const device = row.device || '';
       const region = extractCountryCode(urlPath);
       const consentBool = consentToBool(row.consent);
-      const bounced = 1 - Number(row.engaged || 0);
+      const bounced = row.engaged > 0 ? 0 : 1;
       const pageviews = Number(row.pageviews || 0);
 
-      const key = `${trafficDate}|${host}|${urlPath}|${trfPlatform}|${device}|${region}|${consentBool}|${bounced}`;
+      const key = JSON.stringify([
+        trafficDate, host, urlPath, trfPlatform, device, region, consentBool, bounced,
+      ]);
 
       if (grouped.has(key)) {
         grouped.get(key).pageviews += pageviews;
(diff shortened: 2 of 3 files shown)


akshaymagapu (changes requested):
Hey @ekremney,

Solid two-step StepAudit design - decoupling the RUM fetch trigger from the CSV export avoids Lambda timeouts and matches the established `cdn-logs-report` pattern well. The country pattern compilation at module load time (using a local `pat` variable instead of mutating the destructured param like the weekly handler does) is actually an improvement over the existing code. Test coverage is thorough: grouping/aggregation, consent mapping, region derivation, padded S3 paths, CSV escaping, and SQS dedup-ID wiring are all covered with `esmock` in the correct ESM-aware way. The 12-column CSV layout matches the `optel` registry entry in the per-source schema exactly - `consent`, `trf_type`, `trf_channel`, and `bounced` are top-level columns on `referral_traffic_optel`, not JSONB. Setting `allowCache: false` hardcoded in Step 1 is also the right call to prevent a stale audit result from anchoring Step 2 to the wrong Parquet.

A few things need fixing before this can land:

## Critical (Must Fix)

### 1. `s3_uri` is a file path and the filename is wrong - the RPC will reject both

`s3_import_parse_prefix` (used by `wrpc_import_referral_traffic`) explicitly raises an exception when `p_s3_uri` ends in `.csv`:

```sql
IF v_match[2] ~ '\.(csv|csv\.gz)$' THEN
    RAISE EXCEPTION 'p_s3_uri must point to an S3 bundle prefix, not a file: %', p_s3_uri;
END IF;
```

The current code passes `s3://bucket/.../data.csv` as `s3_uri`. Two fixes needed:

- Rename the uploaded file from `data.csv` to `referral_traffic.csv` - the RPC looks for `ARRAY['referral_traffic.csv.gz', 'referral_traffic.csv']` inside the prefix.
- Pass the trailing-slash directory prefix as `s3_uri`, not the file path - e.g. `s3://bucket/rum-metrics-compact/llmo-daily-csvs/siteid={id}/year={y}/month={m}/day={d}/`.

This error is not in the projector's `classifyError` list, so it retries indefinitely until the SQS visibility timeout expires - silent failure from the audit side.

### 2. No S3 cleanup on SQS dispatch failure

If `sqs.sendMessage` throws after `PutObjectCommand` succeeds, a stale CSV sits in S3 with no retry path from the state machine, and `audit.result` records success. The agentic export pattern wraps upload + dispatch in a try/catch that deletes the S3 object on failure. Same protection needed here.

---

## Important (Should Fix)

### 3. `getQueues()` returning `null` crashes with TypeError

`configuration?.getQueues?.().analytics` is not null-safe when `getQueues()` returns `null`. Fix: `configuration?.getQueues?.()?.analytics || ''`.

### 4. Pipe `|` delimiter in grouping key creates silent collision

The composite key `` `${trafficDate}|${host}|${urlPath}|...` `` collides when any field contains a literal `|`. `url_path` comes from RUM data and is user-controlled - `/foo|bar` with platform `baz` hashes identically to `/foo` with platform `bar|baz`, silently aggregating two distinct rows into one. Fix: use `JSON.stringify([trafficDate, host, urlPath, trfPlatform, device, region, consentBool, bounced])` as the key.

### 5. Bare `\r` not escaped in `escapeCsvValue`

RFC 4180 treats CR as a record terminator. The regex `/["\n,]/` misses `\r`. Fix: `/["\r\n,]/`.

### 6. Non-idempotent SQS deduplication ID

`batchId = uuidv4()` is fresh each invocation - a Lambda retry sends a new dedup ID, defeating FIFO deduplication entirely and causing double-ingestion if the optel pipeline is not idempotent. Use a deterministic ID derived from `siteId + date + pipeline_id` (e.g. a SHA-256 hash truncated to 128 chars).

### 7. No validation on `auditContext.date`

A malformed date like `"2026-13-99"` splits to `['2026', '13', '99']`, sets `month=13, day=99`. The Parquet key misses (NoSuchKey → `rowCount: 0` returned silently), but on the write path `getCsvS3Key` also creates garbage S3 keys if a parquet happened to exist. Add: `if (!/^\d{4}-\d{2}-\d{2}$/.test(date)) throw new Error('Invalid date format: ' + date)` at the top of both step functions.

### 8. `bounced` is a float in the grouping key

`bounced = 1 - Number(row.engaged || 0)` produces a float when `engaged` is a fractional rate (e.g. `0.78`). Rows with `engaged=0.78` and `engaged=0.79` get different grouping keys and never aggregate, inflating row counts. Bin on a boolean threshold: `const bounced = row.engaged > 0 ? 0 : 1`.

---

Note: this is a companion PR with adobe/spacecat-import-worker#731 - they should be reviewed and merged together.

ekremney:
Thanks for the thorough review! Here's what was done for each item:

**Critical 1 — filename + `s3_uri` format: reverted**

The reviewer referenced `s3_import_parse_prefix` raising an exception on `.csv` paths, but that function is not called by `wrpc_import_referral_traffic`. Reading the actual RPC:

```sql
v_match := regexp_match(p_s3_uri, '^s3://([^/]+)/(.+)$');
v_s3_key := v_match[2];
-- ... passed directly to:
PERFORM aws_s3.table_import_from_s3('pg_temp._referral_staging', ...,
    aws_commons.create_s3_uri(v_bucket, v_s3_key, 'us-east-1'));
```

`p_s3_uri` is parsed as a direct file URI and `v_s3_key` goes straight to `aws_s3.table_import_from_s3` — no prefix parsing, no filename restriction. `data.csv` works and is already used in existing migrated data, so the filename and URI format are kept as originally written.

**Critical 2 — S3 cleanup on SQS failure: ✅ fixed**
`DeleteObjectCommand` cleanup added in a try/catch around `sqs.sendMessage`.

**Important 3 — `getQueues()` null-safety: ✅ fixed**
`getQueues?.()?.analytics` now safe when `getQueues()` returns `null`.

**Important 4 — pipe `|` collision in grouping key: ✅ fixed**
Replaced with `JSON.stringify([trafficDate, host, urlPath, ...])`.

**Important 5 — `\r` not escaped: ✅ fixed**
Regex updated to `/["\r\n,]/`.

**Important 6 — non-idempotent dedup ID: ✅ fixed**
Replaced `uuidv4()` with `createHash('sha256').update(\`${siteId}:${date}:referral_traffic_optel\`).digest('hex')`.

**Important 7 — date format validation: ✅ fixed**
`validateDate()` added at the top of both step functions.

**Important 8 — `bounced` float: ✅ fixed**
Changed to `row.engaged > 0 ? 0 : 1`.

**M1 — `bytes.buffer` slicing: ✅ fixed**
Now `bytes.buffer.slice(bytes.byteOffset, bytes.byteOffset + bytes.byteLength)`.

**M2 — duplicated parquet S3 path: ✅ fixed**
`parquetKey` is computed once in step 1 and forwarded through `auditResult` to step 2, which reads it directly instead of rebuilding the string.

akshaymagapu (approved):
Re-reviewed after the fix commits. All previous findings are resolved or correctly justified:

**Critical 1 (filename + s3_uri format) — accepted as reverted.** I had this wrong. The new RPC in mysticat-data-service#483 parses `p_s3_uri` with `regexp_match` and passes the key directly to `aws_s3.table_import_from_s3` — no prefix convention, no filename restriction. `data.csv` + full file URI is the correct shape for the new RPC contract, and matches the existing migrated data. Apologies for the noise here.

**Critical 2 (S3 cleanup on SQS failure) — fixed.** `DeleteObjectCommand` in the catch around `sqs.sendMessage`.

**Important 3 (`getQueues()` null-safety) — fixed.** `configuration?.getQueues?.()?.analytics || ''`.

**Important 4 (pipe-delimiter collision) — fixed.** Composite key now uses `JSON.stringify([...])`.

**Important 5 (`\r` not escaped) — fixed.** Regex updated to `/["\r\n,]/`.

**Important 6 (non-idempotent dedup ID) — fixed.** Replaced `uuidv4()` with `createHash('sha256').update(\`${siteId}:${date}:referral_traffic_optel\`).digest('hex')`.

**Important 7 (date validation) — fixed.** `validateDate()` now called at the top of both step functions.

**Important 8 (bounced float collision) — fixed.** Boolean threshold `row.engaged > 0 ? 0 : 1`.

LGTM, ready to merge with the companion import-worker#731 and data-service#483.

Issue #3, opened 2023-10-09
Migrate franklin-site-status/audit-worker to Helix Core


ekremney:
This task is about migrating [franklin-site-status/audit-worker](https://github.com/adobe-rnd/franklin-site-status/tree/main/audit-worker) to Helix Core as [spacecat-audit-worker](https://github.com/adobe-rnd/spacecat-audit-worker)

- `spacecat-audit-worker` should consume audit tasks from SQS in following format, and then query the `site` collection for all the required data for auditing a website:

```
{ 
  "domain": "{domain}"
  "path": "{path}"
}
```

- business logic in `franklin-site-status/audit-worker` should not change. All the audit steps performed should be transferred to `spacecat-audit-worker` as is
- summary of the PSI check (category scores) response is saved in the audits collection (in dynamoDB) under `audit_results` field
- currently audit-worker follows redirects, then run the PSI check on the final url. The final url should be saved under `final_url` field.
- `error` field should be empty if audit runs successfully. It contain an error message if audit results in error.
- last content publication date should be fetched from the helix admin using the final url, and stored under `content_publication_date` field.
- git hashes in the last 24h should be stored under `git_hashes` field
- tag manager and tag manager version (see the LHS result) should be calculated for known tag managers (ie `gtm@v123`) and stored under `tag_manager` field. 
- only a portion (summary) of the PSI check response is saved in the audits collection (in dynamoDB) under `audit_results` field

![Screen Shot 2023-10-17 at 16 18 35](https://github.com/adobe-rnd/spacecat-audit-worker/assets/1171225/b1a95db6-8558-4285-bd22-87f8335793b4)

- `audit_results` object should also be stored under `last_audit_results_psi` of `site` collection
- raw json response from pagespeed should be first sanitized then stored in S3 for both mobile and desktop strategies

- all data models can be found in [lucidchart](https://lucid.app/lucidchart/f5480d07-626d-42e8-8eb5-39b0e9a1c75f/edit?viewport_loc=-329%2C181%2C2339%2C1150%2Cd~mCYSBSUKf.)

PR #2452, opened 2026-04-30
feat(cdn-logs-report): add daily referral traffic export to analytics pipeline

ekremney:
## Problem

CDN referral traffic logs were being processed for SharePoint weekly reports but were never forwarded to the downstream analytics pipeline (Aurora via SQS/SNS). This meant LLM-driven referral traffic data from CDN logs was invisible to the analytics systems that power referral traffic dashboards and attribution reporting.

The agentic traffic flow already had a daily export path (`runDailyAgenticExport` → `pipeline_id: 'llmo_traffic_cdn'`). Referral traffic had no equivalent.

## Solution

Implements a `runDailyReferralExport` function that mirrors the agentic daily export pattern:

```
Athena query → JS-side classifyTrafficSource filter → 15-column CSV → S3 CDN aggregates bucket → SQS dispatch
```

### Architecture

**New file: `src/cdn-logs-report/referral-daily-export.js`**
- Queries Athena for the previous UTC day's referral data using a new single-day partition SQL query
- Uses `classifyTrafficSource` (from `@adobe/spacecat-shared-rum-api-client`) JS-side to classify each row — only `{ type: 'earned', category: 'llm' }` rows are kept
- Groups rows by `[date, host, urlPath, vendor, device, region]`, summing `pageviews`
- Serializes to a 15-column CSV and uploads to the CDN aggregates S3 bucket at:
  `aggregated-referral-llmo-daily-csvs/<siteId>/<Y>/<M>/<D>/data.csv`
- Dispatches a `batch.completed` SQS message with `pipeline_id: 'referral_traffic_cdn'`
- Uses SHA-256 deterministic dedup ID (`sha256(siteId:trafficDate:referral_traffic_cdn)`) for idempotent SQS delivery
- Cleans up the S3 CSV on upload or SQS failure to avoid orphaned files

**15-column CSV contract:**
```
traffic_date, host, url_path, trf_platform, device, region,
pageviews, referrer, utm_source, utm_medium, tracking_param,
trf_type, trf_channel, cdn_provider, updated_by
```
- `trf_platform` = LLM vendor (e.g. `ChatGPT`, `Perplexity`)
- `trf_type` = `earned`, `trf_channel` = `llm` (constants for this export)
- `referrer`, `utm_*`, `tracking_param`, `cdn_provider` = empty string (cleared — classification already done)
- `updated_by` = `spacecat:cdn`

**New SQL: `src/cdn-logs-report/sql/referral-traffic-daily-report.sql`**
- Adds `host` to the SELECT/GROUP BY of the existing referral report SQL pattern
- Uses a single-day partition filter `(year = 'Y' AND month = 'MM' AND day = 'DD')` instead of the weekly date range
- Retains the same LLM referrer/UTM CTEs as the weekly report

**Modified: `src/cdn-logs-report/utils/query-builder.js`**
- Adds `createReferralDailyReportQuery({ trafficDate, databaseName, tableName, site })` using `buildWhereClauseReferral` with a partition filter clause
- Exported via `weeklyBreakdownQueries`

**Modified: `src/cdn-logs-report/handler.js`**
- Wires `runDailyReferralExport` alongside the existing `runDailyAgenticExport` block
- Both exports are guarded by `!isWeeklyOnlyRun` (skipped when `auditContext.weekOffset` is provided)
- Referral export uses `referralReportConfig` (the `referral` entry from `getConfigs`)
- Errors are caught non-fatally and included in the return payload as `{ success: false, error }`
- `dailyReferralExport` is included in the handler return object

### Key differences from agentic export

| | Agentic | Referral |
|---|---|---|
| `pipeline_id` | `llmo_traffic_cdn` | `referral_traffic_cdn` |
| S3 key prefix | `aggregated-agentic-llmo-daily-csvs/` | `aggregated-referral-llmo-daily-csvs/` |
| Dedup ID | UUID (random) | SHA-256 deterministic |
| Classification | JS-side agentic filter | `classifyTrafficSource` → earned/llm |
| `host` in output | No | Yes |

## Test Coverage

**100% line, branch, and statement coverage** across all modified/new source files.

### New test file: `test/audits/cdn-logs-report/referral-daily-export.test.js` (16 tests)
- Happy path: verifies S3 key, CSV headers/body, SHA-256 dedup ID, SQS message shape
- Omits `org_id` when `getOrganizationId` not set; includes it when set
- Skips gracefully (returns `skipped: true`) when Athena returns no rows
- Aggregates pageviews for rows with the same grouping key
- Falls back to site hostname when `row.host` is missing
- Strips query strings from paths
- Uses `row.date` when present; falls back to `trafficDate`
- Throws `'analytics queue is not configured'` when queue URL is missing
- Cleans up S3 CSV and re-throws on Athena failure
- Cleans up S3 CSV and re-throws on SQS failure
- Warns (does not throw) when S3 cleanup itself fails
- Handles null/undefined optional row fields via `|| ''` fallbacks
- Escapes special characters in CSV values (quotes, newlines, commas)
- Handles null `vendor` (maps to empty `trf_platform`)
- Handles path without leading `/`

### Modified: `test/audits/cdn-logs-report/handler.test.js` (4 new tests)
- Skips daily referral export when referral report config is missing
- Includes successful `dailyReferralExport` in handler return value
- Catches referral export failure non-fatally (logs error, sets `success: false`)
- Passes `referenceDate` from `auditContext.date` when provided

### Modified: `test/audits/cdn-logs-report/query-builder-referral.test.js` (2 new tests)
- Verifies single-day partition filter `(year = 'Y' AND month = 'MM' AND day = 'DD')`
- Verifies site-level host exclusion filters are applied

## Test Plan

- [ ] Run `npm test` — all 178+ tests pass, 100% coverage enforced
- [ ] Verify `npm run lint` passes with no errors
- [ ] In staging: trigger a `cdn-logs-report` audit with `auditContext.date` set and confirm:
  - A CSV appears at `s3://<cdn-aggregates-bucket>/aggregated-referral-llmo-daily-csvs/<siteId>/...`
  - An SQS message is dispatched with `pipeline_id: 'referral_traffic_cdn'`
  - The analytics pipeline ingests the CSV and rows appear in Aurora

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/src/cdn-logs-report/handler.js
+++ b/src/cdn-logs-report/handler.js
@@ -32,6 +32,9 @@ import { generatePatternsWorkbook } from './patterns/patterns-uploader.js';
 import {
   runDailyAgenticExport,
 } from './agentic-daily-export.js';
+import {
+  runDailyReferralExport,
+} from './referral-daily-export.js';
 
 async function runCdnLogsReport(url, context, site, auditContext) {
   const { log } = context;
@@ -58,6 +61,7 @@ async function runCdnLogsReport(url, context, site, auditContext) {
   const siteId = site.getId();
   const reportConfigs = getConfigs(s3Config.bucket, s3Config.siteKey, siteId);
   const agenticReportConfig = reportConfigs.find((config) => config.name === 'agentic');
+  const referralReportConfig = reportConfigs.find((config) => config.name === 'referral');
 
   const results = [];
   const reportsToPublish = [];
@@ -155,6 +159,7 @@ async function runCdnLogsReport(url, context, site, auditContext) {
   }
 
   let dailyAgenticExport;
+  let dailyReferralExport;
   if (!isWeeklyOnlyRun) {
     if (!agenticReportConfig) {
       log.debug(`Skipping daily agentic export for ${siteId}: agentic report config not found`);
@@ -179,6 +184,30 @@ async function runCdnLogsReport(url, context, site, auditContext) {
         };
       }
     }
+
+    if (!referralReportConfig) {
+      log.debug(`Skipping daily referral export for ${siteId}: referral report config not found`);
+    } else {
+      try {
+        dailyReferralExport = await runDailyReferralExport({
+          athenaClient,
+          s3Client,
+          s3Config,
+          site,
+          context,
+          reportConfig: referralReportConfig,
+          ...(auditContext?.date ? { referenceDate: new Date(auditContext.date) } : {}),
+        });
+      } catch (error) {
+        context.log.error(`Failed daily referral export for site ${siteId}: ${error.message}`, error);
+        dailyReferralExport = {
+          enabled: true,
+          success: false,
+          siteId,
+          error: error.message,
+        };
+      }
+    }
   }
 
   // Batch publish all uploaded reports using bulk API
@@ -193,6 +222,7 @@ async function runCdnLogsReport(url, context, site, auditContext) {
   return {
     auditResult: results,
     dailyAgenticExport,
+    dailyReferralExport,
     fullAuditRef: `${site.getConfig()?.getLlmoDataFolder()}`,
   };
 }
--- a/src/cdn-logs-report/referral-daily-export.js
+++ b/src/cdn-logs-report/referral-daily-export.js
@@ -14,6 +14,7 @@
 import { createHash } from 'crypto';
 import { DeleteObjectCommand, PutObjectCommand } from '@aws-sdk/client-s3';
 import { classifyTrafficSource } from '@adobe/spacecat-shared-rum-api-client/src/common/traffic.js';
+import { joinBaseAndPath } from '../utils/url-utils.js';
 import { loadSql } from './utils/report-utils.js';
 import { weeklyBreakdownQueries } from './utils/query-builder.js';
 
@@ -71,12 +72,12 @@ export function mapToReferralCsvRows(rawRows, site, trafficDate) {
       referrer, utm_source, utm_medium, tracking_param,
     } = row;
     const device = row.device || '';
-    const date = row.date || '';
+    const normalizedDate = (row.date || '') || trafficDate;
     const region = row.region || 'GLOBAL';
     const rowPageviews = row.pageviews;
 
     const urlPath = rawPath.split('?')[0];
-    const url = `${baseURL}${urlPath.startsWith('/') ? urlPath : `/${urlPath}`}`;
+    const url = joinBaseAndPath(baseURL, urlPath || '/');
 
     const { type, category, vendor } = classifyTrafficSource(
       url,
@@ -87,17 +88,20 @@ export function mapToReferralCsvRows(rawRows, site, trafficDate) {
     );
 
     if (type === 'earned' && category === 'llm') {
-      const key = JSON.stringify([date, effectiveHost, urlPath, vendor, device, region]);
+      const normalizedVendor = vendor || '';
+      const key = JSON.stringify(
+        [normalizedDate, effectiveHost, urlPath, normalizedVendor, device, region],
+      );
       const pageviews = Number(rowPageviews) || 0;
 
       if (grouped.has(key)) {
         grouped.get(key).pageviews += pageviews;
       } else {
         grouped.set(key, {
-          traffic_date: date || trafficDate,
+          traffic_date: normalizedDate,
           host: effectiveHost,
           url_path: urlPath,
-          trf_platform: vendor || '',
+          trf_platform: normalizedVendor,
           device,
           region,
           pageviews,
@@ -175,7 +179,7 @@ export async function runDailyReferralExport({
   const rows = mapToReferralCsvRows(rawRows, site, trafficDate);
 
   if (rows.length === 0) {
-    log.info(`[cdn-logs-report] No referral daily export rows for ${siteId} on ${trafficDate}`);
+    log.info(`[cdn-logs-report] No LLM referral rows for ${siteId} on ${trafficDate} (Athena returned ${rawRows.length} rows, 0 matched classification)`);
     return {
       enabled: true,
       success: true,
--- a/src/cdn-logs-report/referral-daily-export.js
+++ b/src/cdn-logs-report/referral-daily-export.js
@@ -1,0 +1,242 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/* eslint-disable camelcase */
+import { createHash } from 'crypto';
+import { DeleteObjectCommand, PutObjectCommand } from '@aws-sdk/client-s3';
+import { classifyTrafficSource } from '@adobe/spacecat-shared-rum-api-client/src/common/traffic.js';
+import { loadSql } from './utils/report-utils.js';
+import { weeklyBreakdownQueries } from './utils/query-builder.js';
+
+const CDN_REFERRAL_CSV_COLUMNS = [
+  'traffic_date', 'host', 'url_path', 'trf_platform', 'device', 'region',
+  'pageviews', 'referrer', 'utm_source', 'utm_medium', 'tracking_param',
+  'trf_type', 'trf_channel', 'cdn_provider', 'updated_by',
+];
+
+function escapeCsvValue(value) {
+  if (value === null || value === undefined) {
+    return '';
+  }
+  const normalized = String(value);
+  if (/["\r\n,]/.test(normalized)) {
+    return `"${normalized.replace(/"/g, '""')}"`;
+  }
+  return normalized;
+}
+
+function serializeCsv(rows) {
+  const header = CDN_REFERRAL_CSV_COLUMNS.join(',');
+  const body = rows.map(
+    (row) => CDN_REFERRAL_CSV_COLUMNS.map((col) => escapeCsvValue(row[col])).join(','),
+  );
+  return [header, ...body].join('\r\n');
+}
+
+export function getPreviousUtcDate(referenceDate = new Date()) {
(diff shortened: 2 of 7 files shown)


akshaymagapu (changes requested):
Hey @ekremney,

## Strengths

- Test suite is genuinely thorough - 16 cases covering the happy path, zero-row skip, SQS/S3 failure rollback, aggregation, hostname fallback, query-string stripping, and null/missing field handling. The 100% coverage claim is credible.
- Handler integration is clean and non-fatal. The skip-on-missing-config + try/catch + structured failure result mirrors the agentic export pattern exactly.
- Deterministic SHA-256 dedup ID is the right choice for this pipeline. Content-derived dedup prevents double-ingestion of the same day's data on retries, unlike the agentic export's random UUID.
- S3/SQS atomicity is handled correctly. The cleanup path is safe because DeleteObjectCommand on a non-existent key is a S3 no-op.
- escapeCsvValue in the new file is actually more correct than the agentic version - it handles bare \r via /["\r\n,]/ where the agentic version's /["\n,]/ would miss it.

## Issues

### Critical (Must Fix)

**1. Grouping key uses raw date but output normalizes it - can produce duplicate CSV rows**

referral-daily-export.js, lines 73/89/96:

```js
const date = row.date || '';
const key = JSON.stringify([date, effectiveHost, urlPath, vendor, device, region]);
// ...
grouped.set(key, { traffic_date: date || trafficDate, ... });
```

If Athena returns rows for the same logical group where some have date='2026-03-31' and others have date='' (CDN anomaly, partition issue), they land in two separate map entries but both emit traffic_date='2026-03-31' in the CSV. The downstream analytics pipeline receives two rows for what should be one group, silently doubling the pageview count.

No test covers this mixed-date scenario - the existing "falls back to trafficDate" test only covers the all-empty case.

Fix:
```js
const normalizedDate = date || trafficDate;
const key = JSON.stringify([normalizedDate, effectiveHost, urlPath, vendor, device, region]);
grouped.set(key, { traffic_date: normalizedDate, ... });
```

---

**2. vendor=undefined vs vendor='' produces separate groups for the same LLM platform**

referral-daily-export.js, lines 80/89/99:

```js
const { type, category, vendor } = classifyTrafficSource(...);
const key = JSON.stringify([date, effectiveHost, urlPath, vendor, device, region]);
// ...
trf_platform: vendor || '',   // output normalizes to ''
```

JSON.stringify on a key containing undefined serializes it as null, while '' serializes as an empty string. If classifyTrafficSource is not fully consistent in returning undefined vs '' for unknown vendors, pageviews for the same URL/date/device split into two groups but both emit trf_platform: '' - an invisible duplicate. No test covers mixed undefined/'' vendor values for the same page.

Fix:
```js
const normalizedVendor = vendor || '';
const key = JSON.stringify([normalizedDate, effectiveHost, urlPath, normalizedVendor, device, region]);
```

---

### Important (Should Fix)

**3. SQL and JS are two independent sources of truth for LLM classification**

referral-traffic-daily-report.sql lines 21-35 / referral-daily-export.js line 80:

The SQL hard-codes a regex list of 10 LLM referrers/UTM sources. classifyTrafficSource in @adobe/spacecat-shared-rum-api-client maintains its own equivalent list. These will drift independently.

The dangerous direction: a new LLM added to classifyTrafficSource but not to the SQL means those rows never arrive from Athena - silently dropped, no error, no log. The analytics team interprets coverage drops as signal, not as a classification sync lag.

Consider making the SQL a broad pre-filter only (referrer IS NOT NULL) and letting classifyTrafficSource be the single authoritative classifier. If Athena cost requires SQL pre-filtering, maintain one shared list that both SQL template and JS import rather than two manually-synced copies.

---

**4. baseURL trailing slash produces a malformed URL passed to classifyTrafficSource**

referral-daily-export.js, line 78:

```js
const url = `${baseURL}${urlPath.startsWith('/') ? urlPath : `/${urlPath}`}`;
```

If site.getBaseURL() returns 'https://example.com/' (trailing slash), this produces 'https://example.com//page'. This URL is passed to classifyTrafficSource and could cause mis-classification depending on the classifier's URL parsing. The agentic mapper avoids this with joinBaseAndPath - the referral mapper should do the same. No test covers a site with a trailing-slash baseURL.

---

**5. "Zero rows" log is ambiguous between two distinct failure modes**

referral-daily-export.js, line 178:

When rows.length === 0 after JS filtering, the function logs "No referral daily export rows" and returns { skipped: true }. This conflates two situations:
- Athena returned zero rows: no LLM referral traffic for that day (normal for low-traffic sites).
- Athena returned rows but JS classified all of them out: the SQL regex and classifyTrafficSource lists have drifted (abnormal - indicates a data quality gap).

Add a log of the raw Athena row count before filtering, e.g. "Athena returned N rows; 0 passed LLM classification filter". This makes the two cases distinguishable in production without guessing.

---

**6. Missing test: referral export not called when isWeeklyOnlyRun is true**

handler.js line 163 / handler.test.js:

The agentic export test suite has a "skips daily export when auditContext.weekOffset is provided" test. The four new handler tests cover daily-run scenarios but none assert that passing weekOffset in auditContext causes runDailyReferralExport to be skipped. This is a gap directly analogous to the existing agentic gate test.

## Assessment

Ready to merge? No - with fixes required.

Critical issues 1 and 2 are data correctness bugs that will silently double-count pageviews for certain Athena result shapes (mixed empty/non-empty date fields, or inconsistent vendor values). Both are one-liner fixes. The SQL/JS dual-classification issue (Important 3) is a correctness risk that will manifest silently at the next LLM addition and is cheapest to address while this context is fresh. The rest can be addressed as follow-ups.

## Next Steps

1. Fix the two grouping key normalization bugs (Critical 1 and 2).
2. Address the SQL/JS single-source-of-truth for LLM classification (Important 3).
3. Fix baseURL trailing slash via joinBaseAndPath (Important 4).
4. Add the missing isWeeklyOnlyRun handler test (Important 6).

ekremney:
@akshaymagapu Thanks for the thorough review! All items have been addressed in the latest commit (`25fce794`).

---

### Critical fixes

**1. Grouping key date normalization**

Fixed — `normalizedDate` is now derived before the key is built:
```js
const normalizedDate = (row.date || '') || trafficDate;
const key = JSON.stringify([normalizedDate, effectiveHost, urlPath, normalizedVendor, device, region]);
grouped.set(key, { traffic_date: normalizedDate, ... });
```
Rows with `date: '2026-03-31'` and `date: ''` for the same page now collapse into a single bucket. Added a test (`normalizes mixed-date rows to the same group key`) that puts two rows with the same page but different date values through the mapper and asserts a single output row with the summed pageviews.

**2. `vendor=undefined` vs `vendor=''` key split**

Fixed — `normalizedVendor = vendor || ''` is computed before the key:
```js
const normalizedVendor = vendor || '';
const key = JSON.stringify([normalizedDate, effectiveHost, urlPath, normalizedVendor, device, region]);
grouped.set(key, { trf_platform: normalizedVendor, ... });
```
Added a test (`normalizes undefined and empty-string vendor to the same group key`) that stubs `classifyTrafficSource` to return `vendor: undefined` on the first call and `vendor: ''` on the second, then asserts a single aggregated row.

---

### Important fixes

**3. SQL/JS dual classification**

Addressed — the `llm_referrals` CTE (with its hard-coded LLM regex list) has been removed from `referral-traffic-daily-report.sql`. The daily SQL is now a broad pre-filter only:
```sql
FROM {{databaseName}}.{{tableName}}
{{whereClause}}
  AND (referrer IS NOT NULL OR utm_source IS NOT NULL)
```
`classifyTrafficSource` is now the single authoritative LLM classifier. The weekly `referral-traffic-report.sql` (which has no JS-side classification step) keeps its own regex unchanged.

**4. Trailing-slash baseURL**

Fixed — replaced the manual template literal with `joinBaseAndPath`:
```js
import { joinBaseAndPath } from '../utils/url-utils.js';
// ...
const url = joinBaseAndPath(baseURL, urlPath || '/');
```
Added a test (`handles trailing slash in baseURL without producing a double-slash URL`) with `getBaseURL: () => 'https://www.example.com/'` that asserts `classifyTrafficSource` is called with `https://www.example.com/page` (no double slash).

**5. Ambiguous zero-rows log**

Fixed — the log now includes the raw Athena row count:
```
[cdn-logs-report] No LLM referral rows for <siteId> on <date> (Athena returned N rows, 0 matched classification)
```
Added a test (`includes raw Athena row count in the zero-rows skip log`) that passes one non-LLM Athena row and asserts the log message contains the count. The two failure modes (genuine zero traffic vs. SQL/JS drift) are now distinguishable in production logs.

**6. Missing `isWeeklyOnlyRun` handler test**

Added — `skips daily referral export when auditContext.weekOffset is provided` asserts `runDailyReferralExport` is not called and `result.dailyReferralExport` is `undefined` when `weekOffset` is set in `auditContext`, mirroring the existing agentic gate test.

---

One extra fix bundled in: added `referral-daily-export.js` to the stub map of all `localHandler = await esmock(handler.js, {...})` calls that didn't already include it. Without it, those tests loaded the real `referral-daily-export.js` → `traffic.js` chain and caused a 10s timeout when the test files were run combined.

akshaymagapu:
Thanks for the thorough fixes @ekremney - all 6 items look correct.

One small note on fix #3 (SQL broadening): the new pre-filter `AND (referrer IS NOT NULL OR utm_source IS NOT NULL)` is safe - verified that `classifyTrafficSource` cannot produce `{type:'earned', category:'llm'}` when both fields are null, so no LLM traffic will be missed. However, this filter is significantly broader than the old LLM regex, so it will pull more rows through Athena (all referral/utm traffic, not just LLM patterns) before JS-side classification narrows them back down. Worth keeping an eye on Athena query time and Lambda peak memory on high-traffic sites after this rolls out.

akshaymagapu (approved):

ekremney:
Thanks for validating @akshaymagapu. Good point on the broader scan — acknowledged. The trade-off was intentional: the old CTE pre-filter was narrower but created the drift risk your review flagged (a new LLM added to `classifyTrafficSource` but not to the SQL regex → silent data gap). The broader filter eliminates that class of failure.

That said, you're right that it increases the row volume flowing through to JS. Worth monitoring Athena scan bytes and Lambda memory on the first few high-traffic sites after rollout. If cost or memory becomes a concern, the next step would be maintaining a shared constant (e.g. an exported regex string from `@adobe/spacecat-shared-rum-api-client`) that both the SQL template and `classifyTrafficSource` import — single source of truth with the specificity of the old approach.

ekremney:
Good call — reverted in . The daily SQL now has the same LLM regex CTE as `referral-traffic-report.sql` (with `host` added), so Athena scan volume stays comparable to the weekly report. `classifyTrafficSource` still does the definitive JS-side classification on top of it.

akshaymagapu (approved):

PR #2455, opened 2026-04-30
feat: Broken internal link cross locale

jkf276:
added new functionality to broken internal links.
you can now add a list of element class to ignore in the site.config.handler. Any element (or its ancestor) with the excluded class will be ignored.
you can now ignore pdp links that go across locale.

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/internal-links/scope-utils.js
+++ b/src/internal-links/scope-utils.js
@@ -92,3 +92,53 @@ export function extractLocalePathPrefix(url) {
     return '';
   }
 }
+
+/**
+ * True when RUM pairs a referring page and 404 target whose first path segments are
+ * both locale-like but different (e.g. /fr/... → /de/...). Those rows are often noise
+ * for same-locale broken-link remediation when the site is audited at domain root.
+ * Same number of segments
+ * all segments are identical except for the first one (locale)
+ *
+ * @param {string} urlFrom - Referrer URL from RUM
+ * @param {string} urlTo - 404 target URL from RUM
+ * @returns {boolean}
+ */
+export function isCrossLocalePDP404RumPair(urlFrom, urlTo) {
+  // base case: one/both of the urls are not valid
+  if (!urlFrom || !urlTo) {
+    return false;
+  }
+
+  try {
+    const urlFromParsed = new URL(prependSchema(urlFrom));
+    const urlToParsed = new URL(prependSchema(urlTo));
+    const { pathname: urlFromPathname } = urlFromParsed;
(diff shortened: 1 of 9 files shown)


src/internal-links/scope-utils.js
```diff
@@ -92,3 +92,62 @@ export function extractLocalePathPrefix(url) {
...
+    }
+
+    /*
+    * Functionality was original built for casio.com, which didn't use BCP 47 standard for locales.
+    * locale check disabled for now until since casio is the only site that uses this functionality.
+    * Ideally, casio.com should migrate to BCP 47 standard for locales.
+    if (!isLocaleLikePathSegment(fromSegments[0]) || !isLocaleLikePathSegment(toSegments[0])) {
+      return false;
+    }
+    */
```
noruiz: Can this block be removed?

test/audits/internal-links/config.test.js
```diff
@@ -160,6 +162,27 @@ describe('internal-links config resolver', () => {
...
+    expect(resolver.getExcludedElementClasses()).to.deep.equal(['no-audit', 'editorial']);
+  });
+
+  it('returns filterCrossLocaleRum404 from handler config when set false', () => {
+    const resolver = new InternalLinksConfigResolver(createSite({
+      filterCrossLocaleRum404: false,
+    }), {});
+
+    expect(resolver.getExcludeCrossLocalePDP()).to.equal(false);
+  });
```
noruiz: This test passes but for the wrong reason getExcludeCrossLocalePDP() reads excludeCrossLocalePDP, not filterCrossLocaleRum404. since that key is undefined, it just returns the default false.

src/internal-links/scope-utils.js
jkf276: per the comments, the functionality is currently disabled because casio.com doesn't use standard locale code (EG. https://www.casio.com/latin/). this is something I would really prefer to put back in but the first user doesn't support it.

noruiz: I just think it's not a good practice to leave commented code in the codebase like this. If we really want to re-enable this in the future I think it's better to create a separate ticket with details to add this back when Casio is ready.

jkf276: I removed the commented out code

test/audits/internal-links/config.test.js
jkf276: test corrected.

noruiz (approved):

PR #2456, opened 2026-04-30
feat: Remove 3rd party services from this process

jsanchez07:
[SITES-4379](https://jira.corp.adobe.com/browse/SITES-4379)

There was a concern reported on slack here about CDN detection using 3rd party dependencies and we shouldn't do this. We should remove these 3rd party services for this. The calls were only happening as a last-resort fallback (after headers, system DNS, and DoH CNAME all fail to identify a CDN), so in real world instances, we're not hitting it much, but the call is to remove this and just set the CDN to "unknown" if the other methods don't produce a CDN type.


mikeweis (approved):
LGTM 👍

PR #2461, opened 2026-05-01
feat: migrating llm-error-pages audit to spacecat DB

ymadobe:
## PR Description

This PR introduces a parallel DB-based Opportunity and Suggestion sync alongside the existing Excel/SharePoint flow. Both write paths operate independently on a best-effort basis—failures in one are logged and do not block the other.

Excel/SharePoint remains the canonical source for the current UI, while the DB path enables the upcoming UI migration.

### What’s Included

* **Parallel DB Sync**

  * Adds DB writes for Opportunities and Suggestions without impacting the existing Excel flow.
  * Ensures isolation between Excel and DB failures.

* **Data Modeling**

  * One Opportunity per status code bucket (404, 403, 5xx).
  * One Suggestion per unique URL via `syncSuggestions`.

* **Enhanced URL-Level Data**

  * Captures `hitCount`, `agentTypes`, `userAgents` (normalized), `avgTtfb`, `countryCode`, `product`, `category`, and `periodIdentifier`.

* **Retention Logic**

  * Implements a 4-week retention policy.
  * URLs not seen within this window are marked as `OUTDATED`, even if no new errors occur.

* **Mystique Integration**

  * Includes real DB `opportunityId` in outgoing messages.
  * Uses it to fetch Opportunities, map Suggestions by URL path, and batch update via `Suggestion.saveMany()`.

* **Resilience Improvements**

  * Excel failures no longer abort execution; DB writes still proceed.
  * Graceful handling for missing/invalid `opportunityId` (logs warning, skips DB write).

* **Backward Compatibility**

  * No change to Excel/SharePoint output.
  * Excel path will be deprecated after UI migration.


src/llm-error-pages/handler.js
```diff
@@ -280,7 +425,7 @@ export async function runAuditAndSendToMystique(context) {
                   }))
                   .filter((link) => link.urlFrom.length > 0),
                 alternativeUrls,
-                opportunityId: `llm-404-${periodIdentifier}`,
+                opportunityId: opportunityMap[404]?.getId(),
```
anagarwa, line 429: @ymadobe is it possible that opportunityId could be undefined , if opportunityMap[404] does not exists

src/llm-error-pages/handler.js
```diff
@@ -230,14 +246,143 @@ export async function runAuditAndSendToMystique(context) {
...
+        }
 
         log.info(`[LLM-ERROR-PAGES] Found ${processedResults.totalErrors} total errors across ${processedResults.summary.uniqueUrls} unique URLs`);
 
+        const opportunityMap = {};
+        const { Suggestion, Opportunity } = dataAccess;
+        const retentionCutoff = new Date(Date.now() - RETENTION_MS);
+        let existingOpportunities = [];
+        try {
+          existingOpportunities = await Opportunity.allBySiteIdAndStatus(site.getId(), 'NEW');
```
anagarwa, line 272: @ymadobe can you explain the rational behind fetching just the `NEW` ?

src/llm-error-pages/handler.js
ymadobe: This is done, intentionally. `opportunityMap[404]` will be `undefined` in two scenarios:
1. The 404 bucket had no errors AND no stale opportunity existed for this site
2. The DB sync for the 404 bucket threw an exception (caught by the outer try/catch, so `opportunityMap[404]` was never set)

The `?.` operator handles this - `opportunityId` becomes `undefined` in the Mystique message. When Mystique calls back, `guidance-handler.js` already has:
```javascript
if (!opportunityId) {
  log.warn('[LLM-ERROR-PAGES] No opportunityId in Mystique message — skipping DB update');
}
```
so the DB side degrades gracefully while Excel still runs.

src/llm-error-pages/handler.js
ymadobe: We're fetching only NEW because the retention sweep should only run on opportunities that haven't been touched yet. If someone has already moved an opportunity to another status (FIXED, SKIPPED, etc.) didn't want to interfere with that.

But this does mean FIXED/RESOLVED suggestions won't get cleaned up and will accumulate over time, I'll take that up in a follow up PR ( would require a reliable timestamp to track this )

anagarwa (approved):

PR #2470, opened 2026-05-04
feat: add LLMO config schema-sweep script (SITES-43238 step 3a)

dzehnder:
## Summary

Step 3a of [SITES-43238](https://jira.corp.adobe.com/browse/SITES-43238). One-off operational script that sweeps every \`config/llmo/<siteId>/lmmo-config.json\` in S3 and classifies each against the published Zod schema. **READ-ONLY** — never writes.

The output sizes the population of corrupt configs and breaks down the failure modes by issue code, so step 3b (targeted repair) can be planned with real data instead of guesses.

## Why now

Steps 1 and 2 stop new corruption (writer-side filter + fail-closed writeConfig, both merged). They do nothing for configs already corrupted in S3 since 2026-03-02 (PR #2057). The known broken site \`78d59744-e06c-4d14-a77a-9490c1464116\` is one example; the rest of the population is unknown until this script runs.

## What's in this PR

A single new file: \`scripts/llmo-config-sweep.js\`.

- Lists all keys under \`config/llmo/\` with paginated \`ListObjectsV2\` (handles >1000 keys).
- Classifies each via \`schemas.llmoConfig.safeParse(JSON.parse(text))\` from \`@adobe/spacecat-shared-utils\`.
- Per-key result is one of: \`valid\`, \`schema-invalid\` (with Zod issue list), \`json-error\`, \`fetch-error\`.
- Worker pool with configurable \`--concurrency\` (default 20) for parallel fetches.
- Outputs progress every 100 keys, then a summary block, then the top-20 \`path: code\` patterns by frequency. Optional \`--out report.json\` for machine-readable dump of failures.
- Continues past individual errors (one bad config does not abort the sweep).
- Exit code 0 on completion regardless of failure count (failures are the result, not an error). Exit 2 on fatal sweep error (missing args, list failure, etc.).

## Why a script (not an audit handler)

It's a one-off operational task, not a per-site audit. Wrapping it in the audit framework (which dispatches via SQS per-site) would be a forced fit and require an SQS message per site. A standalone script that takes operator AWS creds via \`klam login\` matches the existing precedent in \`scripts/\` (\`watch-and-copy.js\`, \`populate-env.sh\`, \`trigger-ad-intent-audit.sh\`).

\`scripts/\` is in the eslint global ignore and excluded from coverage — fits this kind of tooling cleanly.

## Why \`schemas.llmoConfig.safeParse\` and not \`llmoConfig.readConfig\`

\`readConfig\` was just upgraded in [spacecat-shared PR #1574](https://github.com/adobe/spacecat-shared/pull/1574) to throw \`LlmoConfigValidationError\`. The audit-worker's installed version of \`@adobe/spacecat-shared-utils\` doesn't have that change yet — the dep bump is part of step 2's Phase 3. Using \`schemas.llmoConfig\` (the underlying Zod schema, exported since 2025) decouples this sweep from the dep bump cycle so it can run **today**.

## Test plan

- [x] \`node --check scripts/llmo-config-sweep.js\` passes.
- [x] No \`npm test\` impact (script is in \`scripts/\`, excluded from coverage by \`.nycrc.json\` \`include: src/**/*.js\`).
- [x] \`npm run lint\` is unaffected (\`scripts/*\` in eslint \`globalIgnores\`).
- [ ] Run against a non-prod bucket first (dev/CI) to validate output shape and concurrency behavior before pointing at prod.
- [ ] Once run against prod (separate operational task, not a CI gate), attach the \`--out\` report to SITES-43238 to size step 3b.

## Out of scope

- The actual repair (step 3b targeted, step 3c batch). That comes after we have the sweep results.
- Tests for the script itself. The validation it relies on (\`schemas.llmoConfig.safeParse\`) is covered by the 1038-test suite in \`spacecat-shared-utils\`. The script's own logic is plain S3 plumbing and one-off-tool simple.


scripts/llmo-config-sweep.js
```diff
@@ -0,0 +1,237 @@
...
+    );
+    process.exit(0);
+  }
+  const bucket = values.bucket || process.env.S3_BUCKET_NAME;
+  if (!bucket) {
+    console.error('error: --bucket or S3_BUCKET_NAME env var is required');
+    process.exit(2);
+  }
+  return {
+    bucket,
```
danieljchuser, line 88: `Math.max(1, NaN)` returns `NaN`, not 1. If an operator typos `--concurrency 2O` (letter O instead of zero), `Array.from({ length: NaN })` produces `[]` and the sweep silently completes with zero results and exit 0 - indistinguishable from "bucket is clean." Same concern applies to `--limit` on line 73 (`Number("50k")` is `NaN`, `keys.length >= NaN` is always false, so the limit never trips). Validate both with `Number.isFinite()` and `process.exit(2)` on bad input.

scripts/llmo-config-sweep.js
```diff
@@ -0,0 +1,237 @@
...
+
+  const result = schemas.llmoConfig.safeParse(parsed);
+  if (!result.success) {
+    return { siteId, status: 'schema-invalid', issues: result.error.issues };
+  }
+  return { siteId, status: 'valid' };
+}
+
+async function runWorkers(queue, concurrency, onResult, onProgress) {
+  const workers = Array.from({ length: concurrency }, async () => {
```
danieljchuser, line 155: `await item()` is not wrapped in try/catch. `classifyKey` is defensive today, but if it ever throws unexpectedly (SDK upgrade, schema change, missing `.error.issues` on a malformed safeParse failure), one worker rejects and `Promise.all` drops every result collected so far. The summary is never printed, the `--out` file is never written, and the operator sees "fatal error" after minutes of S3 GETs. The PR states the script "continues past individual errors" - a try/catch here makes that contract hold regardless of what `classifyKey` does:
```js
let r;
try {
  r = await item();
} catch (e) {
  r = { status: 'unexpected-error', error: e?.message ?? String(e) };
}
```

scripts/llmo-config-sweep.js
```diff
@@ -0,0 +1,237 @@
...
+  const { top, totalDistinct } = topIssueCodes(results);
+  if (top.length > 0) {
+    console.log('\n=== Schema issues (path: code => count) ===');
+    for (const [key, count] of top) {
+      console.log(`  ${String(count).padStart(6)}  ${key}`);
+    }
+    if (totalDistinct > top.length) {
+      console.log(`  (and ${totalDistinct - top.length} more distinct issue patterns)`);
+    }
+  }
```
danieljchuser, line 259: If `outPath` is unwritable (missing parent dir, permissions, read-only fs), `writeFile` throws into `main().catch`, which logs `[sweep] fatal error:` and exits 2. By this point the human-readable summary is already on stdout, but the operator sees exit 2 and "fatal error" for what was actually a successful sweep. Localizing the catch keeps the exit code honest:
```js
if (outPath) {
  try {
    const failures = results.filter((r) => r.status !== 'valid');
    await writeFile(outPath, JSON.stringify({ summary, failures }, null, 2));
    console.log(`\n[sweep] wrote ${failures.length} failure entries to ${outPath}`);
  } catch (e) {
    console.error(`[sweep] WARNING: failed to write ${outPath}: ${e.message}`);
  }
}
```

danieljchuser (changes requested):
Hey @dzehnder,

## Strengths

- Clean separation of failure modes into four distinct buckets (`valid`, `schema-invalid`, `json-error`, `fetch-error`) gives steps 3b/3c actionable data for planning targeted repairs (scripts/llmo-config-sweep.js:96-125).
- Read-only by design - only `ListObjectsV2Command` and `GetObjectCommand` are imported. Cannot mutate S3 state.
- Correct architectural call placing this in `scripts/` as a one-off sweep rather than shoehorning it into the per-site audit framework. Matches existing precedent and the trade-off is well-justified in the PR description.
- Decoupling from the in-flight `readConfig` upgrade by binding directly to `schemas.llmoConfig` is the right choice - the schema is the durable contract, and this avoids gating 3a on a dep bump cycle.
- Per-key error isolation in `classifyKey`: one malformed config cannot crash the sweep before the inventory completes.
- Top-N issue path/code aggregation in `topIssueCodes` will be far more useful for sizing the repair than a flat failure list.

## Issues

### Important (Should Fix)

**1. scripts/llmo-config-sweep.js:73-74 - Non-numeric `--concurrency` / `--limit` silently produce a zero-result or unbounded sweep**

`parseArgs` returns strings. `Number("2O")` (letter O) yields `NaN`, and `Math.max(1, NaN)` is still `NaN`. `Array.from({ length: NaN })` produces an empty array, so `runWorkers` returns instantly with zero results. The summary prints `total: 0, valid: 0` with exit 0 - indistinguishable from "bucket is clean." For a tool whose entire purpose is sizing a corruption population, a silent zero-result run that reads as "no problem found" is the worst possible failure mode. Similarly, `--limit 50k` silently scans the full bucket since `NaN >= NaN` is always false. Validate both with `Number.isFinite()` and exit 2 on bad input.

**2. scripts/llmo-config-sweep.js:133 - Unexpected throw in worker drops all collected results**

`await item()` in the worker loop is not wrapped in try/catch. `classifyKey` is defensive today, but if it ever throws unexpectedly, one worker rejects and `Promise.all` drops every result collected so far - the summary is never printed, the `--out` file is never written. The PR states the script "continues past individual errors"; a try/catch here makes that contract hold regardless of future changes to `classifyKey`. See inline comment for suggested fix.

**3. scripts/llmo-config-sweep.js:225 - `writeFile` failure makes a successful sweep report as fatal**

If `--out` targets an unwritable path, `writeFile` throws into `main().catch`, logging `[sweep] fatal error:` and exiting 2. The stdout summary is already printed, but the exit code and message actively mislead the operator about what happened. A localized try/catch keeps the exit code honest. See inline comment for suggested fix.

### Minor (Nice to Have)

**4. scripts/llmo-config-sweep.js:149 - `elapsedSeconds` is a string in the JSON report**

`.toFixed(1)` returns a string. The same `summary` object goes into the `--out` JSON file, where downstream consumers would expect a number. Either format separately for stdout, or wrap: `Number((elapsedMs / 1000).toFixed(1))`.

**5. scripts/llmo-config-sweep.js:96-125 - `--out` failure entries lack the S3 key**

`classifyKey` returns `siteId` but not the full S3 key. When the report drives step 3b, the exact key is what an operator pastes into the AWS console. Deriving it from `siteId` is mechanical but error-prone. Adding `key` to the returned object is a one-line change.

## Recommendations

1. **Capture AWS error class in fetch-error results** (lines 104, 111): `error: e.message` drops the SDK error code (`AccessDenied`, `NoSuchKey`, `NetworkingError`). Adding `errorName: e.name` helps 3b/3c planners distinguish failure classes.
2. **Record sweep boundary in summary**: include `limitApplied` and the limit value so a `--limit 50` smoke-test report is structurally distinguishable from a full sweep.
3. **Lifecycle cleanup**: once steps 3b/3c complete and the corrupt-config population is repaired, plan to delete this script. A note in SITES-43238 closes the loop.
4. **Comment on `KEY_RE` matching the historical `lmmo-config.json` typo** (line 44): prevents a future maintainer from "fixing" the regex and silently breaking the sweep.

## Assessment

Ready to merge? With fixes

Well-scoped, read-only, correctly placed operational tooling with good staged execution (measure before repair). The three Important items are all cheap fixes that directly protect the script's core value as a reliable evidence-gathering tool: the NaN gap could produce a misleading "all clean" report, the missing worker catch could lose minutes of work on prod, and the writeFile exit code could confuse scripted workflows. Everything else is polish.

dzehnder:
Thanks @danieljchuser — sharp catches, especially on the NaN trap. Pushed [6dcea06](https://github.com/adobe/spacecat-audit-worker/commit/6dcea06d) addressing all three Important items plus the two minors and three of the four recommendations.

## Important — applied

1. **NaN guard on `--concurrency` / `--limit`** (lines 73-74). Replaced `Math.max(1, Number(...))` with a `parsePositiveInt` helper that uses `Number.isFinite` + `Number.isInteger` and exits 2 on bad input. Verified locally:
   ```
   $ node scripts/llmo-config-sweep.js --bucket dummy --concurrency 2O
   error: --concurrency must be a positive integer (got "2O")
   $ node scripts/llmo-config-sweep.js --bucket dummy --limit 50k
   error: --limit must be a positive integer (got "50k")
   ```
   Agreed completely on the framing — for a tool whose whole job is sizing a corruption population, a silent zero-result run is the worst possible failure mode.

2. **Worker try/catch** (line 133). Wrapped `await item()` so any unexpected throw now lands as `{ status: 'unexpected-error', error }` and the sweep completes. Added `unexpectedError` to the summary buckets. Inline comment explains this is a defensive backstop against future changes to `classifyKey`.

3. **Localized `writeFile` try/catch** (line 225). An unwritable `--out` path now logs `WARNING: failed to write …` and the script returns success. The exit code reflects what actually happened (sweep succeeded), not what the operator was trying to do at the end.

## Minor — applied

4. `elapsedSeconds` is now `Number((elapsedMs / 1000).toFixed(1))`, not a string. Downstream JSON consumers don't need to coerce.

5. `classifyKey` results now include the full S3 `key` alongside `siteId`. Operators driving step 3b can paste the exact key directly into the AWS console.

## Recommendations — applied

- (1) `errorName: e.name` on fetch-error results so SDK error classes (`AccessDenied`, `NoSuchKey`, `NetworkingError`) are preserved for triage.
- (2) `limitApplied` (boolean) and `limit` (number or null) added to the summary block so a `--limit 50` smoke-test report is structurally distinguishable from a full sweep.
- (4) Comment on `KEY_RE` explaining that `lmmo-config.json` is the intentional on-disk filename (matching `llmoConfigPath` in `@adobe/spacecat-shared-utils`), with a "do not fix" note. Prevents the regex from getting "corrected" away.

## Recommendation — deferred (not a code change)

- (3) Lifecycle cleanup. Agreed — once steps 3b/3c land and the corrupt-config population is repaired, this script should be deleted. I will note that on SITES-43238 as part of the eventual step 3 closing comment so the cleanup task isn't lost.

## State

`node --check` clean. The script is in `scripts/` so it stays out of the lint and coverage paths. Manual smoke tests of the new error paths pass.

Ready for re-review.

danieljchuser (approved):
Hey @dzehnder,

## Strengths

- All three Important findings from the prior round are now addressed cleanly. `parsePositiveInt` (scripts/llmo-config-sweep.js:53-60) combines `Number.isFinite` + `Number.isInteger` + `n < 1` and exits 2 with the offending value quoted via `JSON.stringify`, so `--concurrency 2O` and `--limit 50k` fail loudly instead of silently producing a zero-result run. The inline rationale at lines 81-83 documents why this matters for an evidence-gathering tool, so a future maintainer won't relax the check.
- `runWorkers` (scripts/llmo-config-sweep.js:142-160) wraps `await item()` and surfaces `{ status: 'unexpected-error', error }` on any future throw, plus an `unexpectedError` counter in the summary. The "defensive backstop" comment is the right level of documentation; it explicitly tells a future reader the catch is intentional, not dead code, so it survives a refactor of `classifyKey`.
- `writeFile` is isolated in its own try/catch (scripts/llmo-config-sweep.js:254-262) with a `WARNING:` log. The sweep's exit code now reflects sweep success rather than report-write success, which is the correct separation for operational tooling.
- Both Minor items also addressed: `elapsedSeconds` is `Number((elapsedMs / 1000).toFixed(1))` (line 178), and every failure result carries the full S3 `key` plus `errorName` (lines 109-119), so a step 3b operator can paste the exact key into AWS without reconstructing it from `siteId`.
- Three of the four Recommendations applied: `errorName: e.name` on fetch-error returns, `limitApplied` / `limit` in the summary so smoke-test reports are structurally distinguishable from full sweeps, and the multi-line block comment on `KEY_RE` (lines 47-50) explains the intentional `lmmo-config.json` typo and references `llmoConfigPath` in spacecat-shared-utils so the regex won't get "corrected" away. Recommendation 3 (lifecycle cleanup after steps 3b/3c) is appropriately deferred to a SITES-43238 ticket note rather than tracked in code.
- Architecture remains sound: read-only by imports (`ListObjectsV2Command` and `GetObjectCommand` only, no Put/Delete/Copy surface), `scripts/` placement matches the precedent for one-off operational tooling, and binding to `schemas.llmoConfig.safeParse` correctly decouples step 3a from the `@adobe/spacecat-shared-utils` bump cycle so it can run today against raw on-disk state.
- Error taxonomy is clean and triageable: `valid` / `schema-invalid` / `json-error` / `fetch-error` / `unexpected-error` map directly to the buckets needed to plan repair, and `topIssueCodes` aggregating by `path: code` is the right signal for deciding between a targeted patch (few codes dominate) and a broader migration (long tail).

## Recommendations

1. `unexpected-error` result is missing `key` and `siteId` (scripts/llmo-config-sweep.js:151). Every other failure status carries both, but the backstop catch in `runWorkers` returns only `{ status: 'unexpected-error', error }`. A triager filtering `status !== 'valid'` in the `--out` report will see `unexpected-error` rows with no object identity, which is exactly when you most want the key to reproduce. Easiest fix: have the queue items carry their `key` (e.g., `queue.push({ key: k, run: () => classifyKey(s3, bucket, k) })` and include `key` in the catch result). Bucket is unlikely to fire, but diagnostic value when it does would be high. Take or leave.
2. Out of scope, worth tracking on SITES-43238: align step 3b on `readConfig` once the `@adobe/spacecat-shared-utils` bump carrying `LlmoConfigValidationError` lands, so the repair path uses the same validation surface as the writer-side filter from steps 1 and 2.

## Assessment

Ready to merge? Yes

Reasoning: All three Important findings and both Minor findings from the prior round are addressed correctly, with comments that document intent so the fixes survive future edits. Architecture remains the right call for a one-off measurement pass. The single recommendation on `unexpected-error` payload shape is a nice-to-have for an unlikely-but-high-value path, not a blocker.

Note: the `ci / build` check is currently pending - worth confirming it passes before merge.

PR #2473, opened 2026-05-05
feat(geo-brand-presence): forward v2 brand_id to DRS (LLMO-4716)

rainer-friederich:
## Summary

Both `geo-brand-presence` and `geo-brand-presence-daily` refresh handlers now resolve the v2 brand once per audit (per site) via the new LLMO-4716 spacecat-api-service endpoint, then thread `brandId` through `drsClient.publishBrandPresenceAnalyze` so the DRS Fargate runner enters the v2 config path for brandalf-true orgs.

## Failure contract

A new `src/utils/brand-resolver.js` encapsulates the HTTP call:

| Result | Caller behavior |
| --- | --- |
| `200` | Use the brand id |
| `404` | Return `null` — deliberate "no v2 brand" (v1 orgs / brandalf-migration / kill-switch / no active brand for site). Caller forwards `null` and the runner takes v1 |
| `5xx / network / timeout / parse error` | **Throw**. For brandalf-true orgs, silently continuing without `brand_id` falls back to v1 reads from a no-longer-published spreadsheet mirror — the BP pipeline keeps running but produces meaningless output. Failing fast lets SQS retry and surface the issue via DLQ instead of silent bad data. |

## Files

- `src/utils/brand-resolver.js` (new) — HTTP client to the LLMO-4716 endpoint
- `src/geo-brand-presence/geo-brand-presence-refresh-handler.js` — resolve + forward
- `src/geo-brand-presence-daily/geo-brand-presence-refresh-handler.js` — resolve + forward
- `test/utils/brand-resolver.test.js` (new, 15 cases) + expansions to both refresh-handler test files

## Context

LLMO-4716 — required for Adobe brandalf flip on 2026-05-06. Adobe's existing audit-worker SNS publishes don't carry `brand_id`; without this change the runner reads stale v1 LLMO config from the no-longer-published spreadsheet mirror.

## Self-review applied

A multi-agent local review surfaced two issues; addressed in a follow-up commit:

- **Resolver errors were silently re-wrapped as misleading "Failed to create S3 folder for audit"** — the brand resolution call ran inside the outer try/catch, so when the resolver fail-closed-threw on a 5xx, DLQ messages pointed at S3 instead of the real cause (the brand endpoint). Hoisted the resolver call above the metadata write and outside the try/catch so the original `[BrandResolver]` error propagates unwrapped to DLQ. Also: when other errors fall into the outer catch (S3 / SharePoint / DRS upload), `throw new Error(msg, { cause: error })` now preserves the original error chain.
- **Brand resolution after S3 metadata write left orphan `metadata.json` files** — when the resolver threw, the metadata file had already been written; SQS retry created a fresh auditId, so orphans accumulated until manual cleanup. Same hoist fixes this — no S3 write fires when the resolver fails. Tests assert `s3Client.send` is never invoked on resolver failure.

## Deferred follow-ups

These items were considered and deliberately deferred:

- **30s fetch timeout** is generous for a single-row DB lookup behind Fastly. Other handlers in this repo use 10s. Worth tightening once we measure dev latency, but not load-bearing for correctness.
- **No in-process retry on 5xx** — relies on SQS retry. A single 250ms-backoff retry would absorb the common transient case without paying the SQS-replay tax (which re-runs SharePoint reads + DRS uploads). Filed as a follow-up.
- **Resolver throws on any non-404 4xx**. 401/403 fail-closed is correct (auth misconfig should be loud); 400 is unlikely in practice. Documented in the file header; classification refinement deferred.
- **`requireSpacecatApi(env)` helper extraction** — this is the fourth direct caller of spacecat-api in the worker. A thin shared client would centralize timeout/auth/retry policy. Architectural follow-up.
- **Cross-org blast radius on 5xx** — when SpaceCat API is briefly unavailable, every BP audit fails (not just brandalf orgs). For v1 orgs the resolver returns 404 (no error) so they're unaffected in steady state; the cross-org risk is only real during a true API outage, when fail-closed is the safer default. Trade-off documented; revisit if false-positive failures become noisy in dev.
- **No structured signal for "v2-path entry"** — adding a CloudWatch metric on `brand_id` set rate would make the 2026-05-06 go/no-go verification minutes-fast instead of log-grep slow. Filed as ops-readiness follow-up.

## Rollout — depends on two PRs

⚠️ **This PR must NOT merge until both of the following ship:**

1. [spacecat-shared#1586](https://github.com/adobe/spacecat-shared/pull/1586) — adds the `brandId` parameter to `publishBrandPresenceAnalyze`. **Release must publish before this PR merges.** Bump `@adobe/spacecat-shared-drs-client` in `package.json` once the release is live.
2. [spacecat-api-service#2334](https://github.com/adobe/spacecat-api-service/pull/2334) — adds the resolver endpoint this PR calls. **Endpoint must be deployed before this PR is enabled in prod.**

## Test plan

- [x] 7751 unit tests passing (audit-worker full suite)
- [x] 100% coverage on `brand-resolver.js` and both refresh handlers
- [x] Lint clean
- [x] Resolver-error preservation test asserts original `[BrandResolver]` message survives the throw
- [x] Orphan-metadata test asserts `s3Client.send` never fires when resolver fails
- [x] Bump `@adobe/spacecat-shared-drs-client` once shared release publishes
- [ ] Manual smoke after deploy on a brandalf-test org:
  - [ ] Trigger `geo-brand-presence-refresh` → confirm `brandId` resolved in audit-worker logs, SNS published with `metadata.brand_id`
  - [ ] DRS-side: confirm Job record has `brand_id`, runner enters v2 path
  - [ ] Trigger same on a non-brandalf org → endpoint returns 404 → SNS published without `brand_id` → runner stays on v1 path
  - [ ] Verify fail-closed: when SpaceCat API is artificially flapping (5xx), audit fails with the resolver error message intact in DLQ

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "@adobe/spacecat-shared-athena-client": "1.9.11",
         "@adobe/spacecat-shared-data-access": "3.56.1",
         "@adobe/spacecat-shared-data-access-v2": "npm:@adobe/spacecat-shared-data-access@2.109.0",
-        "@adobe/spacecat-shared-drs-client": "1.6.0",
+        "@adobe/spacecat-shared-drs-client": "1.7.0",
         "@adobe/spacecat-shared-google-client": "1.5.11",
         "@adobe/spacecat-shared-gpt-client": "1.6.22",
         "@adobe/spacecat-shared-html-analyzer": "1.2.10",
@@ -3039,9 +3039,9 @@
       }
     },
     "node_modules/@adobe/spacecat-shared-drs-client": {
-      "version": "1.6.0",
-      "resolved": "https://registry.npmjs.org/@adobe/spacecat-shared-drs-client/-/spacecat-shared-drs-client-1.6.0.tgz",
-      "integrity": "sha512-MMuyoSlQTR8D2pqymauvTvWgYEXh5chtCzDO9VJmlyZgjcORLMug1wvtIjAKJD/dR8SXu4FGaVGukgIBCj3xSQ==",
+      "version": "1.7.0",
+      "resolved": "https://registry.npmjs.org/@adobe/spacecat-shared-drs-client/-/spacecat-shared-drs-client-1.7.0.tgz",
+      "integrity": "sha512-+lQLN1/OnKzaCrvg4ppsWXitzjuNF1Ady4+hB4vfCEUfhikNuKndRLND4HtkGpckHIAUOJMOGdjGw8G6O11nlA==",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/spacecat-shared-utils": "1.98.1",
--- a/package.json
+++ b/package.json
@@ -93,7 +93,7 @@
     "@adobe/spacecat-shared-athena-client": "1.9.11",
     "@adobe/spacecat-shared-data-access": "3.56.1",
     "@adobe/spacecat-shared-data-access-v2": "npm:@adobe/spacecat-shared-data-access@2.109.0",
-    "@adobe/spacecat-shared-drs-client": "1.6.0",
+    "@adobe/spacecat-shared-drs-client": "1.7.0",
     "@adobe/spacecat-shared-google-client": "1.5.11",
     "@adobe/spacecat-shared-gpt-client": "1.6.22",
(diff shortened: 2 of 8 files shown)


igor-grubic (approved):
my claude said ok

PR #2476, opened 2026-05-05
fix(preflight): report unreachable links as broken (status 0) instead of silently dropping them

clotton:
## Summary

Fixes [SITES-40919](https://jira.corp.adobe.com/browse/SITES-40919)

When both HEAD and GET requests throw (DNS failure, network timeout, unsupported protocol), the GET fallback catch block in `checkLinkStatus()` previously returned `null`, silently dropping the link from the audit result. This meant genuinely unreachable links were never reported as broken.

**Root cause:** `src/preflight/links-checks.js` GET fallback catch block returned `null` and logged at `error` level on network-level failures.

**Fix:** Return a broken-link result with `status: 0` (sentinel for "no HTTP response received") and downgrade the log from `error` to `info` — an unreachable link is a detection outcome, not an application error.

**Confirmed in production:** Coralogix logs from a Walmart DA preflight run (April 22 2026, traceId `1-69e91836-e29ad3f9459ed9e838068e05`) show `getaddrinfo ENOTFOUND` on intranet hostnames — both HEAD and GET threw, links were silently dropped from output.

## Changes

- `src/preflight/links-checks.js` — GET fallback catch block now returns `{ urlTo, href, status: 0, ...elements }` instead of `null`; log level downgraded from `error` to `info`
- `test/preflight/links-checks.test.js` — updated existing test that documented the bug; added internal-link variant

## Test plan

- [x] Updated unit test: `'flags as broken (status 0) when both HEAD and GET throw network errors'` — asserts `brokenExternalLinks` has length 1, `status === 0`, `log.info` called with "unreachable", `log.error` not called
- [x] New unit test: internal link variant with `getaddrinfo ENOTFOUND` — asserts `brokenInternalLinks` has length 1, `status === 0`
- [x] All 27 unit tests passing (`npx mocha --spec test/preflight/links-checks.test.js`)
- [x] Integration smoke test (`scripts/test-links-checks-dns.js`) with real DNS calls — ENOTFOUND on dead external domain and TLS error on dead internal path both surface as `status: 0`; live google.com not flagged; `log.error` never called

## MFE impact

No MFE changes needed. `status: 0` propagates through `links.js` as `issue: "Status 0"` — the MFE renders this via the generic `formatDefaultOpportunity` path with no allowlist or status-code filtering.


clotton:
@johobot ready for re-review — addressed your feedback in commit `a588b2d0`:

- Added a scheme guard at the top of `checkLinkStatus()` that returns `null` for any protocol other than `http:` or `https:`. Non-web schemes (`mailto:`, `tel:`, `javascript:`, `sms:`, etc.) now skip the fetch entirely and cannot be reported as broken.
- Added three regression tests covering `mailto:`, `tel:`, and `javascript:` — each asserts `brokenExternalLinks` is empty and `fetchStub` is never called.
- All 30 unit tests passing.

The SITES-35543 work (filtering non-web schemes earlier in `anchors.each()` so they never enter the link maps) remains a separate PR per the original ticket scope.

GeezerPelletier:
## PR Review: `fix(preflight): report unreachable links as broken (status 0)`

### Overview

This fix addresses a real production bug where DNS/timeout failures during link checking caused links to be silently dropped from audit output rather than surfaced as broken. The change is well-scoped: two lines changed in the source, scheme filtering added in response to review feedback, and 30 tests now passing.

---

### What Changed

| File | Change |
|---|---|
| `src/preflight/links-checks.js` | Scheme guard + broken-link return in GET fallback catch |
| `test/preflight/links-checks.test.js` | 5 new/updated tests |
| `test/audits/preflight.test.js` | 2 existing tests updated to match new behavior |

---

### Correctness

**Positive:**
- The `status: 0` sentinel is a clean, unambiguous signal for "no HTTP response received."
- The return shape `{ urlTo: href, href: pageUrl, status: 0, ...toElementTargets(selectors) }` is consistent with how other broken-link results are structured (confirmed by test assertions).
- The scheme guard correctly short-circuits before `fetch()` is called, so non-web URIs generate zero noise.

**One subtle risk — `new URL(href)` can throw:**
```js
const { protocol } = new URL(href);
```
This is at the top of `checkLinkStatus()` with no try/catch. If a caller ever passes a non-parseable string here, it will throw unhandled. The existing test `'silently skips hrefs that cannot be parsed as URLs'` confirms that path is guarded *upstream*, but there is now a latent fragility: any future caller that bypasses that upstream filter will get an uncaught exception rather than a graceful `null`. Consider wrapping this in a try/catch or noting the invariant in a comment.

---

### Code Quality

- Log level downgrade from `error` → `info` is correct — DNS failures on intranet links are expected detection outcomes, not application faults.
- The comments in the diff are more verbose than typical for this codebase. They explain *why*, which is good, but the block comment on the scheme guard is three lines for something that could be one.
- Minor: the `href` / `urlTo` naming in the return object (`urlTo: href, href: pageUrl`) is confusing at first read. This isn't new in this PR but worth flagging if the struct is ever refactored.

---

### Test Coverage

- External + internal link ENOTFOUND cases: covered.
- `mailto:`, `tel:`, `javascript:` non-web schemes: covered (directly addressing johobot's concern).
- `fetchStub.callCount === 0` assertion on scheme-filtered paths is a strong guard — proves fetch is never even attempted.
- **Gap:** No test for `sms:` or `data:` schemes (mentioned in johobot's comment). Not a blocker, but `sms:` in particular appears in mobile-optimized content.

---

### Security

No security concerns. The change narrows the attack surface slightly by rejecting non-HTTP schemes before any network call.

---

### Summary

**Ready to merge** with one recommendation:

1. **Should address (low risk today, fragile long-term):** Wrap `new URL(href)` in the scheme guard with a try/catch and return `null` on parse failure, so `checkLinkStatus()` is defensively self-contained regardless of what callers pass it.

The core fix is correct, the scope is tight, and the scheme filtering added in response to the johobot review resolves the main concern. The test suite is solid.

---
*Reviewed with [Claude Code](https://claude.com/claude-code)*

GeezerPelletier (approved):

PR #2477, opened 2026-05-05
fix(geo-brand-presence): retry SharePoint XLSX download on non-XLSX response

irenelagno:
## Summary

- Adds `readFromSharePointWithRetry()` to `src/utils/report-uploader.js` that validates XLSX magic bytes (`PK\x03\x04`) after each SharePoint download and retries up to 3 times with a 60-second delay before aborting
- Replaces the bare `readFromSharePoint` call in `refreshGeoBrandPresenceSheetsHandler` (the sheet-upload-to-S3 path) with `readFromSharePointWithRetry`, preventing corrupted SharePoint HTML error pages from being written to `s3://drs-v2-prod-bp/...source.xlsx`
- Adds 4 new unit tests covering: immediate XLSX success, retry-then-succeed, exhausted-retries throw, and network error propagation

## Context

On 2026-05-05, 4 brand presence analyses failed for Astral Pool (b6d70b53) and Zodiac Pool (70f86b2a, 1cd860cc). Root cause: SharePoint returned a transient HTML error page (HTTP 200, `<!DOCTYPE html>`) instead of XLSX content. Lambda wrote the 8,062-byte HTML to S3; DRS Fargate paid full startup cost before failing inside `openpyxl` with "File is not a zip file".

## Test plan

- [ ] `npm run test:spec -- test/utils/report-uploader.test.js` — 4 new tests for `readFromSharePointWithRetry` pass
- [ ] `npm run test:spec -- test/audits/geo-brand-presence-refresh.test.js` — all existing tests still pass with the new stub wired in
- [ ] `npm run lint` — no errors in changed files


irenelagno:
## Related PRs — LLMO-4739 fix set

This is part of a 3-PR fix for the 2026-05-05 incident where SharePoint HTML error pages were written to S3 as `source.xlsx`, causing DRS Fargate tasks to fail.

| Repo | PR | Responsibility |
|------|----|---------------|
| `spacecat-audit-worker` | **this PR** | Retry `readFromSharePoint()` up to 3× with 60s backoff before aborting upload |
| `spacecat-shared` | [#1589](https://github.com/adobe/spacecat-shared/pull/1589) | Last-resort guard in `uploadExcelToDrs()` — reject non-XLSX before `PutObjectCommand` |
| `llmo-data-retrieval-service` | [#1575](https://github.com/adobe-rnd/llmo-data-retrieval-service/pull/1575) | DRS-side safety net: pre-flight magic-byte check before Fargate ECS launch + fail-fast in `parse_excel_to_rows()` |

All three should merge together (or this PR first, since it prevents the bad write at the source).

irenelagno:
The Lambda timeout is **900 seconds (15 minutes)**, set in [`template.yml:11`](https://github.com/adobe/spacecat-audit-worker/blob/fix/sharepoint-xlsx-validation-retry/template.yml#L11). The worst-case blocking sleep from retries is 2 × 60s = **120 seconds per sheet**. Sheets are processed via `Promise.allSettled` concurrently, so parallel retry sleeps don't stack. There is no timeout risk under the current configuration. No change needed.

irenelagno:
Valid point — the name implies general retry behavior but only magic-byte failures are retried; network/SDK throws propagate immediately. Added a `Note:` to the JSDoc in the latest commit clarifying the scope rather than renaming (a rename would touch the import and call site in the handler plus all test stubs for no behavioral change).

src/utils/report-uploader.js
```diff
@@ -83,6 +88,39 @@ export async function readFromSharePoint(filename, outputLocation, sharepointCli
...
+export async function readFromSharePointWithRetry(sheetName, sourceFolder, sharepointClient, log) {
+  for (let attempt = 1; attempt <= XLSX_RETRY_MAX; attempt += 1) {
+    // eslint-disable-next-line no-await-in-loop
+    const buffer = await readFromSharePoint(sheetName, sourceFolder, sharepointClient, log);
+    if (Buffer.from(buffer.slice(0, 4)).equals(XLSX_MAGIC)) {
+      return buffer;
+    }
+    log.warn(
+      `%s: SharePoint returned non-XLSX content on attempt ${attempt}/${XLSX_RETRY_MAX} `
+      + `(size=${buffer.length ?? buffer.byteLength}, magic=${Buffer.from(buffer.slice(0, 4)).toString('hex')})`,
```
dzehnder: **Important.** The retry warn lacks `sheetName` and `sourceFolder`. Under `Promise.allSettled` every sheet retries concurrently, so warns from different sheets interleave and you cannot tell which sheet got the HTML page from logs alone - exactly the question the postmortem asked. Add the path context here and to the thrown Error at line 121.

Two minor things on this same line: `buffer.length ?? buffer.byteLength` is dead code (`.length` is always defined on a Node Buffer) and inconsistent with the rest of the file (line 75 uses `byteLength || length`). And the `Buffer.from(buffer.slice(0, 4))` wrap is redundant - `slice` returns a Buffer already, `subarray(0, 4)` is the non-deprecated form.

src/utils/report-uploader.js
```diff
@@ -83,6 +88,39 @@ export async function readFromSharePoint(filename, outputLocation, sharepointCli
...
+      AUDIT_NAME,
+    );
+    if (attempt < XLSX_RETRY_MAX) {
+      // eslint-disable-next-line no-await-in-loop
+      await sleep(XLSX_RETRY_DELAY_MS);
+    }
+  }
+  throw new Error(`SharePoint returned non-XLSX content after ${XLSX_RETRY_MAX} attempts — aborting S3 upload`);
```
dzehnder: **Minor.** Two things on this line: the message contains U+2014 em-dash which can break ASCII-expecting parsers (alerts, SIEM, ticket titles) - use `--` or `:`. And the message has no structured context - compare to the catch block in `readFromSharePoint` above which logs `filename`, `outputLocation`, `errorCode`, stack. Operators reading this in prod see only the static message; a `log.error` with `sheetName`/`sourceFolder` immediately before the throw would mirror the sibling pattern.

dzehnder (changes requested):
Hey @irenelagno,

Solid, focused fix for the LLMO-4739 incident. Magic-byte validation is the right primitive, and the layered defense across the three companion PRs is sound. Two gaps worth addressing before merge: a same-handler call site that is still unprotected against the exact failure mode this PR exists to prevent, and observability/test gaps around the new throw path under `Promise.allSettled`.

## Strengths

- Magic-byte check (`PK\x03\x04`) is unambiguous and has no false positives against legitimate XLSX (`src/utils/report-uploader.js:108`).
- Loop with explicit `attempt < XLSX_RETRY_MAX` guard avoids a useless trailing 60s sleep after the final attempt (`src/utils/report-uploader.js:115`).
- Retry scope is narrow (only magic-byte mismatches retry; network/SDK throws propagate). The JSDoc note (`src/utils/report-uploader.js:94-95`) and the matching test (`test/utils/report-uploader.test.js:204-216`) lock that contract in.
- Constants at module scope (`XLSX_RETRY_MAX`, `XLSX_RETRY_DELAY_MS`) make the policy tunable without hunting call sites.
- Author already addressed the Lambda-timeout objection in PR comments: 3 attempts x 60s = 120s worst case, well under the 900s ceiling, and `Promise.allSettled` keeps sleeps from stacking.

## Issues

### Important (Should Fix)

**1. `query-index.xlsx` read in the same handler is still unprotected.** `src/geo-brand-presence/geo-brand-presence-refresh-handler.js:101` calls the bare `readFromSharePoint` and feeds the buffer straight to `ExcelJS.Workbook().xlsx.load()` a few lines below. If SharePoint serves the same HTML decoy that triggered the May 5 incident for `query-index.xlsx`, the entire audit fails with the original "File is not a zip file"-class error one stack frame upstream of the path this PR fixes. This is the same incident shape, same handler, just one function-call earlier. Either swap this call to `readFromSharePointWithRetry` as well, or fold the check into `readFromSharePoint` itself.

**2. Retry warn log is missing the sheet path - triage gap under concurrent failures.** `src/utils/report-uploader.js:111-115`. The warn line reports `attempt`, `size`, and `magic` hex, but neither `sheetName` nor `sourceFolder`. Under `Promise.allSettled` with N sheets, every sheet hitting a SharePoint outage logs interleaved warns and you cannot answer "which sheet got the HTML page?" from logs alone - exactly the question the postmortem asked. Add `sheetName` and `sourceFolder` to the warn (and to the thrown Error message at line 121, which loses path context as it propagates).

**3. No handler-level test for the retry-exhaustion failure path.** `test/audits/geo-brand-presence-refresh.test.js` stubs `readFromSharePointWithRetryStub.resolves(...)` (lines 121, 160) but never makes it reject. The whole point of `Promise.allSettled` (which the author cited as the timeout-safety guarantee) is poison-pill resistance, and the new code introduces a new throw site without exercising it through the handler. Add one test: two sheets, first stub rejects with the new error, second resolves; assert the second is queued to DRS, the first is recorded via `writeSheetRefreshResultFailed`, and the handler does not throw out of the `allSettled` block.

### Minor (Nice to Have)

**4. Fixed 60s delay diverges from the existing `fetchWithRetry` pattern in the same file.** `src/utils/report-uploader.js:106` uses fixed 60s, no jitter. `fetchWithRetry` (further down the same file) uses exponential backoff with jitter starting at 10s. With concurrent sheets retrying in lockstep, every sheet wakes at the same two synchronized 60s boundaries and hammers SharePoint together. Either add a small jitter window, or add a one-line comment at the constant explaining why fixed is correct here (e.g. "tuned for the SharePoint outage window observed in LLMO-4739, exponential offers no benefit for cache-flip recovery").

**5. `Buffer.from(buffer.slice(0, 4))` is a redundant wrap.** `src/utils/report-uploader.js:108` and `:113`. `Buffer.prototype.slice` already returns a Buffer; `Buffer.equals` and `.toString('hex')` work directly on it. `subarray(0, 4)` is the documented non-deprecated equivalent. Pure readability.

**6. `buffer.length ?? buffer.byteLength` is dead code and inconsistent with the file.** `src/utils/report-uploader.js:113`. `.length` is always defined on a Node `Buffer`, so the `??` branch can never fire. The rest of the file uses `byteLength || length` (e.g. line 75). Pick one form and use it consistently.

**7. Em-dash in thrown error message.** `src/utils/report-uploader.js:121` contains U+2014 (`—`). Most logging stacks handle UTF-8, but if the message ever lands in an ASCII-expecting parser (alerts, certain SIEM ingestors, ticket titles), it renders as `?` or breaks parsing. Replace with `--` or `:`.

**8. Thrown error has no structured context.** `src/utils/report-uploader.js:121`. Compare to the catch block in `readFromSharePoint` next door which logs `filename`, `outputLocation`, `errorCode`, and stack. The new throw site has no `sheetName`, no `sourceFolder`, no last-seen prefix - operators see only the static message. A `log.error` with structured fields immediately before the throw, mirroring the sibling pattern, would close the gap.

## Recommendations

- Extract a tiny `isXlsx(buffer)` helper. The companion `spacecat-shared#1589` is implementing the same predicate on the upload side, so the concept is now used in two places. A named, greppable invariant beats two inline magic-byte checks that can drift.
- Surface a metric (or structured log field) for `xlsx_magic_retry_attempted`. The May 5 incident was discovered after the fact via openpyxl crashes. A metric lets you alert the next time SharePoint flips back to HTML before downstream damage accumulates.
- If you keep the wrapper rather than folding into `readFromSharePoint`, audit the other call sites of the bare function (other handlers in `src/geo-brand-presence/`, `src/faqs/`, `src/related-urls/`, `src/llm-error-pages/`, etc.) and decide deliberately which ones do/don't need the retry. Otherwise the codebase grows a "guess which one to call" convention that will leak protection over time.

- The companion PRs (`spacecat-shared#1589` upload-side guard, `llmo-data-retrieval-service#1575` DRS-side fail-fast) are the right complementary defenses. Confirm all three land before closing LLMO-4739.
- Pushing extension-aware response validation into `@adobe/spacecat-helix-content-sdk` (so any consumer of `getDocumentContent()` gets the safety) would close this class of bug at the source rather than across N consumers. Worth a follow-up conversation with the SDK owners.

## Assessment

Ready to merge? With fixes.

Reasoning: The core fix is correct and well-tested for the path it covers. The Important items are small additions, not redesigns: protect the `query-index.xlsx` read in the same handler, add sheet path to the retry warn and thrown error, and add a handler-level test that exercises the new throw through `Promise.allSettled`. Minor items are polish.

## Next Steps

1. Address the 3 Important items first - they are all small changes, none requires a redesign.
2. The 5 Minor items are polish; bundling Minor 6, 7, and 8 into a single follow-up commit on the same PR is fine.
3. Confirm the two companion PRs are progressing in parallel.

irenelagno:
All 3 important items and all 5 minor items addressed in the latest commit:

**Important**
1. **query-index.xlsx now protected** — swapped the bare `readFromSharePoint` call in `fetchQueryIndexPaths` to `readFromSharePointWithRetry`. `readFromSharePoint` is no longer imported in the handler at all.
2. **Retry warn includes path context** — `sheetName` and `sourceFolder` added to the warn message; structured `log.error` with `{ sheetName, sourceFolder }` fields added immediately before the throw, mirroring the sibling pattern in `readFromSharePoint`.
3. **Handler-level retry-exhaustion test added** — two sheets (`SHEET_W45`, `SHEET_W46`), chatgpt sheet throws the exhaustion error, gemini sheet resolves normally; asserts the gemini sheet reaches DRS, the failed sheet is recorded via `PutObjectCommand`, and the handler does not throw out of `Promise.allSettled`.

**Minor**
4. **Fixed delay comment added** — one-line comment at the sleep call explaining why fixed 60s is correct for LLMO-4739 recovery cadence.
5. **`buffer.subarray(0, 4)`** — replaced both `Buffer.from(buffer.slice(0, 4))` occurrences.
6. **Dead `?? byteLength` dropped** — warn now uses `buffer.length` directly.
7. **Em-dash replaced with `--`** in both the thrown error message and the `log.error` call.
8. **Structured `log.error` before throw** — covered by item 2 above.

src/utils/report-uploader.js
irenelagno: Fixed in the latest commit — warn message now includes `sheet=${sheetName}` and `folder=${sourceFolder}`, and the thrown Error message carries the same context. Also switched to `buffer.subarray(0, 4)` and dropped the dead `?? byteLength`.

src/utils/report-uploader.js
irenelagno: Fixed in the latest commit — em-dash replaced with `--`, and a `log.error` with structured `{ sheetName, sourceFolder }` fields now fires immediately before the throw, mirroring the catch block in `readFromSharePoint`.

dzehnder (changes requested):
Hey @irenelagno,

## Strengths
All 8 findings from the prior review are confirmed addressed in db9ec3f6:
- `fetchQueryIndexPaths` at `src/geo-brand-presence/geo-brand-presence-refresh-handler.js:101` is now protected
- Retry warn includes `sheet=`/`folder=`; structured `log.error` with `{ sheetName, sourceFolder }` fires before throw at `src/utils/report-uploader.js:85-90`
- Handler-level retry-exhaustion test added at `test/audits/geo-brand-presence-refresh.test.js:303`
- Fixed-delay rationale comment, `buffer.subarray(0,4)` replacing `Buffer.from(buffer.slice(0,4))`, dead `?? byteLength` removed, em-dash replaced with `--` -- all landed

The JSDoc note scoping retry to magic-byte failures only (`src/utils/report-uploader.js:91-92`) is clear and correctly matched by the "propagate without retrying" unit test. The function is correct: fail-closed, structured logging, right primitives.

## Issues

### Important (Should Fix)

**1. The daily-refresh handler has the same unprotected call sites.** `src/geo-brand-presence-daily/geo-brand-presence-refresh-handler.js` imports and calls bare `readFromSharePoint` at line 115 (`query-index.xlsx`) and line 331 (per-sheet reads). The May-5 incident shape -- SharePoint returning an HTTP-200 HTML decoy -- is not exclusive to the weekly handler. The daily path has the same failure mode and would write the same corrupted `source.xlsx` to S3. Swapping both call sites to `readFromSharePointWithRetry` is 2 lines; alternatively, explicitly track this as a follow-up issue in the PR description rather than leave the gap silent.

**2. Handler-level retry-exhaustion test has degenerate failure-path assertions.** `test/audits/geo-brand-presence-refresh.test.js:319`: `expect(s3Client.send).to.have.been.calledWith(sinon.match.instanceOf(PutObjectCommand))` does not verify that `writeSheetRefreshResultFailed` ran for the failed sheet -- `s3Client.send` is called with `PutObjectCommand` in the handler for the gemini sheet's metadata write regardless of whether the failure path fired. The inline comment "Handler does not throw out of Promise.allSettled" is also unverified -- Mocha passes that block whether or not the handler threw internally. To lock the contract: match on the failure-record Key pattern, assert `log.error` was called with the retry-exhaustion message, and assert the handler return value.

### Minor (Nice to Have)

**3. Wrapper name encodes mechanism, not intent.** `src/utils/report-uploader.js`. Audit authors scanning autocomplete see two functions and must read the JSDoc to choose. A name like `readXlsxFromSharePoint` (or annotating the raw one as `readFromSharePointUnsafe`) would steer the right default without changing behavior. Not blocking.

**4. No test for buffers shorter than 4 bytes.** `buffer.subarray(0, 4).equals(XLSX_MAGIC)` correctly returns `false` for a 0-3 byte buffer (length mismatch causes retry), but this edge is not exercised. A SharePoint outage returning an empty 200 body is a plausible real-world failure mode.

## Recommendations
- Audit bare `readFromSharePoint` consumers in `src/related-urls/`, `src/faqs/`, `src/llm-error-pages/`, and `src/prerender/utils/shared.js`. If most need protection, folding the magic-byte check into `readFromSharePoint` itself avoids the growing "guess which one to call" convention.

- Companion PRs (`adobe/spacecat-shared#1589`, `adobe-rnd/llmo-data-retrieval-service#1575`) provide the correct layered defense. All three should land before LLMO-4739 is closed.

## Assessment
Ready to merge? With fixes.
Reasoning: Core function and unit tests are solid. Both Important items are small additions: protect the daily handler (2 call-site swaps), and tighten the failure-path test assertions (3-4 lines). Neither requires a redesign.

## Next Steps
1. Swap `src/geo-brand-presence-daily/geo-brand-presence-refresh-handler.js` lines 115 and 331 to `readFromSharePointWithRetry`
2. Add specific assertions to the retry-exhaustion test: Key pattern match, handler return value, `log.error` call verification
3. Minors 3 and 4 are optional polish

dzehnder:
**Inline note on `test/audits/geo-brand-presence-refresh.test.js:320`**

`s3Client.send.calledWith(sinon.match.instanceOf(PutObjectCommand))` matches any `PutObjectCommand` including the gemini sheet's normal metadata write, so this assertion passes even if `writeSheetRefreshResultFailed` never ran for the failed sheet. The comment on line 321 ("Handler does not throw out of Promise.allSettled") is also unverified -- Mocha passes the block whether or not the handler threw internally.

To lock the contract:
- Match on the failure-record Key pattern (e.g. something that distinguishes the failure metadata from success metadata)
- Assert `log.error` was called with the retry-exhaustion message
- Assert the handler return value to verify it did not throw out of `Promise.allSettled`

irenelagno:
@dzehnder  Both important items and minor 4 addressed in the latest commit:

**Important 1 — daily handler protected:** Swapped both bare `readFromSharePoint` calls in `src/geo-brand-presence-daily/geo-brand-presence-refresh-handler.js` (line 115 query-index, line 331 per-sheet reads) to `readFromSharePointWithRetry`. Updated the daily handler test stub accordingly.

**Important 2 — retry-exhaustion test assertions tightened:**
- Assert handler return value is 200 (proves it doesn't throw out of `Promise.allSettled`)
- Match S3 Key against `/brandpresence-chatgpt.*\.metadata\.json$/` (proves `writeSheetRefreshResultFailed` ran for the correct sheet via the results-processing loop at line 394-403, not a spurious metadata write)
- Assert `log.error` was called with the results-loop pattern `%s:Failed to process sheet` plus the sheet name and retry-exhaustion message (the throw from `readFromSharePointWithRetry` bypasses the inner DRS catch block and lands in the outer rejected-result handler)

**Minor 4 — short-buffer test added:** Empty buffer (0 bytes) triggers a retry and recovers on the next attempt with a valid XLSX buffer.

dzehnder (approved):
**Re-review #3 · commit 24566cb2 · 1 substantive commit since last review**

---

### Prior Findings — Status Update

All four tracked findings are resolved:

- **Important 1** (daily handler called bare `readFromSharePoint`) → Addressed — both call sites in `geo-brand-presence-refresh-handler.js` (lines 115 and 331) now use `readFromSharePointWithRetry`
- **Important 2** (retry-exhaustion test assertions were degenerate) → Addressed — `result.status === 200`, the `brandpresence-chatgpt.*\.metadata\.json$` S3 key regex, and the `log.error` format-string matcher at `geo-brand-presence-refresh.test.js:317-331` make the test non-degenerate
- **Minor 3** (wrapper name encodes mechanism, not intent) → Unresolved per author's discretion; no new evidence to override
- **Minor 4** (no test for short buffer) → Addressed — `Buffer.alloc(0)` case added at `test/utils/report-uploader.test.js:208`

---

### New Issues

#### Minor (Nice to Have)

**Missing symmetric retry-exhaustion test for the daily handler** (`test/audits/geo-brand-presence-daily/geo-brand-presence-refresh-handler.test.js`)

The weekly handler has an explicit test covering "sibling sheet succeeds, failed sheet writes failure record" (`test/audits/geo-brand-presence-refresh.test.js:306`). The daily handler was updated to call `readFromSharePointWithRetry`, but has no corresponding retry-exhaustion path test. If the daily handler's results loop ever diverges from the weekly handler's (different log format string, different `writeSheetRefreshResultFailed` call site, different return shape), there is no test that catches it.

**How to fix:** Mirror the weekly retry-exhaustion test in the daily test file — stub `readFromSharePointWithRetryStub` to throw on one sheet, assert `result.status === 200`, the failure-record S3 Key pattern, and the daily handler's `log.error` format. Approximately 20-30 lines.

---

### Assessment

**Ready to merge.**

Both prior Important findings are properly addressed. The daily-handler symmetry gap is closed, failure-path assertions now genuinely distinguish success from failure metadata writes, and the short-buffer edge case is covered. The single new Minor (no symmetric daily-handler retry-exhaustion test) does not block merge given the unit-level coverage of `readFromSharePointWithRetry` itself — worth a follow-up commit or tracked issue.

*Multi-agent review: staff-engineer, architect, qa, security — all clear.*

PR #2483, opened 2026-05-06
fix(semantic-value-visibility): set tags on opportunity update in convertToOpportunity

yevheniia-severinovska:
## Summary
- In `convertToOpportunity`, tags from the opportunity data mapper were only applied on **CREATE**, not on **UPDATE**
- When an opportunity already existed (re-triggered audit), the update path skipped `setTags()` — so `isElmo` was never set on existing opportunities
- This caused semantic-value-visibility opportunities to be invisible in the LLMO UI (which filters by `isElmo` tag via `isValidElmoOpportunity()`)

## Fix
Add `opportunity.setTags(opportunityInstance.tags)` in the update branch before `save()`, guarded by a length check to avoid clearing tags unnecessarily.

The `isElmo` tag is already correctly defined in `src/semantic-value-visibility/opportunity-data-mapper.js` — this fix ensures it's applied on every run, not just the first one.

### DIFF
--- a/src/common/opportunity.js
+++ b/src/common/opportunity.js
@@ -103,6 +103,9 @@ export async function convertToOpportunity(auditUrl, auditData, context, createO
           dataSources: opportunityInstance.data?.dataSources,
         });
       }
+      if (opportunityInstance.tags?.length) {
+        opportunity.setTags(opportunityInstance.tags);
+      }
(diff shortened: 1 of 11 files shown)


yevheniia-severinovska:
Not needed for PROD for now

PR #2494, opened 2026-05-08
fix(paid-keyword-optimizer): guidance validation error on opportunity create

danieljchuser:
## Summary
**Bug fix 1: `guidance: {}` validation error**
- `mapToKeywordOptimizerOpportunity` was setting `guidance: {}` (empty object), which fails the Opportunity schema validation (`isNonEmptyObject({})` returns `false`)
- Changed to `guidance: null`, which the validator accepts (it allows null/undefined or non-empty objects)
- This bug prevented any ad-intent-mismatch opportunity from being created by the guidance handler

**Bug fix 2: `misalignedClusters` count included good-aligned clusters**
- `misalignedClusters` was counting any cluster with a recommendation, regardless of alignment score
- Changed to only count clusters with `overallAlignmentScore` of "poor" or "fair" (matching Mystique's `MISALIGNED_ALIGNMENT_SCORES` definition)
- This aligns the audit-worker metric with Mystique's portfolio metrics calculation (companion PR: mystique#2185)
- `assignClusterRanks` Tier 0 intentionally keeps the broader "has recommendation" predicate for display ordering — documented in code comment
- Description copy tightened to "significant alignment gaps" to match the new semantics

## Test plan
- [x] All existing tests pass at 100% coverage
- [x] Explicit `expect(result.guidance).to.be.null` regression guard added
- [x] `good` alignment + recommendation → excluded from `misalignedClusters`
- [x] `poor` alignment + recommendation → counted
- [x] `null` alignment score → treated as not misaligned
- [x] Verified fix works via E2E test on DEV (opportunity `c7feba55-3296-441c-95d1-d25c03ead5f8` created successfully after patching the Lambda)
- [x] UI compatibility verified — experience-success-studio-ui handles `guidance: null` and zero-valued spend fields correctly


danieljchuser (commented):
Hey @danieljchuser,

Clean, surgical bug fix - the diagnosis is precise (`guidance: {}` violates the validator's `isNonEmptyObject(value)` check, and `null` matches the schema's `!value || isNonEmptyObject(value)` shape). The E2E verification on DEV with a real opportunity id is great. A few items worth addressing before merge.

## Strengths

- The inline comment in `src/paid-keyword-optimizer/guidance-opportunity-mapper.js:46-50` documenting why `assignClusterRanks` and `misalignedClusters` use different criteria is exactly the kind of in-code rationale that pre-empts future "are these supposed to disagree?" PRs.
- `MISALIGNED_SCORES` is hoisted to a module-level `Set` rather than reconstructed per filter call - the right choice.
- The 3 new tests cover the meaningful slice of the score-filter change (good excluded, poor included, null treated as not-misaligned). Behavior-named, not interaction-named.

## Issues

### Important (Should Fix)

**1. No regression test for the actual bug in the PR title (`guidance: null`).** No test in `test/audits/paid-keyword-optimizer/opportunity-mapper.test.js` asserts on `result.guidance`. The existing happy-path tests do execute the `guidance: null` line, so Codecov reports it as "covered" - but if a future refactor flips it back to `{}` or `undefined`, no assertion fails. The cluster-level `result.data.cluster.guidance` checks are a different field. Add a one-line `expect(result.guidance).to.be.null;` to one of the existing top-level "creates opportunity" tests, or a dedicated `it('sets guidance to null to satisfy Opportunity schema', ...)`. This is the entire bug being fixed; it deserves an explicit guard.

**2. PR scope and title do not match the diff.** The title and Summary describe only `guidance: {}` -> `null`, but the diff also changes the semantics of `misalignedClusters` (now requires `overallAlignmentScore` in `{poor, fair}` in addition to a recommendation). That is a behavioral change that affects the user-facing `description` and any downstream consumer of `data.misalignedClusters`. Two unrelated fixes in one PR makes the git history harder to bisect or revert. Either split the count-refinement into its own PR, or at minimum update the PR description to call out the second change so reviewers and future archeologists know what to look for.

### Minor (Nice to Have)

**3. User-facing `description` text is now inconsistent with `misalignedClusters` semantics.** The `description` (constructed around `src/paid-keyword-optimizer/guidance-opportunity-mapper.js:105-107`) still reads "X of Y clusters show alignment gaps", but the count now excludes good-aligned clusters with recommendations. A user could see, say, 5 clusters ranked first in the UI but a description saying "2 of 8 clusters show alignment gaps." Worth tightening the copy ("show significant alignment gaps") or recomputing the description from the has-recommendation set so it matches what is ranked. Not blocking.

## Recommendations

- **Centralize the alignment-score vocabulary.** `MISALIGNED_SCORES = new Set(['poor', 'fair'])` hard-codes the score enum in the mapper. The same vocabulary already appears in suggestion data (`overallAlignmentScore`) and in Mystique's payload. If Mystique adds `'critical'` or renames `'fair'` to `'weak'`, this count silently drops the affected clusters with no test failure and no log. Either import from an existing constants module, or define `ALIGNMENT_SCORES = { GOOD, FAIR, POOR }` once in `../common/constants.js` and reuse for the count, the suggestion mapping, and any future consumer. Cheap to harden later: a single `log.warn` when a cluster has a recommendation but an `overallAlignmentScore` that is non-null and not in `MISALIGNED_SCORES` surfaces upstream drift instead of silently undercounting.

- **The `guidance: {}` antipattern is systemic, not isolated to this mapper.** The Opportunity validator's asymmetry (accepts `null`/`undefined`, rejects `{}`) is a recurring footgun. Worth a follow-up ticket: either loosen the validator in `spacecat-shared` to treat `{}` as semantically absent, or codify "always use `null`" as a lint rule. Out of scope here.

- **Architectural test gap.** This bug shipped because the mapper's unit tests construct the opportunity object but never round-trip it through the actual `Opportunity` model's validator. Every mapper in this worker has the same blind spot. A shared "mapper output must satisfy `Opportunity.create()`" test helper in `spacecat-shared` would have caught this in CI and will catch the next variant.

## Assessment

Ready to merge? With fixes.

Reasoning: The fix is correct and E2E-verified, but two items deserve attention before merge - add a regression test pinning `result.guidance === null` (one line; the literal bug being fixed has no explicit guard), and either split the `misalignedClusters` semantics change into its own PR or document it in the PR description. The security surface is zero, and the recommendations above are forward-looking, not blocking.

## Next Steps

1. Add the one-line `expect(result.guidance).to.be.null` assertion to lock in the regression guard for the literal bug.
2. Either split the `misalignedClusters` change into a separate PR or update the PR description to document the second behavioral change.
3. Minor: tighten the user-facing `description` copy so it lines up with the new count semantics.

danieljchuser (commented):
Hey @danieljchuser,

Re-review: all three previously flagged items are addressed in `2e35ef3`. Nothing new to flag.

## Previously flagged - now resolved

- **Important: regression test for `guidance: null`** - `expect(result.guidance).to.be.null;` added to the existing "creates opportunity" test (`test/audits/paid-keyword-optimizer/opportunity-mapper.test.js:354`). The literal bug now has an explicit guard.
- **Important: PR scope and title didn't match the diff** - PR description rewritten to call out both fixes as "Bug fix 1" (guidance) and "Bug fix 2" (misalignedClusters), plus reference to the companion `mystique#2185` that documents Mystique's `MISALIGNED_ALIGNMENT_SCORES` definition. The second behavioral change is now explicit for future bisecters.
- **Minor: description text inconsistent with new count semantics** - copy tightened to "significant alignment gaps" (`src/paid-keyword-optimizer/guidance-opportunity-mapper.js:107`) with the corresponding test assertion updated.

## Forward-looking recommendations (unchanged, follow-up tickets)

- Centralizing the alignment-score enum across audit-worker and Mystique - the PR description now confirms the constant lives in Mystique as `MISALIGNED_ALIGNMENT_SCORES`. Worth a shared module in `spacecat-shared` so the two consumers cannot drift independently. Out of scope for this PR.
- The systemic `guidance: {}` footgun in the Opportunity validator and the missing mapper-to-`Opportunity.create()` round-trip test - both still worth follow-up tickets.

## Assessment

Ready to merge? Yes (once CI build completes - currently pending). The fix is correct, regression-guarded, and the scope is now documented. Recommendations above are forward-looking, not blocking.

MysticatBot-Dev (approved):
Approved by MysticatBot (test review for PAT validation)

PR #2495, opened 2026-05-08
chore(ci): grant pull-requests write for DEV deploy status comments

danieljchuser:
## Summary

- Bumps `pull-requests` permission from `read` to `write` in `ci.yaml`
- Enables the sticky PR comment feature from [mysticat-ci#8](https://github.com/adobe/mysticat-ci/pull/8), which posts DEV deploy status (skipped/deployed/failed) directly on the PR

## Test plan

- [ ] Verify CI passes on this PR (no behavioral change until mysticat-ci v3 is released)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,7 +4,7 @@ permissions:
   id-token: write
   contents: write
   issues: read
-  pull-requests: read
+  pull-requests: write
 
 on:
   push:


solaris007 (approved):
Hey @danieljchuser,

Single-line permission bump in `.github/workflows/ci.yaml` from `pull-requests: read` to `pull-requests: write`, coordinated with the parent PR in `adobe/mysticat-ci`. Trivial change, CI is green.

## Strengths

- Necessary and correct: the reusable `service-ci.yaml` workflow declares `pull-requests: write` at the job level for `branch-deploy`. Reusable workflows can only request a subset of the caller's permissions, so granting it here at the workflow level is what makes the sticky-comment feature work end to end.
- Narrow scope: only the `pull-requests` permission moves from read to write. All other permissions (`id-token`, `contents`, `issues`) are untouched.
- Safe trigger model: the workflow uses plain `pull_request` (not `pull_request_target`), so PR runs from forks keep the restricted token regardless of declared permissions. Internal pushes get the elevated permission, which is what the comment poster needs.
- Pre-staged rollout: granting the permission before mysticat-ci v3 ships is a no-op against the current v2 reusable workflow, but it lets the future v3 bump be a clean single-line change.

## Issues

### Minor (Nice to Have)

- `.github/workflows/ci.yaml`: the bumped `pull-requests: write` is declared at the workflow level. In repos that have local jobs alongside the reusable `ci` call (e.g. `guardrails`, `typecheck`, `build`), this grants the permission to those jobs too, even though only the reusable workflow uses it. Per least-privilege you could scope it tighter by moving the permissions block onto the `ci` job that calls the reusable workflow:

  ```yaml
  ci:
    uses: adobe/mysticat-ci/.github/workflows/service-ci.yaml@v2
    permissions:
      id-token: write
      contents: write
      issues: read
      pull-requests: write
    with: ...
    secrets: inherit
  ```

  Practical blast radius today is small because the local jobs do not make GitHub API calls, but explicit scoping is easier to audit. Non-blocking; fine to merge as-is and tighten later.

## Assessment

Ready to merge: Yes
Reasoning: Safe, narrow, well-coordinated permission bump that is a precondition for the documented mysticat-ci v3 rollout. The change is correct, CI is green, and there is no behavioral change until v3 lands.

PR #2497, opened 2026-05-11
feat(rum-config): add periodic rum-config-refresh handler

Sugandhgyl:
Adds a **new audit-worker handler (type: rum-config-refresh)** that re-checks **RUM** domain key availability for a site and persists the result in `site.config.rumConfig`. Sites checked within the last 7 days are skipped to avoid hammering the RUM API.

## **CHANGES**
- Skips sites whose `rumConfig.lastCheckedAt `is within the last 7 days
- Calls `RUMAPIClient.retrieveDomainkey(domain)` with a 3s timeout for stale/unchecked sites
- Persists { hasDomainKey, lastCheckedAt } back via site.getConfig().updateRumConfig() + site.save()
- Returns internalServerError only if the DB write fails — RUM failures are non-fatal and set hasDomainKey=false

The orchestration service sends { type: 'rum-config-refresh', siteId: '<uuid>' } on a daily cadence.

## Changes
- src/rum-config-refresh/handler.js — new handler
- src/index.js — registers 'rum-config-refresh' in HANDLERS
- test/audits/rum-config-refresh.test.js — 9 tests

## Related Issues

Thanks for contributing!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@adobe/mysticat-shared-seo-client": "1.3.0",
         "@adobe/spacecat-helix-content-sdk": "1.4.33",
         "@adobe/spacecat-shared-athena-client": "1.9.11",
-        "@adobe/spacecat-shared-data-access": "https://gist.githubusercontent.com/Sugandhgyl/383af13fb32c5d0307dba87ce88f699c/raw/adobe-spacecat-shared-data-access-3.62.0.tgz",
+        "@adobe/spacecat-shared-data-access": "3.63.0",
         "@adobe/spacecat-shared-data-access-v2": "npm:@adobe/spacecat-shared-data-access@2.109.0",
         "@adobe/spacecat-shared-drs-client": "1.7.1",
         "@adobe/spacecat-shared-google-client": "1.5.11",
@@ -2334,18 +2334,18 @@
       }
     },
     "node_modules/@adobe/spacecat-shared-data-access": {
-      "version": "3.62.0",
-      "resolved": "https://gist.githubusercontent.com/Sugandhgyl/383af13fb32c5d0307dba87ce88f699c/raw/adobe-spacecat-shared-data-access-3.62.0.tgz",
-      "integrity": "sha512-zm/K1aL/vwpUCExUrqeW95UQb0gMmPmz72EuiGZvHWrp0fgMqFZY996f5wn9g4bGFfusOVpGvhLTY6uSTXanZg==",
+      "version": "3.63.0",
+      "resolved": "https://registry.npmjs.org/@adobe/spacecat-shared-data-access/-/spacecat-shared-data-access-3.63.0.tgz",
+      "integrity": "sha512-3ZjwhQig41OuhEhRpAO+LyBmViz31E4nb18rFDVIZ2md1Fn0eyGmCaTAvlO49f2dziMBdJUcHeO7TOkF4+RlGw==",
       "license": "Apache-2.0",
       "dependencies": {
         "@adobe/fetch": "^4.2.3",
         "@adobe/spacecat-shared-utils": "1.105.0",
         "@aws-sdk/client-s3": "^3.940.0",
-        "@supabase/postgrest-js": "2.101.1",
+        "@supabase/postgrest-js": "2.105.4",
         "@types/joi": "17.2.3",
         "aws-xray-sdk": "3.12.0",
-        "joi": "18.1.2",
+        "joi": "18.2.1",
         "pluralize": "8.0.0"
       },
       "engines": {
@@ -3034,9 +3034,9 @@
       }
     },
     "node_modules/@adobe/spacecat-shared-data-access/node_modules/joi": {
-      "version": "18.1.2",
-      "resolved": "https://registry.npmjs.org/joi/-/joi-18.1.2.tgz",
-      "integrity": "sha512-rF5MAmps5esSlhCA+N1b6IYHDw9j/btzGaqfgie522jS02Ju/HXBxamlXVlKEHAxoMKQL77HWI8jlqWsFuekZA==",
+      "version": "18.2.1",
+      "resolved": "https://registry.npmjs.org/joi/-/joi-18.2.1.tgz",
+      "integrity": "sha512-2/OKlogiESf2Nh3TFCrRjrr9z1DRHeW0I+KReF67+4J0Ns+8hBtHRmoWAZ2OFU6I5+TWLEe6sVlSdXPjHm5UbQ==",
       "license": "BSD-3-Clause",
       "dependencies": {
         "@hapi/address": "^5.1.1",
@@ -14209,9 +14209,9 @@
       "license": "MIT"
     },
     "node_modules/@supabase/postgrest-js": {
-      "version": "2.101.1",
-      "resolved": "https://registry.npmjs.org/@supabase/postgrest-js/-/postgrest-js-2.101.1.tgz",
-      "integrity": "sha512-UW1RajH5jbZoK+ldAJ1I6VZ+HWwZ2oaKjEQ6Gn+AQ67CHQVxGl8wNQoLYyumbyaExm41I+wn7arulcY1eHeZJw==",
+      "version": "2.105.4",
+      "resolved": "https://registry.npmjs.org/@supabase/postgrest-js/-/postgrest-js-2.105.4.tgz",
+      "integrity": "sha512-SppIyLo/kTwIlz1qpv2HN1EQqBg0GVktrDDFsXygYROha3MgVn4rT7p5EjFHFqXQm2rdRGb/BI7bc+jr10m91w==",
       "license": "MIT",
       "dependencies": {
         "tslib": "2.8.1"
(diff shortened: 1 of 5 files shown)


sandsinh (changes requested):
Hey @Sugandhgyl,

## Strengths
- The staleness gate logic (`age < STALENESS_MS`) is a clean and correct implementation of a refresh-avoidance pattern - rechecking a site that was verified 5 minutes ago is wasteful, and this prevents it.
- Returning `ok({ skipped: true, reason: '...' })` with a consistent shape for early-exit paths gives the SQS consumer a structured response it can log and observe.
- The handler is registered via the existing `HANDLERS` map in `src/index.js`, keeping the entry-point consistent with all other handlers.
- Test coverage is present for the main code paths using esmock, and the test structure follows the established pattern for this codebase.

## Issues

### Critical (Must Fix)

**1. `site.setConfig(siteConfig.state)` accesses internal state and will not persist correctly - src/rum-config-refresh/handler.js**
`site.setConfig(siteConfig.state)` passes the internal `state` closure variable of the `Config` instance directly to `site.setConfig`. The correct call is `site.setConfig(Config.toDynamoItem(siteConfig))`, which serializes the config through the model's own serialization path. This bug is masked in the tests because the mock explicitly adds `state: {}` to the config object, making the access succeed in tests while it will fail or produce incorrect output in production. Fix the call to use `Config.toDynamoItem(siteConfig)`.

### Important (Should Fix)

**2. `setTimeout` inside `Promise.race` is never cleared on success - src/rum-config-refresh/handler.js**
When `retrieveDomainkey` resolves before the timeout fires, the pending `setTimeout` callback is never cancelled. In a Lambda environment this holds the event loop open past the handler's return, which can delay the next invocation or cause unexpected callback execution. Wrap the timer reference in a variable and call `clearTimeout` in a `finally` block, or use `AbortController` to cancel the pending request instead.

**3. URL extraction via regex is incorrect for non-root base URLs - src/rum-config-refresh/handler.js**
`site.getBaseURL().replace(/^https?:\/\//, '')` will include path segments, ports, query strings, and trailing slashes in the value passed to `retrieveDomainkey`. Use sharedWwwUrlResolver.

**4. No validation that `siteId` is present in the SQS message before querying the database - src/rum-config-refresh/handler.js**
If the SQS message payload omits `siteId` (malformed message, schema drift), the handler calls `Site.findById(undefined)`. Depending on the data access layer this may silently return nothing (no-op) or throw. Add an explicit guard at the top of the handler and return a non-retryable error so the malformed message does not redrive indefinitely.

**5. Timeout and genuine "no domain key" response are conflated into the same outcome - src/rum-config-refresh/handler.js**
A timeout and a real "no domain key found" response both result in `hasDomainKey = false` and update `lastCheckedAt`. This means a transient network issue silently marks a site as not having a domain key for 7 days. A timeout should be logged at a higher severity and should either skip the `lastCheckedAt` update (so the site is rechecked sooner) or set `hasDomainKey` to a distinct "unknown" state rather than `false`.

## Recommendations

- Align the `site.setConfig(Config.toDynamoItem(...))` pattern across all three companion PRs before any of them merge. The correct pattern is in `plg-onboarding.js` in the api-service PR; use that everywhere.
- Consider whether a timeout should skip the `lastCheckedAt` update entirely, so a transient RUM API outage does not suppress checks for 7 days fleet-wide.

## Assessment
Ready to merge? No
The Gist URL dependency is a supply chain risk that must not reach `main`. The `site.setConfig(siteConfig.state)` persistence bug will cause silent data loss in production while all tests pass. Both are blockers that require fixes before this PR is mergeable.

Sugandhgyl:
> Hey @Sugandhgyl,
> 
> ## Strengths
> * The staleness gate logic (`age < STALENESS_MS`) is a clean and correct implementation of a refresh-avoidance pattern - rechecking a site that was verified 5 minutes ago is wasteful, and this prevents it.
> * Returning `ok({ skipped: true, reason: '...' })` with a consistent shape for early-exit paths gives the SQS consumer a structured response it can log and observe.
> * The handler is registered via the existing `HANDLERS` map in `src/index.js`, keeping the entry-point consistent with all other handlers.
> * Test coverage is present for the main code paths using esmock, and the test structure follows the established pattern for this codebase.
> 
> ## Issues
> ### Critical (Must Fix)
> **1. `site.setConfig(siteConfig.state)` accesses internal state and will not persist correctly - src/rum-config-refresh/handler.js** `site.setConfig(siteConfig.state)` passes the internal `state` closure variable of the `Config` instance directly to `site.setConfig`. The correct call is `site.setConfig(Config.toDynamoItem(siteConfig))`, which serializes the config through the model's own serialization path. This bug is masked in the tests because the mock explicitly adds `state: {}` to the config object, making the access succeed in tests while it will fail or produce incorrect output in production. Fix the call to use `Config.toDynamoItem(siteConfig)`.
> 
> ### Important (Should Fix)
> **2. `setTimeout` inside `Promise.race` is never cleared on success - src/rum-config-refresh/handler.js** When `retrieveDomainkey` resolves before the timeout fires, the pending `setTimeout` callback is never cancelled. In a Lambda environment this holds the event loop open past the handler's return, which can delay the next invocation or cause unexpected callback execution. Wrap the timer reference in a variable and call `clearTimeout` in a `finally` block, or use `AbortController` to cancel the pending request instead.
> 
> **3. URL extraction via regex is incorrect for non-root base URLs - src/rum-config-refresh/handler.js** `site.getBaseURL().replace(/^https?:\/\//, '')` will include path segments, ports, query strings, and trailing slashes in the value passed to `retrieveDomainkey`. Use sharedWwwUrlResolver.
> 
> **4. No validation that `siteId` is present in the SQS message before querying the database - src/rum-config-refresh/handler.js** If the SQS message payload omits `siteId` (malformed message, schema drift), the handler calls `Site.findById(undefined)`. Depending on the data access layer this may silently return nothing (no-op) or throw. Add an explicit guard at the top of the handler and return a non-retryable error so the malformed message does not redrive indefinitely.
> 
> **5. Timeout and genuine "no domain key" response are conflated into the same outcome - src/rum-config-refresh/handler.js** A timeout and a real "no domain key found" response both result in `hasDomainKey = false` and update `lastCheckedAt`. This means a transient network issue silently marks a site as not having a domain key for 7 days. A timeout should be logged at a higher severity and should either skip the `lastCheckedAt` update (so the site is rechecked sooner) or set `hasDomainKey` to a distinct "unknown" state rather than `false`.
> 
> ## Recommendations
> * Align the `site.setConfig(Config.toDynamoItem(...))` pattern across all three companion PRs before any of them merge. The correct pattern is in `plg-onboarding.js` in the api-service PR; use that everywhere.
> * Consider whether a timeout should skip the `lastCheckedAt` update entirely, so a transient RUM API outage does not suppress checks for 7 days fleet-wide.
> 
> ## Assessment
> Ready to merge? No The Gist URL dependency is a supply chain risk that must not reach `main`. The `site.setConfig(siteConfig.state)` persistence bug will cause silent data loss in production while all tests pass. Both are blockers that require fixes before this PR is mergeable.

**Here's a summary of what was fix for the review comments:**

**handler.js — all 5 issues addressed:**

- Critical — site.setConfig(Config.toDynamoItem(siteConfig)) replaces the internal siteConfig.state access
- Important — clearTimeout(timeoutId) in a finally block clears the pending timer after the race
- Important — new URL(site.getBaseURL()).hostname replaces the regex (handles ports, paths, trailing slashes correctly)
- Important — Early !siteId guard returns skipped=true before touching the DB
- Important — Timeout now returns { skipped: true, reason: 'timeout' } without writing lastCheckedAt, so the site is rechecked at the next run rather than suppressed for 7 days

**test file:**

- Dropped state: {} from mockConfig (no longer accessed)
- Added toDynamoItemStub via esmock and asserts it's called on success paths
- New missing siteId test block
- Timeout test updated: expects skipped/timeout, no updateRumConfig call, no save, and log.error (higher severity)
- The Gist URL chore commit was dropped during the rebase (also flagged by Sandesh as a supply chain risk)

sandsinh (changes requested):
Hey @Sugandhgyl,

## Strengths

All five numbered issues from the previous review have been addressed in commit 867730c:
- Previously flagged Critical: site.setConfig(siteConfig.state) persistence bug - now correctly calls Config.toDynamoItem(siteConfig). Verified by the test asserting toDynamoItem is called with the config object.
- Previously flagged: clearTimeout in a finally block - the timer reference is tracked and cleared regardless of success or failure path.
- Previously flagged: URL extraction via new URL().hostname - correctly strips protocol, port, and path segments.
- Previously flagged: siteId guard at handler entry - missing or empty siteId returns skipped=true immediately, confirmed by test.
- Previously flagged: timeout vs no-domain-key distinction - timeout now returns skipped=true and does not write lastCheckedAt, so a transient outage does not suppress rechecks for 7 days.

Test coverage is solid: all paths exercised including the fake-timer timeout test verifying updateRumConfig is not called on timeout.

## Issues

### Important (Should Fix)

**deploy-dev npm script still targets a personal label - package.json**
The deploy-dev script still uses -l sugandhg rather than -l latest (or the shared label). Any developer or CI job running npm run deploy-dev after this merges will deploy to the personal slot. This was called out in the previous review assessment as a blocker alongside the setConfig bug. Revert to -l latest before merging.

## Assessment
Ready to merge? With fixes
One item remains: the personal deploy label in package.json. Everything else is clean.

sandsinh (approved):

PR #2502, opened 2026-05-11
docs(llmo-config): step 3b/3c repair plan (SITES-43238)

dzehnder:
## Summary

Step 3b/3c plan for [SITES-43238](https://jira.corp.adobe.com/browse/SITES-43238). Docs-only — no code yet. Implementation will land as a separate PR once this plan is approved.

## What changed since step 3a

Sweep ran today against `spacecat-prod-importer`:

- 7,481 LLMO configs scanned. 22 schema-invalid (0.29%).
- All 22 corrupt configs were written between 2025-11-12 and 2026-04-30 — i.e. before step 1 merged on 2026-05-04. **No new corruption since.** Steps 1 and 2 are holding.
- The bad values are not malformed region codes as the original ticket described. They are prompt-intent / search-stage classifier strings (`"comparative"`, `"instructional"`, `"informational"`, `"comparison & decision"`, `"discovery & research"`, `"usage & troubleshooting"`, `"scheduling & appointments"`, `"online resources & accessibility"`, `"unknown"`) written to the `region` field.

## Per-site classification (Phase 1, done in the plan)

| Bucket | Count | Repair logic |
|---|---|---|
| `auto-us-majority` | 17 | Replace bad regions with `"us"` — every one of these sites already has multiple categories with `region: "us"` correctly, anchoring the inference. |
| `auto-au-majority` | 1 | Replace bad regions with `"au"` — site has unanimous `"au"` on its other categories. |
| `manual-review` | 4 | No internal signal; operator decision required. German-language sites and an Indeed-related site. |

The plan lists every affected siteId with which bucket it lands in and what the specific bad fields are.

## Plan structure

- **Phase 1** — per-site classification. **Done** in the plan body.
- **Phase 2** — implement `scripts/llmo-config-repair.js` (sibling of the merged sweep script). Dry-run by default; `--write` required to mutate; hardcoded allowlist of auto-repairable sites; closed-set of recognized bad-region values; post-write read-back via fail-closed `readConfig`; local backups + S3 versioning as redundant backstops.
- **Phase 3** — step 3b: canonical repair on `78d59744-e06c-4d14-a77a-9490c1464116`. Validation gate is the original SITES-43238 reproducer (`curl` of the api-service LLMO config GET) returning HTTP 200.
- **Phase 4** — step 3c: batch over the remaining 18 auto-repair sites. Validation gate is a post-repair sweep showing `schemaInvalid: 4` (only the manual-review sites left).
- **Phase 5** — manual review of the 4 remaining sites.
- **Phase 6** — cleanup: delete both `scripts/llmo-config-sweep.js` and `scripts/llmo-config-repair.js` since the platform now prevents recurrence; file the DRS sibling ticket; post-mortem note on the missed major-version bump in step 2.

## Validation gates

Every phase has a concrete, runnable validation gate per workspace `CLAUDE.md`. No phase proceeds until its gate passes.

## Why dry-run by default

Defense-in-depth for a tool that modifies a production S3 bucket. Even if the script has a bug, `--dry-run` (the default) prints what would change without writing. To actually mutate, the operator must pass `--write` explicitly.

## Out of scope for this PR

This PR is docs-only. The implementation (Phase 2 onwards) lands in a separate PR after this one is approved.

### DIFF
--- a/docs/plans/2026-05-11-llmo-config-repair.md
+++ b/docs/plans/2026-05-11-llmo-config-repair.md
@@ -1,0 +1,274 @@
+# LLMO Config Repair — Step 3b/3c of SITES-43238
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Repair the 22 LLMO configs in `s3://spacecat-prod-importer/config/llmo/<siteId>/lmmo-config.json` that fail the published Zod schema. Bring the bucket to 100% schema-valid. Step 3b is the canonical-reproducer repair (one site), step 3c is the batch over the remaining 21.
+
+**Architecture:** A one-off operational script `scripts/llmo-config-repair.js` in `spacecat-audit-worker`, sibling to `llmo-config-sweep.js`. Pulls a config, derives the target region per category from existing valid signals in the same config (majority rule), rewrites the bad fields, validates with `schemas.llmoConfig.safeParse`, writes back via `PutObjectCommand`. Supports `--dry-run`, `--site <id>`, and a hardcoded allowlist of sites pre-classified as "auto-repairable" vs "needs manual review".
+
+**Tech Stack:** Node.js ESM (matching the rest of `scripts/`), `@aws-sdk/client-s3`, `@adobe/spacecat-shared-utils` (`schemas.llmoConfig` for validation). No new dependencies.
+
+**Context**
+
+- Step 3a (sweep) ran 2026-05-11 against prod. 7,481 configs scanned, 22 schema-invalid (0.29%).
+- All 22 corrupt configs were written between 2025-11-12 and 2026-04-30, i.e. **before** step 1 (writer-side filter, merged 2026-05-04) and step 2 (fail-closed writeConfig, merged 2026-05-04). No new corruption since the fix.
+- The bad data is **not** what the original SITES-43238 description assumed. Instead of malformed region codes (`"en-us"`, `"global"`), DRS was emitting **prompt-intent / search-stage classifiers** as `region` values: `"comparative"`, `"instructional"`, `"informational"`, `"comparison & decision"`, `"discovery & research"`, `"usage & troubleshooting"`, `"scheduling & appointments"`, `"online resources & accessibility"`, `"unknown"`.
+
+---
+
+## File Map
+
+**New files:**
+
+- `scripts/llmo-config-repair.js` — the repair script, modeled on `scripts/llmo-config-sweep.js`.
+
+**Out of scope:**
+
+- Repair of `customerConfigV2` (different schema, separate concern).
+- Any DRS-side change. That belongs in the upstream DRS team's tracker (separate followup ticket).
+
+All paths below are relative to `spacecat-audit-worker/` unless stated otherwise.
+
+---
+
+## Phase 1: Per-site classification — DONE (data captured here)
+
+Each of the 22 affected sites was inspected against the sweep report and the live S3 config. Classification per site:
+
+| Bucket | Count | Repair logic |
+|---|---|---|
+| `auto-us-majority` | 17 | Every bad region field gets replaced with `"us"`. Justified by ≥1 other valid category on the same site already having `region: "us"`. |
+| `auto-au-majority` | 1 | Same logic, target region `"au"`. |
+| `manual-no-signal` | 2 | All categories on the site have bad regions. No internal signal for inference. Operator decides target. |
+| `manual-mixed-mislabel` | 2 | German-language site mislabeled with `"us"` / `"unknown"` on some fields. Operator decides whether to repair to `"de"` or follow existing (incorrect) `"us"`. |
+
+### Site-by-site assignment
+
+**auto-us-majority** (17 sites — categories with bad regions get `"us"`, prompt regions get the same):
+
+| siteId | bad fields |
+|---|---|
+| `08544afb-4ee4-46f4-bdfa-4e5864efeb33` | 2 categories with `"comparison & decision"` |
+| `205e88e0-8610-4f92-b851-4708bc8166b4` | 1 category named "US" with `"instructional"` |
+| `291e27ca-b0d3-4474-be01-38e0488cc9fc` | 1 category named "US" with `"instructional"` |
+| `5956f693-7bd8-4cc6-a499-877b9c99e083` | 2 categories with `"comparison & decision"` |
+| `624509a5-114f-4864-8b39-2827df8e2e0e` | 1 category named "US" with `["informational","instructional"]` |
+| `646eda46-1bf5-4dde-8836-d7e60e76bb3c` | 1 category named "US" with `"usage & troubleshooting"` |
+| `6ddccace-dcb4-449c-804c-3d894041fad6` | 1 category named "US" with `"informational"` |
+| `6fc5ec43-e358-46a5-8c22-829c3790ad09` | 1 category with `["scheduling & appointments","online resources & accessibility"]` |
+| `78d59744-e06c-4d14-a77a-9490c1464116` | **canonical reproducer**; 1 category named "US" with `["comparative","instructional","informational"]`. 70 issues total when counting nested prompt regions. |
+| `825ce5e4-8afa-4a68-b816-623585b995b9` | 1 category named "US" with `"informational"` |
+| `90644f8c-ab6e-4be6-9275-d7418d0b3a93` | 1 category with mixed array `["comparison & decision","us"]` — drop bad, keep `"us"` |
+| `a4828e25-3901-41d7-8d55-3aa419c496bf` | 1 category with `"comparison & decision"` |
+| `aa557324-bfa3-4606-bbb7-022492dc3305` | 1 mixed array `["us","usage & troubleshooting"]` + 1 `"comparison & decision"` |
+| `b84b60c7-35e0-4e6b-ae6e-826118fa43ef` | 1 category with `"comparison & decision"` |
+| `c0e4bcb0-738f-4380-ab54-a70ad05ac351` | 4 categories with `"usage & troubleshooting"` |
+| `e2b58651-55cc-4a28-882a-6643f21466fe` | 1 category named "US" with `"informational"` |
+| `e3adc80e-b337-4db8-8272-29691dbb61df` | 1 category named "US" with `["informational","comparative","instructional"]` |
+| `ea353752-af8b-43ad-8e20-348e6a94eb80` | 1 category named "US" with `"instructional"` |
+
+**auto-au-majority** (1 site — Australian, all categories have `region: "au"`, bad regions only in prompt `regions[]` arrays):
+
+| siteId | bad fields |
+|---|---|
+| `8b65ba43-1f17-43aa-bf7d-51c571463f85` | bad regions on `aiTopics.*.prompts.*.regions` only; categories are all clean `"au"` |
+
+**manual-no-signal** (2 sites — every category has a bad region, no internal "us" anchor):
+
+| siteId | situation |
+|---|---|
+| `11692c48-6970-419e-819e-15839880cb1b` | 10 categories, all with classifier strings as region. Site is Indeed-related; likely "us" but should be operator-confirmed. |
+| `4ee18233-4cc9-4550-b5f3-91f815b2e013` | German category names ("Produkte", "Dienstleistungen") but `region: "us"` on the German ones and `region: null` on the English ones. Likely "de"; needs operator confirmation. |
+
+**manual-mixed-mislabel** (2 sites — language/region mismatch suggests pre-existing data-quality issue, not just step-2 fallout):
+
+| siteId | situation |
+|---|---|
+| `bd2cbfb7-9137-4f0d-a57f-754cf90c5163` | German categories ("Produkte", "Dienstleistungen") with `region: "unknown"`. Repair target probably "de"; needs operator confirmation. |
+| `4ee18233-...` | listed in `manual-no-signal`; the mixed labels make it fit either bucket. Treat as manual. |
+
+**Validation gate (Phase 1):** PASSED.
+
+- 22 sites accounted for. 18 in auto-repair buckets, 4 routed to manual review (the table double-counts `4ee18233` deliberately; one of the manual buckets is the source of truth).
+- The bad-region values across all 22 sites are a closed set: `"comparative"`, `"instructional"`, `"informational"`, `"comparison & decision"`, `"discovery & research"`, `"usage & troubleshooting"`, `"scheduling & appointments"`, `"online resources & accessibility"`, `"unknown"`. The repair logic recognizes only these as "bad to replace" — anything else triggers a manual-review halt for safety.
+
+---
+
+## Phase 2: Repair script implementation
+
+**Purpose:** Build the tool that does the per-site repair. Script is the implementation of Phase 1's logic, with dry-run by default for safety.
+
+### Task 2.1: Build `scripts/llmo-config-repair.js`
+
+- [ ] CLI flags:
+  - `--bucket <name>` (defaults to `S3_BUCKET_NAME`, required if env missing)
+  - `--site <siteId>` repair a single site
+  - `--all` repair every site listed in the auto-repair allowlist
+  - `--region <xx>` override the default `"us"` target (for `8b65ba43` use `--region au`)
+  - `--dry-run` (default `true`): print before/after diff to stdout, do not write
+  - `--write` (or `--no-dry-run`): actually `PutObject`. Mutually exclusive with `--dry-run`.
+  - `--backup-dir <path>` (default `/tmp/llmo-config-repair-backups/`): always save the pre-repair config locally before writing.
+  - `--concurrency` and `--limit` reused from sweep style.
+
+- [ ] Hardcoded `AUTO_REPAIR_ALLOWLIST` constant in the script: the 18 siteIds from the `auto-us-majority` + `auto-au-majority` buckets in Phase 1, each with its target region. `--all` only processes sites in the allowlist; sites outside the allowlist require explicit `--site <id>` AND `--region <xx>`.
+
+- [ ] Bad-region set: hardcoded `BAD_REGION_VALUES` Set of the 9 known classifier strings. Anything not in this set and not matching the alpha-2 pattern triggers a hard error in repair (operator must extend the set deliberately).
+
+- [ ] Repair logic per config:
+  - Walk `categories.<uuid>.region`: if string, drop iff in `BAD_REGION_VALUES`, replace with target. If array, filter out `BAD_REGION_VALUES` entries, replace empty with `[target]`, dedupe.
+  - Walk `aiTopics.<uuid>.prompts[N].regions`: filter `BAD_REGION_VALUES`, dedupe. If empty, set to `[target]`.
+  - Same walk on `topics.<uuid>.prompts[N].regions` for symmetry (sweep showed this path isn't currently affected, but cover it).
+
+- [ ] After in-memory repair: `schemas.llmoConfig.safeParse(repaired)`. If it still fails, log issues, write nothing, exit 2 (operator must investigate).
+
+- [ ] On success in write mode:
+  - Write the backup to `<backup-dir>/<siteId>-<timestamp>-before.json`.
+  - `PutObjectCommand` the repaired config to S3.
+  - Re-read with `readConfig` (which fails closed since step 2) and confirm parse succeeds.
+  - Print `repaired siteId=<id> versionBefore=<...> versionAfter=<...>`.
+
+- [ ] Exit 0 only when every requested site repaired and re-validated. Any single failure exits non-zero.
+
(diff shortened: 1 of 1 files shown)


solaris007 (commented):
Hey @dzehnder,

## Strengths

- Defense in depth on writes: dry-run default, in-memory `safeParse` before `PutObject`, post-write re-read validation, local backup, and S3 versioning are five independent backstops. The right posture for a destructive one-off.
- Closed-set `BAD_REGION_VALUES`: refusing unknown bad values rather than guessing avoids the "looks weird, normalize it" trap. The hardcoded allowlist + closed bad-value set is the correct shape for hand-curated prod-data ops.
- Hardcoded allowlist with explicit override path: `--all` only touches sites in `AUTO_REPAIR_ALLOWLIST`; non-allowlisted sites require both `--site` AND `--region`. Good guardrail against blind batch reuse later.
- Concrete validation gates per phase. Phase 3's HTTP 200 on the canonical SITES-43238 reproducer + 24h passive watch is the right primitive: the verifier is independent of the mutator.
- Honest pivot from the original ticket. The plan calls out that SITES-43238's filed assumption (malformed region codes) was wrong and the real bad data is classifier strings. That correction is in the Context, not buried.
- Evidence-grounded Phase 1: per-site classification is concrete data, not vague intent.
- Risk analysis is operationally sharp. The concurrent-DRS-write reasoning - that step 1 makes a DRS write produce a clean config either way - shows the author understands the system in motion.

## Issues

### Important (Should Fix)

**1. Location: the runbook is correctly placed; the incident narrative is not.** (`docs/plans/2026-05-11-llmo-config-repair.md`, whole file)

This plan is two documents fused into one: (a) an operational runbook for `scripts/llmo-config-repair.js`, which is genuinely repo-local and correctly placed here, and (b) a cross-system incident narrative covering DRS's classifier-string emission, the spacecat-shared schema, the writer-side filter, fail-closed `writeConfig`, the api-service GET reproducer, and the semantic-release missed major bump on PR 1574. That story spans multiple services.

The Jira ticket (SITES-43238) itself stated the implementation plan would live at `docs/plans/2026-04-17-llmo-config-region-guard-and-repair.md` in the mysticat-workspace repo - a workspace-level location, not audit-worker. That doc was never created (steps 1 and 2 merged without it), and this PR puts the step 3b/3c plan in audit-worker's `docs/plans/` instead. The original Jira intent was for the unified plan to live above any single sub-repo.

By the letter of `mysticat-architecture/DOCUMENTATION-GUIDE.md`, repo-specific implementation plans belong in the repo, so the runbook is correctly placed here. But spacecat-audit-worker is on a deprecation path (audits migrating to mystique); when this repo is archived, the cross-system narrative disappears with it. The Context section, the BAD_REGION_VALUES inventory, the semantic-release lesson, the DRS root-cause story - these are platform-team artifacts that should outlive this repo.

Fix: keep the runbook here. Either (a) trim the cross-system content from this doc and commit to capturing it in `mysticat-architecture/` as an explicit followup, or (b) hold this PR until the incident-narrative companion lands in mysticat-architecture. Splitting is preferable to relocating because the operational steps belong near the code they describe.

**2. Phase 6 Task 6.2 conflates a one-shot artifact with a reusable primitive.** (Phase 6 Task 6.2)

`scripts/llmo-config-sweep.js` is a read-only schema-conformance scanner over an S3 prefix with concurrency control and structured output. That primitive is broadly useful: any future schema drift on any S3-backed config will want exactly this tool. Carrying cost is near-zero. The "platform-level fixes prevent recurrence" argument applies to this specific corruption class, not to all future S3-config schema drift.

`scripts/llmo-config-repair.js` is genuinely single-use because of its hardcoded allowlist and bad-value set - deleting it after use is correct.

Fix: keep the sweep, delete only the repair. If audit-worker is on a real deprecation path, promote the sweep to a tooling location before audit-worker is archived.

**3. Phase 6 Task 6.3 DRS sibling ticket has no coordination plan.** (Phase 6 Task 6.3)

The plan correctly identifies that the writer-side filter (step 1) stops bad data at the spacecat-shared boundary, but the upstream emitter (DRS/Mystique) is not fixed. "File DRS sibling ticket" has no named owner, no Jira target, no agreement with the DRS team, no steady-state position on the filter.

Architecturally: is the writer-side filter a permanent guard, or a temporary measure pending upstream cleanup? If permanent, the cost (every reader/writer round-trip pays a filter price forever) should be accepted explicitly. If temporary, the DRS fix needs an owner and a target date.

Fix: before merging, either name the DRS sibling ticket and the responsible team, or state the steady-state intent. Either is defensible; absence of a position is not.

**4. `--dry-run` vs `--write` interaction is underspecified.** (Task 2.1 flag list)

The plan declares `--dry-run` defaults to `true` and `--write` is "mutually exclusive with `--dry-run`," but does not say what happens if both are passed, or how this is implemented with `node:util parseArgs` (which has no native mutual-exclusion). For a script whose safety story rests on dry-run-by-default, this needs a one-line decision: is passing both an error (exit 2), does `--write` win, or does the plan drop `--write` and rely on `--no-dry-run`? "Either flag works" semantics shipped in operator scripts have caused real prod mutations.

**5. Batch failure policy is unspecified.** (Phase 4 Task 4.1)

`--all` across 17 sites: does a mid-batch failure abort the remaining sites or continue? "Any single failure exits non-zero" does not answer this. Fail-fast and continue-on-error are both defensible but have very different operator ergonomics. Decide explicitly and log the policy in `--help`. If continue-on-error: tie the exit-code-non-zero contract to a tally printed at the end, not the first failure.

**6. `--region` vs allowlist precedence needs to be explicit.** (Task 2.1 flag list)

If an operator runs `--site <allowlisted-au-site> --region us`, does the explicit flag win (operator override) or does the allowlist win (curated truth)? The plan's example (`8b65ba43-...` AU site, `--region au`) is ambiguous because that site is in the allowlist with target `"au"`. Default to "allowlist wins; `--region` is only consulted for sites NOT in the allowlist," or document an alternative explicitly.

**7. Idempotency claim needs a concrete contract.** (Phase 4 Task 4.1)

"A repair on an already-valid config is a no-op and prints a no-diff message" requires the script to deep-equal the repaired object against the original before `PutObject`. If the script instead does "always `PutObject`," every re-run pollutes S3 version history. Spell out the skip-when-unchanged check.

### Minor (Nice to Have)

**8. Phase 4: no review gate between Task 4.1 (US batch) and Task 4.2 (AU site).** The rest of the plan correctly enforces "stop and review" discipline (Phase 3 is one site, Phase 5 is per-site). Adding an explicit gate after 4.1 before 4.2 is consistent with that posture.

**9. Concurrency for 22 writes is over-engineering.** (Task 2.1, `--concurrency` flag) Sweep defaults to 20 because it does 7,481 reads. Repair touches at most 22 objects. Parallelism here adds interleaved log output for the operator watching `--write` runs without wall-time benefit. Recommend `--concurrency 1` default for repair.

**10. `--dry-run` diff format is unspecified.** (Task 2.1, `--dry-run` description) "Print before/after diff to stdout" - full JSON, unified text diff, or only changed fields? For an operator reviewing 22 outputs by eye, "only the fields that change, grouped by site, with the path key" is the only readable option.

**11. Backup-write failure semantics are ambiguous.** (Task 2.1 `--backup-dir` + Risks item 2) If the backup write fails (permission, disk full, dir missing), does the script abort before `PutObject` or proceed? Risks item 2 calls the local backup "convenience," suggesting proceed - but then "always save the pre-repair config locally before writing" is misleading. Lean prerequisite (hard abort on backup-write failure).

**12. `4ee18233-...` is classified into two manual buckets.** (Phase 1 site-by-site assignment) The note "the table double-counts deliberately; one of the manual buckets is the source of truth" leaves the operator following Phase 5 wondering which bucket is authoritative. Pick one and footnote the alternate.

## Recommendations

- **Add `--site-list <file>` flag.** Hardcoded `AUTO_REPAIR_ALLOWLIST` is fine for this run, but Phase 5 will produce 4 more `--site --region` invocations. A JSON/CSV of `{siteId, targetRegion}` pairs would preserve a textual record of what was actually applied per site (good for the SITES-43238 closing comment) and avoid the hardcoded-list pattern that future copy-paste reuses tend to forget to update.
- **Log per-site decisions in a structured trailer.** At the end of an `--all` run, print one structured line per siteId: `{siteId, status: ok|skipped|failed, versionBefore, versionAfter, changedPaths: [...]}`. This becomes the artifact you attach to SITES-43238 without rerunning the sweep.
- **Note WHY `readConfig` is used for the post-write re-read.** It exercises the same code path the api-service uses, so a green re-read mirrors what callers will see. Without that justification, a future reader will reasonably ask "why not just `GetObject` + `safeParse`?" and "simplify" it.
- **Designate a home for future LLMO/SpaceCat incidents now.** The mysticat-architecture docs hierarchy work is the natural place. Naming a permanent location now prevents re-debating it next time.

- Phase 6.3 second followup (semantic-release didn't bump to major on the BREAKING CHANGE footer for PR 1574). The plan correctly identifies this as a separate problem and the right place to file the followup; not the right place to fix it.

## Assessment

Ready to merge? With fixes.

Reasoning: The plan's bones are sound - closed-set bad-value detection, defense-in-depth on writes, runnable validation gates, evidence-grounded Phase 1, and an honest pivot from the original Jira ticket's wrong assumption (the Jira described malformed region codes like "en-us"/"global", but the actual bad data is classifier strings). The PR correctly addresses the Jira's Step 3 (one-shot repair) for the canonical reproducer (site 78d59744) and the bucket-wide cleanup. The Important findings split between two themes: (a) spec gaps the implementer needs nailed down (dry-run/write semantics, batch policy, region precedence, idempotency contract) and (b) doc-lifecycle decisions (where the cross-system incident narrative lives, sweep retention, DRS coordination). Both classes are tighten-the-plan work, not redesign.

## Next Steps

1. Decide on the dry-run/write semantics, batch failure policy, `--region` precedence, and idempotency contract (Important items 4-7). One sentence each in the plan unblocks the implementation PR.
2. Pick a position on sweep retention (Important item 2) and DRS coordination (Important item 3).
3. Resolve the doc-lifecycle question (Important item 1): trim cross-system content with a followup to mysticat-architecture, or hold this PR pending that companion.
4. Address the Minor items as appropriate.

dzehnder:
Thanks @solaris007 — really useful review, especially the doc-lifecycle framing. Pushed [6ee31b9](https://github.com/adobe/spacecat-audit-worker/commit/6ee31b99). All seven Important items applied, plus all Minor items and most recommendations.

## Important — applied

1. **Doc location (runbook vs incident narrative split).** Trimmed the cross-system content from this doc. The Context section now keeps only what an operator needs to drive the repair tool. The DRS root cause, schema asymmetry that step 2 closed, semantic-release BREAKING-CHANGE-footer behavior, and the sweep tool's broader applicability all move to a new **Phase 6.4** followup PR against `mysticat-architecture/` (home TBD between `platform/decisions/` and `products/llmo/`). The runbook stays here next to the code it describes.

2. **Phase 6.2 — keep the sweep, delete only the repair.** Agreed. Sweep is a reusable schema-conformance scanner with near-zero carrying cost; repair is genuinely single-use (hardcoded allowlist + closed-set). Updated the plan to reflect this, with a note that if audit-worker is later archived as part of the mystique migration, the sweep gets promoted to a tooling location rather than deleted.

3. **Phase 6.3 — DRS coordination steady-state.** Declared the writer-side filter as a **permanent defense-in-depth guard**, not a temporary measure pending the DRS fix. The filter caps the trust placed in any single producer; cost is one `Set.has` per region value. The DRS sibling ticket will improve upstream data quality but does not replace the filter. Task now also requires a named DRS-team owner before the implementation PR merges.

4. **`--dry-run` vs `--write`** — declared mutually exclusive, passing both is **hard error exit 2**, default is `--dry-run`. Implemented in code post-`parseArgs` (which has no native mutual-exclusion) via an explicit assert.

5. **Batch failure policy** — **continue-on-error** with a structured trailer line per site. Final exit code is non-zero if any site errored. Rationale: a 17-site batch with one mid-run failure should still produce a complete report so the operator sees the whole picture.

6. **`--region` vs allowlist precedence** — **allowlist wins**. For a site already in the allowlist with a target region, `--region` is ignored (with a log line saying so). `--region` is only consulted for sites NOT in the allowlist. The `8b65ba43-...` Phase 4.2 invocation drops the `--region au` flag accordingly since it is in the allowlist.

7. **Idempotency contract** — deep-equal the repaired object against the original before `PutObject`. If unchanged, skip the write entirely and log `status=unchanged`. Re-runs do not pollute S3 version history.

## Minor — applied

8. **Review gate between Task 4.1 and Task 4.2** added explicitly, with rationale: the AU repair uses a different target region and deserves a deliberate decision-point rather than being folded into the `--all` run.

9. **Concurrency default 1** for repair (was unspecified). Serial output is easier to review for both `--dry-run` and `--write`.

10. **`--dry-run` diff format** — only the fields that change, grouped per site, as `path: <before> -> <after>`. Full-JSON dumps are excluded.

11. **Backup-write failure** — hard abort with exit 2 before any `PutObject`. The local backup is a prerequisite, not convenience. Updated Risks item 2 to match (S3 versioning is the post-write safety net; local backup is the safety net for the write attempt itself).

12. **`4ee18233-...` double-counting** — moved to `manual-no-signal` as the source of truth. `manual-mixed-mislabel` now has one member (`bd2cbfb7-...`). Total still 4 manual-review sites.

## Recommendations — applied

- **`--site-list <file>` flag** added. JSON of `[{siteId, targetRegion}]`. Doubles as the textual record attached to the SITES-43238 closing comment.
- **Structured per-site trailer log** added: `{siteId, status: ok|unchanged|dry-run|error, versionBefore, versionAfter, changedPaths}`. This is the artifact captured at the end of Phase 4.
- **Justification for using `readConfig` for post-write re-read** added inline in Task 2.1: it exercises the same code path the api-service uses, so a green re-read mirrors what callers will see.
- **Home for future LLMO/SpaceCat incidents** named: the postmortem followup in Phase 6.4 establishes `mysticat-architecture/` as the destination.

## Not applied — flagged

- The semantic-release behavior on PR 1574 stays in scope as a footnote of the Phase 6.4 postmortem (your "out-of-scope worth tracking" point). Not in this PR; not lost.

Ready for re-review.

solaris007 (changes requested):
Hey @dzehnder,

Thanks for the thorough rework. All seven Important items from the prior round are addressed cleanly, and the Phase 6 split (runbook here, cross-system narrative in mysticat-architecture) is the right boundary. Two new Important items, four Minor, and a few recommendations below.

## Strengths

- **Idempotency contract via deep-equal before PutObject** (Task 2.1, line 132-133). "Skip the write entirely and log status=unchanged" keeps re-runs safe by construction and S3 version history clean. The contract states the invariant AND the operational consequence.
- **Allowlist-wins precedence with the override-ignored log line** (Task 2.1, `--region` bullet). Visible precedence resolution beats silent precedence; the trailer artifact will surface it.
- **Backup-as-hard-prerequisite framing** (Task 2.1, `--backup-dir` bullet). "S3 versioning is the safety net for a successful write; the local backup is the safety net for the write attempt itself" is the right mental model for two independent failure modes.
- **Phase 6.3 steady-state intent** (lines 253-256). Declaring the writer-side filter a permanent defense-in-depth guard, with the cost argument and the named-DRS-owner gate, settles a decision that would otherwise drift forever.
- **Phase 6.4 doc-lifetime split** (lines 261-265). Runbook lives with the code, cross-system narrative goes to mysticat-architecture. The rationale "co-locating with the repair runbook ties its lifetime to the runbook's, which is wrong" captures the failure mode that drops platform narratives when disposable tools get deleted.
- **Mitigation stack layered with no single point of failure** (Risks item 4). Five independent backstops, and the deliberate choice to use `readConfig` (not `GetObject + safeParse`) for post-write verification so it exercises the api-service code path.

## Issues

### Important (Should Fix)

**1. Phase 1 site counts are internally inconsistent and will fail the Phase 4 validation gate as written.**

The auto-us-majority table at lines 52-69 has 18 rows, but the section header at line 48 says "17 sites". The manual buckets (lines 77, 84) total 2 + 1 = 3 sites, but the Phase 1 validation gate at line 94 says "4 manual review" with a parenthetical that reads like residue from the pre-fix double-counting of 4ee18233. Phase 4 Task 4.1's title at line 190 says "17 auto-us-majority sites (minus 78d59744...)" - implying 18 total, matching the table. Phase 4's validation gate at line 213 says "exactly 4 schema-invalid configs ... 4 `manual-*` siteIds"; Phase 5's header at line 220 says "4 remaining sites" but Phase 5 Task 5.1 lists only 3.

Why it matters: the implementer building `AUTO_REPAIR_ALLOWLIST` will follow the table rows (18 + 1 = 19 sites), but the Phase 4 validation gate expects `schemaInvalid` to drop to 4. After the batch the sweep will report 3 schema-invalid configs and the gate will fail. The operator is at the most stressful moment in the runbook - between a prod batch repair and the manual sweep - left wondering whether the math is wrong or whether a 4th site silently broke.

Fix: pick one source of truth and propagate. Either (a) the table is correct: auto-us=18, auto-au=1, manual=3 (2 no-signal + 1 mixed-mislabel) - update lines 48, 94, 190, 208, 213, 220 to reflect this. Or (b) a 4th manual site is genuinely missing from the tables and Phase 5 Task 5.1 - add it with rationale.

**2. No open-loop verification before `--write`.**

The 22-site inventory was captured 2026-05-11. The plan validates the post-repair state via re-sweep in Phase 4 Task 4.3 (closed-loop), but does not require re-running the sweep immediately before Phase 3 Task 3.1 (open-loop). The implementation PR may land days or weeks after this plan is approved.

Why it matters: if a new schema-invalid config appears outside the allowlist between plan approval and execution, the operator would not notice until after the batch ran. The check is cheap (sweep is read-only) and would catch a step-1 regression that the plan otherwise has no way to detect at execution time.

Fix: add a precondition to Phase 3 Task 3.1: "Re-run `scripts/llmo-config-sweep.js` immediately before `--write`. Validation gate: schema-invalid count matches Phase 1 and affected siteIds match Phase 1's tables exactly. Any new siteId outside the allowlist halts the run as a step-1 regression."

### Minor (Nice to Have)

**3. Task 2.1 safeParse-failure wording "exit 2" conflicts with continue-on-error semantics.**

Line 124 says "If it still fails, log issues, write nothing, exit 2 (operator must investigate)." Two bullets later the batch policy is continue-on-error. A careful reader resolves the apparent conflict, but an implementer might literally `process.exit(2)` inside the per-site handler and break batch behavior.

Fix: reword to "If it still fails, log issues, write nothing, record `status=error` in the trailer. In single-site mode the script exits with code 2; in `--all` mode it continues to the next site and exits 2 at the end."

**4. Structured-trailer schema is under-specified for non-ok statuses.**

Line 128 shows only the `ok` shape. Operators parsing the trailer for audit need the full enum (`ok | unchanged | dry-run | error`) and per-status field presence. For `unchanged`, is `versionAfter` populated equal to `versionBefore`, or omitted? For `dry-run`, what holds the place of `versionAfter`? For `error`, what fields are guaranteed?

Fix: short "Trailer schema" subsection listing the four status values and the fields each carries.

**5. `--site-list` interaction with `--site` is unstated.**

In `--site abc --site-list mylist.json --write`, where does `targetRegion` come from? Phase 4.2 implies allowlist-lookup for the hardcoded case but does not state it for the file-based case.

Fix: one sentence in the `--site` bullet: "When `--site` is used, `targetRegion` is looked up in the effective allowlist (hardcoded or `--site-list`); if the site is not in the allowlist, `--region` must be passed explicitly."

**6. Writer-side filter migration path is unstated.**

Phase 6.3 declares the filter steady-state, but it lives in `src/drs-prompt-generation/drs-config-writer.js` on a deprecation path. The plan does not say whether the filter migrates with the DRS prompt-generation code path or gets promoted to `@adobe/spacecat-shared-utils` alongside the fail-closed read/write boundary.

Fix: one sentence in Phase 6.3. Default (migrates with the code path) is fine, but worth stating either way so the migration does not drop the principle.

## Recommendations

- **Specify the deep-equal implementation.** Key-order sensitivity matters: the script reads JSON (no inherent key order), repairs in memory (preserves traversal order), writes back. A naive `JSON.stringify` compare could flag re-runs as changed purely from key ordering and bypass the idempotency contract. Commit to a structural deep-equal or state the round-trip-stability assumption explicitly.
- **Reference the exact line in `drs-config-writer.js`** when filing the DRS sibling ticket (Phase 6.3). The DRS team will want to point at the artifact when they reproduce the filter.
- **Producer-side vs shared-utils-boundary filter rationale** belongs in the Phase 6.4 postmortem, not this plan. The layered design (producer scrubs with visible WARN, boundary refuses unconditionally) is the kind of decision that gets re-litigated; capturing it once prevents the next round.

- The audit-worker deprecation timeline. The plan handles it gracefully (Phase 6.4 split, sweep-promotion note). How the LLMO config write path lives across the migration is a concern for the mysticat-architecture followup.

## Assessment

Ready to merge? With fixes.

Reasoning: the plan's architecture is sound and the prior round closed cleanly. The Phase 1 count reconciliation is a single-source-of-truth pass through six places in the doc, not a redesign, but it directly impacts the implementer and the Phase 4 validation gate will literally fail as written. The pre-write open-loop check is a small text addition.

## Next Steps

1. Reconcile the Phase 1 site counts (Important item 1) before merge.
2. Add the pre-write open-loop verification gate (Important item 2).
3. Minor items 3-6 are single-sentence tightening worth folding into the same commit.
4. The Recommendations and the Phase 6.4 carry-over are optional.

PR #2510, opened 2026-05-12
fix(prerender): reduce scrape error rate — SQL 404 filter, bot-block detection, skip deployed URLs

anuj-adobe:
## Problem

Error-category analysis from [LLMO-4040](https://jira.corp.adobe.com/browse/LLMO-4040) identified **311,086 scrape failures** across a representative window. The breakdown by root cause:

| Error Category | Failures | % of Total | Root Cause |
|---|---|---|---|
| HTTP 404 (Not Found) | 117,462 | **37.8%** | Agentic URL source includes URLs CDN already serves as 404; scraper queues them anyway |
| HTTP 403 (Forbidden / Bot Block) | 32,333 | **10.4%** | Domains actively blocking the scraper; no per-domain detection before submitting |
| HTTP 429 (Too Many Requests) | ~3,284 URLs / 64 domains | ~1.1% | Scraper retries immediately without respecting `Retry-After` header |
| HTTP 410 (Gone) | Small | <1% | Permanently removed pages re-queued indefinitely |
| TypeError (shadow DOM null) | 611 URLs / 103 domains | ~0.2% | Null-reference in shadow DOM expansion mid-evaluation |
| Already-deployed URLs re-scraped | Hundreds/run per deployed domain | Ongoing | URLs already deployed to the CDN submitted to scraper despite scoring 100% — for sites with stale `status.json` error entries (e.g. ricekrispies.ca: 351 of 417 covered URLs had `scrapingStatus: error`) this triggered fresh failures on every daily run |

**Total addressable by fixes 1, 3, and 4: ~48%+ of failures + ongoing per-run budget waste.**

Full analysis and fix proposals: [Reduce prerender content scraping budget](https://wiki.corp.adobe.com/spaces/AEMSites/pages/3871795321/Reduce+prerender+content+scraping+budget)

---

## Summary

This PR implements fixes 1, 3, and 4 in `spacecat-audit-worker`:

- **Fix 1 — SQL 404 filter**: Switch Athena query to `getTopAgenticLiveUrlsFromAthena` (new `top-agentic-live-urls-by-limit.sql` with `AND status = 200`) — eliminates 404 failures at source. **~117,000 failures (~38%)**
- **Fix 2 — Permanent 410 exclusion**: Covered indirectly by Fix 1 — `getTopAgenticLiveUrlsFromAthena` only returns URLs with `status=200` CDN rows. A 410 Gone URL has no 200 rows in the CDN data, so it never enters the candidate pool and is never submitted for scraping.
- **Fix 3 — Bot-block detection with sticky skip**: Two-stage detection to avoid re-scraping bot-blocked domains. (1) **Pre-scrape sticky skip**: `submitForScraping` reads `scrapeForbidden` from `status.json`; if a prior run confirmed blocking within the last 3 days, the entire batch is skipped immediately (`domainBlocked: true`) — no URLs submitted, no wasted scrape budget. Slack-triggered runs bypass this so operators can force a re-scrape. (2) **Post-scrape reactive detection**: after results arrive, if the 403 ratio exceeds the threshold, `detectBotBlocker()` from `@adobe/spacecat-shared-utils` is called; if it returns a named CDN blocker (Cloudflare, Imperva, Akamai, Fastly, CloudFront) at confidence ≥ 0.99, `scrapeForbidden: true` is written to `status.json` so the next daily run's sticky check fires. Threshold 0.99 avoids false positives from HTTP/2 stream errors, which produce the same 403 codes as real bot blocking. **~32,000 failures (~10%)**
- **Fix 4 — Skip already-deployed URLs**: Filter URLs where `isDeployedAtEdge` is set in `status.json` out of the daily scrape batch. This flag is written per-URL by the content scraper and naturally captures both directly-deployed URLs and domain-wide CDN coverage. Re-scraping these adds no value — they already score 100% in the UI, and for sites with stale error entries in `status.json` they cause fresh failures on every run.

Fix 5 (TypeError null-guard for shadow DOM) is in a separate `spacecat-content-scraper` PR. Fix 6 (429 Retry-After) is deferred.

## Context

This PR is a cherry-picked, non-refactoring extraction from #2488, which bundled the same fixes together with a larger `handler.js` decomposition (splitting into `url-selector.js`, `opportunity-utils.js`, `status-json.js`, `ai-only-handler.js`). Splitting allows the fixes to land and be validated independently before the structural refactor follows.

**Follow-up:** #2488 will be rebased on top of this PR once it merges and serves as the refactoring PR (PR2).

## Test plan
- [ ] All tests pass (`handler.test.js`, `ai-only-mode.test.js`, `query-builder.test.js`)
- [ ] Fix 1: `top-agentic-live-urls-by-limit.sql` only returns status=200 URLs
- [ ] Fix 3: `status.json` with `scrapeForbidden: true` set within 3 days skips the batch (`domainBlocked: true`) in `submitForScraping`; post-scrape high 403 ratio triggers `detectBotBlocker` at 0.99 confidence and writes `scrapeForbidden` to `status.json`
- [ ] Fix 4: URLs with `isDeployedAtEdge: true` in `status.json` are filtered out of the daily batch
- [ ] `buildStatusFilter` rejects non-integer and out-of-range status codes (SQL injection prevention)
- [ ] esmock-heavy tests have `this.timeout(5000)` to prevent CI timeouts on slower runners
- [ ] No `url-selector.js`, `status-json.js`, or `ai-only-handler.js` introduced — all changes in existing `handler.js`

## References
- [LLMO-4040](https://jira.corp.adobe.com/browse/LLMO-4040) — Scrape failure error-category analysis
- [Wiki: Reduce prerender content scraping budget](https://wiki.corp.adobe.com/spaces/AEMSites/pages/3871795321/Reduce+prerender+content+scraping+budget)
- #2488 — Original PR (fixes + refactoring); becomes PR2 after this merges


src/cdn-logs-report/utils/query-builder.js
```diff
@@ -251,9 +251,13 @@ export function buildExcludedUrlSuffixesFilter(suffixes = []) {
   return `AND NOT regexp_like(url, '${pattern}')`;
 }
 
+function buildStatusFilter(statuses) {
+  return `AND status IN (${statuses.join(', ')})`;
```
ssilare-adobe: `buildStatusFilter` calls `.join()` directly on the input array without validating that elements are integers. Today the only caller passes `[200]` (a hardcoded literal), but `createTopUrlsQueryWithLimit` is exported and `statuses` is now part of its public option contract -- any future caller can pass string values and inject arbitrary SQL into this Athena query. The `buildExcludedUrlSuffixesFilter` function directly above uses `sqlEscape`; this function skips all of that.

Fix:
```js
function buildStatusFilter(statuses) {
  const safe = statuses.map((s) => {
    const n = Number(s);
    if (!Number.isInteger(n) || n < 100 || n > 599) {
      throw new Error(`Invalid HTTP status code for SQL filter: ${s}`);
    }
    return n;
  });
  return `AND status IN (${safe.join(', ')})`;
}
```

test/audits/prerender/handler.test.js
```diff
@@ -803,7 +995,7 @@ describe('Prerender Audit', () => {
 
         const makeHandlerWithAgentic = async (agenticUrls) => esmock('../../../src/prerender/handler.js', {
```
ssilare-adobe, line 1005: **Critical:** `makeHandlerWithAgentic` mocks `getTopAgenticLiveUrlsFromAthena` but does not mock `@adobe/spacecat-shared-utils`. Every test in this `daily batching` suite that uses this helper calls the real `detectBotBlocker`, which issues a live HEAD request to `https://example.com`. This makes ~25 tests flaky in restricted-egress CI environments, adds ~5s latency per test, and if the confidence threshold is unexpectedly met, `submitForScraping` returns early with `urls: []` causing URL assertions to pass vacuously.

Fix:
```js
const makeHandlerWithAgentic = async (agenticUrls) => esmock('../../../src/prerender/handler.js', {
  '@adobe/spacecat-shared-utils': {
    detectBotBlocker: sandbox.stub().resolves({ crawlable: true, confidence: 0 }),
  },
  '../../../src/utils/agentic-urls.js': {
    getTopAgenticLiveUrlsFromAthena: sandbox.stub().resolves(agenticUrls),
  },
});
```

**Also:** Fix 2's core filtering behavior (`gonePathnames.has(normalizePathname(url))`) is never exercised in any `submitForScraping` test -- all tests stub S3 to return `NoSuchKey` so `gonePathnames` is always empty. A regression like `p.isGone` instead of `p.gone` would pass every test silently. Add at least one test that stubs `readSiteStatusJson` to return a `status.json` with a `{ url: '...', gone: true }` entry and asserts the URL is excluded from `result.urls`.

src/prerender/handler.js
```diff
@@ -809,6 +853,29 @@ export async function importTopPages(context) {
...
+  } catch (e) {
+    if (e.name !== 'NoSuchKey') {
+      log.warn(`${LOG_PREFIX} Could not read existing status.json for siteId=${siteId}: ${e.message} — starting fresh`);
+    }
+    return { existingStatus: {}, existingPages: [] };
+  }
+}
+
+/**
+ * Computes the next domain-level 403 block entry, or null if threshold not met (auto-restore).
```
ssilare-adobe: This JSDoc block is a cherry-pick artifact describing the removed `computeDomainBlock` function. The backoff algorithm it documents (2d -> 4d -> 8d) does not exist in this file. It currently sits immediately before `submitForScraping`'s own JSDoc, which creates a misleading double-comment. Delete these 4 lines.

src/prerender/handler.js
```diff
@@ -867,6 +936,22 @@ export async function submitForScraping(context) {
   // When triggered from Slack, skip agentic sources and daily batching
   const isSlackTriggered = !!(auditContext?.slackContext?.channelId);
 
+  // Proactive bot-block check before expensive URL fetches (not applicable for Slack)
+  if (!isSlackTriggered && site.getBaseURL()) {
+    const { crawlable, confidence } = await detectBotBlocker({ baseUrl: site.getBaseURL() });
```
ssilare-adobe: `detectBotBlocker` throws synchronously (`throw new Error('Invalid baseUrl')`) when `isValidUrl` returns false -- which can happen for truthy-but-malformed DB values like `"https://"` or `"example"`. The `site.getBaseURL()` truthiness check above only guards against null/empty string, so a malformed URL aborts the entire `submitForScraping` call for that site with no retry. Wrap in try/catch:

```js
try {
  const { crawlable, confidence } = await detectBotBlocker({ baseUrl: site.getBaseURL() });
  if (!crawlable && confidence >= 0.95) {
    // ... return domainBlocked result
  }
} catch (e) {
  log.warn(`${LOG_PREFIX} Bot block detection failed: ${e.message}. baseUrl=${site.getBaseURL()} - proceeding`);
}
```

src/prerender/handler.js
```diff
@@ -353,7 +397,7 @@ async function getScrapedHtmlFromS3(url, context) {
   } = context;
 
   try {
-    const bucketName = env.S3_SCRAPER_BUCKET_NAME;
+    const bucketName = env?.S3_SCRAPER_BUCKET_NAME;
```
ssilare-adobe: Changing `env.S3_SCRAPER_BUCKET_NAME` to `env?.S3_SCRAPER_BUCKET_NAME` silently produces `undefined` when `env` is misconfigured. In `readSiteStatusJson`, the downstream S3 error is caught and treated identically to `NoSuchKey` -- meaning a configuration failure is indistinguishable from a fresh site with no history. The `gone` exclusion set and all `existingPages` merge state silently vanish every run with no observable error.

If `env` is never null here (which test patterns suggest), revert to `env.S3_SCRAPER_BUCKET_NAME` to preserve the loud failure. The same applies to the 3 other `env?.S3_SCRAPER_BUCKET_NAME` occurrences added in this PR.

src/prerender/handler.js
```diff
@@ -1346,8 +1458,9 @@ export async function uploadStatusSummaryToS3(auditUrl, auditData, context) {
         scrapedAt,
         scrapeJobId: wasSubmitted
           ? (scrapeJobId || null)
-          : (existingPageMap.get(normalizePathname(result.url))?.scrapeJobId ?? null),
+          : (existingEntry?.scrapeJobId ?? null),
         ...(result.scrapeError && { scrapeError: result.scrapeError }),
+        ...((isGone || wasGone) && { gone: true }),
```
ssilare-adobe: Once `gone: true` is written, it propagates indefinitely via `wasGone` regardless of the page's actual HTTP status. A site that temporarily 410s (CMS migration, deployment incident, CDN misconfiguration) will have those URLs permanently blacklisted -- even when they return 200 in Athena data, which Fix 1 now accurately tracks. There is no TTL, no self-healing, and no operator command to clear the flag short of editing S3 directly.

Minimum viable fix -- clear `gone` when the current scrape succeeds:
```js
const isNowSuccessful = !result.error && !result.scrapeError;
...((isGone || (wasGone && !isNowSuccessful)) && { gone: true }),
```
Alternatively, add a `goneSince` timestamp and auto-expire after N days. At minimum, document the operator recovery path in CLAUDE.md.

src/prerender/handler.js
```diff
@@ -287,6 +288,49 @@ async function getRecentlyProcessedPathnames(context, siteId) {
...
+  const { dataAccess, log } = context;
+  try {
+    const opportunities = await dataAccess?.Opportunity?.allBySiteIdAndStatus?.(siteId, 'NEW') ?? [];
+    const opportunity = opportunities.find((o) => o.getType() === AUDIT_TYPE) ?? null;
+    if (!opportunity) {
+      return new Set();
+    }
+
+    const suggestions = await opportunity.getSuggestions();
+    const domainWideDeployed = !!suggestions.find((s) => {
```
ssilare-adobe: `domainWideDeployed` is set when any domain-wide suggestion has `getStatus() !== OUTDATED` plus `edgeDeployed`. This includes FIXED, APPROVED, SKIPPED, and PENDING_VALIDATION states. If a domain-wide deployment was rolled back and its suggestion moved to FIXED, `coveredByDomainWide` URLs would still be permanently excluded from scraping. This is asymmetric with the per-URL filter below which correctly checks `=== NEW`.

If FIXED/APPROVED domain-wide deployments should still suppress per-URL scraping, add a comment explaining why. If the intent is only active deployments, change to `s.getStatus() === Suggestion.STATUSES.NEW`.

test/audits/prerender/handler.test.js
```diff
@@ -6609,6 +7131,51 @@ describe('Prerender Audit', () => {
       expect(page2.scrapedAt).to.equal('2025-01-01T00:00:00.000Z'); // preserved
     });
 
+    it('should preserve gone: true from prior status when current scrape succeeds (wasGone branch)', async () => {
```
ssilare-adobe: This test verifies `wasGone` persistence, but there is no corresponding test in `submitForScraping` that verifies the `gonePathnames` exclusion actually filters URLs from `result.urls`. All `submitForScraping` tests stub S3 to return `NoSuchKey`, so `gonePathnames` is always empty. Fix 2's filtering logic is untested end-to-end.

ssilare-adobe (changes requested):
Hey @anuj-adobe,

## Strengths
- The PR description is exemplary - a data-driven breakdown of 311k failures by root cause with expected impact per fix is exactly how multi-vector changes should be justified.
- `readSiteStatusJson` extraction (`src/prerender/handler.js`) - collapsing the duplicated S3 read pattern into a single helper with consistent NoSuchKey handling is a clean refactoring step.
- `getTopAgenticLiveUrlsFromAthena` API design (`src/utils/agentic-urls.js`) - the `runTopAgenticUrlsQuery` private helper with a backward-compatible public wrapper and a new live-only export is a clean extension that does not break existing callers across other audit types.
- `isDomainBlocked` fast-path reuses `scrapeForbidden` infrastructure - routing domain-blocked sites through the existing `scrapeForbidden` opportunity pathway rather than inventing a new code path is the right call, and the `skippedReason: 'domainBlocked'` propagation through `auditContext` across SQS steps is well-executed.
- Fix 4 test coverage (`test/audits/prerender/handler.test.js`) - nine dedicated tests for `getDeployedOrCoveredPathnames` covering edgeDeployed on organic/agentic, coveredByDomainWide with/without active domain-wide, error paths, malformed URLs, and root pathname normalization is thorough for the behavioral complexity.
- New SQL CTE structure (`src/cdn-logs-report/sql/top-agentic-urls-by-status-and-limit.sql`) - selecting `status` in the filtered CTE and filtering in the aggregated CTE before GROUP BY is valid Presto SQL and consistent with the existing `top-agentic-urls-by-limit.sql` pattern.

## Issues

### Critical (Must Fix)
Also posted as inline comments on the specific lines.

**item 1 - `buildStatusFilter` constructs Athena SQL via unvalidated string interpolation**

`statuses.join(', ')` emits array elements directly into SQL with no type checking. Today the only caller passes the hardcoded literal `[200]` so there is no active injection risk. But `createTopUrlsQueryWithLimit` is exported and the `statuses` option is now part of its public contract - any future caller passing string values from an untrusted source produces exploitable Athena SQL. The `buildExcludedUrlSuffixesFilter` function immediately above it uses `sqlEscape` and regex-escaping for the same reason; `buildStatusFilter` skips all of that.

Fix:
```js
function buildStatusFilter(statuses) {
  const safe = statuses.map((s) => {
    const n = Number(s);
    if (!Number.isInteger(n) || n < 100 || n > 599) {
      throw new Error(`Invalid HTTP status code for SQL filter: ${s}`);
    }
    return n;
  });
  return `AND status IN (${safe.join(', ')})`;
}
```

**item 2 - `makeHandlerWithAgentic` causes live network calls to `detectBotBlocker` in ~25 daily-batching tests**

The helper mocks `getTopAgenticLiveUrlsFromAthena` but does not mock `@adobe/spacecat-shared-utils`. Every test in the `daily batching` describe block therefore executes the real `detectBotBlocker`, issuing a live HEAD request to `https://example.com` per invocation. This makes ~25 tests flaky in CI environments with restricted egress, adds ~5 seconds of latency per test, and - critically - if `detectBotBlocker` unexpectedly meets the confidence threshold, `submitForScraping` returns early with `urls: []`, causing URL assertions to pass vacuously while the actual filtering code never ran.

Fix (3-line change to `makeHandlerWithAgentic`):
```js
const makeHandlerWithAgentic = async (agenticUrls) => esmock('../../../src/prerender/handler.js', {
  '@adobe/spacecat-shared-utils': {
    detectBotBlocker: sandbox.stub().resolves({ crawlable: true, confidence: 0 }),
  },
  '../../../src/utils/agentic-urls.js': {
    getTopAgenticLiveUrlsFromAthena: sandbox.stub().resolves(agenticUrls),
  },
});
```

### Important (Should Fix)
Also posted as inline comments on the specific lines.

**item 3 - Orphaned JSDoc block describes a removed function**

The diff adds a JSDoc block immediately before `submitForScraping`'s own JSDoc describing a `computeDomainBlock` backoff algorithm (2d -> 4d -> 8d) that does not exist in this file. It is a cherry-pick artifact from the removed function. Delete it.

**item 4 - `gone: true` is permanent with no automated recovery path**

Once `gone: true` is written, it propagates indefinitely via the `wasGone` check regardless of future HTTP status. A site that temporarily 410s during a CMS migration or deployment incident will have those URLs silently excluded forever - even when they return 200 in Athena data (which Fix 1 now tracks accurately). There is no TTL, no self-healing, and no operator command to clear the flag short of editing S3 directly.

Minimum viable fix - clear `gone` when the current scrape succeeds:
```js
const isNowSuccessful = !result.error && !result.scrapeError;
...((isGone || (wasGone && !isNowSuccessful)) && { gone: true }),
```
Alternatively, add a `goneSince` timestamp and auto-expire after N days. At minimum, document the operator recovery path in CLAUDE.md.

**item 5 - `detectBotBlocker` call is unguarded against exceptions from malformed base URLs**

The `if (site.getBaseURL())` guard only screens for falsy values. `detectBotBlocker` throws synchronously (`throw new Error('Invalid baseUrl')`) when `isValidUrl` returns false - which can happen for truthy-but-malformed strings like `"https://"` or `"example"` plausible from DB values. This exception propagates and aborts the entire audit for that site. Wrap the call:

```js
if (!isSlackTriggered && site.getBaseURL()) {
  try {
    const { crawlable, confidence } = await detectBotBlocker({ baseUrl: site.getBaseURL() });
    if (!crawlable && confidence >= 0.95) {
      // ... return domainBlocked result
    }
  } catch (e) {
    log.warn(`${LOG_PREFIX} Bot block detection failed: ${e.message}. baseUrl=${site.getBaseURL()} - proceeding`);
  }
}
```

**item 6 - `env?.S3_SCRAPER_BUCKET_NAME` converts loud misconfiguration into silent data loss**

Changing `env.S3_SCRAPER_BUCKET_NAME` to `env?.S3_SCRAPER_BUCKET_NAME` silently produces `undefined` when `env` is misconfigured. In `readSiteStatusJson`, the downstream S3 error is caught and treated identically to `NoSuchKey` - meaning a configuration failure is indistinguishable from a fresh site with no history. The `gone` exclusion set and all `existingPages` merge state silently vanish every run with no observable error. If `env` is never null here (which test patterns suggest), revert to `env.S3_SCRAPER_BUCKET_NAME` to preserve the loud failure.

**item 7 - Fix 2 (`gonePathnames` filtering) has no test in `submitForScraping`**

All existing `submitForScraping` tests stub the S3 client to return `NoSuchKey`, so `gonePathnames` is always an empty Set and the `.filter((url) => !gonePathnames.has(...))` branch is never exercised. A regression (e.g., `p.isGone` instead of `p.gone`) would pass every test silently. With 100% coverage enforcement this should be failing CI unless exempted. Add:
- A test that stubs S3 to return a `status.json` with one `{ url: '...', gone: true }` entry and asserts it is excluded from `result.urls`
- A test that confirms a non-gone URL in the same `status.json` is not filtered

**item 8 - `domainWideDeployed` allows FIXED/RESOLVED domain-wide suggestions to activate URL exclusion**

The `domainWideDeployed` flag is set when any domain-wide suggestion has `getStatus() !== OUTDATED` plus `edgeDeployed`. This includes FIXED, APPROVED, SKIPPED, and PENDING_VALIDATION states. If a domain-wide deployment was rolled back and the suggestion transitioned to FIXED, `coveredByDomainWide` URLs would still be permanently excluded from scraping. This is asymmetric with the per-URL filter which correctly checks `=== NEW`. If the intent is that FIXED/APPROVED domain-wide deployments should still suppress per-URL scraping, document why. If the intent is only active deployments, change the check to `s.getStatus() === Suggestion.STATUSES.NEW`.

### Minor (Nice to Have)
- `src/cdn-logs-report/sql/top-agentic-urls-by-status-and-limit.sql` - name describes the mechanism rather than the invariant; `top-agentic-live-urls-by-limit.sql` would align with the function name `getTopAgenticLiveUrlsFromAthena`.
- `src/prerender/handler.js` - the comment "before expensive URL fetches" on the `detectBotBlocker` check is misleading since `getTopOrganicUrlsFromSeo` and `getIncludedURLs` already ran above it; moving the check before those calls would save work for bot-blocked domains.
- `src/prerender/handler.js` - `log?.warn?.` uses double optional chaining inconsistently with the rest of the file where `log.warn` is called without guards.
- `test/audits/prerender/handler.test.js` - `this.timeout(5000)` is added to ~15 individual tests; setting it at the `describe` block level for esmock-heavy suites would be cleaner.
- `src/prerender/handler.js` - `getScrapeJobStats`'s `isDomainBlocked` boolean control parameter could be an early-return at the callsite rather than branching inside the function.

## Recommendations
- The `gone` flag's permanent-blacklist semantics warrant either a TTL field (`goneSince` + configurable max age) or documentation in CLAUDE.md. This is an operational decision with long-term implications - the default "gone forever" behavior will silently grow a dead set of excluded URLs across site lifetimes.
- Consider caching `detectBotBlocker` results in `status.json` with a 24-hour TTL. As written it issues a live outbound HEAD request to each customer site on every daily scheduled run - unnecessary load on customer infrastructure for what is a stable signal.
- Per CLAUDE.md guidelines, the 410 permanent blacklist behavior and domain-blocked treatment are architectural decisions with long-term operational impact that warrant ADR entries.

## Assessment
Ready to merge? No / With fixes
The two Critical issues are both 3-line fixes with no logic risk. The Important items - particularly the unguarded `detectBotBlocker` throw (item 5) and the missing `gonePathnames` test (item 7) - should also be resolved before production volume reaches these code paths.

## Next Steps
1. Address the 2 critical issues (SQL validation guard, `makeHandlerWithAgentic` mock fix)
2. Review the Important items, prioritizing the `detectBotBlocker` error handling and the `gone` recovery path
3. Minor items are optional

anuj-adobe:
## Review comments addressed

**Critical**
- **SQL injection in `buildStatusFilter`** — added integer validation (100–599) before interpolating status codes into SQL; throws on invalid input.
- **`detectBotBlocker` mock missing in `makeHandlerWithAgentic`** — added mock via `@adobe/spacecat-shared-utils`; was triggering live HTTP calls in CI causing the test suite to take 13+ minutes and time out.

**Important**
- **`domainWideDeployed` condition** — changed `!== OUTDATED` to `=== NEW` so only actively-new domain-wide suggestions suppress per-URL scraping.
- **`gone` self-healing** — introduced `isNowSuccessful` flag; `gone: true` is now cleared when a previously-gone URL scrapes successfully again instead of being set permanently.
- **`detectBotBlocker` error guard** — wrapped call in `try/catch` so malformed base URLs don't abort the entire scrape job; logs a warning and proceeds.
- **`env.S3_SCRAPER_BUCKET_NAME` optional chaining** — removed `?.` so misconfigured Lambda environments fail loudly rather than silently passing `undefined`.

**Minor**
- **Orphaned JSDoc** — removed stale 4-line block describing the old `computeDomainBlock` backoff algorithm (cherry-pick artifact, no longer relevant).

**Test coverage**
- Added `gonePathnames` describe block with tests for gone-URL exclusion and non-gone URL passthrough.
- Added `detectBotBlocker` error test (confirms warn log + scraping continues).
- Updated `wasGone` test to validate self-healing (flag cleared on successful re-scrape) and added companion test for the still-failing case.

anuj-adobe:
## 410 gone URL fix removed

After further analysis, decided to **remove the 410/gone-URL handling entirely** rather than add a TTL.

**Why it doesn't pull its weight:**
- **Agentic URLs** already get 200 status URLs (Fix 1 in this PR) — 410s cannot appear here at all
- **Organic URLs** come from SEO top-pages data — a page returning 410 loses organic traffic and naturally drops out of the feed over time without any special handling
- **Error rate was <1%** — the complexity of maintaining a `gone` flag lifecycle (write, read, filter, TTL, self-heal) isn't justified

**What was removed:**
- `gonePathnames` set and filtering in `submitForScraping`
- `isGone` / `wasGone` / `isNowSuccessful` logic in `uploadStatusSummaryToS3`
- `s3Client` / `env` destructure from `submitForScraping` (no longer needed there)
- The redundant `readSiteStatusJson` call in `submitForScraping`
- All related tests

src/cdn-logs-report/utils/query-builder.js
```diff
@@ -251,9 +251,20 @@ export function buildExcludedUrlSuffixesFilter(suffixes = []) {
   return `AND NOT regexp_like(url, '${pattern}')`;
 }
 
+function buildStatusFilter(statuses) {
+  const safe = statuses.map((s) => {
+    const n = Number(s);
+    if (!Number.isInteger(n) || n < 100 || n > 599) {
```
ssilare-adobe, line 257: I think we can exclude other statuses such as 404, 400, etc.

ssilare-adobe: Discussed

ssilare-adobe (changes requested):

ssilare-adobe (changes requested):
Hey @anuj-adobe,

## Strengths

All 8 previously-flagged issues (2 Critical, 6 Important) are confirmed resolved in this revision. Specific callouts:

- All prior critical issues fixed: `buildStatusFilter` now validates each status as `Number.isInteger(n)` + range `[100, 599]` with a throw on failure (`src/cdn-logs-report/utils/query-builder.js:254`), and `makeHandlerWithAgentic` now stubs `@adobe/spacecat-shared-utils` with `detectBotBlocker: sandbox.stub().resolves({ crawlable: true, confidence: 0 })` (`test/audits/prerender/handler.test.js:1041`).
- `getDeployedOrCoveredPathnames` (`src/prerender/handler.js:300`) correctly uses `=== Suggestion.STATUSES.NEW` for the domain-wide deployed check, addressing the previous status-asymmetry concern within the new function.
- The 410 gone-URL exclusion logic is cleanly removed in full - no stale `gonePathnames`, `gone` flag, or related status.json fields remain.
- `detectBotBlocker` integration is well-structured: guarded against Slack runs, threshold at 0.99 (with rationale documented), wrapped in `try/catch` with warn-and-proceed, and the `skippedReason` signal propagates cleanly through `auditContext` between steps.
- New test coverage is thorough: 6 `detectBotBlocker` branch tests, 9 deployed/covered filtering tests, `buildStatusFilter` SQL injection and range tests, and `getTopAgenticLiveUrlsFromAthena` delegation tests.
- `runTopAgenticUrlsQuery` private/public wrapper API design is clean - `getTopAgenticUrlsFromAthena` remains backward-compatible with an added optional `statuses` param, and `getTopAgenticLiveUrlsFromAthena` is a single-line delegation (`src/utils/agentic-urls.js:132-169`).

## Issues

### Important (Should Fix)

**item 1 - Stale `/* c8 ignore next 12 */` suppresses real coverage on the `scrapeForbidden` branch**

`src/prerender/handler.js`, around the `else if (scrapeForbidden)` block.

The new `domainBlocked` test in `processContentAndGenerateOpportunities` exercises this exact branch: `isDomainBlocked=true` produces `scrapeForbidden: true` from `getScrapeJobStats`, which flows into `else if (scrapeForbidden)` and calls `createScrapeForbiddenOpportunity`. The comment predates the test and now suppresses a covered and tested branch from the coverage report. This masks coverage regressions if the test is later modified, and trains future reviewers to assume this path is uncoverable.

Fix: remove `/* c8 ignore next 12 */` before the `else if (scrapeForbidden)` block and confirm `npm test` still passes at 100% coverage.

**item 2 - Status check asymmetry between `getDeployedOrCoveredPathnames` and `getDomainWideSuggestionDeployedAtEdge` defeats Fix 4 for non-NEW deployed domain-wide suggestions**

`src/prerender/handler.js` - `getDomainWideSuggestionDeployedAtEdge` function vs. `getDeployedOrCoveredPathnames`.

`getDeployedOrCoveredPathnames` (this PR, step 2) checks `s.getStatus() === Suggestion.STATUSES.NEW` before treating a domain-wide suggestion as active. `getDomainWideSuggestionDeployedAtEdge` (step 3, pre-existing) checks `s.getStatus() !== Suggestion.STATUSES.OUTDATED` - which includes FIXED, APPROVED, SKIPPED, and PENDING_VALIDATION.

Practical consequence: if a domain-wide suggestion with `edgeDeployed` set is in FIXED state (e.g. operator closed it), step 3 still calls `markNewSuggestionsAsCovered`, writing `coveredByDomainWide` on all NEW per-URL suggestions. But step 2's `getDeployedOrCoveredPathnames` will NOT see the domain-wide deployment as active (it fails the `=== NEW` check), so those `coveredByDomainWide` URLs are still submitted for scraping - the scrape budget waste Fix 4 is designed to prevent.

`detectWrongEdgeDeployedStatus` explicitly warns about non-NEW suggestions with `edgeDeployed` set, so this state is treated as a data anomaly that "should not happen". If that invariant holds, the divergence has no runtime impact. The fix is either to align `getDomainWideSuggestionDeployedAtEdge` to `=== NEW` (stricter, matches the new function), or add a JSDoc comment explaining the intentional breadth and why it differs from `getDeployedOrCoveredPathnames`.

**item 3 - `detectWrongEdgeDeployedStatus` fires before the `isDomainBlocked` short-circuit, wasting 2 DB round-trips per blocked domain per run**

`src/prerender/handler.js:1627`.

`detectWrongEdgeDeployedStatus` calls `Opportunity.allBySiteIdAndStatus` + `opportunity.getSuggestions()` unconditionally, before the `isDomainBlocked` check that follows. When `isDomainBlocked=true`, all subsequent URL fetching and comparison is skipped - but the diagnostic has already consumed 2 DB calls. With a 200-connection pool shared across 10 ECS tasks, per-call efficiency matters for domains that trigger this path on every daily run.

Fix: guard the call with `if (!isDomainBlocked)`:

```js
if (!isDomainBlocked) {
  await detectWrongEdgeDeployedStatus(dataAccess, siteId, site.getBaseURL(), log);
}
```

The diagnostic's own comment says it "runs unconditionally so audits with no prerender findings still catch pre-existing issues" - `domainBlocked` runs are not audits with no findings, they are skipped audits, so the intent does not require unconditional execution.

### Minor (Nice to Have)

- `src/prerender/handler.js` - `getScrapeJobStats` JSDoc is missing `@param {boolean} [isDomainBlocked=false]` and `submittedUrlSet` is absent from the `@returns` type. The function is exported; stale docs for exported functions erode the contract over time.
- `test/audits/prerender/handler.test.js` - The `makeContext` helper omits `Opportunity` from `dataAccess`. `getDeployedOrCoveredPathnames` uses optional chaining so tests relying on `makeContext` silently resolve to an empty Set and skip all deployed/covered URL filtering. Adding a default `Opportunity: { allBySiteIdAndStatus: sandbox.stub().resolves([]) }` to `makeContext` would make those tests exercise the real code path.
- `src/cdn-logs-report/sql/top-agentic-urls-by-status-and-limit.sql` - name describes the mechanism rather than the invariant; `top-agentic-live-urls-by-limit.sql` would align with `getTopAgenticLiveUrlsFromAthena`. (Flagged in prior review as Minor, still unaddressed.)

## Recommendations

- The security reviewer identified two pre-existing SQL injection surfaces adjacent to the `buildStatusFilter` surface reviewed in this PR: (1) `buildSiteFilters` in `src/cdn-logs-report/utils/query-builder.js` interpolates `rootHost` into a `REGEXP_LIKE` call without sanitization - a customer domain containing an apostrophe (legal in IDNA hostnames) would break the SQL string literal; (2) the `limit` parameter in `createTopUrlsQueryWithLimit` is interpolated as `LIMIT {{limit}}` without integer validation similar to `buildStatusFilter`. Neither is introduced by this PR. They should be filed as separate follow-up issues.
- Per CLAUDE.md guidelines, the `domainBlocked` treatment as a first-class audit outcome (long-lived signal, daily re-detection, feeds into the `scrapeForbidden` opportunity pathway) has operational implications worth an ADR entry - specifically the decision to detect at confidence >= 0.99, the absence of a per-domain suppression TTL, and the rationale for routing through `scrapeForbidden` rather than a distinct opportunity type.

## Assessment

Ready to merge? No / With fixes.

Items 1-3 are all 2-5 line changes. Item 1 (remove stale c8 ignore) and item 3 (guard the diagnostic call) are mechanical. Item 2 (status check asymmetry) requires a judgment call - align `getDomainWideSuggestionDeployedAtEdge` to `=== NEW` or document the intentional difference. Once those three are addressed, this PR is ready to merge.

## Next Steps

1. Address the 3 Important issues: remove the c8 ignore comment, guard `detectWrongEdgeDeployedStatus` behind `!isDomainBlocked`, resolve the status check asymmetry.
2. The Minor items (JSDoc, `makeContext` Opportunity stub, SQL filename) are optional but worth addressing before the branch.
3. File separate issues for the two pre-existing SQL surfaces identified in the security review.

nit23uec (approved):

ssilare-adobe (approved):

PR #2517, opened 2026-05-13
feat(broken-links): entity replacement detection and parent path fallback 

jsanchez07:
Description:                                                                                                                                                                                       
  When all suggested URLs are filtered out and no prior suggestions exist,                                                                                                                           
  fall back to the nearest live parent path instead of the root domain.                                                                                                                              
                                                                                                                                                                                                     
  Also drops suggestions that substitute one person/entity page for another                                                                                                                          
  under the same parent (e.g. /contact/john-smith → /contact/jeremy-williams),                                                                                                                       
  which is semantically wrong.                                                                                                                                                                       
                                                                                                                                                                                                     
  Changes:                                                  
  - `isEntityReplacementSuggestion()` — detects sibling entity substitutions via person-name slug heuristic + known section names (contact, team, authors, etc.)                                     
  - `resolveParentPathFallback()` — walks up the broken URL's path hierarchy, returning the first parent that responds 200                                      
  - Guidance handler wires both into the suggestion pipeline                                                                                                                                         
                                                                                                                                                                                                     
  Closes [SITES-44158](https://jira.corp.adobe.com/browse/SITES-44158)


src/utils/url-utils.js
```diff
@@ -17,6 +17,151 @@ import {
...
+ * A slug segment matching one of these is not a person-name slug.
+ */
+const SLUG_STOP_WORDS = new Set([
+  'how', 'the', 'for', 'and', 'with', 'from', 'into', 'about',
+  'new', 'old', 'top', 'best', 'your', 'our', 'get', 'use', 'its',
+  'what', 'why', 'when', 'where', 'who', 'which', 'are', 'was',
+  'part', 'page', 'post', 'blog', 'news', 'tips', 'guide', 'series',
+  'intro', 'overview', 'review', 'update', 'release', 'launch',
+  'all', 'key', 'see', 'now', 'via',
+]);
```
mikeweis, line 49: For both the new `ENTITY_SECTIONS` as well as the `SLUG_STOP_WORDS` -- do non-English websites always use **English** path names?  
* Or could a German website (for example) include the German equivalent forms of each of these words?
   * If so, how should we solve this? 🤔

mikeweis (approved):
* Approved, but I did add an i18n related comment. 👀

PR #2519, opened 2026-05-15
fix: canonical / broken-backlinks / metatags - update opportunity when no issues detected

oiatsiuk:
For canonical, broken-backlinks and metatags audits: when a run finds no issues, any existing NEW opportunity is now set to RESOLVED and its NEW/PENDING_VALIDATION suggestions marked OUTDATED. Previously, these opportunities stayed open indefinitely after a clean audit run.

Suggestions in terminal states (FIXED, SKIPPED, REJECTED, APPROVED, IN_PROGRESS) are preserved, consistent with handleOutdatedSuggestions in the shared utils.

### DIFF
--- a/src/backlinks/handler.js
+++ b/src/backlinks/handler.js
@@ -14,6 +14,7 @@ import { tracingFetch as fetch, prependSchema, stripWWW } from '@adobe/spacecat-
 import SeoClient from '@adobe/mysticat-shared-seo-client';
 import {
   Audit,
+  Opportunity as OpportunityModel,
   Suggestion as SuggestionModel,
   FixEntity as FixEntityModel,
 } from '@adobe/spacecat-shared-data-access';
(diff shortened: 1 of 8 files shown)


src/backlinks/handler.js
```diff
@@ -294,6 +295,27 @@ export const generateSuggestionData = async (context) => {
...
+        .allBySiteIdAndStatus(site.getId(), OpportunityModel.STATUSES.NEW);
+      const opportunity = opportunities
+        .find((oppty) => oppty.getType() === Audit.AUDIT_TYPES.BROKEN_BACKLINKS);
+      if (opportunity) {
+        log.info(`Resolving existing broken-backlinks opportunity ${opportunity.getId()} for ${site.getId()}`);
+        await opportunity.setStatus(OpportunityModel.STATUSES.RESOLVED);
+        const suggestions = await opportunity.getSuggestions();
+        if (suggestions?.length > 0) {
+          await dataAccess.Suggestion
+            .bulkUpdateStatus(suggestions, SuggestionModel.STATUSES.OUTDATED);
```
mikeweis: 🤔  What if a suggestion was marked as `SKIPPED` or as `FIXED` <-- do we have any existing suggestion states that we would want to preserve?  Or do we **always** want to mark every suggestion as `OUTDATED`?

src/canonical/handler.js
```diff
@@ -681,6 +681,27 @@ export async function processScrapedContent(context) {
...
+      const opportunities = await dataAccess.Opportunity
+        .allBySiteIdAndStatus(site.getId(), OpportunityModel.STATUSES.NEW);
+      const opportunity = opportunities.find((oppty) => oppty.getType() === auditType);
+      if (opportunity) {
+        log.info(`[canonical] Resolving existing canonical opportunity ${opportunity.getId()} for ${baseURL}`);
+        await opportunity.setStatus(OpportunityModel.STATUSES.RESOLVED);
+        const suggestions = await opportunity.getSuggestions();
+        if (suggestions?.length > 0) {
+          await dataAccess.Suggestion
+            .bulkUpdateStatus(suggestions, SuggestionModel.STATUSES.OUTDATED);
```
mikeweis: (*same question as for the "backlinks" audit*) ...
🤔  What if a suggestion was marked as `SKIPPED` or as `FIXED` <-- do we have any existing suggestion states that we would want to preserve?  Or do we **always** want to mark every suggestion as `OUTDATED`?

src/backlinks/handler.js
oiatsiuk: thanks, fixed.

src/canonical/handler.js
oiatsiuk: thanks, fixed.

src/canonical/handler.js
```diff
@@ -681,6 +681,31 @@ export async function processScrapedContent(context) {
...
+        if (suggestionsToOutdate.length > 0) {
+          await dataAccess.Suggestion
+            .bulkUpdateStatus(suggestionsToOutdate, SuggestionModel.STATUSES.OUTDATED);
+        }
+        opportunity.setUpdatedBy('system');
+        await opportunity.save();
+      }
+    } catch (e) {
+      log.error(`[canonical] Failed to resolve opportunity for ${baseURL}: ${e.message}`);
+    }
```
noruiz: This section and the one in the backlinks handler can probably be moved to a shared function

noruiz (approved):
Approved with a small comment

src/canonical/handler.js
oiatsiuk: done.

noruiz (approved):

PR #2520, opened 2026-05-15
feat(semantic-value-visibility): use generic-autofix-edge as opportunity type

yevheniia-severinovska:
## Summary

- Changes `OPPORTUNITY_TYPE` from `'semantic-value-visibility'` to `'generic-autofix-edge'` so created opportunities are editable in the back-office and routable in elmo-ui via `GenericTokawakaPatchOpportunitySection`
- Audit trigger (`semantic-value-visibility` Slack command) and Mystique routing (`guidance:semantic-value-visibility`) are unaffected — only the stored opportunity type in SpaceCat changes

## Fixes (addressing review feedback)

- **Type collision**: added `additionalMetrics: [{ key: 'subtype', value: 'semantic-value-visibility' }]` to opportunity data as a stable per-audit discriminator. Both the stale-scan and `convertToOpportunity` now use an `isSvvOpportunity` predicate keyed on this field — prevents SVV from silently resolving or overwriting foreign `generic-autofix-edge` opportunities on the same site
- **Legacy orphans**: stale-scan treats both the old `semantic-value-visibility` type and the new `generic-autofix-edge` + subtype as SVV during a transition window — existing prod rows get resolved naturally on the next SVV run with no suggestions, no zombie rows

## Test plan

- [x] 30 SVV unit tests pass locally
- [x] Regression test: foreign `generic-autofix-edge` opp on same site is not touched
- [x] Regression test: legacy `semantic-value-visibility` opp is resolved during transition
- [x] `comparisonFn` passed to `convertToOpportunity` and validated for correct/incorrect subtype
- [x] `additionalMetrics` subtype discriminator asserted in mapper test
- [ ] Trigger audit via Slack `run audit semantic-value-visibility` — confirms audit still runs
- [ ] Verify created opportunity has `type: generic-autofix-edge` in SpaceCat API response
- [ ] Verify opportunity is editable in back-office
- [ ] Verify opportunity renders correctly in elmo-ui via `GenericTokawakaPatchOpportunitySection`

### DIFF
--- a/src/semantic-value-visibility/constants.js
+++ b/src/semantic-value-visibility/constants.js
@@ -12,4 +12,4 @@
 
 export const AUDIT_TYPE = 'semantic-value-visibility';
 export const GUIDANCE_TYPE = 'guidance:semantic-value-visibility';
-export const OPPORTUNITY_TYPE = 'semantic-value-visibility';
+export const OPPORTUNITY_TYPE = 'generic-autofix-edge';
--- a/src/semantic-value-visibility/guidance-handler.js
+++ b/src/semantic-value-visibility/guidance-handler.js
@@ -14,7 +14,15 @@ import { ok, notFound, badRequest } from '@adobe/spacecat-shared-http-utils';
 import { convertToOpportunity } from '../common/opportunity.js';
 import { syncSuggestions } from '../utils/data-access.js';
 import { createOpportunityData } from './opportunity-data-mapper.js';
-import { OPPORTUNITY_TYPE } from './constants.js';
+import { AUDIT_TYPE, OPPORTUNITY_TYPE } from './constants.js';
+
+// Matches SVV opportunities by subtype discriminator, regardless of which bucket type they carry.
+// Needed because OPPORTUNITY_TYPE ('generic-autofix-edge') is shared across multiple audits.
+const isSvvOpportunity = (oppty) => {
+  const metrics = oppty.getData?.()?.additionalMetrics;
+  const isSvvSubtype = (m) => m.key === 'subtype' && m.value === AUDIT_TYPE;
+  return Array.isArray(metrics) && metrics.some(isSvvSubtype);
+};
 
 /**
  * Guidance handler for semantic value visibility.
@@ -68,7 +76,9 @@ export default async function handler(message, context) {
     // Check if there's an existing opportunity to clean up
     const existing = await Opportunity.allBySiteIdAndStatus(siteId, 'NEW');
     const staleOpportunity = existing.find(
-      (oppty) => oppty.getType() === OPPORTUNITY_TYPE,
+      // Match new (generic-autofix-edge + subtype) and legacy (semantic-value-visibility) rows
+      (oppty) => oppty.getType() === AUDIT_TYPE
+        || (oppty.getType() === OPPORTUNITY_TYPE && isSvvOpportunity(oppty)),
     );
 
     if (staleOpportunity) {
@@ -91,6 +101,7 @@ export default async function handler(message, context) {
     createOpportunityData,
     OPPORTUNITY_TYPE,
     { guidance },
+    isSvvOpportunity,
   );
 
   if (!opportunity) {
--- a/src/semantic-value-visibility/opportunity-data-mapper.js
+++ b/src/semantic-value-visibility/opportunity-data-mapper.js
@@ -11,6 +11,7 @@
  */
 
 import { DATA_SOURCES } from '../common/constants.js';
+import { AUDIT_TYPE } from './constants.js';
 
 export function createOpportunityData(props = {}) {
   const { guidance = {} } = props;
@@ -28,6 +29,10 @@ export function createOpportunityData(props = {}) {
     tags: ['isElmo', 'content', 'edgeOptimize'],
     data: {
       dataSources: [DATA_SOURCES.SITE],
+      // Stable per-audit discriminator — required because OPPORTUNITY_TYPE is a shared bucket
+      // (generic-autofix-edge). Without this, stale-scan and upsert would match any
+      // generic-autofix-edge opportunity on the site, not just SVV ones.
+      additionalMetrics: [{ key: 'subtype', value: AUDIT_TYPE }],
     },
   };
 }
--- a/test/audits/semantic-value-visibility/guidance-handler.test.js
+++ b/test/audits/semantic-value-visibility/guidance-handler.test.js
@@ -316,7 +316,7 @@ describe('Semantic Value Visibility Guidance Handler', () => {
     it('should mark stale opportunity as RESOLVED when no new suggestions', async () => {
       const staleOpportunity = {
         getId: sinon.stub().returns('stale-oppty-123'),
-        getType: sinon.stub().returns('semantic-value-visibility'),
+        getType: sinon.stub().returns('generic-autofix-edge'),
         setStatus: sinon.stub(),
         setUpdatedBy: sinon.stub(),
         save: sinon.stub().resolves(),
--- a/test/audits/semantic-value-visibility/guidance-handler.test.js
+++ b/test/audits/semantic-value-visibility/guidance-handler.test.js
@@ -313,10 +313,11 @@ describe('Semantic Value Visibility Guidance Handler', () => {
   });
 
   describe('stale opportunity cleanup', () => {
-    it('should mark stale opportunity as RESOLVED when no new suggestions', async () => {
+    it('should mark stale SVV opportunity as RESOLVED when no new suggestions', async () => {
       const staleOpportunity = {
         getId: sinon.stub().returns('stale-oppty-123'),
         getType: sinon.stub().returns('generic-autofix-edge'),
+        getData: sinon.stub().returns({ additionalMetrics: [{ key: 'subtype', value: 'semantic-value-visibility' }] }),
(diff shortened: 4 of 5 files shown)


src/semantic-value-visibility/constants.js
```diff
@@ -12,4 +12,4 @@
 
 export const AUDIT_TYPE = 'semantic-value-visibility';
 export const GUIDANCE_TYPE = 'guidance:semantic-value-visibility';
-export const OPPORTUNITY_TYPE = 'semantic-value-visibility';
+export const OPPORTUNITY_TYPE = 'generic-autofix-edge';
```
danieljchuser, line 15: Important - two concerns on this line, see the consolidated review for full context:

1. **Type-only matching collides with the shared `generic-autofix-edge` bucket.** Both `guidance-handler.js` call sites match by type alone (stale-scan `find` and the default predicate inside `convertToOpportunity` in `src/common/opportunity.js`). The moment any other audit writes a `generic-autofix-edge` opportunity for the same `siteId` in status `NEW`, SVV will silently `RESOLVE` or overwrite that foreign opportunity. Fix: pass an SVV-specific `comparisonFn` to `convertToOpportunity` (and use the same predicate in the stale-scan), keyed on a field SVV owns (e.g. `data.origin: 'semantic-value-visibility'` or the existing `auditId` linkage). Add a regression test with one SVV and one foreign `generic-autofix-edge` opportunity on the site, asserting only the SVV one is mutated on each code path.

2. **Legacy `semantic-value-visibility` opportunities are orphaned on deploy.** Existing prod rows with the old type in status `NEW` are unreachable by both code paths after this change, leaving zombie rows and duplicate UI tiles. Either include a one-shot resolve/relabel, support both type strings during a transition window, or document the manual cutover. Add test coverage for the legacy-coexistence case.

danieljchuser (changes requested):
Hey @yevheniia-severinovska,

## Strengths

- Tight, focused diff. The `AUDIT_TYPE` / `GUIDANCE_TYPE` / `OPPORTUNITY_TYPE` factoring in `src/semantic-value-visibility/constants.js` paid off here - the bucket swap is a one-line change rather than a refactor.
- Test stub updated in lockstep at `test/audits/semantic-value-visibility/guidance-handler.test.js:319`, and codecov reports 100% patch coverage.
- PR description correctly scopes the blast radius (Slack trigger and Mystique routing unaffected), which is exactly the framing reviewers need.

## Issues

### Important (Should Fix)

**1. `src/semantic-value-visibility/constants.js:15` - type-only matching collides with the shared `generic-autofix-edge` bucket**

`OPPORTUNITY_TYPE` is now a shared bucket (already consumed by `spacecat-shared-tokowaka-client`'s generic-mapper and routed generically in elmo-ui). Both call sites in `src/semantic-value-visibility/guidance-handler.js` treat the type as a unique identity key:

1. Stale-scan: `existing.find((oppty) => oppty.getType() === OPPORTUNITY_TYPE)` followed by `setStatus('RESOLVED')` and `save()`.
2. Upsert: `convertToOpportunity(url, auditData, context, createOpportunityData, OPPORTUNITY_TYPE, { guidance })` - in `src/common/opportunity.js` this falls through to the default `(oppty) => oppty.getType() === auditType` predicate because no `comparisonFn` is passed.

Once any other audit on the same `siteId` writes a `generic-autofix-edge` opportunity in status `NEW`, SVV will either silently flip a foreign opportunity to `RESOLVED` or overwrite its `data`/`auditId` on the next run. No exception, no log signal. The current 24 SVV tests use SVV-shaped fixtures only and would not catch this.

Fix in this PR: pass an SVV-specific `comparisonFn` to `convertToOpportunity` and use the same predicate in the stale-scan - discriminating on a stable field that SVV owns (for example `data.origin: 'semantic-value-visibility'`, or the existing `auditId` linkage). Add a regression test that puts two `generic-autofix-edge` opportunities on the site - one SVV, one foreign - and asserts only the SVV one is mutated on each code path.

**2. `src/semantic-value-visibility/constants.js:15` - legacy `semantic-value-visibility` opportunities are orphaned on deploy**

Any prod opportunity created before this PR with `type === 'semantic-value-visibility'` and status `NEW` becomes unreachable after deploy: the stale-scan skips it (type mismatch), and the upsert creates a parallel `generic-autofix-edge` opportunity alongside it. Net result is zombie `NEW` rows per site and duplicate tiles in the UI until something cleans them up.

Either include a one-shot resolve/relabel for legacy rows, or have the handler treat both type strings as SVV during a transition window. At minimum, call out the cutover expectation in the PR description so ops knows to expect duplicates until cleanup runs. Cover the legacy-coexistence case with a test.

### Minor

**3. PR description - stated manual checks are mostly unchecked**

Three of four items in the test plan are unticked (Slack trigger, back-office editability, elmo-ui rendering). Those are exactly the checks that exercise the upsert and cleanup paths against real opportunity data and would surface the issues above before merge.

## Recommendations

- Add JSDoc on `OPPORTUNITY_TYPE` noting it is a shared/generic bucket and that any type-based lookup against it must include a per-audit discriminator. Saves the next audit author from copy-pasting the type-only `find` pattern.
- Add structured context to the `Removing stale opportunity` log line in `guidance-handler.js` (oppty type, origin/source if available). If a cross-source resolve ever slips through, this is what makes it greppable in Coralogix post-deploy.

`convertToOpportunity`'s default lookup contract in `spacecat-shared` should accept (or require) a discriminator when the type is a shared bucket. The disambiguation needs to land there before the third or fourth audit adopts `generic-autofix-edge`. Not fixable in this repo.

## Assessment

Ready to merge? **No, with fixes.**

Reasoning: The one-line constant flip is correct in intent (aligning with the shared `generic-autofix-edge` bucket so opportunities are editable in the back-office and routable in elmo-ui), but it silently removes the unique identifier the handler had for "its own" opportunity. Two existing site-level lookups widen from "SVV opportunities" to "anything in this generic bucket," and existing prod rows with the old type are orphaned. Add a discriminator on both lookups with a regression test, and decide on legacy-row handling, before merging.

## Next Steps

1. Add an SVV-specific discriminator and pass it through both the stale-scan predicate and a `comparisonFn` argument to `convertToOpportunity`, with a regression test.
2. Decide legacy-row handling (one-shot resolve, dual-type read window, or a documented manual cleanup) and reflect the choice in the PR.
3. Manual test-plan items remaining in the description should be ticked off once the fixes land.

danieljchuser (approved):
Hey @yevheniia-severinovska,

Re-reviewing after 3 new commits since my last pass. Both prior Important findings are substantively addressed - one fully, one with a remaining sharp edge worth a brief note.

## Strengths

- **Discriminator design is clean.** Putting the subtype on `data.additionalMetrics` in `src/semantic-value-visibility/opportunity-data-mapper.js:33` and reusing the existing `comparisonFn` parameter on `convertToOpportunity` (rather than inventing a new contract) keeps the change localized. The inline comment on the mapper explains exactly why this exists, which is the right place for that note.
- **Stale-scan predicate is correct.** `guidance-handler.js:81` (`oppty.getType() === AUDIT_TYPE || (oppty.getType() === OPPORTUNITY_TYPE && isSvvOpportunity(oppty))`) correctly disambiguates SVV from foreign `generic-autofix-edge` opportunities and still picks up legacy rows.
- **Regression tests cover the right ground.** The new "foreign generic-autofix-edge opportunity is not touched" case in `test/audits/semantic-value-visibility/guidance-handler.test.js` is exactly the test the prior review asked for, plus a legacy-coexistence case and a comparisonFn shape assertion.

## Issues

### Minor

**1. `src/semantic-value-visibility/guidance-handler.js:97` - upsert path does not migrate legacy rows**

The stale-scan correctly handles legacy `semantic-value-visibility` rows, but `convertToOpportunity` is called with `auditType = OPPORTUNITY_TYPE` (`generic-autofix-edge`), and its internal find at `src/common/opportunity.js:43` keys on `oppty.getType() === auditType` first. Legacy rows fail that match and are skipped, so the upsert creates a fresh `generic-autofix-edge` row alongside them.

The PR description's claim that legacy rows "get resolved naturally on the next SVV run with no suggestions" is technically true, but it's conditional on a zero-suggestion run happening - which may never occur for active sites that consistently produce marketing-image suggestions. In that case the legacy row persists in `NEW` indefinitely beside the new row.

Two reasonable resolutions, neither blocking this PR:

- Have the stale-scan reuse its predicate as a *filter* (not `find`) and resolve all legacy + SVV-subtype rows whenever it runs, so the first SVV run on a site - regardless of suggestion count - cleans up.
- Or run a one-shot ops cleanup outside this PR and accept that until then a small number of sites will carry a duplicate row.

Either is fine. I'd just call out the choice in the PR description so the ops side knows what to expect post-deploy.

## Recommendations

- Consider adding a small log line when the stale-scan resolves a *legacy*-typed row (vs the new-typed one), e.g. `log.info('[svv] resolved legacy opportunity ...')`. Makes the transition window observable in Coralogix and lets you confirm cleanup is happening.

## Assessment

Ready to merge? **Yes.**

Reasoning: The prior Important findings are addressed in the way the review asked for, with the right shape of test coverage. The remaining legacy-upsert sharp edge is a transition-only concern - it does not affect steady-state correctness, the stale-scan eventually catches it, and it's reasonable to handle in a follow-up or ops sweep rather than expanding this PR. The Slack/back-office/elmo-ui manual checks still in the test plan should be ticked off before merge but do not need to gate the review.

PR #2529, opened 2026-05-18
fix: summarization - use updateByKeys to persist audit result in send-to-mystique step

noruiz:
## Summary
- Related to https://adobe.enterprise.slack.com/archives/C09UYE3U48P/p1778862991328429
- `audit.setAuditResult()` does not exist on Audit model instances because the Audit schema is configured with `.allowUpdates(false)`, which prevents the base model from auto-generating setter methods
- This caused `TypeError: audit.setAuditResult is not a function` at the `send-to-mystique` step of the summarization audit
- Replaced with `dataAccess.Audit.updateByKeys()`, which bypasses the schema restriction — the same pattern already used in `image-alt-text/handler.js`

## Test plan

- [ ] All existing summarization handler tests pass (`npm run test:spec -- test/audits/summarization/handler.test.js`)
- [ ] Test assertions updated to verify `dataAccess.Audit.updateByKeys` is called with the correct audit ID and result payload

### DIFF
--- a/src/summarization/handler.js
+++ b/src/summarization/handler.js
@@ -207,7 +207,7 @@ export async function submitForScraping(context) {
  */
 export async function sendToMystique(context) {
   const {
-    site, audit, auditContext, log, sqs, env, scrapeResultPaths, s3Client,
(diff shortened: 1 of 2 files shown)


mikeweis (approved):
LGTM 👍

PR #2533, opened 2026-05-19
docs(spec): preflight links — unverifiable result category (SITES-44776)

clotton:
Adds a spec at `docs/specs/2026-05-19-preflight-links-unverifiable-category.md` proposing a third audit-result category — **unverifiable** — for preflight links that can't be probed for understandable reasons.

Tracking: [SITES-44776](https://jira.corp.adobe.com/browse/SITES-44776)

## Background

Walmart reported that preflight flags 14 corp-network-only domains (timesheet.wal-mart.com, internal.walmart.com, etc.) as broken on their authoring environment. Root cause: [PR #2476](https://github.com/adobe/spacecat-audit-worker/pull/2476) correctly classifies genuinely unreachable links as broken, but corp-network-only services hit the same network-error path. Authors see false-positive "broken link" warnings on legitimate internal services.

## What this spec proposes

A three-category model:

| Category | Trigger | Behavior |
|---|---|---|
| OK | 2xx / 3xx / 4xx-auth-gated | not reported (unchanged) |
| Broken | 404 / 410 / 5xx | red, "broken" (unchanged) |
| **Unverifiable** *(new)* | network error (`ENOTFOUND`, `ECONNREFUSED`, `ETIMEDOUT`, etc.) | yellow, "could not verify — may require auth or corp network" |

The MFE renders unverifiable entries distinctly so authors recognize their internal services and dismiss without confusion; genuinely broken external links continue to surface as before.

## Cross-repo coordination

This is a wire-format change. The spec calls out the MFE dependency as a **hard gate on rollout** in a dedicated section — SAW emission must not reach prod until the MFE supports the new check type. Backward compatibility with the current MFE is an open question that determines whether a feature flag is needed.

## What this PR is and isn't

- **Is**: spec for team review, capturing the design decision, alternatives considered, MFE coordination requirements, and task-by-task implementation plan.
- **Isn't**: implementation. SAW + MFE PRs come after spec approval.
- **Replaces a separate ADR**: the spec includes "Alternatives Considered" + "What would change this decision" sections, so no separate `docs/decisions/` entry needed for now.

## Open questions for the team

1. Does the current MFE handle unknown `check` types gracefully (silent ignore) or error? Determines feature-flag need.
2. Who owns the MFE-side companion PR in `aem-sites-optimizer-preflight-mfe`?
3. Internal-link unverifiability — keep as broken (current proposal) or symmetric "unverifiable" bucket?
4. Reason-code coverage — is `tls-error` worth a distinct category?
5. Customer-facing `reasonHuman` copy — first-draft strings ready for content review.

## Suggested reviewers

I have not auto-assigned reviewers — wanted to leave that to the SAW codeowners + preflight team. Likely candidates: Johnson Ho, Lucian Felix Ghita, Guy Pelletier on the preflight side; MFE team rep for the cross-repo question.

🤖 Drafted with [Claude Code](https://claude.com/claude-code)


GeezerPelletier:
Good spec — the three-category model is the right call and the Walmart repro as acceptance criteria is exactly what's needed. Four things worth resolving before the implementation PR opens:

**1. `tls-error` contradiction (taxonomy table vs. open question 4)**
The reason taxonomy table already defines `tls-error` as a distinct reason with its own `reasonHuman` copy. Open question 4 then asks whether it should be distinct or folded into `unreachable`. These two sections conflict — the question is already answered by the table. My lean: keep it distinct. TLS errors are the most actionable failure mode (the cert is broken — someone needs to know), and `reasonHuman` copy for cert failures is meaningfully different from "server may be temporarily unavailable." Resolve the open question and remove the ambiguity.

**2. `unverifiableInternalLinks` optional bucket in Task 3 contradicts the non-goals**
Non-goals explicitly state: "Internal-link unverifiability — AEM author URLs we control should always resolve from our infrastructure. Treat internal-link network errors as broken." Task 3 then introduces `(optional) unverifiableInternalLinks` with "the bucket may be empty." That's a contradiction. Either remove it from Task 3, or promote the decision out of non-goals. I'd remove it — internal network failures from our infra IS a real problem and should stay broken.

**3. Rollout step ordering needs an explicit merge constraint**
Step 2 says "SAW implementation PR opens — code change behind a feature flag if Task 5 determines one is needed." Step 5 is "SAW PR merges." The ordering is correct (MFE deploys at step 4, SAW merges at step 5), but if Task 5 concludes no feature flag is needed, there's an implicit constraint that the SAW PR must not merge until after step 4. That constraint isn't stated anywhere. Worth adding a note to step 2 or step 5: "If no feature flag, SAW PR must not merge until MFE is confirmed deployed (step 4)."

**4. Wire format field name: `selectors` vs `elementTargets`**
The example JSON uses `elementTargets` but Task 2 refers to "existing element selectors." The current broken-links wire format is the reference point — whichever field name it uses, the new `unverifiable-external-links` entries should explicitly match it here so the MFE author isn't left guessing when they pick up the companion PR.

clotton:
Thanks @GeezerPelletier — all four addressed in `879aa2f3`:

1. **`tls-error` contradiction**: Open Question 4 removed. Taxonomy table answer (keep distinct) stands. Agreed on your lean — TLS errors are the most actionable failure mode and the `reasonHuman` copy is genuinely different.

2. **`unverifiableInternalLinks` contradiction**: dropped the optional bucket from Task 3 — internal-link network errors stay classified as broken per non-goals. Also removed the related Open Question 3 since it had the same contradiction baked in.

3. **Rollout merge constraint**: step 5 now explicitly says the SAW PR must not merge until step 4 (MFE deploy) is confirmed live, when no feature flag is in use.

4. **Wire-format field name**: the canonical field is `elements` (produced by `toElementTargets(selectors)` in `src/preflight/utils/dom-selector.js`). Updated the JSON example to use `elements`, and tightened Task 2 to explicitly reference the helper so the implementation matches the broken-link wire format exactly.

Re-requesting your review.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,291 @@
...
+3. **MFE companion PR opens** in `aem-sites-optimizer-preflight-mfe` — adds render path for `unverifiable-external-links`.
+4. **MFE PR merges + deploys** to prod (or stage with verification).
+5. **SAW PR merges**. If feature-flagged, flag remains off in prod. **If no feature flag was needed (Task 5)**, the SAW PR must not merge until step 4 (MFE deploy) is confirmed live — otherwise prod will emit a `check` type the MFE doesn't know how to render.
+6. **Verification on stage** — Lauren's example page (or equivalent staged repro): confirm the 14 corp-only domains now render as "unverifiable," not broken.
+7. **Enable in prod** (flip flag or roll deploy) once MFE is confirmed live.
+8. **Customer ack**: notify SITES-44776 reporter (Lauren) with link to test against; close the ticket once confirmed in their environment.
+
+**Hard gate**: do not let the SAW PR's emission reach prod before the MFE renders correctly. If the feature flag is in use, leave it off until step 7.
+
+---
```
johobot, line 284: This section reads more like a plan for AI agents. To me, it doesn't read as if it belongs in the spec.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Proposed Solution
+
+### The three-category model
+
+| Category | Trigger | Current behavior | Proposed behavior |
+|---|---|---|---|
+| **OK** | 2xx, 3xx, 401, 403, 405, 429, 451 | not reported | unchanged |
+| **Broken** | 404, 410, 5xx (after HEAD + GET fallback) | reported via `brokenExternalLinks` / `brokenInternalLinks` | unchanged |
+| **Unverifiable** | Network error (`ENOTFOUND`, `ECONNREFUSED`, `ETIMEDOUT`, `ECONNRESET`, `EAI_AGAIN`, etc.) | reported via `brokenExternalLinks` w/ `status: 0` (post-#2476) | reported via NEW `unverifiableExternalLinks` w/ `reason` and `reasonHuman` |
```
johobot: ```reported via NEW `unverifiableExternalLinks` w/ `reason` and `reasonHuman` ```

Is this a new shape in the response? I wonder why use `reason` and `reasonHuman` and not `code` and `message`? Nitpicking but `reasonHuman` looks weird to me 🤷

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,291 @@
...
+  - `issue: "Links that could not be verified"`
+  - `seoImpact: "Low"` (not broken; informational)
+  - `seoRecommendation: "Review whether these links need to remain. They may require authentication or be on a private network."`
+  - `links: [...]` — entries with `urlTo`, `href`, `reason`, `reasonHuman`, and the `elements` array (per the existing broken-link wire format — produced by `toElementTargets(selectors)` in `src/preflight/utils/dom-selector.js`).
+
+Example entry:
+
+```json
+{
+  "urlTo": "https://timesheet.wal-mart.com/gtaapp/etm/defaultHomePage/welcomePage.jsp",
```
johobot: The array key `links` and the top-level `seoImpact`/`seoRecommendation` scalars don't match the shape the rest of the links audit emits. All three existing check types use `{ check, issue: [...] }` where `issue` holds the per-link array and `seoImpact`/`seoRecommendation` are fields on each element inside it. The per-entry link field is also `url` not `urlTo` by the time it reaches the opportunity object (see `links.js:163`). An implementation following this example verbatim would produce a shape inconsistent with every other check type.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,291 @@
...
+}
+```
+
+### Reason taxonomy
+
+`classifyNetworkError(err)` maps Node error codes to reason values:
+
+| `err.code` | `reason` | `reasonHuman` |
+|---|---|---|
+| `ENOTFOUND`, `EAI_AGAIN`, `ENOTRESOLVED` | `dns-failure` | "DNS lookup failed — may require corporate network or authentication" |
```
johobot: `ENOTRESOLVED` is not a valid Node.js / POSIX error code. The correct codes for DNS lookup failure are `ENOTFOUND` and `EAI_AGAIN`, both of which are already listed in this row. Any `classifyNetworkError` implementation matching on `err.code === 'ENOTRESOLVED'` will never fire, so DNS failures would silently fall through to the catch-all bucket instead of the `dns-failure` category.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,291 @@
...
+
+- [ ] In `src/preflight/handler.js`, where the audit envelope is assembled, add a new opportunity object for `unverifiable-external-links` populated from `unverifiableExternalLinks`.
+- [ ] Match the existing opportunity-object shape (`check`, `issue`, `seoImpact`, `seoRecommendation`, `links` / opportunities array).
+- [ ] Update tests to assert the new opportunity appears in the envelope when unverifiable links are present, and is absent (or empty) when none are.
+
+### Task 5: Feature flag (conditional on MFE backward compat)
+
+- [ ] Confirm MFE backward-compat behavior — see [Cross-repo coordination](#cross-repo-coordination-mfe-side).
+- [ ] If MFE handles unknown check types gracefully: skip flag.
+- [ ] If MFE errors on unknown check types: add `PREFLIGHT_LINKS_UNVERIFIABLE_ENABLED` env var; when false, fall back to existing post-#2476 behavior (status: 0 in `brokenExternalLinks`).
```
johobot: The flag-off "fall back to existing post-#2476 behavior" isn't implementable as described. Task 3 restructures `runLinksChecks` to return a separate `unverifiableExternalLinks` bucket — so by the time `handler.js` sees the results, the two categories are already split. For the flag-off path to restore the old behavior, either: (a) `links-checks.js` needs to check the flag and skip the split entirely, keeping everything in `brokenExternalLinks`, or (b) `handler.js` needs to explicitly merge `unverifiableExternalLinks` back into `brokenExternalLinks` before assembling the audit envelope. Neither is described. Without this, flag-off silently drops the unverifiable links rather than reporting them as broken.

johobot (changes requested):

clotton:
Thanks @johobot — all five items addressed in `ae3368c1`. Notes per item:

**Wire-format example (line 116) — fixed.** You're right, the prior shape was inconsistent with the existing `broken-internal-links` / `broken-external-links` opportunities. Confirmed against `handler.js` (line 163 area where it does the `urlTo` → `url` rename and assembles each broken-link entry with per-element `issue` / `seoImpact` / `seoRecommendation` / `elements`). Rewrote the bullet and JSON example to match: `{ check, issue: [...] }` where `issue` is the array of per-link entries with `url`, per-link `issue` string ("Could not verify", parallel to "Status 404"), `code`, `message`, `seoImpact`, `seoRecommendation`, `elements`.

**`ENOTRESOLVED` (line 130) — removed.** Good catch — that's not a real Node error code, and `ENOTFOUND` / `EAI_AGAIN` were already covering the case. Dead match removed.

**Flag-off path (line 244) — clarified.** You're right that the prior framing wasn't implementable; once Task 3 splits the buckets, `handler.js` can't undo that without an extra merge step the spec didn't describe. Updated Task 5 so the flag is consulted in `links-checks.js` **before** the split: flag-on splits as Task 3 describes; flag-off skips the split entirely so everything stays in `brokenExternalLinks` with `status: 0` and the existing `"Status 0"` `issue` shape — exact restoration of post-#2476 behavior, and `unverifiableExternalLinks` stays empty so `handler.js` doesn't emit the new check type. Gating at the producer is the cleaner fallback than merging in the handler.

**`reason` / `reasonHuman` → `code` / `message` (line 95) — taken.** Agreed `reasonHuman` looked weird; `code` / `message` is more conventional. Swap propagated through the wire-format bullet, JSON example, taxonomy table headers, Task 2 return-shape, MFE rendering note, the verification criterion, the customer-copy open question, and the future-work note.

**Task list style (line 267) — keeping the checkbox format.** Matches the established SAW spec convention (e.g., `docs/specs/2026-04-15-cookie-consent-seo-importer-fix.md` uses the same agentic-workers note + checkbox steps). Fine to revisit as a broader spec-format conversation, but didn't want to diverge from how recent specs in this folder are organized.

Re-requesting your review.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
clotton: Swapped `reason` → `code` and `reasonHuman` → `message` throughout the spec in `ae3368c1` (wire-format bullet, JSON example, taxonomy table headers, Task 2 return-shape, MFE rendering note, verification criterion, customer-copy open question, future-work note). Agreed `reasonHuman` was weird; `code` / `message` is the conventional pairing. Resolving.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
clotton: Fixed in `ae3368c1`. Confirmed against `handler.js` (around line 163 where it does the `urlTo` → `url` rename and assembles each broken-link entry). Rewrote both the bullet and the JSON example to match the existing `{ check, issue: [...] }` shape — per-link entries now carry `url`, per-link `issue` string (`"Could not verify"`, parallel to `"Status 404"`), `code`, `message`, `seoImpact`, `seoRecommendation`, and `elements`. Resolving.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
clotton: Removed in `ae3368c1`. Good catch — `ENOTFOUND` and `EAI_AGAIN` already cover DNS failure, so the `ENOTRESOLVED` entry was dead match. Resolving.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
clotton: Clarified in `ae3368c1`. Task 5 now specifies that the flag is consulted in `links-checks.js` **before** the split: flag-on splits as Task 3 describes; flag-off skips the split entirely so everything stays in `brokenExternalLinks` with `status: 0` and the existing `"Status 0"` `issue` shape — exact post-#2476 behavior preserved. `handler.js` doesnt emit the new check type because `unverifiableExternalLinks` is empty. Gating at the producer is cleaner than merging in the handler. Resolving.

GeezerPelletier:
All four original items addressed, and johobot's follow-on fixes (wire-format alignment, `ENOTRESOLVED` removal, flag-off path, `code`/`message` rename) are all the right calls. Spec is solid — good to merge. 🟢

GeezerPelletier (approved):

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,305 @@
...
+
+### Reason taxonomy
+
+`classifyNetworkError(err)` maps Node error codes to a `(code, message)` pair:
+
+| `err.code` | `code` | `message` |
+|---|---|---|
+| `ENOTFOUND`, `EAI_AGAIN` | `dns-failure` | "DNS lookup failed — may require corporate network or authentication" |
+| `ECONNREFUSED` | `connection-refused` | "Connection refused — may require corporate network or VPN" |
+| `ETIMEDOUT`, `ECONNRESET`, `UND_ERR_CONNECT_TIMEOUT` | `timeout` | "Request timed out — may require authentication or corporate network" |
```
johobot: `ECONNRESET` is a TCP connection reset by peer — the remote actively closed the connection — not a timeout. Grouping it with `ETIMEDOUT` under `code: "timeout"` and displaying "Request timed out" to authors is factually wrong and will send them down the wrong debugging path. It should either have its own row (`connection-reset`) or be moved to `connection-refused` with a more accurate message.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,305 @@
...
+### Task 3: Update `runLinksChecks` to bucket results
+
+- [ ] Split the result accumulators into `brokenExternalLinks`, `unverifiableExternalLinks`, and `brokenInternalLinks`. Per [non-goals](#non-goals), internal-link unverifiability stays as broken — no `unverifiableInternalLinks` bucket.
+- [ ] Update the return object signature.
+- [ ] Update tests to assert each bucket.
+
+### Task 4: Emit the new opportunity entry in the handler
+
+- [ ] In `src/preflight/handler.js`, where the audit envelope is assembled, add a new opportunity object for `unverifiable-external-links` populated from `unverifiableExternalLinks`.
+- [ ] Match the existing opportunity-object shape (`check`, `issue`, `seoImpact`, `seoRecommendation`, `links` / opportunities array).
```
johobot: The `links` field name here is a leftover from the earlier draft. The wire format section was correctly updated to use `issue: [...]` as the array key — this parenthetical should match: `(`check`, `issue: [...]`)`.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,305 @@
...
+  - When the flag is **off**: skip the bucket split entirely in `links-checks.js`. The `catch (finalErr)` block still calls `classifyNetworkError` for logging, but the returned object goes into `brokenExternalLinks` with `status: 0` and the existing `issue: "Status 0"` shape — restoring post-#2476 behavior exactly. `handler.js` doesn't emit the new opportunity entry because `unverifiableExternalLinks` is empty.
+- Rationale: the split happens at `links-checks.js`, before `handler.js` ever sees the buckets. Trying to merge them back in `handler.js` would be more invasive (and would still emit the new check type in the envelope even when flag-off, which the MFE may not handle). Gating at the producer is the cleaner fallback path.
+
+### Task 6: Tests, lint, coverage
+
+- [ ] `npm test` passes with 100% coverage maintained.
+- [ ] `npm run lint` passes.
+- [ ] All Walmart-domain test cases from the existing `scripts/test-links-checks-dns.js` debug script demonstrate the new categorization (DNS failure → `unverifiable`).
```
johobot: `scripts/test-links-checks-dns.js` does not exist in the repo. This validation gate is unrunnable as written — either the file needs to be created as part of the implementation, or this step needs to be reworded to describe how to verify the Walmart-domain cases differently.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,305 @@
...
+- [ ] Update tests to assert the new opportunity appears in the envelope when unverifiable links are present, and is absent (or empty) when none are.
+
+### Task 5: Feature flag (conditional on MFE backward compat)
+
+- [ ] Confirm MFE backward-compat behavior — see [Cross-repo coordination](#cross-repo-coordination-mfe-side).
+- [ ] If MFE handles unknown check types gracefully: skip flag.
+- [ ] If MFE errors on unknown check types: add `PREFLIGHT_LINKS_UNVERIFIABLE_ENABLED` env var, and make `runLinksChecks` consult it BEFORE splitting the buckets:
+  - When the flag is **on**: split as Task 3 describes — `unverifiableExternalLinks` becomes its own bucket; `handler.js` emits the new `unverifiable-external-links` opportunity entry.
+  - When the flag is **off**: skip the bucket split entirely in `links-checks.js`. The `catch (finalErr)` block still calls `classifyNetworkError` for logging, but the returned object goes into `brokenExternalLinks` with `status: 0` and the existing `issue: "Status 0"` shape — restoring post-#2476 behavior exactly. `handler.js` doesn't emit the new opportunity entry because `unverifiableExternalLinks` is empty.
```
johobot: `checkLinkStatus` in `links-checks.js` has never returned an `issue` field — the catch block returns `{ urlTo, href, status: 0, ...elements }`. The `issue: \`Status ${status}\`` string is assembled downstream in `links.js`. An implementer reading "the existing `issue: \"Status 0\"` shape" here could add that field to the `links-checks.js` return object, which would change the existing contract with `links.js`.

clotton: Good catch — that was a wrong description of the existing contract. Fixed in `bc4ab11c`. The flag-off path text now correctly describes the existing return shape:

> The `catch (finalErr)` block in `checkLinkStatus` still calls `classifyNetworkError` for logging, but the returned object stays at the existing post-#2476 shape — `{ urlTo, href, status: 0, ...elements }` (no `issue` field on the return). The `issue: "Status 0"` string is assembled downstream in `links.js`; flag-off doesn't touch that path.

Prevents an implementer from accidentally adding an `issue` field to `checkLinkStatus`'s return and breaking the existing contract with `links.js`.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
clotton: Right — that was a real mis-classification. Fixed in `bc4ab11c`: `ECONNRESET` now has its own row as `connection-reset` with message *"Connection reset by remote server — may require authentication or corporate network"*. The `timeout` row is now correctly only `ETIMEDOUT` and `UND_ERR_CONNECT_TIMEOUT`. An RST hitting the same author-facing message as a true read timeout would have sent debugging in the wrong direction.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
clotton: Fixed in `bc4ab11c`. The script was a local debug aid I wrote while drafting the spec — never committed, so the gate as written was unrunnable. Reworded the verification to be concrete and live in the existing test suite:

> Unit tests in `test/preflight/links-checks.test.js` cover the Walmart-domain cases that motivated this spec: every `err.code` value in the [reason taxonomy](#reason-taxonomy) table — `ENOTFOUND`, `EAI_AGAIN`, `ECONNREFUSED`, `ECONNRESET`, `ETIMEDOUT`, `UND_ERR_CONNECT_TIMEOUT`, plus at least one TLS cert error — produces the expected `(code, message)` and routes the link to `unverifiableExternalLinks`. Use `nock` to simulate each error.

Concrete, runnable, lands in the existing harness.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
clotton: Fixed in `bc4ab11c`. Dropped the stale `links / opportunities array` reference. Now reads: *"Match the existing opportunity-object shape: `{ check, issue: [...] }` with per-link entries containing `url`, `seoImpact`, `seoRecommendation`, `elements`."* — aligned with the wire-format section.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,306 @@
...
+- `src/preflight/handler.js` — emit the new `unverifiable-external-links` opportunity entry
+- `test/preflight/links-checks.test.js` — new test cases for each error code category
+
+### Add
+
+- (none — all changes are localized to existing files)
+
+### Possibly Modify (TBD pending MFE decision)
+
+- `src/preflight/handler.js` — feature flag (env var) to gate the new behavior, if MFE backward compatibility requires it
```
johobot, line 224: `links.js` is missing from this inventory. The broken-external-links opportunity assembly (the analogous existing code) lives in `src/preflight/links.js`, not `handler.js` — `handler.js` is an orchestrator that delegates to it. Task 4 points an implementer to the wrong place, and `links.js` is the third file that needs to change.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,306 @@
...
+```
+
+### Reason taxonomy
+
+`classifyNetworkError(err)` maps Node error codes to a `(code, message)` pair:
+
+| `err.code` | `code` | `message` |
+|---|---|---|
+| `ENOTFOUND`, `EAI_AGAIN` | `dns-failure` | "DNS lookup failed — may require corporate network or authentication" |
+| `ECONNREFUSED` | `connection-refused` | "Connection refused — may require corporate network or VPN" |
```
johobot, line 144: `ECONNRESET` is a TCP RST — it fires at the transport layer before any HTTP exchange happens. Auth failures return HTTP 401/403; they do not produce RST packets. "May require authentication or corporate network" is wrong for this error and will mislead authors into trying auth solutions. Suggest: "Connection was reset — the server or an intermediary forcibly closed the connection."

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,306 @@
...
+
+### Task 2: Update `checkLinkStatus` to return categorized failure objects
+
+- [ ] In the `catch (finalErr)` block of `checkLinkStatus`, call `classifyNetworkError(finalErr)` and return an object with `status: 0`, `category: 'unverifiable'`, `code`, `message`, `urlTo`, `href`, and the spread of `toElementTargets(selectors)` (i.e. the `elements` array, matching the broken-link wire format). Note: the function returns `urlTo` internally; `handler.js` performs the `urlTo` → `url` rename when assembling the opportunity entry, same as for broken links today.
+- [ ] Preserve the existing log line; consider lowering severity from `info` to `debug` for `unverifiable` since these are now expected.
+- [ ] Update tests to assert the new return shape on each error code.
+
+### Task 3: Update `runLinksChecks` to bucket results
+
+- [ ] Split the result accumulators into `brokenExternalLinks`, `unverifiableExternalLinks`, and `brokenInternalLinks`. Per [non-goals](#non-goals), internal-link unverifiability stays as broken — no `unverifiableInternalLinks` bucket.
```
johobot, line 244: Task 2 says the catch block unconditionally returns the new shape with `category`/`code`/`message`. Task 5 says the flag-off path must keep the existing post-#2476 shape without those fields. Satisfying both requires a conditional inside `checkLinkStatus` that is never specified — an implementer following Task 2 literally will always return the new shape and cannot satisfy Task 5 without inferring undocumented logic.

johobot (approved):
Conditional approval — good to merge once the three inline comments are resolved (wrong file in inventory/Task 4, ECONNRESET message copy, Task 2/Task 5 return shape contradiction). Core design is solid.

solaris007 (approved):
LGTM

clotton:
@johobot — addressed all three items from your conditional approval in 72efbd15:

1. **Wrong file in Task 4 inventory** — swapped `handler.js` → `links.js` in the Modify list, Task 4, and the Task 5 flag-on path.
2. **ECONNRESET message copy** — removed the incorrect auth/corp-network framing; now reads "server may be temporarily unavailable or rejecting connections."
3. **Task 2 / Task 5 contradiction** — clarified that `checkLinkStatus` always returns the enriched shape (unconditionally); the feature flag is consulted by `runLinksChecks` for bucketing only.

Ready for another look when you have a moment. Thanks!

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
MysticatBot, line 144: ECONNRESET is a TCP RST at the transport layer - auth failures produce HTTP 401/403, not RST packets. The message "may require authentication or corporate network" will mislead authors into trying auth solutions for a transport-level problem.

**Suggested message:** "Connection was reset - the server or an intermediary forcibly closed the connection."

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,306 @@
...
+- `src/preflight/links.js` — emit the new `unverifiable-external-links` opportunity entry (this is where `broken-external-links` assembly lives today; the analogous new-bucket assembly goes here)
+- `test/preflight/links-checks.test.js` — new test cases for each error code category
+
+### Add
+
+- (none — all changes are localized to existing files)
+
+### Possibly Modify (TBD pending MFE decision)
+
+- `src/preflight/handler.js` — feature flag (env var) to gate the new behavior, if MFE backward compatibility requires it
```
MysticatBot, line 224: `src/preflight/links.js` is missing from this inventory. The broken-external-links opportunity assembly lives in `links.js`, not `handler.js`. Task 4 should point implementers here for the analogous pattern.

docs/specs/2026-05-19-preflight-links-unverifiable-category.md
```diff
@@ -0,0 +1,306 @@
...
+
+- [ ] In the `catch (finalErr)` block of `checkLinkStatus`, call `classifyNetworkError(finalErr)` and return an object with `status: 0`, `category: 'unverifiable'`, `code`, `message`, `urlTo`, `href`, and the spread of `toElementTargets(selectors)` (i.e. the `elements` array, matching the broken-link wire format). Note: the function returns `urlTo` internally; `links.js` performs the `urlTo` → `url` rename when assembling the opportunity entry, same as for broken links today. **This enriched shape is returned unconditionally** — the feature flag (Task 5) is consulted by `runLinksChecks` when deciding how to bucket the result, not by `checkLinkStatus` when building it.
+- [ ] Preserve the existing log line; consider lowering severity from `info` to `debug` for `unverifiable` since these are now expected.
+- [ ] Update tests to assert the new return shape on each error code.
+
+### Task 3: Update `runLinksChecks` to bucket results
+
+- [ ] Split the result accumulators into `brokenExternalLinks`, `unverifiableExternalLinks`, and `brokenInternalLinks`. Per [non-goals](#non-goals), internal-link unverifiability stays as broken — no `unverifiableInternalLinks` bucket.
```
MysticatBot, line 244: Task 2 says the catch block unconditionally returns `{ category, code, message, ... }`. Task 5 says flag-off must keep the existing shape without those fields. An implementer cannot satisfy both without undocumented conditional logic.

**Recommended fix:** Have Task 2 always return the enriched shape (classification is cheap), then let Task 3/5 bucketing logic decide disposition based on the flag. Separates classification from routing - both become independently testable.

MysticatBot (dismissed):
Hey @clotton,

## Strengths

- **Excellent problem statement** (lines 22-70): Names the exact customer, the exact page, the exact 14 domains, and traces the causal chain back to PR #2476. An implementer can reproduce the problem without guessing.
- **Three-category model is the right abstraction** (lines 95-101): OK/Broken/Unverifiable maps cleanly to the signal fidelity available from Lambda's network position. Does not attempt to infer intent or collapse distinct failure modes.
- **Wire format as additive-only change** (lines 104-138): New `check` type means existing consumers that iterate by check value ignore the new type by default. Safest evolution strategy for a multi-consumer wire format.
- **Feature flag gating at the producer** (lines 258-265): Gating in `links-checks.js` rather than merging buckets back in the handler is the right call. Rationale is stated clearly.
- **Rollout plan acknowledges cross-repo ordering** (lines 275-286): The hard gate and explicit merge-ordering constraint is the correct way to manage this. Many specs hand-wave this; this one does not.
- **Iterative review quality**: The author has responded substantively to every round of feedback with code references and clear rationale. The spec improved materially across rounds.

## Issues

### Important (Should Fix)

**1. ECONNRESET message copy is misleading** - `docs/specs/2026-05-19-preflight-links-unverifiable-category.md`:142

The `message` for `connection-reset` says "may require authentication or corporate network." ECONNRESET is a TCP RST - it fires at the transport layer before any HTTP exchange. Auth failures produce HTTP 401/403, not RST packets. This message will send authors down the wrong debugging path.

**Fix:** Change to something like "Connection was reset - the server or an intermediary forcibly closed the connection." The `code: "connection-reset"` already tells what happened; the message should not attribute causation it cannot know.

**2. File inventory is incomplete** - `docs/specs/2026-05-19-preflight-links-unverifiable-category.md`:222

`src/preflight/links.js` is missing from the "Modify" inventory. The broken-external-links opportunity assembly lives in `links.js`, not `handler.js`. Task 4 points an implementer to the wrong file for the analogous pattern.

**Fix:** Add `src/preflight/links.js` to the Modify list. Update Task 4 to point at `links.js` rather than `handler.js`.

**3. Task 2 / Task 5 return-shape contradiction** - `docs/specs/2026-05-19-preflight-links-unverifiable-category.md`:242

Task 2 says `checkLinkStatus`'s catch block should unconditionally return `{ status: 0, category: 'unverifiable', code, message, urlTo, href, ...elements }`. Task 5 says the flag-off path must keep the existing shape `{ urlTo, href, status: 0, ...elements }` (no `category`/`code`/`message`). Both cannot be true without a conditional that Task 2 never mentions.

**Fix:** Recommend Task 2 always return the enriched shape (classification is cheap), then have Task 3/5's bucketing logic decide disposition based on the flag. Separates classification from routing - both become independently testable.

### Minor (Nice to Have)

**4. Wire format section references wrong file** - `docs/specs/2026-05-19-preflight-links-unverifiable-category.md`:110

Line 110 says "see `src/preflight/handler.js` for the assembly" but the actual assembly is in `links.js`. This predates the file-inventory fix and should be updated for consistency.

**5. Historical data discontinuity not documented** - `docs/specs/2026-05-19-preflight-links-unverifiable-category.md`:101

Once this ships, existing audit records will still contain `status: 0` entries in `brokenExternalLinks`. Consumers that aggregate broken-link counts over time will see a discontinuity. A one-line note ("Historical audit records retain the pre-cutover classification; no backfill is planned") would prevent a "why did broken links drop by 14 on tenant X" investigation.

## Recommendations

- Consider documenting whether `unverifiable-external-links` should appear in aggregate audit health metrics or SLI definitions. Since `seoImpact` is "Low," it likely should not contribute to site-health scores - a one-sentence note would prevent downstream implementers from accidentally including it.

## Assessment

Ready to merge? With fixes
Reasoning: The core design is architecturally sound and the spec has improved materially across three review rounds. The three Important items (ECONNRESET message, file inventory, Task 2/5 contradiction) align with johobot's conditional approval and are all fixable in a small commit. Once addressed, this is ready to guide implementation.

## Next Steps

1. Address the 3 Important items (all raised by johobot's conditional approval, validated here).
2. Minor items are optional but easy wins in the same commit.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 8s | <b>Cost:</b> $2.17 | <b>Commit:</b> <code>bc4ab11c7f89a428e88b082d61701fc279e5f04d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2534, opened 2026-05-20
feat(cdn-reports-bulk-publish): consolidator audit for bulk preview+publish

akshaymagapu:
## Summary

New audit `cdn-reports-bulk-publish` that consolidates `admin.hlx.page` preview+publish for all sites with `cdn-logs-report` enabled in a single batch.

## Why

Per-site `cdn-logs-report` audits do best-effort bulk publish with a 3-min preview cap and fire-and-forget publish ([PR #2531](https://github.com/adobe/spacecat-audit-worker/pull/2531)). On bad days, some sites' previews time out and skip publish; on new-site onboarding the per-site call may not have run yet. This audit acts as a safety net: one call, all sites, awaited preview with a longer 10-min cap, fire-and-forget publish.

Intended to be run manually (no scheduler entry).

## Behavior

For each run:
1. Fetch all sites; keep those where `Configuration.isHandlerEnabledForSite('cdn-logs-report', site)` is true and the LLMO data folder is set
2. Build target periods: always current ISO week; if today is Monday, also previous week (to catch the rollover)
3. For each site × period, add both report paths:
   - `/{llmoFolder}/agentic-traffic/agentictraffic-w{N}-{year}.json`
   - `/{llmoFolder}/referral-traffic-cdn/referral-traffic-w{N}-{year}.json`
4. Call `bulkPublishToAdminHlx(paths, log, { pollTimeoutMs: 10 * 60_000 })` — preview awaited up to 10 min, publish fire-and-forget

## Changes

`src/cdn-reports-bulk-publish/handler.js` (new) — the audit handler (~50 LOC)

`src/utils/report-uploader.js` — add `pollTimeoutMs` option to `runBulkJob` and `bulkPublishToAdminHlx` so callers can override the default 3-min preview cap. Existing default behavior unchanged.

`src/index.js` — register the new audit type.

## Test plan

- [x] `npx mocha test/utils/report-uploader.test.js` — 23 passing (existing tests cover the option propagation)
- [x] `npx eslint src/` — clean
- [ ] Trigger manually in dev via SQS message of type `cdn-reports-bulk-publish`; confirm in Coralogix:
  - `bulk-publishing N paths across M sites for periods [...]` log line
  - `preview status: stopped` for the consolidated job
  - `publish job fire-and-forget, not polling for completion`


src/cdn-reports-bulk-publish/handler.js
```diff
@@ -0,0 +1,101 @@
...
+    result.error = error.message;
+  }
+
+  return {
+    auditResult: result,
+    fullAuditRef: `${AUDIT_TYPE}/${periods.join(',')}`,
+  };
+}
+
+export default new AuditBuilder()
```
ctinp: Because this handler is registered with the default `AuditBuilder`, a `cdn-reports-bulk-publish` message still goes through the normal per-site `isAuditEnabledForSite(type, site, context)` gate before `runCdnReportsBulkPublish()` is ever called. In practice that means this “cross-site” manual safety-net only runs if the arbitrary `siteId` on the message also has the new `cdn-reports-bulk-publish` handler configured with product codes and enabled for that site.

That seems at odds with the PR description: if ops sends the manual SQS message for a site that has `cdn-logs-report` enabled but does not also have this new audit type enabled, the whole job will be skipped and no cross-site publish will happen. Can we bypass the normal per-site audit enablement for this handler, or explicitly key it off `cdn-logs-report` instead?

src/cdn-reports-bulk-publish/handler.js
```diff
@@ -0,0 +1,101 @@
...
+    log.info(`%s: bulk-publishing ${reports.length} paths across ${llmoFolders.length} sites for periods [${periods.join(', ')}]`, AUDIT_TYPE);
+  }
+
+  const result = {
+    sites: llmoFolders.length,
+    paths: reports.length,
+    periods,
+  };
+
+  try {
```
ctinp, line 59: Catching `bulkPublishToAdminHlx()` errors and converting them into `auditResult.success = false` means the audit invocation itself still succeeds even when the consolidated preview times out or the preview request fails. Since this job is supposed to be the safety-net for missed publishes, swallowing the exception here makes that failure easy to miss and also suppresses the normal failed-run/redrive signal from the worker.

If the 10-minute preview still times out on a bad day, we will persist a failed-looking audit row, but the manual trigger path itself will appear successful. Can we either rethrow after recording the failure, or add an explicit failure signal that the caller/operator will actually see?

ctinp (approved):

src/cdn-reports-bulk-publish/handler.js
akshaymagapu: Good catch, thanks. Switched to a plain HANDLERS entry (no `AuditBuilder`), matching the pattern used by `rum-config-refresh`, `llmo-config-db-sync`, `delivery-config-writer`, etc.

That bypasses `siteProvider` / `urlResolver` / `isAuditEnabledForSite` entirely. The SQS message no longer needs a `siteId`, and the audit type doesn't need to be enabled per-site in Configuration before triggering. The trigger surface is now just: send `{ type: 'cdn-reports-bulk-publish' }` to the audit-jobs queue.

See commit 57ad2aea.

src/cdn-reports-bulk-publish/handler.js
akshaymagapu: Fair point - the original try/catch was silent. Tightened it: we still catch (see below), but the catch now logs at `error` level with site/path counts so a Coralogix alert on `bulk publish failed` surfaces it, and the response body has `success: false, error: <message>` so the operator who triggered the manual run sees the failure in their Slack response.

Reason we still catch instead of throwing: `spacecat-audit-jobs` has `maxReceiveCount = 1` with a DLQ, so a thrown error would send the message to `spacecat-dead-letter-queue`, which isn't actively monitored (see the Apr 21 incident from the SKYSI-79147 thread - 1,500 messages sat in DLQ for 4 days). Coralogix is the path that has eyes on it.

See commits c65150ac and c920c0db.

PR #2538, opened 2026-05-20
fix(prerender): remove fallback that wrote phantom error entries to status.json

anuj-adobe:
## Summary

- When all submitted URLs had FAILED status in the scraper, `scrapeResultPaths` was empty and the old fallback re-fetched ~153 top-page URLs, then ran `compareHtmlContent` on all of them
- Since no S3 data existed for those URLs under the current `scrapeJobId`, all ~153 appeared as errors in `status.json` — even though only a handful were ever submitted to this scrape job
- `getScrapeJobStats` already handles FAILED-status URLs via the `ScrapeUrl` DB (`missingPages`), making the fallback both redundant and harmful
- Replaced fallback with a warning log; `status.json` now only records URLs actually submitted to scraping

## Root Cause

`getScrapeResultPaths` only returns COMPLETE-status URLs. When all submitted URLs fail (FAILED status), `scrapeResultPaths.size === 0`, which triggered the fallback. The fallback pre-dates the `getScrapeJobStats` mechanism that was later added to handle this case correctly.

## Test Plan

- [x] 238 tests passing (`npm run test:spec -- test/audits/prerender/handler.test.js`)
- [x] Deleted tests that covered the removed fallback behavior
- [x] Added test: "should log a warning and skip comparison when scrapeResultPaths is empty"
- [x] Existing `getScrapeJobStats` tests cover the correct path for FAILED URLs


ssilare-adobe (changes requested):
Hey @anuj-adobe,

## Strengths

The root cause analysis is accurate and clearly communicated. The fallback was genuinely harmful - running `compareHtmlContent` against ~153 top-page URLs with no matching S3 data under the current `scrapeJobId` would necessarily produce `error` entries in `status.json` for URLs never submitted to the scrape job. Replacing it with a warning log and a pointer to `getScrapeJobStats` is exactly right (`src/prerender/handler.js`, ~line 1636-1643).

The dual-track accounting design - COMPLETE-status URLs via `compareHtmlContent`/`comparisonResults`, FAILED-status URLs via the `ScrapeUrl` DB/`missingPages` - is the correct architectural pattern. This PR reinforces that boundary cleanly instead of blurring it.

Removing the `/* c8 ignore start */` and `/* c8 ignore stop */` annotations improves coverage integrity. Those guards were shielding ~40 lines of untested code; their removal now makes the true coverage picture visible.

The test fixups across ~14 contexts that now supply `scrapeResultPaths: new Map([...])` correctly reflect the new contract. Tests that were relying on the deleted fallback as a side-entry were testing the wrong invariant.

## Issues

### Critical (Must Fix)

**CI failing - branch coverage below 100% at `src/prerender/handler.js` line 457**

The build fails at "Lint, Test, Coverage Upload". The uncovered branch is in `getScrapedHtmlFromS3`:

```js
const scrapeJsonData = results[2].status === 'fulfilled' ? results[2].value : null;
```

The `'rejected'` branch (`: null`) is no longer exercised after the fallback tests are removed. Those tests previously exercised this branch as a side effect by mocking S3 operations in ways that caused the third `Promise.allSettled` slot to settle as rejected. With those tests gone, the branch is uncovered and the project's required 100% branch threshold is not met.

Fix options:

- **Option A (preferred):** Add a targeted test that stubs the `getObjectFromKey` call for the scrape-json key to reject, exercising `results[2].status === 'rejected'`.
- **Option B:** If the false branch is genuinely unreachable by design, simplify the expression and add a `/* c8 ignore next */` with a comment explaining why.

### Important (Should Fix)

**`src/prerender/handler.js`, ~line 1641 - Warning log is missing `scrapeJobId`, making production triage harder**

```js
log.warn(`${LOG_PREFIX} No COMPLETE scrape results for baseUrl=${site.getBaseURL()}, `
  + `siteId=${siteId}. Skipping comparison; failed URLs recorded via ScrapeUrl DB.`);
```

When this fires in production, the operator's first move is to look up which scrape job had all failures. `siteId` gets you to the site but not the specific run. `auditContext?.scrapeJobId` is in scope:

```js
const { scrapeJobId } = auditContext ?? {};
log.warn(`${LOG_PREFIX} No COMPLETE scrape results for baseUrl=${site.getBaseURL()}, `
  + `siteId=${siteId}, scrapeJobId=${scrapeJobId ?? 'unknown'}. `
  + `Skipping comparison; failed URLs recorded via ScrapeUrl DB.`);
```

**`test/audits/prerender/handler.test.js`, ~line 4849 - Test description no longer matches behavior**

The test titled `"should handle sheet load failures gracefully even when log.warn is missing"` now provides `warn: sinon.stub()` in its context - meaning it no longer tests the "log.warn is missing" scenario. The test body was updated but the description was not. Update the title to match what the test actually does.

### Minor (Nice to Have)

**New test builds a raw context object instead of using `MockContextBuilder`**
`test/audits/prerender/handler.test.js`, ~line 1717: Every other test in this describe block uses `MockContextBuilder`. The new test constructs a raw context object. This works now but won't pick up future structural changes to the context shape automatically.

**New test does not assert that S3 is written in the all-FAILED path**
The new test verifies `result.status === 'complete'` and the warning message but does not assert that `uploadStatusSummaryToS3` fires. A future change that silently skips the S3 write when `urlsToCheck` is empty would not be caught.

## Recommendations

1. After resolving the coverage gap, consider including the count of FAILED URLs in the warning log ("N URLs had FAILED status") for more actionable production signals.
2. A manual staging run against a domain that fails all scrapes would confirm `status.json` contains only `missingPages` entries and no phantom error entries - worth doing before deploying given this was a production incident.

## Assessment

Ready to merge? No

The change is correct - the fallback was causing real data corruption and removing it is the right fix. The code comment and PR description are clear. The test suite cleanup is thorough.

The blocker is the CI coverage failure at `src/prerender/handler.js` line 457. This is a direct consequence of removing the `c8 ignore` guards - those guards were hiding an existing coverage gap in `getScrapedHtmlFromS3`. One targeted test addition will resolve it.

## Next Steps

1. Address the coverage gap at `src/prerender/handler.js` line 457 - this is the only CI blocker.
2. Add `scrapeJobId` to the warning log message.
3. Fix the test description mismatch.
4. Minor items are optional.

Note: CI checks are currently failing - the coverage issue is the root cause and must be resolved before merge.

ssilare-adobe (approved):
Hey @anuj-adobe,

## Previously Flagged Issues Now Addressed

The Critical coverage gap at `src/prerender/handler.js:457` was resolved by adding `scrapeResultPaths` to the two test contexts that exercise the scrape-json rejection path - CI is now pending instead of failing.

Both Important items are addressed in this commit:
- `scrapeJobId` is correctly added to the warning log. `auditContext` is destructured at the top of `processContentAndGenerateOpportunities` so it is properly in scope, and `auditContext?.scrapeJobId ?? 'unknown'` handles the absent-context edge case without emitting `undefined` in the log string.
- The test title "should handle sheet load failures gracefully and continue scraping" accurately describes the behavior now that the context provides a `warn` stub.

## Remaining Minor Items

The two Minor items from the prior review are still open - the new test constructs a raw context object rather than using `MockContextBuilder`, and the all-FAILED path test does not assert that `uploadStatusSummaryToS3` fires. These were labeled optional and are not blocking.

## Assessment

Ready to merge? Yes, pending CI green.

The change is correct, the test suite is clean, and all Critical and Important feedback has been addressed.

PR #2539, opened 2026-05-20
feat(cdn-logs-report): decouple URL classification from LLMO customer config (LLMO-4748)

akshaymagapu:
## What this PR does

Two things, in one branch:

1. **Decouples URL classification from LLMO customer config.** Previously the `cdn-logs-report` audit read `config.llmo.categories` from S3 and used it both as the LLM steering prompt AND as a strict post-filter that dropped any auto-discovered category not in the LLMO list. That coupling produced poor categorization. This PR removes the LLMO-config reads and lets the LLM derive categories purely from the site's own URL structure.

2. **Enriches auto-derived rules with provenance + sample URLs.** Each rule the audit-worker writes now carries `source='ai'`, `sample_urls` (up to 20 URLs from the customer's site that the regex actually matches), and `derivation_method='llm'`. The read path also widens to surface these fields downstream.

Also bundled: a small punch list from earlier review — a regex compile-test, an alternation-branch cap (<=12 branches) and a 40%-path-coverage catch-all check that together drop over-broad / uncompilable LLM regex, query-string stripping before feeding URLs to the LLM, prompt cleanup, and dead-export deletion — plus a defensive clamp so a legacy/junk DB row can't abort the all-or-nothing replace RPC: `sample_urls` is clamped to <=50 string elements of <=2048 chars before the payload is posted.

## Why

**Why decouple from LLMO config?** Validated across 15 LLMO customer sites: categories steered by the LLMO config were thin (coverage often 0-30%) and generic ("publish", "mypayinfo", "usage & troubleshooting"). The LLMO config is a prompt-management concept that doesn't map to actual site sections. Letting the LLM derive from the URL structure alone produces materially better categorization (e.g. business.adobe.com gets meaningful umbrella categories like marketing / analytics / commerce).

**Why enrich the rules now?** The next step in this workstream is a customer-facing UI where customers review, edit, and confirm rules. For that to work:
- The UI needs concrete sample URLs to show ("matches /products/photoshop, /products/photoshop/cc.html, ..."), not just opaque regex
- The system needs to know which rules are auto vs customer so they can be displayed with the right badge and treated correctly on subsequent runs
- Customers shouldn't have to write regex — they should paste URLs and have the regex computed

This PR populates the data side; the UI / customer-edit endpoints come later.

## Auto-derivation is a one-time bootstrap

Important design point: auto-derivation runs only on **first-time onboarding** (when no DB rules exist for the site). The existing reuse short-circuit in `patterns-uploader.js` means scheduled runs never re-derive once any rules exist — existing rules (auto or customer) round-trip through the replace RPC unchanged. Customer edits made later are safe by inertia; the audit-worker doesn't touch them.

To reset a site to fresh-bootstrap state, delete all its rules (via the forthcoming customer-edit API) → the next scheduled run sees no existing rules and re-derives from scratch.

## Regex guards & the Athena engine (corrected)

An earlier revision on this branch added a **512-char length cap** and a **lookaround rejection** (`(?=`, `(?!`, `(?<=`, `(?<!`) on LLM-emitted regex, on the assumption that Athena runs a POSIX / RE2-style engine that can't handle them. **That assumption is wrong** — verified directly against Athena, whose `regexp_like` uses the JONI (Oniguruma) engine:

| Athena query | Result |
|---|---|
| `regexp_like('foobar', 'foo(?=bar)')`   | `true`  |
| `regexp_like('foobaz', 'foo(?=bar)')`   | `false` |
| `regexp_like('foobar', '(?<=foo)bar')`  | `true`  |
| `regexp_like('foobar', 'foo(?!bar)')`   | `false` |
| `regexp_like('foobaz', 'foo(?!bar)')`   | `true`  |
| `regexp_like('page.html', '[.]html')`   | `true`  |
| `regexp_like('page.html', '\.html')`    | `true`  |

So lookaround **and** `\.` both work in Athena. The length cap also mirrored no real DB constraint (regex is plain `TEXT`). Both guards were therefore removed — the alternation-branch cap + 40%-coverage check already drop the pathological / catch-all output they were meant to catch, without falsely rejecting valid patterns.

The one *real* backslash hazard is **upstream**: a raw `\` in the LLM's JSON output is an invalid string escape that breaks `JSON.parse`. So the prompts still tell the model "no backslash escapes" — but the rationale is corrected from the false "Athena breaks on them" to the true "it breaks parsing of your JSON output." Consistent with that, the forced homepage default now matches a literal dot with the backslash-free character class `([.]html)?` instead of `(\.html)?`, so it round-trips through the JSON->Athena pipeline with no escaping ambiguity.

## What writes look like after this PR

```jsonc
{
  "name": "Photoshop",
  "regex": "(?i)(photoshop|adobe-photoshop)",
  "sort_order": 3,
  "source": "ai",
  "sample_urls": [
    "/products/photoshop",
    "/products/photoshop/cc.html",
    "/products/photoshop/features"
    // up to 20 paths from the audit URL list that the regex matches
  ],
  "derivation_method": "llm"
}
```

`source`, `sample_urls`, and `derivation_method` are read by the `wrpc_replace_agentic_url_classification_rules` RPC (extended on data-service `main`). `source` ('ai'|'human') and `derivation_method` (enum) are DB CHECKs and the RPC's own guards — a bad value aborts the whole-site replace, which is why the worker only ever emits the validated `'ai'`/`'llm'` pair.

## Verified

- Full test suite passing (cdn-logs-report + llmo-customer-analysis suites), 100% coverage gate green
- Lint clean
- Athena engine behavior confirmed directly via `regexp_like` (table above)
- Manually validated on 5 real LLMO sites (business.adobe.com, stryker.com, motilaloswal.com, hm.com, blog.adobe.com) — categories materially improved, no regressions

## Caveats

- `derivation_method` is uniformly `'llm'` today because the production codebase only has LLM-based product-analysis + page-type-analysis. Earlier exploration on this branch (sitemap clustering, breadcrumb/path tiered derivation) is not wired in. If we ever ship tiered clustering, swap the uniform value for per-tier values.

## Test plan
- [ ] CI build passes
- [ ] Manual stage run on one site: confirm rules persist with `source='ai'` and `sample_urls` populated
- [ ] Confirm a B2B site (e.g. business.adobe.com) no longer mislabels `/products/<single-offering>/` as PDP


src/cdn-logs-report/patterns/page-type-analysis.js
```diff
@@ -408,19 +402,26 @@ export async function analyzePageTypes(domain, paths, context) {
...
+
+    // Hoist homepage to front so the cap never drops it.
+    const entries = [['homepage', homepageRegex], ...Object.entries(orderedRegexes)];
+    if (entries.length > MAX_LLM_PAGE_TYPES + 1) {
+      log.info(`Capping ${entries.length} → ${MAX_LLM_PAGE_TYPES + 1} page types (hard cap, before adding defaults)`);
     }
+    const capped = Object.fromEntries(entries.slice(0, MAX_LLM_PAGE_TYPES + 1));
 
-    // Add default patterns at the end
     const defaultPatterns = {
```
ctinp, line 421: **`robots` / `sitemap` defaults are now broad substring matches and can misclassify normal content URLs**

In `src/cdn-logs-report/patterns/page-type-analysis.js` around lines 418-421, the auto-injected defaults changed from file-shaped patterns to:

- `(?i)robots.txt`
- `(?i)sitemap`

Those are later used as plain `REGEXP_LIKE(url, ...)` checks in the report query, so they will match any URL containing those substrings, not just actual `robots.txt` / sitemap files. For example, `/docs/meta-robots-txt` or `/blog/how-to-build-a-sitemap` would now be classified as `robots` / `sitemap`.

This looks like a regression from the previous anchored patterns (`.../robots.txt$`, `.../sitemap.*.xml$`).

src/common/agentic-url-classification-rules.js
```diff
@@ -20,6 +20,9 @@ function normalizeRuleRows(rows = []) {
     name: row.name,
     regex: row.regex,
     sort_order: Number.isInteger(row.sort_order) ? row.sort_order : index,
+    source: row.source ?? 'ai',
```
ctinp, line 23: **Defaulting missing `source` to `'ai'` can relabel legacy human rules as auto-derived**

In `src/common/agentic-url-classification-rules.js` around line 23, `source` is normalized with:

```js
source: row.source ?? 'ai',
```

That means any legacy row with NULL source will now be surfaced downstream as AI-derived, even if it was originally customer-authored or otherwise unknown. Since this PR’s goal is to distinguish auto vs human rules, changing missing historical data into 'ai' feels risky: once these rows round-trip through the replace flow, they’ll keep that incorrect provenance.

ctinp (approved):

src/cdn-logs-report/patterns/page-type-analysis.js
akshaymagapu: Good catch — anchored both to path-end and switched to `[.]` (matching the homepage regex convention) in e790076d.

src/common/agentic-url-classification-rules.js
akshaymagapu: The `source` column is `NOT NULL DEFAULT 'ai'` with `CHECK (source IN ('ai','human'))`, so the value is always populated and the `?? 'ai'` never fires on real rows — keeping it as a defensive default.

PR #2544, opened 2026-05-21
feat: add spacecatOrgId as optional as optional parameter for slackbot command

rarescheseli:
Add an optional `spacecatOrgId` parameter to the slackbot command for the `offsite-brand-presence` audit. If present, use it when submitting jobs to DRS.

## Related Issues

Thanks for contributing!


tb-adbe (approved):

amitruluimadalina (dismissed):
Hey @rarescheseli,

## Strengths

- `messageData?.spacecatOrgId` extraction in `handler.js:623-624` follows the existing `messageData` passthrough idiom - no new SQS contract plumbing required.
- `submitWithRetry` (`handler.js:449-474`) is a cleanly isolated, independently testable helper with the right shape.
- The defensive `r.response?.job_id` change in `notifyDrsResults` (`handler.js:595`) prevents a crash when the direct path returns `null` on a non-JSON 200 response.
- The config-missing guard (`handler.js:494-505`) is fail-closed and audible - logs and returns `[]` instead of crashing.
- Test coverage on the new direct path is thorough: success, retry-then-succeed, missing URL and KEY separately, trailing-slash normalization, multi-domain fan-out, `days_back` only for reddit, non-JSON response, and drsClient passthrough when org id is absent.

## Issues

### Critical (Must Fix)

**`submitDrsJobDirect` bypasses `DrsClient` to add one body field, permanently forking the DRS HTTP transport**

`src/offsite-brand-presence/handler.js` (new `submitDrsJobDirect` function)

The only delta from `DrsClient.submitScrapeJob` is one top-level field: `spacecat_org_id`. The fork loses:

- `tracingFetch` from `@adobe/spacecat-shared-utils` - APM and Coralogix trace chains break for every org-id audit; you cannot stitch a DRS submission on this path back to its audit-worker trace.
- Input validation - `datasetId` checked against `VALID_SCRAPE_DATASET_IDS`, `urls` must be non-empty, `daysBack` only valid for `reddit_comments`. The direct path sends whatever it receives; a future refactor of `OFFSITE_DOMAINS` or a new caller will silently submit malformed requests that surface as 4xx from DRS rather than a local error.
- Canonical log shapes - two "Submitting DRS job" formats now exist, requiring union queries during incident investigations.

Every future DRS contract change has two homes in two repos, and the audit-worker copy will silently fall behind.

Fix: add `spacecatOrgId` as an optional parameter to `DrsClient.submitScrapeJob` in `@adobe/spacecat-shared-drs-client` (a one-line body-assembly addition). The audit-worker change collapses to `drsClient.submitScrapeJob({ ...params, spacecatOrgId })`. The entire `submitDrsJobDirect` function, the duplicate config check, and the conditional `submitFn` branch disappear.

---

**Retry silently applied to the existing `DrsClient` path; `Promise.all` replaced by a serial loop; terminal-failure test deleted**

`src/offsite-brand-presence/handler.js:533-541`, `test/audits/offsite-brand-presence.test.js:1086`

The PR description says "add an optional parameter." The change also rewires all DRS submission through `submitWithRetry`, so every failure on the non-org-id path now retries automatically - including 4xx responses that cannot succeed. The test that documented the old contract (`'should handle DRS network error gracefully'`) was replaced in this same PR, leaving no test for the terminal-failure case on the existing path.

The parallel fan-out (`Promise.all`) was also replaced by a serial `for...await` loop. Wall-clock time goes from `max(jobLatency)` to `sum(jobLatency)`. With retry on top, worst case is approximately 2x `sum(jobLatency)` for every audit - not just org-id ones. On a site near Lambda or scheduler timeout budget, this is a silent production regression for all existing callers.

Fix: scope retry to the direct-HTTP path only, or document the behavior change in the PR description and get DRS-side sign-off on doubled load during incidents. Restore `Promise.all` - per-job retry is self-contained: `Promise.all(jobs.map((job) => submitWithRetry(job, submitJob, log)))`. Restore a terminal-failure test: both attempts fail, result is `{ status: 'error', error: '...' }`.

---

**`spacecatOrgId` forwarded to DRS without format validation**

`src/offsite-brand-presence/handler.js:623-624`

`messageData?.spacecatOrgId` flows from Slack input through SQS into the DRS request body as `spacecat_org_id` with no shape check or UUID format validation. The unvalidated value is also interpolated into a Slack message string (`orgSuffix`, line ~530): a value with newlines or Slack formatting characters can forge lines in engineering channels.

Fix: validate format at the runner boundary before the value is used anywhere:

```js
const raw = messageData?.spacecatOrgId;
const spacecatOrgId = (typeof raw === 'string' && /^[a-zA-Z0-9-]+$/.test(raw.trim()))
  ? raw.trim() : undefined;
```

Additionally, `spacecat-api-service` should validate the caller has access to the supplied org before publishing the SQS message, so the audit-worker receives a pre-validated claim.

### Important (Should Fix)

**Retry catches all errors including non-retriable 4xx, with no backoff and no idempotency**

`src/offsite-brand-presence/handler.js:449-474`

`submitWithRetry` retries immediately on any thrown `Error`. `POST /jobs` is not idempotent - no `Idempotency-Key` header is sent. A request that times out client-side but already landed server-side will produce two BrightData scrape jobs, double-billing and double-counting downstream results. 4xx responses are also retried even though they cannot succeed.

Fix: narrow retries to clearly retriable conditions - pre-send network errors (`TypeError` from `fetch` before a response arrives) and 5xx responses. Skip retry on 4xx. Add a short backoff (e.g. 250ms + jitter). Until DRS exposes an idempotency key, document the duplicate-job risk in a comment.

---

**Direct-HTTP path uses global `fetch` (no distributed tracing) and has no request timeout**

`src/offsite-brand-presence/handler.js:417-424`

`DrsClient` uses `tracingFetch` from `@adobe/spacecat-shared-utils`. The new path uses the global `fetch`, breaking trace chain propagation for org-id audits. There is also no `AbortSignal.timeout()` on the call - if DRS hangs, the promise never rejects and the retry helper cannot help; the call hangs until the Lambda is killed.

Fix: import `tracingFetch` from `@adobe/spacecat-shared-utils` and use it in place of global `fetch`. Add `signal: AbortSignal.timeout(30_000)` and handle `AbortError` as a retriable condition.

---

**`spacecat_org_id` has no documented server-side contract**

`src/offsite-brand-presence/handler.js:392-401`

There is no reference to `spacecat_org_id` in `@adobe/spacecat-shared-drs-client`, and the PR description does not link a DRS-side PR or spec. It is unclear whether DRS currently accepts and uses this field or silently ignores it. The field also sits at the request envelope level while all other parameters (`dataset_id`, `site_id`, `urls`, `days_back`) live inside `parameters` - an asymmetry worth clarifying.

Fix: link the DRS-side change or spec in the PR description. Confirm whether `spacecat_org_id` belongs at the envelope level or inside `parameters`. If the server-side change has not yet landed, document this explicitly as a coordinated two-PR change.

---

**No tests for network-error retry or terminal-failure on the direct path**

`test/audits/offsite-brand-presence.test.js:1195-1213`

The error test covers only `ok: false` (server-returned error). There is no test where `fetch` throws a network error (different code path), no test confirming the retry warn log fires once followed by the error log on the final attempt, and no test where both attempts throw producing `status: 'error'`.

Fix: add (a) `fetchStub.onCall(0).rejects(new Error('ECONNRESET')).onCall(1).resolves(...)` to verify retry fires on network errors, and (b) both calls throwing to verify error propagation and correct log output.

### Minor (Nice to Have)

- `handler.js:449-474`: the `/* c8 ignore next 4 */` fallback at the end of `submitWithRetry` is unreachable. Restructure with an explicit try/catch and `let lastErr` to remove the dead code and the `eslint-disable` comment.
- `handler.js:389-392`: `drsApiUrl.replace(/\/+$/, '')` is clearer than the `while` loop.
- `handler.js:595`: `r.response?.job_id` evaluates to `undefined` when `response` is `null`, rendering ``job_id: `undefined``` in Slack. Use `r.response?.job_id ?? 'n/a'` or omit the field when null.
- `handler.js:383`: JSDoc return type is `Promise<object>` but the function can return `null`. Update to `Promise<object|null>`.
- `handler.js:481-489`: JSDoc for `triggerDrsScraping` documents the new param but does not mention that retry now applies to both paths.
- `handler.js:410-414` and `~530`: two different log conventions used in the same PR - value in structured kwarg in one place, in the message string in another. Pick one.

## Recommendations

1. **Extend `DrsClient` upstream rather than forking it here.** A one-line addition to `submitScrapeJob` in `@adobe/spacecat-shared-drs-client` eliminates the two critical design issues, keeps tracing and validation consistent, and gives every future consumer org-scoping automatically.
2. **Separate the retry change from the org-id feature.** They are independent improvements with different risk profiles; a rollback of one should not lose the other.
3. **Add an idempotency key to DRS job submission before enabling client-side retry.** Until `POST /jobs` is idempotent, retry on network timeout creates duplicate scrape jobs.
4. **Link the DRS-side change before merge.** If `spacecat_org_id` is not yet accepted by DRS, this PR ships a feature that depends on an unwritten server-side promise.

## Assessment

Ready to merge? No - with fixes
Reasoning: The functional intent is sound, but the implementation routes around the shared DRS client to add one field, silently changes retry and serialization behavior for all existing callers, and forwards an unvalidated Slack input to a privileged downstream service - three compounding issues that individually would be "should fix" but together represent the wrong architectural layer for this change.

## Next Steps
1. Address the 3 critical issues first - extend `DrsClient` upstream, scope or document the retry change and restore the terminal-failure test, and add `spacecatOrgId` format validation at the runner boundary.
2. Review the 5 important items - particularly idempotency, `tracingFetch`, request timeout, and the undocumented server contract.
3. Minor items are optional.

rarescheseli:
@amitruluimadalina very valid feedback, but this PR will be reverted if we notice that sending `spacecat_org_id` in the request works. It will be added in the `DrsClient` class, as suggested in your review.

PR #2551, opened 2026-05-22
feat(paid-keyword-optimizer): pass resolvedPageHeading and pageTopics through to opportunity.data

danieljchuser:
## Summary

Mystique now emits two page-level context fields in the ad-intent-mismatch SQS body: `resolvedPageHeading` (the page's vision-resolved primary heading) and `pageTopics` (the full topic list extracted from the landing page, not the per-cluster filtered subsets). Read both from the guidance body in `mapToKeywordOptimizerOpportunity` and append them to `opportunity.data` verbatim.

Backwards compatible:
- Both fields default to `null` / `[]` when the SQS message lacks them or sets them to `null`.
- Uses `??` (nullish coalescing) instead of ES destructure defaults so `undefined` and explicit `null` both collapse to safe defaults.
- No existing `opportunity.data` field is renamed, removed, or retyped.
- The UI will start consuming the new fields in a follow-up; until then it ignores them.

## Spec / Cross-repo contract

- Implementation plan (mysticat-architecture PR): https://github.com/adobe/mysticat-architecture/pull/83
- Wire-schema fixture: https://github.com/adobe/mysticat-architecture/blob/spec/ad-intent-page-context-fields/products/aso/fixtures/ad-intent-mismatch-sqs-body.json
- Companion mystique PR: see https://git.corp.adobe.com/experience-platform/mystique (Adobe GHE)

## Test plan

- [x] 5 new tests added under `'Paid Keyword Optimizer opportunity mapper (cluster format)'` covering: pass-through, `undefined` SQS keys (old mystique), explicit `null` regression guard, existing-field regression, and top-level `title`/`guidance` preservation.
- [x] `npm test` — 8220 passing, 0 failing.
- [x] `.nycrc.json` 100/100/100 coverage gate passes.
- [ ] Manual smoke once mystique companion PR is deployed to ephemeral and a fresh audit produces a new opportunity.

### DIFF
--- a/src/paid-keyword-optimizer/guidance-opportunity-mapper.js
+++ b/src/paid-keyword-optimizer/guidance-opportunity-mapper.js
@@ -87,6 +87,21 @@ export function mapToKeywordOptimizerOpportunity(siteId, audit, message) {
     clusterResults = [],
     portfolioMetrics = {},
   } = guidanceBody;
+  /**
+   * Page-level context fields emitted by mystique (additive, both optional).
+   *
+   * Distinct from per-cluster
+   * `gapAnalysis.{keywordToPageGap,adToPageGap}.relevantExtractedTopics`,
+   * which are LLM-filtered subsets of `pageTopics` filtered for cluster-
+   * specific intent relevance. `pageTopics` here is the full page-level set.
+   *
+   * Use `??` (nullish coalescing) instead of ES destructure defaults so that
+   * a future regression emitting `pageTopics: null` (instead of omitting the
+   * key) still results in an empty array on the consumer side. ES destructure
+   * defaults only fire for `undefined`, not `null`.
+   */
+  const resolvedPageHeading = guidanceBody.resolvedPageHeading ?? null;
+  const pageTopics = guidanceBody.pageTopics ?? [];
   const { langfuseTraceId, langfuseTraceUrl } = guidanceBody?.observability || {};
 
   const hasConflictingHeadlineRecommendations = clusterResults.filter(
@@ -132,6 +147,8 @@ export function mapToKeywordOptimizerOpportunity(siteId, audit, message) {
       totalClusters,
       misalignedClusters,
       totalMisalignedSpend,
+      resolvedPageHeading,
+      pageTopics,
     },
     status: 'NEW',
     tags: [
--- a/test/audits/paid-keyword-optimizer/opportunity-mapper.test.js
+++ b/test/audits/paid-keyword-optimizer/opportunity-mapper.test.js
@@ -58,6 +58,7 @@ function createClusterMessage({
   portfolioMetrics = { totalSpend: 1000 },
   langfuseTraceId = 'trace-123',
   langfuseTraceUrl = 'https://langfuse.example.com/trace/123',
+  extraBody,
 } = {}) {
   return {
     auditId: 'audit-id-123',
@@ -72,6 +73,7 @@ function createClusterMessage({
             langfuseTraceId,
             langfuseTraceUrl,
           },
+          ...(extraBody || {}),
         },
       }],
     },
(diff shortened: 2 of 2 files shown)


danieljchuser (commented):
Hey @danieljchuser,

Reviewed by a team of agents: senior staff engineer, distinguished architect, security researcher, and QA. All four converge: this is a clean, trivial-tier additive change that is ready to merge. Findings are limited to one Minor and a handful of forward-looking recommendations.

## Strengths

- The inline JSDoc in `src/paid-keyword-optimizer/guidance-opportunity-mapper.js:90-104` disambiguates page-level `pageTopics` from the per-cluster `gapAnalysis.{keywordToPageGap,adToPageGap}.relevantExtractedTopics` subsets, and justifies `??` over ES destructure defaults so `null` from a future mystique regression still collapses to a safe default. That comment will save the next reader a round trip through three repos.
- The choice of `??` over `||` correctly preserves falsy-but-valid scalar values for `resolvedPageHeading` and treats `null` like missing on a cross-repo wire boundary where producer and consumer evolve on different cadences.
- Change discipline is tight: additive only, no renames or retypes, contained to one mapper and its test file.
- Test coverage is genuinely behavioral, not coverage-padding. The five new `it()` blocks exercise both branches of each `??` (present, absent, explicit `null`), the existing-keys preservation, and the top-level `title`/`guidance` fields. The "explicit null regression guard" test in particular is the one most reviewers skip - it catches a future refactor back to `const { pageTopics = [] } = guidanceBody` that would silently break.
- No new dependencies, no secrets, no new endpoints, no auth surface change. CI green; codecov reports all modified lines covered.

## Issues

### Minor (Nice to Have)

- `test/audits/paid-keyword-optimizer/opportunity-mapper.test.js`, in the `passes resolvedPageHeading and pageTopics through to opportunity.data` block: the first array element is verified with `to.deep.equal({ topic, context, supportingQuote, paragraphIndex })` but the second element only spot-checks `topic` and `paragraphIndex`. If the mapper ever silently drops `context` or `supportingQuote` on later elements (e.g. via a `.map(({ topic, paragraphIndex }) => ...)`), this test still passes. Either deep-equal both elements or deep-equal the entire `pageTopics` array in one assertion.

## Recommendations

1. Hardening against malformed `pageTopics` from mystique. `?? []` only fires on `null`/`undefined`, so a non-array value (string, object, array-of-nulls) would pass verbatim into `opportunity.data.pageTopics` and surface as a UI error or DynamoDB write failure. A one-line `Array.isArray(guidanceBody.pageTopics) ? guidanceBody.pageTopics : []` would close the gap without growing scope.

2. Bound the length of LLM-extracted strings. `pageTopics[].supportingQuote` and `resolvedPageHeading` go straight through with no cap. DynamoDB items are limited to 400 KB; a degenerate mystique payload could fail the opportunity write and stall the pipeline for that URL. Either truncate at this seam or assert shape with the validation helpers already in the repo.

3. Pattern consistency at this seam. The new fields use `??`, while `clusterResults` and `portfolioMetrics` on lines 91-93 still use destructure-with-defaults that only fire on `undefined`. The JSDoc you added justifies `??` as the safer choice on a cross-repo wire boundary - worth either lifting the existing two to the same style in a small follow-up, or adding a one-line note that `??` is the new default for fields arriving from mystique, so the next contributor does not have to re-derive the decision.

4. Watch the shape of `opportunity.data`. It is now mixing url/page identifiers, portfolio metrics, observability ids, aggregate cluster counts, and page-level context at the top level. With this PR it stays manageable, but the next two or three additions will push it past the point where a flat bag is a good idea. Before more page-level fields land, consider grouping under `data.pageContext: { resolvedPageHeading, pageTopics }` (mirroring how `portfolioMetrics` is already nested). Cheaper to introduce now while there are two such fields than after there are six.

- UI-side escaping of `supportingQuote` and `resolvedPageHeading` when rendered in studio-ui or backoffice. Content is LLM-extracted from a customer landing page and can legitimately contain HTML-like substrings; XSS defense lives with the rendering surface.
- A contract test or shared JSON Schema on the mystique side asserting that `pageTopics` and `resolvedPageHeading` are emitted with the expected shape. Closes the loop end-to-end - this PR covers the consumer half, the producer half lives in the mystique repo.
- Keep `mysticat-architecture#83` and its SQS-body fixture as the single source of truth for these field semantics across mystique, audit-worker, and the UI.

## Assessment

Ready to merge: Yes.

Reasoning: 17 LOC of additive forwarding logic with thorough behavioral tests, defensible design choices documented inline, no security/auth/dependency surface change, CI green. The one Minor is a small tightening on a test assertion; everything else is forward-looking hardening that belongs in follow-up PRs.

## Next Steps

1. Optional: tighten the asymmetric `pageTopics[1]` assertion (the single Minor item) - one-line edit.
2. Merge.
3. Consider filing follow-up tickets for the four Recommendations (type guard, length cap, `??` pattern consistency, `data.pageContext` grouping).

MysticatBot-Dev (approved):
Approved. See the consolidated team review on this PR (staff engineer, distinguished architect, security researcher, QA) for details: one Minor on test assertion symmetry and four forward-looking recommendations, all non-blocking. Ready to merge.

PR #2552, opened 2026-05-22
feat: refactor to send spacecatOrgId to submitScrapeJob

rarescheseli:

## Related Issues

Thanks for contributing!


tb-adbe (approved):

MysticatBot-Dev (approved):
Hey @rarescheseli,

## Strengths

- **Clean elimination of the parallel HTTP path** (handler.js, removal of `submitDrsJobDirect`): Collapsing from two integration paths (raw HTTP + DRS client) to a single path through `drsClient` eliminates a maintenance fork where any DRS API evolution (auth changes, headers, rate limiting) had to be implemented twice. This is the right architectural move.
- **Single code path regardless of spacecatOrgId presence** (handler.js:530): Removing the conditional `spacecatOrgId ? null : DrsClient.createFrom(context)` eliminates a class of bugs where the two paths drift in behavior.
- **Selective retry with explicit non-idempotency acknowledgment** (handler.js JSDoc on `submitWithRetry`): Documenting that POST /jobs lacks an idempotency key and capping retries at one attempt shows good judgment. This prevents someone from "improving" this with exponential backoff and three retries later.
- **Clean constant extraction** (constants.js:59-60): `RETRIABLE_STATUSES` and `RETRY_DELAY_MS` make the retry policy visible and testable without digging into control flow.
- **Well-tested retry boundaries** (test file, Selective Retry section): Testing both retriable (502, TypeError) and non-retriable (400, 422) cases exercises the actual decision boundary, not just the happy path.

## Issues

### Minor (Nice to Have)

1. **handler.js:450 - TypeError check is broader than network errors.** `isRetriable` treats all `TypeError` instances as retriable. In the context of `fetch`, this is correct (DNS failure, connection refused), but a programming bug in the DRS client (e.g., accessing a property of undefined) would also trigger a retry. Given the single-retry cap and narrow scope, this is a pragmatic tradeoff - the retry is harmless and the error still surfaces. A comment noting "fetch throws TypeError on network failures" would help a future reader understand the intent.

2. **test file - No test for plain Error without `.status` property.** The `isRetriable` logic correctly rejects a generic `new Error('something')` (no status, not TypeError), but there is no test asserting this path fails immediately. The existing 400/422 tests always attach a numeric `.status`. Adding one test with a plain Error without status would guard against a future contributor accidentally broadening the retry scope.

## Recommendations

- The DRS client now implicitly needs to accept `spacecatOrgId` in the params object and translate it to `spacecat_org_id` in the request body. Worth confirming this contract is handled in the client (outside this repo's scope, but a deployment dependency).
- As retry tests accumulate, consider using `sinon.useFakeTimers` to avoid real 500ms delays in the test suite.

## Assessment

Ready to merge? Yes

This is a clean simplification that removes accidental complexity, consolidates integration logic into the shared DRS client, and introduces well-scoped selective retry. Full coverage confirmed by Codecov, already approved by a human reviewer.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 41s | <b>Cost:</b> $2.54 | <b>Commit:</b> <code>c3ede20d94f0dd4b17a7ee02d6fd2f238632d961</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2558, opened 2026-05-25
feat(preflight): excludedElementClasses to skip auth-gated link subtrees (SITES-44776)

clotton:
## Summary

Quick-fix mitigation for [SITES-44776](https://jira.corp.adobe.com/browse/SITES-44776) — Preflight is reporting valid SSO-authenticated / corp-network-only links as broken on customer pages (Walmart's homepage flags 14 such links; Micron showing similar symptoms).

Mirrors James Fong's [PR #2455](https://github.com/adobe/spacecat-audit-worker/pull/2455) pattern for the broken-internal-links audit. Customers list CSS class tokens under their site's `handlers.preflight.config.excludedElementClasses`; any anchor inside a matching subtree is pruned before link probing and therefore never reported as broken or insecure.

- New helper `filterExcludedElements` in `src/preflight/links-checks.js` (in-memory cheerio mutation only — the scraped HTML and customer's page are untouched)
- `runLinksChecks` accepts an `excludedElementClasses` option (defaults to `[]`, fully backward-compatible)
- `src/preflight/links.js` reads from `site.getConfig().getHandlers().preflight.config.excludedElementClasses`, normalizes via the helper Fong already exported, and threads through to **both** the broken-links extraction and the insecure-link (http→https) scan — single consistent ignore semantics across the audit's opportunity types

This is parallel to (and intentionally lighter than) the broader \"unverifiable\" category in spec PR #2533 — that's the systemic improvement; this is the same-week customer mitigation that requires no MFE coordination.

## Wire-format example

Customer adds to their SpaceCat site config:
\`\`\`json
{
  \"handlers\": {
    \"preflight\": {
      \"config\": {
        \"excludedElementClasses\": [\"cmp-feature-apps\"]
      }
    }
  }
}
\`\`\`

Aliases accepted: \`[\".cmp-feature-apps\"]\` (leading dot stripped), comma-separated strings, multiple tokens. No code deploy needed per customer; next preflight run honors the field.

## Test plan

- [x] Unit tests for filter behavior: nested wrappers, multi-class tokens, empty/whitespace class attribute, no-op when unset (`test/preflight/links-checks.test.js`)
- [x] Unit tests for site-config plumbing: missing handler, present handler, missing \`getConfig\`, undefined return (`test/preflight/links.test.js`)
- [x] Unit tests for insecure-link scan filtering
- [x] Lint clean
- [x] 100% line + branch coverage on changed files
- [x] Full suite passes (8233 tests)
- [ ] Pre-merge: Layer 1 verification script against Guy's captured uswire homepage HTML (Jira attachment 18132239) to confirm the 14 corp-only links are pruned on real DOM
- [ ] Post-deploy dev env: PATCH a test site, trigger preflight, verify before/after
- [ ] Production canary: Walmart uswire site with \`cmp-feature-apps\` token, watch Coralogix, before/after delta on the reported homepage URL

## Notes for reviewers

- The in-memory cheerio mutation is fully scoped to the audit's own working tree — other preflight audits (canonical, metatags, headings, a11y, readability) re-parse the same raw HTML string into their own independent cheerio trees and are completely unaffected.
- Did **not** apply this to internal-links opportunity \`per-check\` granularity — kept the ignore semantics consistent across both opportunity types in this audit. If a customer wants per-check control later, easy to add as a nested option without breaking the current contract.


src/preflight/links.js
```diff
@@ -89,9 +97,22 @@ export default async function links(context, auditContext) {
...
+  const excludedHrefPatterns = normalizeHrefPatterns(preflightHandlerConfig?.excludedHrefPatterns);
   const { auditResult } = await runLinksChecks(auditUrls, scrapedObjects, context, {
     pageAuthToken,
+    excludedElementClasses,
+    excludedHrefDomains,
+    excludedHrefPatterns,
   });
+  // Pre-compile patterns once for the insecure-link scan below — keeps both
+  // passes on the same compiled list and surfaces bad-regex warnings only once.
+  const compiledHrefPatterns = compileHrefPatterns(excludedHrefPatterns, log);
```
MysticatBot-Dev: Patterns are compiled here a second time (they were already compiled inside `runLinksChecks` above). This causes duplicate `log.warn` messages for invalid patterns, contradicting the comment on lines 112-113. Consider having `runLinksChecks` return its compiled patterns, or compile once here and pass the compiled `RegExp[]` into `runLinksChecks` via options.

src/preflight/links-checks.js
```diff
@@ -15,6 +15,160 @@ import { load as cheerioLoad } from 'cheerio';
...
+}
+
+/**
+ * Compile customer-supplied regex strings once per audit run. Bad patterns
+ * are logged and dropped so subsequent matches don't repeatedly throw.
+ *
+ * @param {string[]} patterns - Raw regex strings from site config
+ * @param {{ warn: Function }} log - Logger
+ * @returns {RegExp[]}
+ */
```
MysticatBot-Dev: Customer-supplied regex is compiled and later executed against every href on every page. A pathological pattern with catastrophic backtracking (e.g. `(a+)+$`) could block the Lambda event loop. Quick mitigation: cap the number of patterns (max 10) and individual pattern length (max 200 chars) in `normalizeHrefPatterns`. A full solution (e.g. `safe-regex2` or timeout wrapper) can follow.

MysticatBot-Dev (dismissed):
Hey @clotton,

## Strengths

- **Clean function decomposition** (`src/preflight/links-checks.js`): Each concern (DOM subtree pruning, domain suffix-matching, regex compilation, composite check) gets its own small, testable, exported function with a clear single responsibility. Mirrors the established pattern from PR #2455 and makes the code easy to reason about.
- **Adversarial boundary safety** (`src/preflight/links-checks.js:matchesExcludedDomain`): The suffix-match using `hostname === d || hostname.endsWith('.'+d)` correctly prevents `evilwal-mart.com` from matching `wal-mart.com`. The test suite explicitly covers this case.
- **Defensive normalization** (`normalizeHrefDomains`, `normalizeHrefPatterns`): Accepts loose customer config (strings, arrays, protocol prefixes, trailing slashes, case variations, duplicates) and produces a consistent internal representation. Right approach for customer-facing config set via API PATCH.
- **Fail-safe regex handling** (`compileHrefPatterns`): Invalid regexes are caught, warned, and skipped rather than crashing the audit. Patterns are compiled once per call site and reused.
- **Deep-to-shallow removal order** (`filterExcludedElements`): Sorting by DOM depth before removal prevents Cheerio detached-node issues with nested excluded subtrees.
- **Excellent test coverage**: ~650 lines of tests covering edge cases (nested exclusions, empty class attrs, adversarial hostnames, invalid regex, combined filter composition, insecure-link scan path). Codecov confirms 100% on changed files.

## Issues

### Important (Should Fix)

**1. Double compilation of regex patterns produces duplicate log warnings**

`src/preflight/links.js` line ~114: `excludedHrefPatterns` is passed as raw strings into `runLinksChecks`, which calls `compileHrefPatterns(excludedHrefPatterns, log)` internally. Then immediately after, `links.js` calls `compileHrefPatterns(excludedHrefPatterns, log)` again for the insecure-link scan. If a customer supplies an invalid regex, `log.warn` fires twice with the identical message. The code comment says "surfaces bad-regex warnings only once" but that is incorrect - both calls share the same raw input and the same logger.

Fix: Either (a) have `runLinksChecks` return its compiled patterns so the caller can reuse them, or (b) compile once in `links.js` before calling `runLinksChecks` and pass compiled `RegExp[]` in via options instead of raw strings. Option (a) is the smaller change.

**2. No ReDoS mitigation on customer-supplied regex patterns**

`src/preflight/links-checks.js` (`compileHrefPatterns` / `matchesExcludedPattern`): Customer-supplied regex strings are compiled with `new RegExp(p)` and then executed via `.test(href)` against every link on every page. A catastrophic-backtracking pattern (e.g. `(a+)+$`) matched against a long input would block the event loop. In a Lambda running audits with hundreds of links per page, this could hit the Lambda timeout.

The current defense (catch compile errors) only protects against invalid syntax, not valid-but-pathological patterns. The blast radius is limited to the customer's own audit run (not a privilege escalation), which is why this is Important rather than Critical.

Quick mitigation (same PR): Cap the number of patterns (e.g. max 10) and cap pattern length (e.g. 200 chars) in `normalizeHrefPatterns`. This eliminates the worst-case scenarios with a two-line addition. A full ReDoS-safe solution (e.g. `safe-regex2` validation or timeout wrapper) can be a fast-follow.

### Minor (Nice to Have)

**3. `normalizeHrefDomains` does not strip port numbers**

`src/preflight/links-checks.js` (`normalizeHrefDomains`): The function splits on `/` to strip paths but does not strip port numbers. If a customer writes `"internal.corp.com:8443"`, the domain entry retains the port, but `matchesExcludedDomain` uses `new URL(href).hostname` which strips ports. So `matchesExcludedDomain('https://internal.corp.com:8443/x', ['internal.corp.com:8443'])` would fail because hostname is `internal.corp.com` but the entry is `internal.corp.com:8443`. Not likely in practice (most customers write bare hostnames), but worth a `.split(':')[0]` in the normalization chain or a note in the README.

## Recommendations

- Consider extracting the compile-once pattern into a small factory (`buildHrefFilters(domains, patterns, log)`) that both `runLinksChecks` and the insecure-link scan can share - ensures single compilation and removes the deceptive comment.
- Log a single summary line when exclusion filters are active (e.g. `[preflight-audit] Exclusion filters active: 2 classes, 3 domains, 1 pattern`) so operators can confirm the config took effect without digging through debug logs.
- Consider validating customer config at write time (SpaceCat PATCH endpoint) rather than only at audit time. A customer setting an invalid regex currently only discovers the problem on the next audit run. Out of scope for this PR but worth tracking.

## Assessment

Ready to merge? With fixes.

The feature solves a real customer pain point (Walmart homepage flagging 14 false-positive broken links) with a clean, composable config surface that mirrors the existing broken-internal-links audit pattern. The code is well-structured, thoroughly tested, and defensively written.

The Important fixes are both straightforward: double-compilation is a small refactor to the call interface, and the ReDoS cap is a two-line guard in the normalization function. Neither is a correctness bug - they are operational quality improvements that should land before this ships to production.

## Next Steps

1. Fix the double compilation - compile patterns once and share the compiled list between `runLinksChecks` and the insecure-link scan.
2. Add count/length caps to `normalizeHrefPatterns` to mitigate ReDoS risk.
3. Minor port-number fix is optional.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 43s | <b>Cost:</b> $4.90 | <b>Commit:</b> <code>ba5bd06e05c742465a13b7fc4e31a8828f5153e0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

GeezerPelletier (approved):
Quick-fix is well-executed. Three composable knobs, security-correct dot-boundary suffix match, adversarial lookalike hostname test explicitly present, patterns compiled before the loops. Good defensive README docs on the GET/splice/PATCH config update pattern. A few minor observations for follow-up if desired:

1. Two compileHrefPatterns calls per run: The comment says "pre-compile once for the insecure-link scan below" but the function is also called inside runLinksChecks. Not a bug, but the two call sites undercut the "compile once" intent - could hoist both to a single call before passing into the functions.

2. normalizeHrefPatterns vs normalizeHrefDomains single-string handling: Domains splits on comma; patterns wraps a bare string as a one-element array. Probably intentional (regex strings can legitimately contain commas) - a brief comment in normalizeHrefPatterns explaining why it does not split would help future readers.

3. filterExcludedElements sort rationale: The deepest-first sort is the right choice for nested excluded wrappers, but a one-liner comment explaining why (i.e., avoid double-removal of already-removed ancestors) would help the next person reading it.

None of these are blockers. Good fix for SITES-44776.

GeezerPelletier (approved):
Thanks for the updates @clotton - narrowing to excludedElementClasses only is the right call, and the sort rationale comment is a nice touch. All good from my end.

PR #2500, opened 2026-05-11
feat: generate path-level prerender suggestions in audit worker

ssilare-adobe:
## Summary

Implements path-level prerender suggestion generation in the audit worker — the third tier between per-URL and domain-wide suggestions. Groups URLs by first path segment (e.g. `/products/*`, `/blog/*`), scores each group using the same formula as the rcv-scoring-dashboard, and emits deployable path suggestions.

### What it does

- **Generates path suggestions** — groups per-URL audit results by first path segment, qualifies via pluggable strategy (default: `RcvPathQualificationStrategy`), persists qualifying paths with `allowedRegexPatterns`
- **Opt-in per site** — controlled by `prerender.pathSuggestionsEnabled` config flag (defaults `false`)
- **Skips when domain-wide is deployed** — no path suggestions generated if `/*` is already deployed (redundant)
- **Preserves deployed/active paths** — re-audits keep deployed or active path suggestions; only metrics are refreshed
- **Marks coverage** — after sync, marks matching per-URL `NEW` suggestions as `coveredByPattern` for deployed paths (including exact root segment match, e.g. `/products`)
- **Marks path suggestions as coveredByDomainWide** — when domain-wide is deployed, existing path suggestions get `coveredByDomainWide` set
- **Self-heals** — clears stale `coveredByPattern` references pointing to undeployed/deleted paths on every audit run
- **Error resilient** — all `saveMany` calls wrapped in try-catch with error logging

### Path suggestion data shape (stored in DB)

```json
{
  "type": "CONFIG_UPDATE",
  "rank": 0,
  "status": "NEW",
  "data": {
    "url": "https://example.com/products/*",
    "allowedRegexPatterns": ["/products/*"],
    "score": 2.84,
    "contentGainRatio": 1.8,
    "wordCountBefore": 9600,
    "wordCountAfter": 19200,
    "aiReadablePercent": 78.5
  }
}
```

Detection: `Array.isArray(data.allowedRegexPatterns) && !data.isDomainWide`
Key: `{allowedRegexPatterns[0]}|prerender` (e.g. `/products/*|prerender`)

### Changes summary

| Category | Description | What changed |
|----------|-------------|--------------|
| **New file** | Core path suggestion logic | `src/prerender/features/path-suggestions/path-suggestions.js` — `buildPathTypeSuggestions`, `findPreservablePathSuggestions`, `markSuggestionsAsCoveredByPaths` |
[...]

PR #2559, opened 2026-05-25
feat(prerender): scope domain-wide suggestion and top-page filtering to site subpath

dhavkuma:
## Problem

For subpath sites (e.g. `nba.com/kings` registered as its own site object), the prerender audit had two issues:

1. **Critical — wrong CDN rule scope**: `allowedRegexPatterns` and `pathPattern` in the domain-wide suggestion were hardcoded to `'/*'`. When a customer deploys the domain-wide suggestion for `nba.com/kings`, the CDN prerender rule would apply to **all of nba.com**, not just `/kings/*`.

2. **Safety — organic URL bleed**: `SiteTopPage.allBySiteIdAndSourceAndGeo` is scoped by `siteId`, but GSC can return domain-wide data. Top-page URLs not belonging to the subpath would get submitted for scraping under the wrong site.

### `src/prerender/handler.js`

**`getSitePathPattern(baseUrl)`** — new helper that derives the correct glob pattern from the site's base URL pathname:
- `https://nba.com` → `'/*'`
- `https://nba.com/kings` → `'/kings/*'`

**`isUrlUnderSiteBase(url, baseUrl)`** — new helper that returns `true` when a URL falls within the site's registered path; no-op for root sites.

**`prepareDomainWideAggregateSuggestion`** — uses `getSitePathPattern(baseUrl)` for both `allowedRegexPatterns` and `pathPattern` instead of hardcoded `'/*'`.

**`getDomainWideSuggestionUrl`** — labels subpath sites as `'All Subpath URLs'` instead of `'All Domain URLs'`.

**`getTopOrganicUrlsFromSeo`** — filters top-page URLs through `isUrlUnderSiteBase` before returning.

## What was already working

- Agentic URLs from Athena: `buildSiteFilters` in `cdn-utils.js` already generates a `REGEXP_LIKE` SQL clause scoped to the site's pathname — no change needed.
- All DB queries (`Opportunity`, `Suggestion`, `PageCitability`): already scoped by `siteId`, which is unique per subpath site.
- Guidance handler Mystique URL matching: `toPathname` keying works correctly with subpath pathnames.
- S3 status.json key: uses `siteId`, naturally isolated per site.

## Testing

- [ ] Unit tests for `getSitePathPattern` and `isUrlUnderSiteBase`
- [ ] Unit test for `prepareDomainWideAggregateSuggestion` with a subpath `baseUrl`
- [ ] Verify domain-wide suggestion for `nba.com/kings` has `pathPattern: '/kings/*'`
- [ ] Verify root site `nba.com` still gets `pathPattern: '/*'`

### DIFF
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -354,7 +354,10 @@ async function getTopOrganicUrlsFromSeo(context, limit = TOP_ORGANIC_URLS_LIMIT)
 
 async function getTopAgenticUrls(site, context, limit = TOP_AGENTIC_URLS_LIMIT) {
   try {
-    return await getTopAgenticLiveUrlsFromAthena(site, context, limit);
+    const urls = await getTopAgenticLiveUrlsFromAthena(site, context, limit);
+    // Filter to subpath scope — Athena query is domain-wide so results may include
+    // URLs outside this site's baseURL path (e.g. nba.com/scores for nba.com/timberwolves).
+    return urls.filter((url) => isUrlUnderSiteBase(url, site.getBaseURL()));
   } catch (e) {
     context.log.warn(`${LOG_PREFIX} Failed to fetch agentic URLs: ${e.message}. baseUrl=${site.getBaseURL()}`);
     return [];
@@ -741,7 +744,7 @@ async function sendPrerenderGuidanceRequestToMystique(auditUrl, auditData, oppor
     const queue = env.QUEUE_SPACECAT_TO_MYSTIQUE;
     await sqs.sendMessage(queue, {
       type: 'guidance:prerender',
-      url: baseUrl, //todo:dhavkuma
+      url: baseUrl,
       siteId,
       auditId,
       deliveryType,
@@ -1110,7 +1113,7 @@ export async function createScrapeForbiddenOpportunity(auditUrl, auditData, cont
  */
 async function prepareDomainWideAggregateSuggestion(
   preRenderSuggestions,
-  baseUrl,//todo:dhavkuma
+  baseUrl,
   context,
 ) {
   const { log } = context;
@@ -1510,7 +1513,7 @@ export async function uploadStatusSummaryToS3(auditUrl, auditData, context) {
     ).length;
 
     const statusSummary = {
-      baseUrl: auditUrl,//todo:dhavkuma
+      baseUrl: auditUrl,
       siteId,
       auditType: AUDIT_TYPE,
       scrapeJobId: scrapeJobId || existingStatus.scrapeJobId || null,
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -48,7 +48,52 @@ const { AUDIT_STEP_DESTINATIONS } = Audit;
 const AUDIT_ERROR_MESSAGE = 'Audit failed';
 
 // Domain-wide suggestion URL format (sync scrapedUrlsSet + prepareDomainWideAggregateSuggestion)
-const getDomainWideSuggestionUrl = (baseUrl) => `${baseUrl}/* (All Domain URLs)`;
+const getDomainWideSuggestionUrl = (baseUrl) => {
+  try {
+    const { pathname } = new URL(baseUrl);
+    const label = pathname === '/' ? 'All Domain URLs' : 'All Subpath URLs';
+    return `${baseUrl}/* (${label})`;
+  } catch {
+    return `${baseUrl}/* (All Domain URLs)`;
+  }
+};
+
+/**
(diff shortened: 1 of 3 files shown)


ssilare-adobe:
Duplicated PR. 

Original PR : https://github.com/adobe/spacecat-audit-worker/pull/2500

MysticatBot-Dev (changes requested):
Hey @dhavkuma,

## Strengths

- `src/prerender/handler.js:45-55` - `isUrlUnderSiteBase` has clean path-prefix matching. The `startsWith(`${normalized}/`)` check with the trailing slash correctly prevents false positives like `/kingston` matching `/kings`. The early return for root sites makes the no-op case explicit.
- `src/prerender/handler.js:27-35` - `getSitePathPattern` handles trailing-slash normalization correctly, preventing double-slash patterns like `/kings//*`.
- `src/prerender/handler.js:341-346` - Filter placement before `.slice(0, limit)` is correct. Subpath sites get the top N URLs *that belong to this site*, rather than top N with out-of-scope ones removed afterward (which could silently return fewer results).
- The approach of deriving scope from the existing `baseUrl` data model (rather than introducing new config or site properties) is architecturally sound - zero new state, zero migration needed.
- PR description is excellent - clearly articulates both bugs, explains what was already working, and provides a concrete test plan.

## Issues

### Important (Should Fix)

**1. Tests absent for new production logic**

The PR adds three new helper functions (`getSitePathPattern`, `isUrlUnderSiteBase`, rewritten `getDomainWideSuggestionUrl`) with non-trivial URL parsing and path-matching logic, plus integration changes in `getTopOrganicUrlsFromSeo` and `prepareDomainWideAggregateSuggestion`. No test files are modified or added. The PR description itself lists four unchecked test items, confirming these are needed but not yet written.

This is a bug-fix PR addressing a critical production issue (wrong CDN rule scope). Shipping without tests means the fix is unverified and could regress silently.

**How to fix:** Add unit tests covering at minimum:
- `getSitePathPattern`: root URL, subpath URL, trailing-slash subpath, invalid URL
- `isUrlUnderSiteBase`: URL matching subpath, URL outside subpath, root site (no-op), URL exactly equal to basePath, invalid URLs
- `prepareDomainWideAggregateSuggestion` with a subpath baseUrl, asserting `pathPattern` and `allowedRegexPatterns` reflect the subpath

**2. `//todo:dhavkuma` debug markers committed on unchanged lines**

`src/prerender/handler.js:741`, `src/prerender/handler.js:1110`, `src/prerender/handler.js:1510`, `src/prerender/handler.js:1724` - Four personal TODO comments are appended to lines with no functional change. They communicate nothing actionable to future readers and make the PR look unfinished. If these mark pending work, capture it in a follow-up ticket. If they're leftover investigation markers, remove them.

### Minor (Nice to Have)

**3. Silent catch blocks in new helpers**

`src/prerender/handler.js:15-17`, `src/prerender/handler.js:32-34`, `src/prerender/handler.js:52-54` - All three helpers catch URL parsing failures and return safe defaults without logging. The fallback choices are defensively correct (fail-open), but if `site.getBaseURL()` ever returns an invalid URL, you'll have no signal that a site record is corrupted until someone notices the CDN rule is too broad. The helpers don't have access to `log` so this is a design tradeoff, not a blocker.

## Recommendations

- Coordinate with PR #2500 (flagged as a duplicate by ssilare-adobe, still open). Whichever lands second will have merge conflicts in the same functions. Resolve which PR moves forward before either merges.
- Validate that the downstream CDN consumer interprets `allowedRegexPatterns` values (e.g., `/kings/*`) as globs rather than raw regex. Characters valid in URL pathnames (parentheses, dots) could produce overly-permissive patterns if the field is interpreted as regex.

## Assessment

Ready to merge? No

The core logic is correct and well-implemented, but the PR ships zero tests for critical bug-fix code, CI is failing, and the `//todo:dhavkuma` markers indicate unfinished state. Address the Important items and get CI green.

## Next Steps

1. Remove all four `//todo:dhavkuma` comments.
2. Add unit tests for the new helpers and the updated integration points.
3. Resolve the duplicate-PR situation with PR #2500.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 59s | <b>Cost:</b> $2.41 | <b>Commit:</b> <code>91ddad8c02109744619cddec5b19432ee3d2c783</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dhavkuma:
changes scoped with prerender moved to https://github.com/adobe/spacecat-audit-worker/pull/2605

PR #2569, opened 2026-05-28
feat(paid-keyword-optimizer): pass adAnalysis through to suggestion.data

danieljchuser:
## Summary

Reads the new `cluster.adAnalysis` field (emitted by mystique PR #2420 — see [`mysticat-architecture#97`](https://github.com/adobe/mysticat-architecture/pull/97) for the cross-repo wire-schema contract) in `mapClusterToSuggestion` and passes it through verbatim to `data.cluster.adAnalysis` on each suggestion.

Per the spec:
- `cluster.adAnalysis = { isBranded, intentType, adPromise } | null`
- `null` is a deliberate value (mystique emits it when the cluster lacks ad-copy signal), not a missing-key default. The UI relies on the null to hide the "What the ad promises" block.

## Why `?? null` and not `|| {}`

The mapper uses nullish coalescing so both cases collapse to the same wire shape:
- **`undefined`** — older mystique deployments don't emit the key. Coerces to `null`.
- **`null`** — explicit; mystique declined to produce a block for this cluster (insufficient ad-copy signal per the spec's Decision #3). Stays `null`.

`|| {}` would silently coerce the deliberate `null` to `{}` and break the UI's hide-when-missing branch. The pre-existing `keywordAnalysis: keywordAnalysis || []` line in this same function is a known latent oddity (defaults an object to an array); explicitly not touching it here to keep this PR's diff scoped — see the mystique PR #2420 spec comment on the same point.

## Files changed

- **`src/paid-keyword-optimizer/guidance-opportunity-mapper.js`** — destructure `adAnalysis` from the SQS cluster body and append it to `data.cluster.adAnalysis` with the `?? null` guard. 7 lines added.

## Tests (3 new + 1 existing extended)

- `"includes all cluster fields"` extended to assert `adAnalysis` pass-through with a populated `{isBranded, intentType, adPromise}` block (spec B.3-(1)).
- `"defaults missing optional fields"` extended to assert `adAnalysis` defaults to `null` when the cluster omits the key — older mystique on a rolling-deploy window (spec B.3-(2)).
- New `"preserves explicit null adAnalysis without coercion (spec B.3-(3))"` — pins the `?? null` vs. `|| {}` contract.
- The extended `"includes all cluster fields"` assertion also serves as the B.3-(4) regression guard for existing keys.

52 tests passing in `opportunity-mapper.test.js` (was 51, +1).

## Test plan

- [x] `npx mocha test/audits/paid-keyword-optimizer/opportunity-mapper.test.js` — 52 passing
- [x] `npm run lint` on the touched files — clean (pre-commit lint-staged passed on the diff)
- [ ] Full suite + 100% coverage — left to CI. **Local verification was constrained** by a stale `.claude/worktrees/page-context-fields/` git worktree that c8's `"all": true` config picks up as duplicate `src/**/*.js` (inflating the coverage denominator) and ESLint walks into (inflating the error count). CI runs against a clean checkout so neither effect applies there.

## Spec / cross-repo coordination

- Spec: [`mysticat-architecture#97`](https://github.com/adobe/mysticat-architecture/pull/97) — Part B (this PR), Decision #3 (producer-internal threshold), B.2 (no shared validator).
- Producer side: [`mystique#2420`](https://git.corp.adobe.com/experience-platform/mystique/pull/2420) (Adobe GHE).
- Upstream wire data: [`spacecat-import-worker#760`](https://github.com/adobe/spacecat-import-worker/pull/760) (merged) — surfaces the `serpDescription` + `visibleUrl` per-keyword fields mystique now feeds to its prompt.

## Backwards compatibility

Purely additive. No existing field is renamed, removed, or retyped. Old opportunities created before this ships will simply lack `adAnalysis`; UI consumers MUST treat as optional (same hide-when-missing pattern v3 already uses for `resolvedPageHeading`).

### DIFF
--- a/src/paid-keyword-optimizer/guidance-opportunity-mapper.js
+++ b/src/paid-keyword-optimizer/guidance-opportunity-mapper.js
@@ -181,6 +181,7 @@ export function mapClusterToSuggestion(context, opportunityId, cluster) {
     gapAnalysis,
     overallAlignmentScore,
     keywordAnalysis,
+    adAnalysis,
     recommendation,
     rank,
   } = cluster;
@@ -214,6 +215,12 @@ export function mapClusterToSuggestion(context, opportunityId, cluster) {
         gapAnalysis: gapAnalysis || {},
         overallAlignmentScore: overallAlignmentScore || null,
         keywordAnalysis: keywordAnalysis || [],
+        // `?? null` (not `|| {}`): mystique emits an explicit `null` when the
+        // cluster lacks sufficient ad-copy signal, and the UI hides the
+        // "What the ad promises" block on that signal. Coercing null to {}
+        // would silently break the hide-when-missing branch. See spec
+        // products/aso/ad-intent-cluster-ad-analysis.md (B.1 + Decision #3).
+        adAnalysis: adAnalysis ?? null,
       },
       ...(recommendationType && { recommendationType }),
       ...(cleanedRecommendation && { recommendation: cleanedRecommendation }),
--- a/test/audits/paid-keyword-optimizer/opportunity-mapper.test.js
+++ b/test/audits/paid-keyword-optimizer/opportunity-mapper.test.js
@@ -786,6 +786,11 @@ describe('Paid Keyword Optimizer opportunity mapper (cluster format)', () => {
           overallAlignmentScore: 'fair',
           keywordAnalysis: [{ keyword: 'buy shoes', score: 0.5 }],
         }),
+        adAnalysis: {
+          isBranded: true,
+          intentType: 'navigational',
+          adPromise: 'Promotes the Twilio brand',
+        },
         rank: 2,
       };
 
@@ -802,6 +807,12 @@ describe('Paid Keyword Optimizer opportunity mapper (cluster format)', () => {
       expect(result.data.cluster.gapAnalysis).to.deep.equal({ severity: 'medium' });
       expect(result.data.cluster.overallAlignmentScore).to.equal('fair');
       expect(result.data.cluster.keywordAnalysis).to.deep.equal([{ keyword: 'buy shoes', score: 0.5 }]);
+      // adAnalysis passed through verbatim (spec B.3-(1))
+      expect(result.data.cluster.adAnalysis).to.deep.equal({
+        isBranded: true,
+        intentType: 'navigational',
+        adPromise: 'Promotes the Twilio brand',
+      });
     });
 
     it('lifts recommendation.type to recommendationType and removes type from nested object', () => {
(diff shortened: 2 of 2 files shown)


MysticatBot-Dev (approved):
Hey @danieljchuser,

## Strengths

- `src/paid-keyword-optimizer/guidance-opportunity-mapper.js:217-222` — The inline comment documenting `?? null` over `|| {}` (with spec citation and the downstream UI consequence) captures the deliberate operator choice at the point of decision. A future maintainer who tries to "simplify" this to `|| {}` will hit the warning before doing damage.
- `src/paid-keyword-optimizer/guidance-opportunity-mapper.js:184` — Conforms to the established destructure-then-forward pattern already used for `gapAnalysis`, `overallAlignmentScore`, `keywordAnalysis`. The mapper stays a thin transport boundary; no field-specific logic accretes here.
- `test/audits/paid-keyword-optimizer/opportunity-mapper.test.js` — Test coverage is symmetric across the three cluster states that matter: populated block (spec B.3-(1)), missing key on rolling-deploy → null (B.3-(2)), and explicit null preserved (B.3-(3)). Each carries a spec citation in its comment.
- The B.3-(2) test uses a hand-rolled minimal cluster rather than `makeCluster()`, so `adAnalysis` is genuinely `undefined` — pinning the `undefined → null` branch independent of factory defaults. Both branches of `?? null` are independently covered.
- Both new/extended tests assert `to.have.property('adAnalysis')` *and* `to.be.null`. This catches the case where someone refactors the mapper to omit the key entirely (which would break the UI's hide-branch if it uses `in` checks).
- Change discipline is solid. The author explicitly leaves the sibling `keywordAnalysis: keywordAnalysis || []` line alone (acknowledged latent oddity). Out of scope for a contract pass-through PR; correct call.
- Honors spec Decision #3 (producer-internal threshold). spacecat-audit-worker stays out of the business of judging ad-copy signal quality, keeping the responsibility boundary clean.

## Recommendations

- **Forward-looking, not for this PR**: The `data.cluster` object now carries four fields with three different null-handling strategies — `gapAnalysis || {}`, `keywordAnalysis || []`, `overallAlignmentScore || null`, and `adAnalysis ?? null`. Each individual choice is defensible (the new one most clearly so, since it preserves a semantic null), but the cluster contract is drifting toward "every field has its own coercion rule." When a fifth field arrives, the next author will read four precedents and pick one. Worth considering a small `normalizeCluster(cluster)` helper that owns these decisions in one place. Flagging for the next time this surface is touched.
- **Optional defensive test**: Consider pinning current behavior for `adAnalysis: {}` (empty object), which would pass through verbatim under `?? null`. Wire contract is `{ isBranded, intentType, adPromise } | null`, so `{}` is technically off-contract. Producer-side responsibility primarily; a one-line test in this repo would surface any future regression intentionally. Optional.

- `adAnalysis.adPromise` is LLM-generated free text. XSS responsibility sits on the UI rendering the "What the ad promises" block (escape-on-render, not escape-on-store). UI-repo concern.
- If `adPromise` is later fed back into a downstream LLM prompt (follow-up audit, autofix), that consumer needs prompt-injection guards. Belongs to whichever service does the second-stage prompt construction.
- Producer-side structural validation / length caps on LLM output belong in mystique, not here.

## Assessment

Ready to merge? **Yes**

Reasoning: A tight, well-scoped 7-line pass-through that conforms to the established pattern in the same file, honors the spec's producer-owned-threshold decision, protects the null-vs-empty semantic with both code (`?? null`) and an inline comment, and pins both relevant branches with independent tests. No security-relevant change, no architectural debt, no test gap on the contract under test.

## Next Steps

No required changes. Recommendations above are optional / forward-looking.

— Reviewed by a 4-agent panel (staff-engineer, architect, security, qa). All four independently concluded ready to merge.

PR #2574, opened 2026-05-28
feat(semantic-value-visibility): dispatch top 50 URLs to Mystique per audit run

yevheniia-severinovska:
## Summary

- Replaces single base-URL dispatch with multi-URL selection using the fleet-standard `getMergedAuditInputUrls` (agentic Athena CDN logs as primary, SEO top pages as fallback, capped at 50 URLs)
- Pre-filters candidates via HTTP GET: only pages containing an `` tag are dispatched to Mystique, avoiding costly agent runs on text-only pages
- Sends one SQS message per qualifying URL — each message triggers an independent Lambda/Mystique invocation, providing automatic parallelism at no extra infrastructure cost

## Test plan

- [x] `auditRunner` returns only URLs whose fetched HTML contains `` tags
- [x] URLs where `fetch` throws are skipped (warning logged, not dispatched)
- [x] `sendToMystique` sends exactly one SQS message per qualifying URL
- [x] `sendToMystique` logs warning and exits early when URL list is empty
- [x] `getTopPages` callback queries `SiteTopPage` with correct args
- [x] `topOrganicLimit` is set to 50
- [x] 100% statement/branch/function/line coverage on `handler.js`


MysticatBot-Dev (changes requested):
Hey @yevheniia-severinovska,

## Strengths

- Well-scoped change: one handler, one test file, clear intent, no unrelated changes mixed in
- Good adoption of fleet-standard `getMergedAuditInputUrls` - aligns this audit with the platform pattern and makes future URL-sourcing improvements available automatically
- Fan-out via individual SQS messages leverages existing Lambda concurrency without new infrastructure
- Per-URL error handling in the pre-filter (catch + log.warn + skip) is the correct pattern for a best-effort filter step
- Empty-URL guard in `sendToMystique` (line 77-79) prevents unnecessary SQS work with observable logging
- Comprehensive test coverage with behavior-focused assertions covering happy path, filtering, errors, and edge cases

## Issues

### Important (Should Fix)

1. **src/semantic-value-visibility/handler.js:53 - HTTP response status not checked before reading body**

   `fetch(url)` does not check `response.ok` before parsing the body. A 403, 404, or 500 response will have its error page HTML regex-tested for ``. Many error pages contain images (logos, illustrations), so inaccessible pages get dispatched to Mystique as "qualifying."

   This produces false positives: Mystique receives URLs that return error content rather than the actual page, wasting agent invocations and generating garbage audit results.

   Fix:
   ```js
   const response = await fetch(url);
   if (!response.ok) {
     log.warn(`[semantic-value-visibility] ${url} returned ${response.status}, skipping`);
     return null;
   }
   ```

2. **src/semantic-value-visibility/handler.js:51 - Unbounded concurrent outbound fetches (up to 50 in parallel)**

   `Promise.all(result.urls.map(...))` fires up to 50 HTTP GETs concurrently, often against the same origin. This can trigger WAF/CDN rate limiting on target sites, exhaust Lambda's outbound socket pool under cold-start conditions, or cause all fetches to fail if the target applies per-IP throttling - resulting in the audit silently returning zero URLs.

   Fix: Add a concurrency limiter (p-limit or chunked batches of 10):
   ```js
   import pLimit from 'p-limit';
   const limit = pLimit(10);
   const fetchResults = await Promise.all(
     result.urls.map((url) => limit(() => fetchAndCheck(url, log)))
   );
   ```

3. **src/semantic-value-visibility/handler.js:54 - No response body size limit on `response.text()`**

   With 50 concurrent fetches, each calling `response.text()` without a size guard, a misconfigured or malicious page returning a multi-MB response could cause the Lambda to OOM. The worst case is 50 large responses buffered simultaneously in memory.

   Fix: Check Content-Length before reading, or stream a limited prefix:
   ```js
   const contentLength = parseInt(response.headers.get('content-length') || '0', 10);
   if (contentLength > 5 * 1024 * 1024) {
     log.warn(`[semantic-value-visibility] Skipping oversized response: ${url}`);
     return null;
   }
   const html = await response.text();
   ```

### Minor (Nice to Have)

4. **src/semantic-value-visibility/handler.js:82 - Promise.all for SQS sends rejects entire batch on single failure**

   If any single `sqs.sendMessage` call fails (transient network error, queue throttling), the entire `Promise.all` rejects. Messages already in-flight may land, but the function throws without reporting which URLs succeeded and which failed. On Lambda retry, all URLs are re-dispatched including ones that already succeeded.

5. **src/semantic-value-visibility/handler.js:31 - Deep optional chaining silently degrades audit quality**

   `dataAccess?.SiteTopPage?.allBySiteIdAndSourceAndGeo?.(...)` silently returns undefined if `dataAccess` or `SiteTopPage` is misconfigured, which coerces to an empty array via `|| []`. The audit proceeds with agentic-only URLs and no warning that the SEO fallback data source is broken. The log line shows `topPages=0` but that is indistinguishable from a site that genuinely has no top pages.

## Recommendations

- Add `AbortSignal.timeout(10_000)` to each fetch call to prevent a slow-responding server from holding the Lambda invocation open for its full execution timeout.
- For the SQS fan-out (finding 4), consider `Promise.allSettled` with per-message error logging as a future improvement. Or batch sends (`SendMessageBatch`, 10 per call) for fewer API round-trips.

## Assessment

Ready to merge? No - with fixes

The response status check (finding 1) is the most impactful: without it, error pages containing images produce false positives in the audit pipeline. The concurrency limit (finding 2) and response size guard (finding 3) are operational hardening that matter at the 50-URL scale this PR introduces. All three are small, self-contained fixes.

## Next Steps

1. Add `response.ok` check before reading body (finding 1)
2. Add concurrency limiting on outbound fetches (finding 2)
3. Add response size guard (finding 3)
4. Minor items are optional but worth considering before production rollout

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 5s | <b>Cost:</b> $3.21 | <b>Commit:</b> <code>a99d7c29afed5502f1359cd42e7c9d60c451bd34</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot-Dev (commented):
Hey @yevheniia-severinovska,

## Strengths

- Previously flagged issues now addressed: all four operational hardening concerns from the prior review are resolved in the fix commit (`dab6976`): HTTP status check before body read (handler.js:23), concurrency batching with FETCH_CONCURRENCY=10 (handler.js:71-77), 5MB response size cap (handler.js:27-30), and Promise.allSettled for SQS fan-out (handler.js:112)
- Good adoption of the fleet-standard `getMergedAuditInputUrls` pattern (handler.js:44-58) - aligns this audit with the platform contract and picks up future improvements automatically
- `fetchAndCheck` (handler.js:20-37) is a well-structured guard: clear responsibility, predictable return contract (url or null), every exit path logs a reason, and AbortSignal.timeout prevents hangs
- SQS failure handling (handler.js:124-128) is exactly right: individual failures logged at error level with reason, sent/total summary at info level for operational visibility
- Fan-out via individual SQS messages leverages Lambda's built-in concurrency model without new infrastructure
- Comprehensive test coverage with behavior-focused assertions covering filtering, error skipping, status codes, oversized responses, empty results, and partial SQS failures

## Issues

### Important (Should Fix)

1. **handler.js:84 - auditResult shape change from `url` (string) to `urls` (array) may break downstream consumers**

   The persisted `auditResult` shape changed from `{ siteId, url, status }` to `{ siteId, urls, status }`. Per the repo's architecture, this object is stored via `Audit.create()` and may be read by downstream systems (guidance handlers, reporting, dashboards). Any consumer reading `auditResult.url` (singular) will now get `undefined`.

   Unable to verify cross-file consumers from this review, but the structural change from a single string field to an array field is a contract change by definition. If the downstream guidance-handler or reporting paths read `auditResult.url`, they will silently produce empty results.

   Fix: Either confirm all downstream readers have been updated (in this PR or a companion), or include both fields for backward compatibility:
   ```js
   auditResult: { siteId, url: qualifyingUrls[0] || auditUrl, urls: qualifyingUrls, status: 'pending-mystique' },
   ```

### Minor (Nice to Have)

2. **handler.js:50-56 - Deep optional chaining silently degrades to agentic-only without differentiated signal**

   `dataAccess?.SiteTopPage?.allBySiteIdAndSourceAndGeo?.(...)` returns `undefined` both when the data layer is broken and when there are genuinely no top pages. The log line reports `topPages=0` in both cases. If `SiteTopPage` is ever removed or renamed, this audit silently loses its fallback URL source with no alert.

   Fix: Add a differentiated log when the accessor itself is missing:
   ```js
   if (!dataAccess?.SiteTopPage?.allBySiteIdAndSourceAndGeo) {
     log.warn('[semantic-value-visibility] SiteTopPage accessor unavailable, falling back to agentic URLs only');
     return [];
   }
   ```

3. **handler.js:27-30 - Content-Length guard is bypassable on chunked responses**

   When a server uses chunked transfer encoding (no Content-Length header), `parseInt(response.headers.get('content-length') || '0', 10)` evaluates to `0`, bypassing the size check. `response.text()` then buffers the full body regardless of size. In practice this is bounded by the 10s timeout and Lambda memory limit, and URLs come from internal data sources, so risk is low. The guard provides incomplete protection but is not blocking.

## Recommendations

- Consider logging total fetch duration for the pre-filter loop. When 50 URLs are checked in batches of 10, wall-clock time varies significantly based on target site latency. A single elapsed-ms log after the loop would answer "why did this audit take 45s?"
- SSRF defense-in-depth: while URLs come from internal data sources, a scheme validation (`url.startsWith('https://')`) before fetch would prevent accidental fetches against non-HTTPS or internal-network targets at negligible cost
- The `fetchAndCheck` pattern (fetch HTML, apply heuristic, batch with guards) is reusable. If other audits benefit from similar pre-filtering, consider extracting a shared utility once a second consumer appears

## Assessment

Ready to merge? With fixes

The code is well-hardened and cleanly structured. The one material question is the `auditResult` shape change (url to urls) - if downstream consumers have already been updated or this is a coordinated change with the Mystique-side handler, this is ready to go. Please confirm whether any downstream readers depend on `auditResult.url` (singular). The minor items are non-blocking quality improvements.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 6s | <b>Cost:</b> $2.98 | <b>Commit:</b> <code>c4412f9f340691085d960855d8869bc2eceb4520</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

yevheniia-severinovska:
Addressing the `auditResult` shape change concern: `guidance-handler.js` never reads `auditResult` from the DB — it receives Mystique's response via a separate SQS message with its own payload. `sendToMystique` (the only consumer of `auditResult` in this repo) already reads `auditResult.urls` correctly. No other downstream consumer depends on `auditResult.url` for this audit type. The shape change is intentional and safe.

Also addressed minor finding 2: added a differentiated `log.warn` when the `SiteTopPage` accessor is unavailable, making silent degradation observable.

MysticatBot-Dev (commented):
Hey @yevheniia-severinovska,

## Strengths

- Previously flagged issues now addressed: all four operational hardening concerns from the first review remain resolved (HTTP status check, concurrency batching, 5MB Content-Length guard, Promise.allSettled for SQS sends). The SiteTopPage accessor warning (second review, Minor finding 2) is now fixed in commit `4082ac5` with a clear differentiated `log.warn` and corresponding test coverage.
- Good adoption of the fleet-standard `getMergedAuditInputUrls` pattern (handler.js:44-58) - maintains cross-domain consistency with other audit types rather than inventing a bespoke URL-resolution path.
- Batched concurrency model (handler.js:71-77) with `FETCH_CONCURRENCY = 10` prevents unbounded parallelism without requiring a new dependency.
- `sendToMystique` using `Promise.allSettled` (handler.js:100-111) with per-failure error logging and a sent/total summary is a solid resilience pattern. Partial SQS failures do not abort the batch, and observability is maintained.
- Comprehensive test coverage: behavior-focused assertions covering filtering, error skipping, HTTP status codes, oversized responses, empty results, partial SQS failures, and the new SiteTopPage accessor warning path.
- Prior concern about `auditResult` shape change (`url` to `urls`) resolved by author response: `sendToMystique` is the sole consumer in this repo and already reads `auditResult.urls`; the guidance-handler receives Mystique's response via a separate SQS message with its own payload. No re-raise.

## Issues

### Important (Should Fix)

1. **src/semantic-value-visibility/handler.js:37 - `response.text()` buffers unbounded body when Content-Length is absent**

   When `Content-Length` is missing (common with chunked transfer encoding or HTTP/2), `contentLength` parses to `0`, passes the size guard, and `response.text()` buffers the entire response. With 50 URLs processed in batches of 10, a few multi-MB chunked responses could cause memory pressure in a Lambda with limited memory.

   This was noted as Minor in the prior review ("bounded by 10s timeout and Lambda memory limit, URLs come from internal data sources"). Escalating because the change from 1 URL to 50 URLs per invocation multiplies the exposure - what was a tolerable edge case for one fetch becomes realistic with 10 concurrent fetches.

   Fix - stream and abort after `MAX_RESPONSE_BYTES`:
   ```js
   const reader = response.body.getReader();
   let received = 0;
   const chunks = [];
   while (true) {
     const { done, value } = await reader.read();
     if (done) break;
     received += value.length;
     if (received > MAX_RESPONSE_BYTES) {
       reader.cancel();
       log.warn(`[semantic-value-visibility] ${url} response exceeded size limit while streaming, skipping`);
       return null;
     }
     chunks.push(value);
   }
   const html = Buffer.concat(chunks).toString();
   ```

   Alternatively, a simpler (though less efficient) post-read check: `if (html.length > MAX_RESPONSE_BYTES) return null;` after `.text()` - this still buffers but at least prevents downstream processing of huge pages.

2. **test/audits/semantic-value-visibility/handler.test.js - No test exercises multi-batch concatenation**

   With only 3 URLs in the primary test, the batching loop (`FETCH_CONCURRENCY = 10`) executes a single iteration. A bug like `fetchResults = batchResults` (assignment instead of push) would pass with fewer than 10 URLs but silently drop results from subsequent batches.

   Fix: Add a test with 11+ URLs (spanning two batches) where qualifying URLs exist in both batches, and assert that all qualifying URLs from both batches appear in the result.

### Minor (Nice to Have)

3. **src/semantic-value-visibility/handler.js:35 - Timeout errors indistinguishable from network errors in logs**

   When `AbortSignal.timeout(10000)` fires, the catch block logs it as "Failed to fetch" - the same message as a DNS failure or connection reset. In log queries filtering for slow targets vs. unreachable hosts, this distinction matters for operational debugging.

   Fix: Check `err.name === 'TimeoutError' || err.name === 'AbortError'` before the generic message and log a distinct "timed out after 10s" warning.

4. **src/semantic-value-visibility/handler.js:38 - Regex matches `` or JavaScript string literals. For a pre-filter (not correctness-critical), false positives mean wasted Mystique invocations. Acceptable for now since Mystique does its own analysis downstream, but tightening to `/<img\s/i` would reduce false positives at negligible cost.

## Recommendations

- Add a `User-Agent` header to the fetch calls in `fetchAndCheck`. Some CDNs and WAFs block requests without one, which would cause otherwise-valid pages to be silently filtered out.
- Scheme allowlist: adding `if (!url.startsWith('https://')) return null;` inside `fetchAndCheck` would eliminate any theoretical internal-network fetch (SSRF defense-in-depth) for near-zero cost.
- Stub `globalThis.fetch` in `beforeEach` with a default that throws (`sandbox.stub(globalThis, 'fetch').rejects(new Error('unstubbed fetch call'))`) so any test that forgets to configure the stub fails fast rather than making real network calls.

## Assessment

Ready to merge? With fixes

The response body size guard (finding 1) is the primary concern - the risk profile has materially changed from 1 URL per invocation to 50, and chunked responses without Content-Length headers are common in the wild. The multi-batch test gap (finding 2) exercises a distinct code path boundary that the current suite does not cover. The code is otherwise well-structured, well-hardened, and the prior review findings have been thoroughly addressed. Minor items and recommendations can be handled in follow-up.

MysticatBot-Dev (approved):
Hey @yevheniia-severinovska,

## Strengths

- Previously flagged issues now addressed: all operational hardening concerns from prior reviews are resolved - streaming body reader with byte cap (handler.js:39-57), HTTPS-only guard (handler.js:21-24), HTTP status check (handler.js:30-33), concurrency batching with FETCH_CONCURRENCY=10 (handler.js:106-111), 5MB Content-Length pre-check (handler.js:34-38), Promise.allSettled for SQS fan-out (handler.js:146-157), timeout/abort error discrimination (handler.js:61-65), User-Agent header (handler.js:28), and SiteTopPage accessor warning (handler.js:80-83)
- Multi-batch concatenation test gap addressed with a dedicated test exercising 12 URLs across two batches
- Good default-reject pattern in tests: `fetchStub` defaults to `.rejects(new Error('unstubbed fetch call'))`, catching any test that forgets to configure a specific URL stub
- Fleet-standard `getMergedAuditInputUrls` integration (handler.js:74-92) maintains cross-domain consistency with other audit types
- Comprehensive test coverage across the full guard chain (HTTPS, HTTP status, Content-Length, streaming size, img regex, timeout, generic error) plus SQS dispatch edge cases

## Issues

### Minor (Nice to Have)

1. **handler.js:59 - Regex matches ``, `<noscript></noscript>`, or JS string literals. For a pre-filter where false positives only result in extra Mystique dispatches (not incorrect audit results), this is acceptable. If tightening later, stripping comments and noscript before testing would reduce unnecessary dispatches.

2. **handler.js:51 - `reader.cancel()` is not awaited**

   `ReadableStreamDefaultReader.cancel()` returns a Promise. Not awaiting it means the underlying connection may not release immediately. Under sustained load with many large-response skips, this could temporarily hold open TCP sockets. Consider `await reader.cancel()`.

3. **handler.js:34-38 - Content-Length check is redundant when header is absent**

   When Content-Length is missing, `parseInt(null || '0', 10)` yields `0`, so the early-exit guard is always skipped for chunked responses. Correctness is maintained by the streaming cap below, but the intent would be clearer with:
   ```js
   const contentLength = response.headers.get('content-length');
   if (contentLength && parseInt(contentLength, 10) > MAX_RESPONSE_BYTES) { ... }
   ```

## Recommendations

- Add `redirect: 'manual'` to the fetch options. Currently fetch follows redirects by default, which means a 3xx redirect from an HTTPS URL to an internal resource would bypass the HTTPS-only check on the initial URL. Using manual redirect mode and validating redirect targets would close this vector as defense-in-depth.
- Log total fetch-phase duration after the batch loop completes. With up to 50 URLs in batches of 10 (each with 10s timeout), wall-clock time can vary significantly. A single elapsed-ms log answers "why did this audit take 45s?" without requiring correlation across individual URL logs.
- Enhance the `makeFetchResponse` test helper to support multi-chunk delivery (e.g., split a large payload across multiple `read()` calls). The streaming loop and accumulation logic are the primary reason for using `getReader()` over `response.text()`, but current tests always deliver the full body in a single chunk.

## Assessment

Ready to merge? Yes

All prior Critical and Important findings have been thoroughly addressed across three rounds of review. The remaining items are all Minor - the regex false-positive concern is conservative (errs toward dispatching more URLs, not fewer), and the other findings are polish. The streaming reader, concurrency batching, and Promise.allSettled dispatch are solid implementations. The author's confirmation that no downstream consumer depends on `auditResult.url` (singular) resolves the shape-change concern. Ship it.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 7s | <b>Cost:</b> $2.61 | <b>Commit:</b> <code>bab17426df9aa96a71cb41ff83cb2d8ebac98caf</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

yevheniia-severinovska:
Thanks for the thorough multi-round review @MysticatBot-Dev! Remaining minor items noted as follow-up.

PR #2578, opened 2026-05-29
fix(rum-config-refresh): try overrideBaseURL first for domain key lookup, fall back to baseURL

Kanishkavijay39:
## Summary

- `rum-config-refresh` was querying the RUM admin API against `site.getBaseURL()` only, ignoring `fetchConfig.overrideBaseURL`
- Sites like `volkswagen.lt` (base URL) that have their RUM domain key registered under `www.volkswagen.lt` (override URL) were incorrectly getting `hasDomainKey=false`
- Fix: build a candidate list `[overrideBaseURL hostname, baseURL hostname]`, try each in order, stop at first hit, set `hasDomainKey=false` only if all candidates fail

## Root cause

Confirmed via Coralogix logs for `volkswagen.lt` (site `1ca88472-14c2-4a0c-8808-61213847ee58`):
```
[rum-config-refresh] RUM check failed for volkswagen.lt: Error during fetching domainkey for domain 'volkswagen.lt' using admin key. Status: 404
[rum-config-refresh] Updated rumConfig for site 1ca88472-...: hasDomainKey=false
```
The site has `fetchConfig.overrideBaseURL: https://www.volkswagen.lt` and RUM data confirmed present under `www.volkswagen.lt`.

## Test plan
- [ ] All existing tests pass (13 total)
- [ ] New: tries `overrideBaseURL` first when set
- [ ] New: falls back to `baseURL` when `overrideBaseURL` lookup fails
- [ ] New: sets `hasDomainKey=false` only when all candidates fail


Sugandhgyl (changes requested):
**Strengths**
The shared-timeout semantics are correct and well-documented (src/rum-config-refresh/handler.js:73-74). Racing the full candidate loop against a single timer - rather than resetting the budget per domain - keeps total RUM check time bounded. The inline comment explains a non-obvious design choice right where a future maintainer would question it.
Set deduplication (handler.js:62-65) prevents a redundant second RUM call when overrideBaseURL and baseURL resolve to the same hostname - exactly the right guard for the case where an operator sets overrideBaseURL to the same value as baseURL.
Malformed overrideBaseURL handling (handler.js:57-61) degrades gracefully: the try/catch logs a WARN, sets overrideHostname to null, and the candidate list collapses to just baseURL. Test "falls back to baseURL only when overrideBaseURL is malformed" pins the behavior.
Test coverage is proportional and thorough: override-first, fallback-on-404, malformed URL, hostname deduplication, timeout-before-fallback, and all-candidates-fail. The fake-timer timeout test is the one most commonly skipped - it correctly asserts both that www.example.com is never reached and that save is never called.
The siteConfig variable hoisting is a legitimate consolidation, not cosmetic churn - it is reused for getFetchConfig() and the downstream updateRumConfig save.
Issues

**Important (Should Fix)**

1. The candidate loop is not cancelled when the timeout fires; RUM API calls keep running detached.

src/rum-config-refresh/handler.js, line 77-88.

When Promise.race rejects via the timeout, the handler returns { skipped: true, reason: 'timeout' } and clearTimeout runs in finally. The async IIFE is still executing - it will attempt retrieveDomainkey on any remaining candidates against a function that has already returned, consuming Lambda execution context resources. The comment at line 73 says the timeout bounds total RUM check time; that is true for the handler's response, but not for the outbound RUM work.

**Fix: thread a cancellation flag into the loop:**


```
let cancelled = false;
// ...in the IIFE:
for (const domain of domains) {
  if (cancelled) break;
  // existing try/await/catch
}
// in the catch block, before returning:
cancelled = true;
return ok({ skipped: true, reason: 'timeout' });
```

2. `site.getBaseURL()` parsing is unguarded while overrideBaseURL parsing is guarded in the same hunk.

src/rum-config-refresh/handler.js, line 64.

The PR introduces `try/catch` for new URL(overrideBaseURL) and correctly falls back to baseURL on parse failure. The line immediately after builds new URL(site.getBaseURL()).hostname without the same protection. This was pre-existing behavior, but having both URL constructions in the same block creates an inconsistency: a reader naturally expects the second one to be guarded too. A malformed baseURL would throw an unhandled exception propagating out of the function rather than following the ok({ skipped: true }) pattern used everywhere else.

Fix: wrap both constructions in the same try block, or add a test asserting the expected behavior for an unparseable baseURL.

**Minor (Nice to Have)**
1. WARN is too noisy for expected per-candidate 404s (handler.js:85). The volkswagen.lt scenario - the exact case this PR fixes - emits a WARN on every successful refresh (override 404s, baseURL succeeds). WARN means "act on this." A 404 on a candidate you might miss is the design working. Use INFO or DEBUG for per-candidate misses and reserve WARN for the all-candidates-failed terminal case. This pairs naturally with the fix for Important issue 1.

2. Candidate ordering is undocumented (handler.js:62-65). The override-first order is the entire point of the fix, encoded by array element position alone. A one-line comment would protect it from a future cleanup pass.

Kanishkavijay39:
Thanks for the thorough review @Sugandhgyl! All four points addressed in the latest commits:

  Important fixes:
  1. Cancellation flag — added cancelled = false before the race, set cancelled = true in the timeout callback alongside
  timedOut, and added if (cancelled) { break; } at the top of each loop iteration so the IIFE stops consuming Lambda resources
  after the timeout fires.
  2. Unguarded site.getBaseURL() parsing — split the two URL constructions into separate try/catch blocks. A malformed
  overrideBaseURL still degrades gracefully (WARN + fall back to baseURL). A malformed baseURL now logs an error and returns
  ok({ skipped: true, reason: 'malformed baseURL' }) consistent with the rest of the handler. New test covers the baseURL
  failure path.

  Minor improvements:
  3. Log level — per-candidate RUM misses now log at info; warn is reserved for the terminal all-candidates-failed case.
  4. Candidate ordering comment — added an inline comment explaining the override-first ordering.

MysticatBot (approved):
Hey @Kanishkavijay39,

## Strengths

- All 4 findings from the prior review have been cleanly addressed: cancellation flag (handler.js:64-66,80), guarded baseURL parsing (handler.js:60-64), log level correction to info (handler.js:73), and candidate ordering comment (handler.js:67).
- The candidate-list pattern with Set deduplication (handler.js:67) is the right abstraction. `[...new Set([overrideHostname, baseHostname].filter(Boolean))]` prevents redundant RUM API calls when both URLs resolve to the same hostname, and the pattern generalizes naturally if a third URL source appears later.
- Shared timeout budget across the full candidate loop (handler.js:71-74) keeps total Lambda execution time bounded regardless of candidate count, avoiding the common mistake of per-item timeouts that multiply wall-clock time.
- Test suite is thorough and behavior-focused: 9 new tests cover the full decision tree (override-first, fallback on 404, malformed override/base, deduplication, cancellation race, hanging timeout, all-fail), with assertions on observable HTTP response bodies and call ordering rather than implementation details.
- The fake-timer cancellation test (test lines 203-221) correctly advances past timeout then past the slow rejection, asserting the second domain was never reached. This directly pins the cancellation contract.

## Issues

### Minor (Nice to Have)

1. **handler.js:86-88 - `c8 ignore` on else branch without a test documenting the intended fallthrough behavior**

   The `/* c8 ignore next 3 */` pragma covers the case where `Promise.race` rejects with a non-timeout error. This branch falls through after logging, which means `hasDomainKey` stays `false` and the handler proceeds to `updateRumConfig(false)` and `save()`. If that fallthrough is the intended contract for unexpected errors, a single test that forces a non-timeout rejection into the outer catch would pin the behavior and remove the pragma. If the intended behavior is instead `return ok({ skipped: true, ... })`, the missing test hides the ambiguity.

   How to fix: Add a test where the async IIFE rejects for a non-timeout reason (e.g., a synchronous throw before the loop), assert the response is `{ hasDomainKey: false, updated: true }` and `log.warn` was called with 'Unexpected error', then remove the pragma. Alternatively, if CI accepts c8 ignore pragmas for genuinely unreachable defensive branches, document that convention.

2. **handler.js:58, 66 - Malformed URL values logged verbatim**

   The `overrideBaseURL` and `baseURL` values are logged as-is when URL parsing fails. These are internal database config values (not user input) so exploitability is very low, but if a corruption event placed unexpected content in these fields, arbitrarily long or sensitive strings would be written to logs. Consider truncating to a reasonable length (e.g., first 100 chars) as a defense-in-depth measure.

## Assessment

Ready to merge? Yes

The fix is well-scoped, correctly addresses the reported bug (volkswagen.lt scenario), handles all edge cases cleanly, and the test coverage is thorough. No architectural, security, or correctness concerns remain. All prior review feedback has been incorporated.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 40s | <b>Cost:</b> $2.88 | <b>Commit:</b> <code>5fde99349572c132eedf0c7704afc8af85c4b494</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Sugandhgyl (approved):

PR #2585, opened 2026-06-01
fix: canonical - exclude noindex url and excludeUrl

oiatsiuk:
Reduces canonical audit noise by skipping two categories of pages:
- noindex — pages with <meta name="robots/googlebot" content="noindex"> are skipped using cheerio to parse the scraped rawBody
- Site config exclusions — reads excludedURLs from handlers.canonical in site config, same pattern as broken-backlinks. Configure via PATCH /sites/{siteId}/canonical

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/canonical/handler.js
+++ b/src/canonical/handler.js
@@ -541,8 +541,9 @@ export async function processScrapedContent(context) {
       const rawBody = scrapedObject?.scrapeResult?.rawBody || '';
       if (rawBody) {
         const $page = cheerio.load(rawBody);
-        const robotsMeta = $page('meta[name="robots"], meta[name="googlebot"]').attr('content') || '';
-        if (robotsMeta.toLowerCase().includes('noindex')) {
+        const hasNoindex = $page('meta[name="robots"], meta[name="googlebot"]').toArray()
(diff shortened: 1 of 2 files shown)


src/canonical/handler.js
```diff
@@ -516,6 +530,23 @@ export async function processScrapedContent(context) {
...
+      if (excludedURLs.includes(finalUrl)) {
+        log.info(`[canonical] Skipping ${finalUrl} - excluded by site config`);
+        return null;
+      }
+
+      // Skip pages that explicitly opt out of indexing via meta robots noindex tag
+      const rawBody = scrapedObject?.scrapeResult?.rawBody || '';
+      if (rawBody) {
+        const $page = cheerio.load(rawBody);
+        const robotsMeta = $page('meta[name="robots"], meta[name="googlebot"]').attr('content') || '';
```
noruiz: `.attr('content')` only returns the content of the first matched element, so if there's more than one element that contains the attribute they will not be detected. Should this iterate all matches instead?

oiatsiuk: fixed.

noruiz (approved):

PR #2586, opened 2026-06-02
fix: preserve query params in URL dedup for CSV and included URLs

ssilare-adobe:
## Summary

When URLs are submitted via CSV upload or configured as included URLs, query parameters are now preserved during deduplication. Previously, `mergeAndGetUniqueHtmlUrls` deduped by pathname only, causing URLs like `/applecare?filter=iphone` and `/applecare?filter=mac` to collapse into one — even though they may render distinct content.

### Changes

**`src/prerender/utils/utils.js`** — Added `{ includeQueryParams }` option to `mergeAndGetUniqueHtmlUrls`

When `includeQueryParams: true`, the dedup key uses the raw query string as-is (e.g. `?b=2&a=1` and `?a=1&b=2` are treated as distinct URLs since param order is not normalized). The default (`false`) preserves existing behavior for organic/agentic URL paths.

**`src/prerender/handler.js`** — Enabled `includeQueryParams` for:
- CSV URL path (`auditContext.urls`) — always enabled since CSV URLs are explicitly chosen by the user
- Included URLs path — each source is now deduped independently before merging: organic/agentic use pathname-only dedup, included use raw query string dedup so CSV query-param variants are preserved exactly as submitted

The regular organic + agentic path keeps the default pathname-only dedup since those URLs typically carry tracking params that shouldn't create distinct scrape jobs.

### Example

**Before:** 846 apple.com URLs from CSV → 655 submitted (191 dropped as same-pathname dupes)
**After:** 846 URLs → 846 submitted (query param variants like `?filter=iphone` preserved)

## Test plan

- [x] 5 new unit tests for `includeQueryParams` option (raw query string comparison, multiple arrays, etc.)
- [x] Existing `mergeAndGetUniqueHtmlUrls` tests still pass (default behavior unchanged)
- [x] Handler tests pass (359 passing)
- [x] All affected test files pass

### DIFF
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -999,7 +999,11 @@ export async function submitForScraping(context) {
       urls: includedSlackDeduped,
       filteredCount: includedSlackFiltered,
     } = mergeAndGetUniqueHtmlUrls(rebasedIncludedURLs, { includeQueryParams: true });
-    finalUrls = [...organicSlackDeduped, ...includedSlackDeduped];
+    const { urls: crossSlackDeduped } = mergeAndGetUniqueHtmlUrls(
+      [...organicSlackDeduped, ...includedSlackDeduped],
+      { includeQueryParams: true },
+    );
+    finalUrls = crossSlackDeduped;
     filteredCount = organicSlackFiltered + includedSlackFiltered;
     currentOrganic = organicSlackDeduped.length;
     currentIncludedUrls = includedSlackDeduped.length;
@@ -1041,9 +1045,11 @@ export async function submitForScraping(context) {
     } = mergeAndGetUniqueHtmlUrls(filteredAgenticUrls);
     filteredCount = organicFiltered + includedFiltered + agenticFiltered;
 
-    const batchedUrls = [
-      ...organicDeduped, ...includedDeduped, ...agenticDeduped,
-    ].slice(0, DAILY_BATCH_SIZE);
+    const { urls: crossDeduped } = mergeAndGetUniqueHtmlUrls(
+      [...organicDeduped, ...includedDeduped, ...agenticDeduped],
+      { includeQueryParams: true },
+    );
+    const batchedUrls = crossDeduped.slice(0, DAILY_BATCH_SIZE);
 
     const organicUrlSet = new Set(organicDeduped);
     const includedUrlSet = new Set(includedDeduped);
@@ -1203,16 +1209,15 @@ export async function processOpportunityAndSuggestions(
   const { auditResult, scrapedUrlsSet: rawScrapedUrlsSet } = auditData;
   const { urlsNeedingPrerender } = auditResult;
 
-  // Normalize scrapedUrlsSet to match by pathname so that domain shifts
-  // (e.g. www.example.com → example.com) don't prevent existing suggestions
-  // from being correctly marked as outdated. The set uses pathname-based
-  // lookups: both stored values and .has() arguments are normalized to pathnames.
+  // Normalize scrapedUrlsSet to pathname+search so query-param variants are treated
+  // as distinct pages. Domain shifts only affect hostname so migration tolerance
+  // (e.g. www.example.com → example.com) is preserved.
   const scrapedUrlsSet = rawScrapedUrlsSet ? (() => {
-    const pathnames = new Set(
-      [...rawScrapedUrlsSet].map(toPathname),
+    const keys = new Set(
+      [...rawScrapedUrlsSet].map(normalizePathnameWithQuery),
     );
     return {
-      has: (url) => pathnames.has(toPathname(url)),
+      has: (url) => keys.has(normalizePathnameWithQuery(url)),
     };
   })() : null;
 
@@ -1465,7 +1470,9 @@ export async function uploadStatusSummaryToS3(auditUrl, auditData, context) {
     const existingStatus = await readSiteStatusJson(siteId, context);
     const existingPages = Array.isArray(existingStatus.pages) ? existingStatus.pages : [];
 
-    const existingPageMap = new Map(existingPages.map((p) => [normalizePathname(p.url), p]));
+    const existingPageMap = new Map(
+      existingPages.map((p) => [normalizePathnameWithQuery(p.url), p]),
+    );
 
     const currentPages = (auditResult.results ?? []).map((result) => {
       // Only stamp the current scrapeJobId for URLs actually submitted to this job.
@@ -1483,7 +1490,7 @@ export async function uploadStatusSummaryToS3(auditUrl, auditData, context) {
         scrapedAt,
         scrapeJobId: wasSubmitted
           ? (scrapeJobId || null)
-          : (existingPageMap.get(normalizePathname(result.url))?.scrapeJobId ?? null),
+          : (existingPageMap.get(normalizePathnameWithQuery(result.url))?.scrapeJobId ?? null),
         ...(result.isErrorPage && { isErrorPage: true }),
         ...(result.scrapeError && { scrapeError: result.scrapeError }),
       };
@@ -1500,10 +1507,10 @@ export async function uploadStatusSummaryToS3(auditUrl, auditData, context) {
       );
     }
 
-    const currentUrlSet = new Set(currentPages.map((p) => normalizePathname(p.url)));
+    const currentUrlSet = new Set(currentPages.map((p) => normalizePathnameWithQuery(p.url)));
     const mergedPages = [
       ...currentPages,
-      ...existingPages.filter((p) => !currentUrlSet.has(normalizePathname(p.url))),
+      ...existingPages.filter((p) => !currentUrlSet.has(normalizePathnameWithQuery(p.url))),
     ];
 
     // Derive aggregate metrics from the full merged page set and latest audit metadata.
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -410,7 +410,7 @@ function isNotRecentUrl(url, recentPathnames) {
 function sanitizeImportPath(importPath) {
   return importPath
     .replace(/^\/+|\/+$/g, '')
-    .replace(/[/._]/g, '-')
+    .replace(/[/._?=&]/g, '-')
     .replace(/-+/g, '-')
     .replace(/^-|-$/g, '');
 }
@@ -425,8 +425,8 @@ function sanitizeImportPath(importPath) {
 * @returns {string} The S3 path to the file
 */
 function getS3Path(url, id, fileName) {
-  const rawImportPath = new URL(url).pathname;
-  const sanitizedImportPath = sanitizeImportPath(rawImportPath);
+  const { pathname, search } = new URL(url);
+  const sanitizedImportPath = sanitizeImportPath(pathname + search);
   const pathSegment = sanitizedImportPath ? `/${sanitizedImportPath}` : '';
   return `${AUDIT_TYPE}/scrapes/${id}${pathSegment}/${fileName}`;
 }
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -897,7 +897,10 @@ export async function submitForScraping(context) {
   if (Array.isArray(auditContext?.urls) && auditContext.urls.length > 0) {
     const preferredBase = getPreferredBaseUrl(site, context);
     const rebasedCsvUrls = auditContext.urls.map((url) => rebaseUrl(url, preferredBase, log));
-    const { urls: explicitUrls, filteredCount } = mergeAndGetUniqueHtmlUrls(rebasedCsvUrls);
+    const { urls: explicitUrls, filteredCount } = mergeAndGetUniqueHtmlUrls(
+      rebasedCsvUrls,
+      { includeQueryParams: true },
+    );
 
     log.info(`
     ${LOG_PREFIX} prerender_submit_scraping_metrics:
@@ -954,10 +957,10 @@ export async function submitForScraping(context) {
   let edgeDeployedPathnames = new Set();
 
   if (isSlackTriggered) {
-    ({ urls: finalUrls, filteredCount } = mergeAndGetUniqueHtmlUrls([
-      ...rebasedTopPagesUrls,
-      ...rebasedIncludedURLs,
-    ]));
+    ({ urls: finalUrls, filteredCount } = mergeAndGetUniqueHtmlUrls(
+      [...rebasedTopPagesUrls, ...rebasedIncludedURLs],
+      { includeQueryParams: rebasedIncludedURLs.length > 0 },
+    ));
     currentOrganic = rebasedTopPagesUrls.length;
     currentIncludedUrls = rebasedIncludedURLs.length;
     isFirstRunOfCycle = true;
@@ -999,7 +1002,10 @@ export async function submitForScraping(context) {
       (url) => !organicUrlSet.has(url) && !includedUrlSet.has(url),
     ).length;
 
-    ({ urls: finalUrls, filteredCount } = mergeAndGetUniqueHtmlUrls(batchedUrls));
+    ({ urls: finalUrls, filteredCount } = mergeAndGetUniqueHtmlUrls(
+      batchedUrls,
+      { includeQueryParams: filteredIncludedURLs.length > 0 },
+    ));
   }
 
   log.info(`${LOG_PREFIX} prerender_submit_scraping_metrics:
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -313,6 +313,32 @@ async function getTopAgenticUrls(site, context, limit = TOP_AGENTIC_URLS_LIMIT) 
   }
 }
 
+function normalizePathname(url) {
+  try {
+    const { pathname } = new URL(url);
+    return pathname.length > 1 ? pathname.replace(/\/+$/, '') : pathname;
+  } catch {
+    return url;
+  }
+}
+
+/**
+ * Normalizes a URL to its pathname + search string.
+ * Trailing slashes on the pathname are removed (except for the root path).
+ * Falls back to the raw string when the URL is not parseable.
+ * @param {string} url
+ * @returns {string} pathname+search, or the original string on parse failure
+ */
+function normalizePathnameWithQuery(url) {
+  try {
+    const { pathname, search } = new URL(url);
+    const normalized = pathname.length > 1 ? pathname.replace(/\/+$/, '') : pathname;
+    return search ? `${normalized}${search}` : normalized;
+  } catch {
+    return url;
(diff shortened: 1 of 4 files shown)


anuj-adobe:
@ssilare-adobe while we are working on further adding code to `submitForScraping` can you pls check to improve `submitForScraping` step using this refactoring PR https://github.com/adobe/spacecat-audit-worker/pull/2548 ?

MysticatBot (commented):
Hey @ssilare-adobe,

## Strengths

- **Backward-compatible API evolution** (`src/prerender/utils/utils.js:69-73`): The options-object-as-last-arg pattern with rest params keeps every existing caller working unchanged. The detection logic (`!Array.isArray(lastArg) && typeof lastArg === 'object'`) is solid and the default `false` preserves current behavior.
- **Intent-driven activation** (`src/prerender/handler.js:960, 1005`): The conditional `{ includeQueryParams: rebasedIncludedURLs.length > 0 }` ties the dedup strategy to business intent (user explicitly chose these URLs) rather than being a global switch. This prevents organic/agentic URLs from accumulating tracking-param variants.
- **Good test coverage** (`test/audits/prerender/utils.test.js:268-310`): Tests verify behavior (output URLs), not implementation details. The test for param-order behavior is especially valuable because it documents an intentional design choice.
- **Clean stub update** (`test/audits/prerender/handler.test.js:4438`): `args.filter(Array.isArray).flat()` correctly adapts the stub to the new signature shape.

## Issues

### Important (Should Fix)

**PR description contradicts implementation on query param ordering** - `src/prerender/utils/utils.js:102-103`

The PR body states: "the dedup key includes sorted query params (so `?b=2&a=1` and `?a=1&b=2` still dedup correctly)." But the implementation appends `urlObj.search` directly (the raw query string), and the test at utils.test.js:280-290 explicitly confirms these are treated as distinct URLs. The test comment even says "Raw query strings differ, so both are kept."

The code is internally consistent and the test passes, but the PR description is actively misleading. On squash-merge this will become the commit message, and someone doing `git blame` will believe param-order normalization is happening when it is not.

**Fix:** Update the PR body to remove the "sorted query params" claim. State that raw query strings are compared as-is, preserving exactly what the user submitted in the CSV.

### Minor (Nice to Have)

**Mixed-batch dedup behavior coupling** - `src/prerender/handler.js:960, 1005`

When `includeQueryParams` is conditionally enabled because included URLs are present, it applies to the entire merged array (organic + included URLs together). If organic/top-pages URLs ever carry tracking parameters (e.g., `?utm_source=...`), they would no longer dedup against their parameterless equivalents in batches that happen to include at least one included URL. The risk is low if the organic source doesn't currently emit params, but the coupling between "included URLs exist" and "organic dedup behavior changes" is worth a brief inline comment explaining the trade-off.

## Recommendations

- **Consider sorting query params before constructing the dedup key.** The current raw-string comparison means `?a=1&b=2` and `?b=2&a=1` produce two separate scrape jobs for what most web servers render as the same page. If preserving exact CSV input is the goal, the current behavior is fine - but document the trade-off in a JSDoc note on the function (e.g., "param order is not normalized; callers who need order-independent dedup should pre-sort"). If dedup accuracy matters more, `[...urlObj.searchParams].sort().map(([k,v]) => k+'='+v).join('&')` would collapse order variants while still treating `/page?a=1` and `/page?b=2` as distinct.

## Assessment

Ready to merge? With fixes
Reasoning: The code is correct, backward-compatible, and well-tested. CI passes with full coverage. The one material fix is correcting the misleading PR description before squash-merge so the commit message accurately reflects what shipped. The minor and recommendation items are optional improvements.

## Next Steps

1. Update the PR description to remove the "sorted query params" claim and accurately describe the raw-string comparison behavior.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 5s | <b>Cost:</b> $2.71 | <b>Commit:</b> <code>22573cff803d6de49bdb9a8cccffa7a3665ed56e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anuj-adobe (commented):
## Code Review

The core fix is correct and well-motivated — CSV and included URLs with distinct query params (e.g. `/applecare?filter=iphone` vs `/applecare?filter=mac`) should not be collapsed. The `mergeAndGetUniqueHtmlUrls` change and the CSV path in `handler.js` are clean. However, the pipeline has two more pathname-keyed chokepoints that would silently undo the benefit for included/CSV URLs, plus one correctness issue with how the flag is applied.

---

### 🔴 Gap 1 — `buildKey` collapses distinct suggestions

`buildKey` in `processOpportunityAndSuggestions` (handler.js ~L1212) uses `toPathname(data.url)` as the sync key:

```js
return toPathname(data.url); // /applecare?filter=iphone → /applecare
```

If `/applecare?filter=iphone` and `/applecare?filter=mac` are submitted as separate scrape jobs, `syncSuggestions` maps both to `/applecare` and creates **one** suggestion — dropping one of the two. Same issue at ~L1783 where another `buildKey` also calls `toPathname`.

The key needs to include query params when the URL has them, at minimum for suggestions that originated from included/CSV paths.

---

### 🔴 Gap 2 — PageCitability 7-day dedup blocks query-param siblings

`getRecentlyProcessedPathnames` (~L338) stores pathnames only:

```js
return new URL(r.getUrl()).pathname; // drops query params
```

And `isNotRecentUrl` (~L383) checks against that set. Once `/applecare?filter=iphone` is scraped and written to PageCitability, `/applecare?filter=mac` will be suppressed for 7 days — even though it's a distinct page. The PageCitability record (and its lookup) needs to key on `pathname + search` for included/CSV URLs.

---

### 🟡 Gap 3 — `includeQueryParams` bleeds into organic/agentic URLs

In both the Slack and Normal paths the flag is activated when included URLs are present, but it applies to the **entire merged batch** including organic/agentic URLs. Organic URLs can carry tracking params (`?gclid=`, `?utm_source=`) that should still be treated as the same page.

The PR description explicitly says *"The regular organic + agentic path keeps the default pathname-only dedup"* — but the implementation doesn't enforce this. Fix: dedup each source independently before merging:

```js
const { urls: organicDeduped } = mergeAndGetUniqueHtmlUrls(filteredOrganicUrls);
const { urls: includedDeduped } = mergeAndGetUniqueHtmlUrls(filteredIncludedURLs, { includeQueryParams: true });
const { urls: agenticDeduped } = mergeAndGetUniqueHtmlUrls(filteredAgenticUrls);
const batchedUrls = [...organicDeduped, ...includedDeduped, ...agenticDeduped].slice(0, DAILY_BATCH_SIZE);
// final mergeAndGetUniqueHtmlUrls call at the end can be dropped — already deduped per source
```

---

### 🟡 PR description inconsistency

The description says *"dedup key includes sorted query params (so `?b=2&a=1` and `?a=1&b=2` still dedup correctly)"* but the implementation uses raw `urlObj.search` without sorting. The test at utils.test.js expects `lengthOf(2)` for those two URLs (both kept), which contradicts the description. The implementation behavior (preserve exact CSV input) seems intentional — just update the description to match.

---

### 🟢 What's good

- CSV path fix (`includeQueryParams: true` unconditionally) is clearly correct.
- Backward compat fully preserved — default is `false`, no existing behavior changes.
- Stub fix in handler.test.js (`args.filter(Array.isArray).flat()`) is correct.
- 5 new unit tests cover the main cases.

MysticatBot (changes requested):
Hey @ssilare-adobe,

This is a re-review after the 3 new commits addressing anuj-adobe's Gaps 1-3. All three gaps are properly resolved and the per-source dedup strategy is sound. One item from the prior reviews remains unresolved.

## Strengths

- **All three gaps from anuj-adobe's review are cleanly resolved:**
  - Gap 1: `buildKey` now uses `normalizePathnameWithQuery` so query-param variants produce distinct suggestions
  - Gap 2: `getRecentlyProcessedPathnames` and `isNotRecentUrl` use `normalizePathnameWithQuery` so the 7-day dedup window correctly distinguishes query-param siblings
  - Gap 3: Each source (organic, included, agentic) is deduped independently before merging - `includeQueryParams` no longer bleeds into organic/agentic paths
- **Mixed-batch dedup coupling resolved** (prior MysticatBot finding): the single merged `mergeAndGetUniqueHtmlUrls` call is replaced by per-source calls with appropriate strategies
- **Backward-compatible API evolution** (`src/prerender/utils/utils.js:65-73`): The options-object-as-last-arg pattern with `!Array.isArray(lastArg)` detection keeps all existing callers working unchanged
- **Consistent normalization across the pipeline**: `normalizePathnameWithQuery` is threaded through `getRecentlyProcessedPathnames`, `isNotRecentUrl`, `buildKey`, `writeToCitabilityRecords`, and `processContentAndGenerateOpportunities` - keys match at read and write boundaries
- **Good test coverage**: Codecov confirms all modified lines are covered, with targeted tests for each gap fix

## Issues

### Important (Should Fix)

**PR description still claims "sorted query params"** - PR body

The PR description states: "the dedup key includes sorted query params (so `?b=2&a=1` and `?a=1&b=2` still dedup correctly)." The implementation uses `urlObj.search` (raw query string, no sorting), and the test at `test/audits/prerender/utils.test.js:280-290` explicitly confirms these are treated as distinct URLs (expects `lengthOf(2)` with the comment "Raw query strings differ, so both are kept"). This was raised in both the prior MysticatBot review and anuj-adobe's review but remains unresolved.

On squash-merge this becomes the commit message. A developer doing `git blame` will believe param-order normalization is happening when it is not.

**Fix:** Update the PR description to remove the "sorted query params" claim and state that raw query strings are compared as-is, preserving exactly what the user submitted in the CSV.

### Minor (Nice to Have)

**Cross-source overlap can produce duplicates in final batch** - `src/prerender/handler.js:1024-1026`

After per-source dedup, the merge is a naive concatenation: `[...organicDeduped, ...includedDeduped, ...agenticDeduped].slice(0, DAILY_BATCH_SIZE)`. The old code ran a final `mergeAndGetUniqueHtmlUrls(batchedUrls)` which would catch cross-source duplicates. If a URL appears in both organic top-pages and site config's includedURLs (with no query params), it could appear twice in the final batch. The probability is low and impact is just a wasted scrape slot, but worth noting.

**`normalizePathname` retained for edge-deployed filter deserves a clarifying comment** - `src/prerender/handler.js:314-320`

The diff retains `normalizePathname` (pathname-only) for the edge-deployed filter while all other callsites use `normalizePathnameWithQuery`. This is correct (edge deployment applies to the pathname regardless of query params), but without a comment a future contributor will wonder whether this is a missed migration. A one-liner explaining the intentional divergence would prevent that.

**`mergeAndGetUniqueHtmlUrls` options-sniffing is fragile long-term** - `src/prerender/utils/utils.js:66-68`

The `hasOptions` check (`lastArg && !Array.isArray(lastArg) && typeof lastArg === 'object'`) would match unintended values like a Date, Map, or any non-null object. Current callers are all correct so this is not blocking. If the API evolves, consider migrating to a two-arg signature `(urlArrays, options?)`.

## Recommendations

- Consider documenting the uniqueness-key change (pathname-only to pathname+search for PageCitability, suggestions, and recently-processed filtering) as an ADR in `/docs/decisions/`. The CLAUDE.md states: "Technical or architectural decisions with long-term impact must be recorded as ADRs." The key change affects existing data lookups and is worth recording for future reference.

## Assessment

Ready to merge? With fixes
Reasoning: The code is correct, backward-compatible, well-tested, and CI passes. All three prior gaps are properly resolved. The one blocking fix is trivial - update the PR description to remove the misleading "sorted query params" claim before squash-merge so the commit history is accurate.

## Next Steps

1. Update the PR description to accurately describe raw query-string comparison behavior (the "sorted" claim is incorrect).
2. Optionally add a brief comment on the `normalizePathname` usage explaining why it intentionally differs from `normalizePathnameWithQuery` at the edge-deployed filter.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 13s | <b>Cost:</b> $4.24 | <b>Commit:</b> <code>a4d377cc2c3ed42017d1ac192fd2edfdb99c2e1e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @ssilare-adobe,

**Verdict: Approve** - prior blocking finding resolved, no new concerns.
_Changes: Preserves query parameters during URL deduplication for CSV and included URLs by adding per-source dedup strategies and threading `normalizePathnameWithQuery` through all key-generation points (4 files)._

<details>

- suggestion: Consider a test for the cross-source overlap scenario (a URL present in both organic and included lists appearing twice in the final batch) to document the accepted behavior - `src/prerender/handler.js:1037`
</details>

- PR description now correctly states raw query string comparison (previously claimed "sorted query params")

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 32s | <b>Cost:</b> $2.29 | <b>Commit:</b> <code>19064530b4c15e620fabc6f2a560f072bc7e714c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @ssilare-adobe,

**Verdict: Comment** - clean incremental commit with no blocking concerns.
_Changes: Adds cross-source dedup step after per-source dedup and aligns all pipeline key-generation points to normalizePathnameWithQuery (2 files)._

<details>

- nit: Slack path sets `currentOrganic`/`currentIncludedUrls` from per-source counts before cross-source dedup, so logged metrics may sum to more than `finalUrls.length` when a URL exists in both sources. The normal path handles this correctly by counting against the final `batchedUrls` - `src/prerender/handler.js:1003`
- suggestion: Cross-source dedup is verified by `callCount(4)` but the stub does not actually deduplicate or check options; a non-stubbed integration test with a shared URL across sources would verify end-to-end dedup behavior - `test/audits/prerender/handler.test.js:4689`
</details>

- Cross-source overlap scenario eliminated by adding the cross-source dedup step

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 49s | <b>Cost:</b> $2.96 | <b>Commit:</b> <code>6269105cf28b52b15b1866392f902cf509d88384</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anuj-adobe (commented):
## PR Review: Preserve query params in URL dedup for CSV and included URLs

**Overall: LGTM.** The feature works correctly, the three `submitForScraping` code paths are all updated consistently, and data model lookups (PageCitability, status.json, suggestion keys, scrapedUrlsSet) are systematically upgraded to `pathname+search`. No regression in the regular organic/agentic flow.

---

### ✅ What works well

**KISS/DRY/YAGNI**
- `normalizePathnameWithQuery` is a clean utility that mirrors `normalizePathname` with a single addition (`search`). Not over-engineered.
- `{ includeQueryParams }` option on `mergeAndGetUniqueHtmlUrls` is backward-compatible — default `false` preserves existing behavior for all callers.
- The variadic options detection (`args.filter(Array.isArray)`) is a reasonable JS idiom here.

**All three `submitForScraping` paths correctly updated**
- CSV path: `mergeAndGetUniqueHtmlUrls(rebasedCsvUrls, { includeQueryParams: true })` ✓
- Slack path: included URLs deduped with `includeQueryParams: true`, organic without, cross-merge with `includeQueryParams: true` ✓
- Normal path: independent per-source dedup (organic/agentic = pathname-only, included = pathname+search) then cross-merge ✓

**Data model lookups consistently updated**
- `getRecentlyProcessedPathnames` + `isNotRecentUrl` → `normalizePathnameWithQuery` ✓
- `writeToCitabilityRecords` map keys → `normalizePathnameWithQuery` ✓
- `uploadStatusSummaryToS3` page map + current URL set → `normalizePathnameWithQuery` ✓
- `processOpportunityAndSuggestions` `scrapedUrlsSet` normalization → `normalizePathnameWithQuery` ✓
- `buildKey` in `syncSuggestions` → `normalizePathnameWithQuery` ✓

**`sanitizeImportPath` + `getS3Path` correctly updated**
- `?=&` added to replacement regex so `/page?filter=iphone` → `page-filter-iphone` in S3 path ✓
- `getS3Path` now uses `pathname + search` instead of `pathname` only ✓

**`scrapeJobId` fallback chain preserved**
- S3 path format is `prerender/scrapes/{jobId}/{sanitizedPath}/file` — the `jobId` is still at index 2. New sanitized paths just have `-` instead of `?=&`. Fallback extraction via `path.split('/')[2]` still works ✓

---

### ⚠️ Minor concerns (non-blocking)

**1. `filteredCount` metric no longer captures cross-source duplicates**

After the refactor, `filteredCount` is the sum of per-source non-HTML URL counts only:
```js
filteredCount = organicFiltered + includedFiltered + agenticFiltered;
```

If organic has `/page` and included has `/page` (exact match), the cross-source `mergeAndGetUniqueHtmlUrls` call drops one, but that's not counted. The `filteredOutUrls=` metric in the submit log would be slightly low. Not a functional issue, but worth noting if you later wonder why the number seems off.

**2. `sanitizeImportPath` collision: `/page-filter=iphone` and `/page?filter=iphone` produce the same S3 key**

Both become `page-filter-iphone` after sanitization. This is a theoretical collision — real-world URLs don't have `-filter=` in their pathname — but it's worth a comment near the regex in case the pattern is widened later.

**3. Intentional asymmetry: edge-deployed filtering is still pathname-only**

Lines 1001/1004/1007 (unchanged):
```js
.filter((url) => !edgeDeployedPathnames.has(normalizePathname(url)));
```

So if `/page?filter=iphone` is marked edge-deployed in `status.json`, both `/page?filter=iphone` AND `/page?filter=mac` are filtered out (because both have pathname `/page`). This is intentional — edge deployment is per-pathname, not per-query-param variant — but the asymmetry with the now-granular PageCitability dedup is worth a comment explaining the design choice.

---

### No regression in regular organic/agentic flow ✓

The regular scheduled path:
1. `getRecentlyProcessedPathnames` now keys on `pathname+search`, but organic URLs from GSC are clean canonical URLs (no query params) — their key is still just `/page`
2. `isNotRecentUrl` uses `normalizePathnameWithQuery` — for no-query-param URLs, returns the same result as before
3. `mergeAndGetUniqueHtmlUrls` called without `includeQueryParams` for organic/agentic → behavior identical to pre-PR
4. Suggestion `buildKey` on `data.url` — existing suggestions have clean URLs, pathname+search = pathname for them

The only behavioral difference for existing sites is the first run after deployment: if any included URL with query params was previously scraped (unlikely since CSV query params were previously collapsed), those suggestions may be re-keyed. One-time, acceptable.

PR #2587, opened 2026-06-02
fix(llm-error-pages): drop malformed URLs from categorized results

ymadobe:
## Summary
- Bot impersonators that pass the aggregated-logs UA filter occasionally emit malformed paths (e.g. `/brandshttps://...`, `/),`, `/%2522`). These add noise to the 404 opportunity without being actionable.
- Adds a conservative `isValidUrlPath` check in `utils.js` and applies it to `categorizedResults` right after categorization, so Excel exports, opportunity sync, and the Mystique payload all see the filtered set.
- Drops are logged per status bucket for visibility.

Context: [Slack thread](#) - confirmed that the upstream aggregated-logs table is already UA-filtered; the impersonators slip through because we don't store IPs/TLS fingerprints. This is a downstream noise filter, not a replacement for upstream verification.

## Test plan
- [x] `npm run lint`
- [x] Targeted coverage: 100% on `src/llm-error-pages/utils.js` and `src/llm-error-pages/handler.js`
- [x] 8 new unit tests for `isValidUrlPath` (valid URLs, non-strings, empty, embedded scheme, quote chars, trailing comma, unparseable)
- [x] New handler test verifying malformed URLs are dropped from categorized results and the drop is logged

### DIFF
--- a/src/llm-error-pages/handler.js
+++ b/src/llm-error-pages/handler.js
@@ -23,7 +23,6 @@ import {
   categorizeErrorsByStatusCode,
   SPREADSHEET_COLUMNS,
   toPathOnly,
-  isValidUrlPath,
 } from './utils.js';
 import { wwwUrlResolver } from '../common/index.js';
 import { createLLMOSharepointClient, saveExcelReport } from '../utils/report-uploader.js';
@@ -187,15 +186,12 @@ export async function runAuditAndSendToMystique(context) {
         const processedResults = processErrorPagesResults(results);
         const categorizedResults = categorizeErrorsByStatusCode(processedResults.errorPages);
 
-        // Drop malformed URLs (e.g. '/brandshttps://...', '/),') emitted by
-        // bot impersonators that slip through the user-agent filter upstream.
-        for (const code of Object.keys(categorizedResults)) {
-          const before = categorizedResults[code].length;
-          categorizedResults[code] = categorizedResults[code].filter((e) => isValidUrlPath(e.url));
-          const dropped = before - categorizedResults[code].length;
-          if (dropped > 0) {
-            log.info(`[LLM-ERROR-PAGES] Filtered ${dropped} malformed URL(s) from status ${code}`);
-          }
+        // Malformed URLs (e.g. '/brandshttps://...', '/),') are dropped inside
+        // processErrorPagesResults so every downstream view stays consistent.
+        // Log a sample so false positives are diagnosable from Coralogix.
+        if (processedResults.droppedUrls?.length > 0) {
+          const sample = processedResults.droppedUrls.slice(0, 5);
+          log.info(`[LLM-ERROR-PAGES] Filtered ${processedResults.droppedUrls.length} malformed URL(s); sample: ${JSON.stringify(sample)}`);
         }
 
         const buildFilename = (code) => `agentictraffic-errors-${code}-${periodIdentifier}.xlsx`;
--- a/src/llm-error-pages/handler.js
+++ b/src/llm-error-pages/handler.js
@@ -23,6 +23,7 @@ import {
   categorizeErrorsByStatusCode,
   SPREADSHEET_COLUMNS,
   toPathOnly,
+  isValidUrlPath,
 } from './utils.js';
 import { wwwUrlResolver } from '../common/index.js';
 import { createLLMOSharepointClient, saveExcelReport } from '../utils/report-uploader.js';
@@ -185,6 +186,17 @@ export async function runAuditAndSendToMystique(context) {
 
         const processedResults = processErrorPagesResults(results);
         const categorizedResults = categorizeErrorsByStatusCode(processedResults.errorPages);
+
+        // Drop malformed URLs (e.g. '/brandshttps://...', '/),') emitted by
+        // bot impersonators that slip through the user-agent filter upstream.
+        for (const code of Object.keys(categorizedResults)) {
+          const before = categorizedResults[code].length;
+          categorizedResults[code] = categorizedResults[code].filter((e) => isValidUrlPath(e.url));
+          const dropped = before - categorizedResults[code].length;
+          if (dropped > 0) {
+            log.info(`[LLM-ERROR-PAGES] Filtered ${dropped} malformed URL(s) from status ${code}`);
+          }
+        }
 
         const buildFilename = (code) => `agentictraffic-errors-${code}-${periodIdentifier}.xlsx`;
 
--- a/src/llm-error-pages/utils.js
+++ b/src/llm-error-pages/utils.js
@@ -452,6 +452,41 @@ export function toPathOnly(maybeUrl, baseUrl) {
   }
 }
 
+/**
+ * Conservative validity check for URLs surfaced from CDN logs. Bot impersonators
+ * that pass the user-agent regex sometimes request malformed paths (e.g.
+ * `/brandshttps://...`, `/),`, `/%2522`). Those add noise to the 404 opportunity
+ * without being actionable, so we drop them before building suggestions.
+ */
+export function isValidUrlPath(url) {
+  if (typeof url !== 'string') {
+    return false;
+  }
+  const trimmed = url.trim();
+  if (!trimmed) {
+    return false;
+  }
+
+  try {
+    // eslint-disable-next-line no-new
+    new URL(trimmed, 'https://example.com');
+  } catch {
+    return false;
+  }
+
+  if (/.+https?:\/\//i.test(trimmed)) {
+    return false;
+  }
+  if (/["']|%2[27]|%252[27]/i.test(trimmed)) {
+    return false;
+  }
+  if (/,$/.test(trimmed)) {
+    return false;
+  }
+
+  return true;
+}
+
 export const SPREADSHEET_COLUMNS = [
   'Agent Type',
   'User Agent',
--- a/src/llm-error-pages/utils.js
+++ b/src/llm-error-pages/utils.js
@@ -319,11 +319,85 @@ export function generateReportingPeriods(referenceDate = new Date(), weekOffsets
 // PROCESSING RESULTS
 // ============================================================================
 
+/**
+ * Conservative validity check for URLs surfaced from CDN logs. Bot impersonators
+ * that pass the user-agent regex sometimes request malformed paths (e.g.
+ * `/brandshttps://...`, `/),`, `/%2522`). Those add noise to the 404 opportunity
+ * without being actionable, so we drop them before building suggestions.
+ *
+ * Tracked stopgap: the right long-term fix is upstream UA / IP-fingerprint
+ * hardening so impersonators are rejected before ingestion. See LLMO-5282.
+ *
+ * Rejects:
+ *   - non-strings / empty / whitespace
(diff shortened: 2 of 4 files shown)


src/llm-error-pages/utils.js
```diff
@@ -452,6 +452,41 @@ export function toPathOnly(maybeUrl, baseUrl) {
...
+  }
+
+  try {
+    // eslint-disable-next-line no-new
+    new URL(trimmed, 'https://example.com');
+  } catch {
+    return false;
+  }
+
+  if (/.+https?:\/\//i.test(trimmed)) {
```
akshaymagapu: **Critical:** `/.+https?:\/\//i` matches a scheme anywhere after one character, so it silently drops legitimate 404s like `/redirect?to=https://example.com`, `/share?url=https://...`, and OAuth `/auth/callback?return=https://...` - exactly the broken redirects a customer needs surfaced. The leading `.+` also lets a bare absolute URL (`https://evil.com/foo`) pass entirely. Tighten to a scheme glued to a path segment, e.g. `/[^/?&=#]https?:\/\//i`, which still catches `/brandshttps://...` but allows query-embedded schemes.

src/llm-error-pages/utils.js
```diff
@@ -452,6 +452,41 @@ export function toPathOnly(maybeUrl, baseUrl) {
...
+    return false;
+  }
+  if (/["']|%2[27]|%252[27]/i.test(trimmed)) {
```
akshaymagapu: **Critical:** This rejects `'` or `"` anywhere in the URL, dropping legitimate paths like `/products/o'reilly`, `/authors/o'brien`, `/it's-friday`, and percent-encoded quotes (`%27`). Raw quotes generally can't survive `new URL()` parsing anyway, so the real signal is the double-encoded form. Recommend keeping only `%2522`/`%2527` and dropping the raw-quote and single `%22`/`%27` rejection. If single-encoded quotes are intentionally rejected, document why and add a dropped-URL sample to the log so false positives are visible.

src/llm-error-pages/utils.js
```diff
@@ -452,6 +452,41 @@ export function toPathOnly(maybeUrl, baseUrl) {
...
+    return false;
+  }
+  const trimmed = url.trim();
+  if (!trimmed) {
+    return false;
+  }
+
+  try {
+    // eslint-disable-next-line no-new
+    new URL(trimmed, 'https://example.com');
```
akshaymagapu: **Important:** `new URL(trimmed, 'https://example.com')` is far more permissive than "valid URL" implies - it accepts `javascript:alert(1)`, `data:...`, null bytes (`%00`), and CRLF (`%0d%0a`), all of which flow into the Mystique LLM payload and Excel. Restrict to `parsed.protocol === 'http:' || parsed.protocol === 'https:'`, or correct the comment to describe what the gate actually does.

src/llm-error-pages/utils.js
```diff
@@ -452,6 +452,41 @@ export function toPathOnly(maybeUrl, baseUrl) {
...
+    return false;
+  }
+
+  if (/.+https?:\/\//i.test(trimmed)) {
+    return false;
+  }
+  if (/["']|%2[27]|%252[27]/i.test(trimmed)) {
+    return false;
+  }
+  if (/,$/.test(trimmed)) {
```
akshaymagapu: **Minor:** `/,$/` catches a single character and misses adjacent junk like `/)` or `/];`. Either widen to a trailing-punctuation cluster (e.g. `/[,.;)\]}>]+$/`) or drop the clause - a one-char rule is the worst of both.

src/llm-error-pages/handler.js
```diff
@@ -186,6 +187,17 @@ export async function runAuditAndSendToMystique(context) {
         const processedResults = processErrorPagesResults(results);
         const categorizedResults = categorizeErrorsByStatusCode(processedResults.errorPages);
 
+        // Drop malformed URLs (e.g. '/brandshttps://...', '/),') emitted by
+        // bot impersonators that slip through the user-agent filter upstream.
+        for (const code of Object.keys(categorizedResults)) {
+          const before = categorizedResults[code].length;
+          categorizedResults[code] = categorizedResults[code].filter((e) => isValidUrlPath(e.url));
```
akshaymagapu: **Important:** The natural boundary for dropping malformed CDN-log rows is `processErrorPagesResults`, not the middle of the handler. Filtering here also leaves the persisted `errorPages` (line 310) and `totalErrors`/`summary` (lines 308-309) carrying the unfiltered set - a dual source of truth. Filter once upstream so `errorPages`, `categorizedResults`, Excel, and Mystique are all consistent by construction, and the inline loop disappears.

src/llm-error-pages/handler.js
```diff
@@ -186,6 +187,17 @@ export async function runAuditAndSendToMystique(context) {
...
+          const dropped = before - categorizedResults[code].length;
+          if (dropped > 0) {
+            log.info(`[LLM-ERROR-PAGES] Filtered ${dropped} malformed URL(s) from status ${code}`);
```
akshaymagapu: **Important:** This logs only a count, giving no way to diagnose false positives in production ("why is my 404 count lower this week?"). Log a sample of the first ~5 dropped URLs per bucket so the regex tightening in #1/#2 is verifiable and the next false-positive class is catchable from Coralogix.

akshaymagapu (changes requested):
Hey @ymadobe,

Thanks for the focused, well-documented fix. The seam is right (filter runs after categorization and before all three consumers - Excel, opportunity sync, Mystique), the per-bucket drop logging fits the module's conventions, and the unit coverage on the new function is thorough. All four reviewers (staff engineer, architect, security, QA) converged on the same core concern: two of the regex rules silently drop legitimate customer URLs.

## Strengths
- Tight scope - one utility plus one filter loop, no drive-by refactors (`src/llm-error-pages/utils.js`, `src/llm-error-pages/handler.js:192`).
- Filter placement is correct: it runs before Excel writes (`handler.js:234`), the 50-row cap and consolidation, and the Mystique payload (`handler.js:242`), so all sinks see one filtered set.
- `new URL()` parse-first, then cheap string checks - sensible ordering. No ReDoS: all four regexes are linear-time on bounded CDN-log input.
- JSDoc explains the why (bot impersonators, 404 noise), matching the repo's comment style.
- Drop count is logged as an integer, not the URL - no log-injection vector introduced.

## Issues

### Critical (Must Fix)

**1. Embedded-scheme regex drops legitimate redirect/share URLs - `utils.js:477`**
`/.+https?:\/\//i` matches `https://` anywhere after one character, so it silently drops legitimate 404s like `/redirect?to=https://example.com`, `/share?url=https://...`, and OAuth `/auth/callback?return=https://...`. These are exactly the broken redirects a customer needs surfaced. Separately, the leading `.+` means a bare absolute URL (`https://evil.com/foo`) passes the check entirely. Tighten to match a scheme glued to a path segment only, e.g. `/[^/?&=#]https?:\/\//i` - this still catches `/brandshttps://...` but allows query-embedded schemes through.

**2. Quote regex drops any URL containing an apostrophe - `utils.js:480`**
`/["']|%2[27]|%252[27]/i` rejects `'` or `"` anywhere in the URL. That drops common legitimate paths: `/products/o'reilly`, `/authors/o'brien`, `/it's-friday`, and any percent-encoded quote (`%27`). Raw quotes generally can't survive `new URL()` parsing in a path anyway, so the high-value signal is the double-encoded form. Recommend keeping only `%2522`/`%2527` and dropping the raw-quote and single `%22`/`%27` rejection. If single-encoded quote rejection is intentional, document why and add a dropped-URL sample to the log so false positives are detectable.

### Important (Should Fix)

**3. Persisted `errorPages` keeps the unfiltered set - `handler.js:310`**
The filter mutates `categorizedResults`, but the audit record still stores `errorPages: processedResults.errorPages` (line 310) and `totalErrors`/`summary` (lines 308-309) from the pre-filter data. Any future consumer reading `audit.getAuditResult()[i].errorPages` gets the noise back - a dual source of truth that will cause "dashboard disagrees with the spreadsheet" confusion. Filter once at the data boundary (in `processErrorPagesResults`) so every downstream view is consistent by construction, which also removes the inline handler loop.

**4. URL parser is far more permissive than "valid URL" implies - `utils.js:472`**
`new URL(trimmed, 'https://example.com')` accepts `javascript:alert(1)`, `data:...`, null bytes (`%00`), and CRLF (`%0d%0a`). Since these flow into the Mystique LLM payload and Excel, restrict to `parsed.protocol === 'http:' || 'https:'`, or correct the comment to describe what the gate actually does.

**5. CSV/formula injection into Excel - `handler.js:214` (pre-existing, natural to fix here)**
URLs beginning with `=`, `+`, `-`, or `@` pass `isValidUrlPath` and are written verbatim to spreadsheet cells. When an analyst opens the xlsx, Excel/Sheets/Calc re-interpret these as formulas (e.g. `=WEBSERVICE(...)` exfiltration). Given the PR's own threat model names adversarial CDN logs and Excel as a sink, either prepend a `'` sentinel on cell write or reject a leading `=+-@` in `isValidUrlPath`.

**6. Drop log has no URL sample - `handler.js:197`**
`Filtered N malformed URL(s) from status <code>` gives no way to diagnose false positives in production. Log a sample (first ~5 dropped URLs) so the regex tightening above is verifiable and the next false-positive class is catchable from Coralogix.

### Minor (Nice to Have)

**7. Trailing-comma rule is too narrow - `utils.js:483`** `/,$/` catches one character and misses adjacent junk (`/)`, `/];`). Either widen to a trailing-punctuation cluster or drop the clause; a single-char rule is the worst of both.

**8. Use `URL.canParse` - `utils.js:471`** Node 24 supports `URL.canParse(trimmed, 'https://example.com')`, removing the `try/catch` and the `eslint-disable no-new`.

**9. Test gaps** Add "must accept" guardrail tests for `/products/o'reilly` and `/redirect?to=https://...` (both fail today, exposing #1/#2); a non-404 bucket case (loop is only exercised on 404); and an all-dropped/empty-bucket case.

## Recommendations
- Move the filter into `processErrorPagesResults` so `errorPages`, `categorizedResults`, Excel, and Mystique all share one filtered, consistent set.
- This is a denylist that will accrete patterns under incident pressure. Link the upstream UA/IP-fingerprint hardening ticket in the code comment and treat this as a tracked stopgap, or flip to an allowlist-shape check (RFC 3986 path charset) that does not grow with each new bot shape.
- Consider surfacing an aggregate `malformedUrlsDropped` metric/summary field rather than only a log line, so a regression where most rows get dropped is visible.

## Assessment
Ready to merge? With fixes. The approach and placement are sound, but issues 1 and 2 silently drop legitimate customer 404s into the void - the regexes need tightening before this lands. Issues 3-6 are small and worth folding in; the long-term denylist trajectory can be a tracked follow-up.

## Next Steps
1. Fix the two regex false positives (#1, #2) - these are the blockers.
2. Add the dropped-URL log sample (#6) and the "must accept" tests (#9) so false positives are catchable.
3. Address the persisted-data inconsistency (#3) and parser/Excel hardening (#4, #5).
4. Minor items are optional.

akshaymagapu (approved):

PR #1999, opened 2026-02-17
feat: Adding auto-detect for publish during broken backlinks audits run

absarasw:
This PR replaces https://github.com/adobe/spacecat-audit-worker/pull/1730

Implements these -

1. Automated detection of Publish for Broken Backlinks opportunity
2. Updates common utility data-access#syncSuggestions method to:
  -  optionally receive `isIssueFixedByAISuggestion` & `isIssueResolvedOnProduction` from opportunity handlers
  - update a NEW Suggestion to Fixed & creating its FixEntity as Published, if `isIssueFixedByAISuggestion` is true
  - updated a Deployed FixEntity to Published, if `isIssueResolvedOnProduction` is true for all Suggestions of the FixEntity
  - not ever update already Fixed Suggestions (whether published or not), if the same issue re-appears
  - create a New Suggestion, if same issue re-appears and previous Suggestions were Deployed+Published

----

Strategies for publish auto-detection is mentioned in https://wiki.corp.adobe.com/pages/viewpage.action?spaceKey=AEMSites&title=AEM+Sites+Optimizer+%7C+Auto-detect+Publish#AEMSitesOptimizer|AutodetectPublish-Strategies

Please ensure your pull request adheres to the following guidelines:
- [ ] make sure to link the related issues in this description
- [ ] when merging / squashing, make sure the fixed issue references are visible in the commits, for easy compilation of release notes
- [ ] If data sources for any opportunity has been updated/added, please update the [wiki](https://wiki.corp.adobe.com/display/AEMSites/Data+Sources+for+Opportunities) for same opportunity.

## Related Issues

Thanks for contributing!

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Kanishka", "Dominique Jäggi"]
rationale: "Auto-added baseline (no PR assessment present at CI time); refine via the cmr skill if this change is higher-impact."
```

PR #2593, opened 2026-06-03
fix(sync): prevent spurious updatedAt bumps on terminal suggestions

absarasw:
## Problem

SpaceCat's `syncSuggestions()` function unconditionally bumps `updatedAt` on every suggestion it processes during audit runs, regardless of whether status or data actually changed. This makes `updatedAt` useless as a lifecycle indicator.

**Impact:**
- REJECTED suggestions: ~30-38% weekly `updatedAt` refresh despite no changes
- ERROR: ~80% weekly refresh
- PENDING_VALIDATION: ~60% weekly refresh
- Dashboard "Moved to Rejected" metrics count total rejected suggestions instead of new transitions

## Solution

Only add suggestions to `toUpdate` array if data or status actually changed:
- Added `deepEqual` check before saving
- Only save when `newStatus !== null` OR `dataChanged`
- Reduces database write load

## Changes

**File**: `src/utils/data-access.js`
1. Import `deepEqual` from `@adobe/spacecat-shared-utils`
2. Modified `syncSuggestions` to only save when changes detected

## Testing

- ✅ All tests updated to match new behavior
- ✅ Linting passes
- Behavior change is backward compatible for all statuses

## Expected Outcomes

- `updatedAt` becomes reliable indicator of last meaningful change
- Dashboard metrics accurately track new transitions
- Reduced database write load (~30-80% reduction depending on status)

## JIRA

Closes [SITES-44646](https://jira.corp.adobe.com/browse/SITES-44646)


MysticatBot (changes requested):
Hey @absarasw,

## Strengths

- **Correct optimization strategy** (`src/utils/data-access.js:492-520`): Using `deepEqual` to gate saves eliminates write amplification on unchanged suggestions, making `updatedAt` a meaningful lifecycle indicator and reducing PostgREST/PostgreSQL load proportional to stable suggestions per audit run.
- **FIXED/OUTDATED unification is well-reasoned** (`src/utils/data-access.js:358-366`): Treating FIXED the same as OUTDATED on reappearance correctly closes a gap where regressions on previously-fixed issues would be invisible to customers.
- **Bulk save pattern preserved** - The change accumulates into `toUpdate` and calls `Suggestion.saveMany` once, maintaining the N+1 avoidance contract documented in CLAUDE.md.
- **Test adaptations validate new semantics** - The `llm-blocked` test correctly asserts that identical data does NOT trigger a save; the `cwv` test introduces a `pageviews` delta to meaningfully exercise the update path.
- **No new attack surface** - The change operates entirely on internal audit data within a trusted zone; `deepEqual` is imported from an existing team dependency (`@adobe/spacecat-shared-utils`) already in the lockfile.

## Issues

### Important (Should Fix)

**1. Missing dedicated test for FIXED suggestion regression path (Phase 2 headline behavior)**

`test/utils/data-access.test.js` - The existing test "should reopen fixed suggestions" (around line 1365) actually sets `getStatus: () => 'OUTDATED'` and asserts the OUTDATED log message. There is no test that creates a suggestion with `getStatus() === 'FIXED'` and verifies it transitions to NEW or PENDING_VALIDATION with the correct log message `"FIXED suggestion found in audit. Regression detected - reopening."`.

CI passes with 100% coverage, so the branch is being hit, but coverage does not equal intentional behavioral verification. This is the headline behavior of Phase 2 and deserves an explicit test proving the contract.

**Fix:** Add a sibling test:
```js
it('should reopen FIXED suggestions as regression when they reappear in audit', async () => {
  // existingSuggestions with getStatus: () => 'FIXED'
  // newData with changed values
  // Assert: setStatus called with NEW (or PENDING_VALIDATION for validation sites)
  // Assert: log.warn called with 'FIXED suggestion found in audit. Regression detected - reopening.'
});
```

**2. CLAUDE.md descriptions now stale**

Per the rule: "When implementation changes, update README in the same PR."

- `syncSuggestions` is described as "diffs current audit findings against stored suggestions; adds new ones, removes resolved ones." It now also skips no-op updates via deep equality and handles FIXED status regressions.
- `syncSuggestionsWithPublishDetection` is described as "extends syncSuggestions with fix-regression detection." Basic FIXED regression handling has moved into the base `defaultMergeStatusFunction`, so this description is partially misleading about what the extension actually provides.

**Fix:** Update the "Opportunities and Suggestions Pattern" section in `CLAUDE.md` to reflect the new responsibilities.

### Minor (Nice to Have)

**3. `getData()` called redundantly** (`src/utils/data-access.js:~499-505`)

`existing.getData()` is called in the filter (for `buildKey`), then again in the forEach for `buildKey`, `mergeDataFunction`, and `deepEqual` - 4 calls total. Caching the result once at the top of the forEach would be clearer and avoids depending on `getData()` always returning the same reference:

```js
const existingData = existing.getData();
const existingKey = buildKey(existingData);
const mergedData = mergeDataFunction(existingData, newDataItem);
const dataChanged = !deepEqual(existingData, mergedData);
```

**4. Misleading test name** (`test/utils/data-access.test.js:~1365`)

"should reopen fixed suggestions" actually tests OUTDATED status. Now that FIXED is a real code path, the name actively misleads. Consider renaming to "should reopen OUTDATED suggestions as regression".

**5. Stale debug log** (`src/utils/data-access.js` - last debug line after saveMany)

`Updated existing suggestions = ${existingSuggestions.length}` reports the total matched count, not the actual update count. With Phase 1 skipping unchanged suggestions, this is now confusing alongside the new `log.info` that correctly reports `toUpdate.length`. Consider removing or updating to `Processed ${existingSuggestions.length} matched suggestions, updated ${toUpdate.length}`.

## Assessment

Ready to merge? With fixes

The optimization is correct and well-scoped. The FIXED/OUTDATED unification closes a real gap. Two items should be addressed: add an explicit behavioral test for the FIXED regression path (Phase 2's headline behavior) and update CLAUDE.md descriptions that are now inaccurate. Minor items are optional polish.

## Next Steps

1. Add a dedicated test for FIXED suggestion reopening behavior.
2. Update CLAUDE.md's syncSuggestions/syncSuggestionsWithPublishDetection descriptions.
3. Minor items (cache getData, rename test, fix debug log) are optional.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 18s | <b>Cost:</b> $4.24 | <b>Commit:</b> <code>9f0604b87c6fcc8362ae1199334d6e4be123c4f6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/utils/data-access.js
```diff
@@ -334,7 +334,7 @@ const defaultMergeDataFunction = (existingData, newData) => ({
  * Default merge function for determining the status of an existing suggestion.
  * This function encapsulates the default behavior for status transitions:
  * - REJECTED suggestions remain REJECTED
- * - OUTDATED suggestions transition to PENDING_VALIDATION or NEW (possible regression)
+ * - FIXED and OUTDATED suggestions transition to PENDING_VALIDATION or NEW (regression detected)
```
habansal: why this change?

@absarasw didn't you already implement this:
> not ever update already Fixed Suggestions (whether published or not), if the same issue re-appears
create a New Suggestion, if same issue re-appears and previous Suggestions were Deployed+Published

via https://github.com/adobe/spacecat-audit-worker/pull/1999

absarasw: Yes, missed that.

MysticatBot (commented):
Hey @absarasw,

## Strengths

- **All prior review findings addressed cleanly.** CLAUDE.md descriptions updated (commit 56ab7c3), FIXED regression behavior correctly reverted per habansal's feedback (commit dd71df0), redundant `getData()` call eliminated, debug log updated to report actual vs processed counts.
- **Core fix is well-reasoned** (`src/utils/data-access.js:492-525`): The `deepEqual` guard before calling `setData`/`setUpdatedBy` directly addresses the stated problem. The condition `newStatus !== null || dataChanged` correctly handles all four cases: data-only change, status-only change, both, and neither (skip).
- **No new attack surface**: The change operates entirely within an internal data sync path using an existing team dependency (`@adobe/spacecat-shared-utils`). No new endpoints, auth changes, or credential handling.
- **Test updates validate new semantics correctly** (`test/audits/llm-blocked.test.js:325-330`): Asserts that when data and status are unchanged, `setData`, `setUpdatedBy`, and `saveMany` are NOT called.
- **CWV test adaptation is honest** (`test/audits/cwv.test.js:506`): Returns data with a different `pageviews` value so `deepEqual` detects a genuine change, ensuring the test still exercises the update path.
- **CLAUDE.md proactively updated** to document the new deep equality behavior in `syncSuggestions`.

## Issues

### Minor (Nice to Have)

**1. Redundant O(N) re-filter in debug log** - `src/utils/data-access.js` (debug log after the save block)

```js
log.debug(`Processed ${existingSuggestions.filter((e) => newDataKeys.has(buildKey(e.getData()))).length} matched suggestions, updated ${toUpdate.length}`);
```

This re-iterates `existingSuggestions` and re-calls `buildKey(e.getData())` for every element on every sync invocation, just to produce a debug message. The matched count is already implicitly available from the `.filter(...).forEach(...)` chain above. Consider storing the intermediate filtered array in a variable and using its `.length`:

```js
const matched = existingSuggestions.filter((existing) => {
  const existingKey = buildKey(existing.getData());
  return newDataKeys.has(existingKey);
});

matched.forEach((existing) => { /* ... */ });

// later:
log.debug(`Processed ${matched.length} matched suggestions, updated ${toUpdate.length}`);
```

This also eliminates the double `getData()` + `buildKey()` call per matched suggestion (once in filter, once in forEach).

**2. "Phase 1" comment may mislead future readers** - `src/utils/data-access.js:491`

The comment `// Phase 1: Only save suggestions where data or status actually changed` implies there is a Phase 2 coming. If this is the final design, the label adds confusion. If there is a planned Phase 2, a brief note of what it entails would help. Same applies to "Phase 1" references in test comments.

## Recommendations

- **Document the mergeDataFunction contract.** The `deepEqual(existingData, mergedData)` comparison assumes `mergeDataFunction` returns a new object rather than mutating `existingData` in place. The default (`defaultMergeDataFunction`) uses spread and is safe, but a brief inline comment noting this invariant would protect future contributors from introducing a subtle bug where deepEqual compares a mutated object to itself.

## Assessment

The core behavioral change is correct, well-tested, and addresses a real problem. The redundant O(N) computation in the debug log is easy to clean up but not blocking. No security concerns, no conventions violations, and all prior findings have been addressed.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 15s | <b>Cost:</b> $3.40 | <b>Commit:</b> <code>34f853439f09c140752dcf173a190f7a967942c6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @absarasw,

## Strengths

- **All prior Important findings addressed cleanly.** CLAUDE.md descriptions updated (commit 56ab7c3) to accurately reflect the new `syncSuggestions` deep-equality optimization and the corrected `syncSuggestionsWithPublishDetection` responsibilities. The FIXED/OUTDATED regression path was correctly reverted per habansal's feedback (commit dd71df0).
- **Core optimization is correct** (`src/utils/data-access.js:504-519`): The guard `if (newStatus !== null || dataChanged)` correctly gates both `setData` and `setStatus` mutations behind an actual-change check, eliminating write amplification on unchanged suggestions.
- **Prior code-level feedback incorporated:** `getData()` result is now cached as `existingData` and reused, eliminating the redundant calls flagged earlier. The stale debug log that dumped all `existingSuggestions` regardless of update count is replaced with a focused message reporting `matchedSuggestions.length` vs `toUpdate.length`.
- **Test adaptations validate new semantics correctly** - `llm-blocked.test.js` asserts that identical data/status does NOT trigger a save; `cwv.test.js` introduces a `pageviews` delta to exercise the update path honestly.
- **No new attack surface** - the change operates entirely on internal audit data within a trusted zone, using an existing team dependency (`@adobe/spacecat-shared-utils`). No new endpoints, auth changes, or credential handling.
- **Conventions compliance** - CLAUDE.md proactively updated to document new behavior, testing patterns preserved.

## Issues

### Minor (Nice to Have)

**1. "Phase 1" comment in test may mislead future readers** - `test/audits/llm-blocked.test.js:~329`

The comment `// Phase 1: With same robotsTxtHash (same data) and SKIPPED status unchanged, no save should occur` references "Phase 1" without context. There is no Phase 2 in this PR or linked work. The source code comment was cleaned up, but this test comment still carries the label.

Suggestion: `// Same data + unchanged SKIPPED status -> no save should occur`

**2. `buildKey(existingData)` called twice per matched suggestion** - `src/utils/data-access.js:~497-503`

The key is computed once in the `.filter()` but not carried forward - it is recomputed in the `.forEach()`. For typical batch sizes this is negligible, but if the separation is kept, consider using a `Map` or tuple to carry the key from filter to forEach rather than recomputing.

## Recommendations

- **Defensive guard for `mergeDataFunction` contract.** The comment on line ~502 notes that `mergeDataFunction` must return a new object (not mutate `existingData`) for `deepEqual` to work correctly. If any caller's merge function mutates in place, `deepEqual` would always return true, silently suppressing all updates. A cheap runtime guard (`if (mergedData === existingData) log.warn(...)`) would catch contract violations early without blocking the hot path.
- **Explicit test for the status-only-change branch.** When `newStatus !== null` but `dataChanged === false`, the code enters the update path and calls `setData` with identical data plus `setStatus`. This branch is exercised (100% coverage confirms it), but a named test case for this specific combination would make the behavioral contract self-documenting.

## Assessment

The core behavioral change is correct, well-tested, and addresses the stated problem. All prior Important findings have been addressed. The remaining items are optional polish that do not affect correctness or safety. No security concerns, no conventions violations.

## Next Steps

1. Optional: clean up the "Phase 1" test comment.
2. Optional: consider the defensive guard for the mergeDataFunction contract as a follow-up.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 12s | <b>Cost:</b> $2.52 | <b>Commit:</b> <code>1374d3078566961b8f6a8c76107ffa7d88d2d1e7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/utils/data-access.js
habansal: What PR #1999 did NOT deliver (despite saying so in the description):

  The intermediate commits in the PR branch (39bcfe66, 790687c6) had the logic:
  1. skipStatuses = [FIXED, IN_PROGRESS] — skip updating FIXED suggestions' data
  2. fullyPublishedFixedKeys — allow new suggestion creation when FIXED+PUBLISHED issue reappears
  
  But those commits were not carried into the final squash-merge (b24e8fe7). The merged syncSuggestions has a simple
  existingSuggestionKeys Set with no FIXED+PUBLISHED exception.

src/utils/data-access.js
```diff
@@ -489,33 +489,51 @@ export async function syncSuggestions({
...
+    // Check if data actually changed using deep equality
+    const dataChanged = !deepEqual(existingData, mergedData);
+
+    // Only update if something actually changed
+    if (newStatus !== null || dataChanged) {
+      existing.setData(mergedData);
+      if (newStatus !== null) {
+        existing.setStatus(newStatus);
+      }
+      existing.setUpdatedBy('system');
```
tarunsinghdev: Could we scope the `setUpdatedBy('system')` so it only fires when the audit actually changes the **status**?

```diff
     if (newStatus !== null || dataChanged) {
       existing.setData(mergedData);
       if (newStatus !== null) {
         existing.setStatus(newStatus);
+        existing.setUpdatedBy('system');   // only when the audit changes the status
       }
-      existing.setUpdatedBy('system');
       toUpdate.push(existing);
     }
```

This keeps everything your PR does (a data-only change still calls `setData` → `updatedAt` bumps, row still pushed to `toUpdate`), but stops overwriting the actor on a pure data refresh — preserving attribution. The alt-text audit already behaves this way (it never re-stamps existing suggestions, only the opportunity).

tarunsinghdev (changes requested):

src/utils/data-access.js
tarunsinghdev: Context: https://jira.corp.adobe.com/browse/SITES-46233

MysticatBot (commented):
Hey @absarasw,

**Verdict: Comment** - all prior findings addressed cleanly; only minor polish remains.
_Changes: adds a `deepEqual` gate to `syncSuggestions()` so suggestions are only persisted when data or status actually changed, preventing spurious `updatedAt` bumps and reducing database write load (23 files)._

<details>

- suggestion: Defensive guard warns but proceeds to silently skip the save - consider updating the warning message to say "data change will be skipped" or failing open (`dataChanged = true`) so a broken merge function doesn't silently suppress updates - `src/utils/data-access.js:511-513`
- nit: Comment says "Filter matched suggestions once to avoid redundant buildKey() calls" but `buildKey(existingData)` is still computed in both the filter and the forEach - update comment to "Separate filter from processing for clarity" - `src/utils/data-access.js:496-504`
</details>

- `setUpdatedBy('system')` scoped to status changes only (tarunsinghdev feedback addressed)
- "Phase 1" test comment cleaned up
- Defensive guard for mergeDataFunction contract added with test coverage
- Explicit tests for status-only and data-only change branches added

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 52s | <b>Cost:</b> $5.09 | <b>Commit:</b> <code>79a45100239566bb81629b9112b3b6149a75d294</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @absarasw,

**Verdict: Approve** - all prior suggestions addressed cleanly, no new concerns.
_Changes: makes the `mergeDataFunction` defensive guard fail-open (`dataChanged = true`) instead of warn-only, and corrects an inaccurate comment (23 files)._

- Defensive guard now forces `dataChanged = true` on same-reference detection instead of silently continuing after warning
- Comment updated from "Filter matched suggestions once to avoid redundant buildKey() calls" to "Separate filter from processing for clarity"

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 23s | <b>Cost:</b> $2.07 | <b>Commit:</b> <code>5ceb2d577d7d300467b51ecb3ff5d2afac5e9b0b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

tarunsinghdev (approved):

PR #2594, opened 2026-06-03
fix(csp): skip backward-compatible CSP findings

bisswanger:
## Problem

Lighthouse's `csp-xss` audit emits a `Medium`-severity finding suggesting that `'unsafe-inline'` be added to `script-src` to remain *backward compatible with older browsers*:

> Consider adding `'unsafe-inline'` (ignored by browsers supporting nonces/hashes) to be backward compatible with older browsers.

All modern browsers support CSP nonces/hashes and `'strict-dynamic'`, so this is not an actionable improvement for our AEM Edge customers — it just adds noise to the CSP opportunity.

## Change

`cspOpportunityAndSuggestions` now filters out any flattened CSP item whose `description` contains `"backward compatible"` before building the opportunity and suggestions.

## Test

Added a `should skip backward compatible findings` test that feeds a realistic `csp-xss` `details.items` array (host-allowlist High finding, the backward-compat Medium finding, and a `<meta>`-tag Medium finding) and asserts:
- only **2** suggestions are produced (`mainMetric.value === 2`), the backward-compat item dropped;
- `"nonces or hashes"` is still normalized to `"nonces"`;
- the `<meta>`-tag finding flows through with `directive: undefined`.

### DIFF
--- a/src/csp/csp.js
+++ b/src/csp/csp.js
@@ -141,6 +141,9 @@ export async function cspOpportunityAndSuggestions(auditUrl, auditData, context,
   csp = flattenCSP(csp);
   log.debug(`[${AUDIT_TYPE}] [Site: ${site.getId()}] CSP information from lighthouse report: ${JSON.stringify(csp)}`);
 
+  // all modern browsers support `strict-dynamic` directive, we don't need related suggestions
+  csp = csp.filter((item) => !item.description?.includes('backward compatible'));
+
   csp.forEach((item) => {
     if (item.description && item.description.includes('nonces or hashes')) {
       // eslint-disable-next-line no-param-reassign
--- a/test/audits/csp/csp.test.js
+++ b/test/audits/csp/csp.test.js
@@ -476,6 +476,81 @@ describe('CSP Post-processor', () => {
     ));
   });
 
+  it('should skip backward compatible findings', async () => {
+    sinon.replace(configuration, 'isHandlerEnabledForSite', (toggle) => toggle === 'security-csp');
+    // Mirrors a real Lighthouse `csp-xss` audit `details.items` array.
+    auditData.auditResult.csp = [
+      {
+        directive: 'script-src',
+        description: 'Host allowlists can frequently be bypassed. Consider using CSP nonces or hashes instead, along with `\'strict-dynamic\'` if necessary.',
+        severity: 'High',
+      },
+      {
+        directive: 'script-src',
+        description: 'Consider adding `\'unsafe-inline\'` (ignored by browsers supporting nonces/hashes) to be backward compatible with older browsers.',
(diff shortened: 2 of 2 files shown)


MysticatBot (approved):
Hey @bisswanger,

## Strengths

- `src/csp/csp.js:143-144`: The filter is placed at exactly the right point in the pipeline - after `flattenCSP` (so subitems are already expanded) and before the `nonces or hashes` normalization and the `csp.length` check. All downstream consumers (count metric, auto-suggest, suggestion sync) see the correct, filtered set.
- The change follows the existing pattern already established in the same function (the `nonces or hashes` string match). No new architectural surface introduced.
- `test/audits/csp/csp.test.js:478-556`: The test uses a realistic 3-item input array that exercises discrimination: one item filtered out, one High-severity item with nonces normalization, and one Medium item without a directive. This validates the filter is selective rather than over-broad.
- `test/audits/csp/csp.test.js:508`: Asserting `mainMetric.value: 2` directly proves the count dropped from 3 inputs to 2 suggestions, which is the core contract of this change.
- The change is well-scoped: one filter line, one test. No unrelated refactoring mixed in.

## Issues

### Minor (Nice to Have)

1. `src/csp/csp.js:144` - The filter uses a case-sensitive substring match (`'backward compatible'`). If Lighthouse ever ships a capitalization variant (e.g., `"Backward compatible"` at the start of a sentence), the filter would miss it. The existing `includes('nonces or hashes')` on line 148 has the same characteristic, so this is consistent with the codebase pattern. A case-insensitive alternative would be `!item.description?.toLowerCase().includes('backward compatible')`, but purely optional given the existing convention.

## Recommendations

- The function now has two distinct string-match-based transformations on Lighthouse output (this filter and the `nonces or hashes` normalization). If a third or fourth arrives, consider extracting a declarative filter/transform table rather than accumulating inline string checks. Not needed today, but a useful decision point for the future.

## Assessment

Ready to merge? Yes

The change is minimal, well-placed, and tested. The filter correctly removes noise before it reaches count calculations, auto-suggest, and suggestion sync. The coupling to Lighthouse's output phrasing already exists in the same function; if Lighthouse changes wording, the worst outcome is the noise returns - no data loss, no broken functionality. No security concerns (the filter operates on trusted Lighthouse output, not user-supplied input).

Note: CI checks were still pending at time of review.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 33s | <b>Cost:</b> $2.04 | <b>Commit:</b> <code>12356a6704c0b9c5bf22834755e93e0657449f73</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2595, opened 2026-06-03
feat(csp): run CSP audit for Crosswalk sites with valid hlxConfig

bisswanger:
## Problem

The CSP opportunity was gated solely on the `AEM_EDGE` delivery type:

```js
if (site.getDeliveryType() !== Site.DELIVERY_TYPES.AEM_EDGE) {
  return { ...auditData };
}
```

Some **Crosswalk** environments are labelled `AEM_CS` while still being edge-delivered (helix). Those sites were incorrectly skipped, so they never got a CSP opportunity.

## Change

Introduced `isCspApplicable(site)` which gates the audit on either:
- delivery type `AEM_EDGE` (unchanged behavior), **or**
- a valid `hlxConfig` — an `rso` carrying both `owner` and `site`, which identifies the backing Edge Delivery repo for Crosswalk (`AEM_CS`) sites.

## Tests

- `should extract opportunity for a Crosswalk (AEM_CS) site with a valid hlxConfig` — `AEM_CS` + valid `rso` → opportunity created.
- `should not extract opportunity for AEM_CS site without a valid hlxConfig` — `AEM_CS` + empty `rso` → skipped.
- Existing "other delivery types" test (no `hlxConfig`) still skips as before.

100% coverage on `src/csp/csp.js`; full lint clean.

### DIFF
--- a/src/csp/csp.js
+++ b/src/csp/csp.js
@@ -55,6 +55,26 @@ export function buildKey(data) {
   return key;
 }
 
+/**
+ * Determines whether the CSP opportunity applies to a site.
+ *
+ * The opportunity targets Edge Delivery (helix) sites. These are normally
+ * flagged with the AEM_EDGE delivery type, but some Crosswalk environments are
+ * labelled AEM_CS while still being edge-delivered. Those sites carry a valid
+ * hlxConfig (an rso identifying the backing repo), so accept them as well.
+ *
+ * @param {Object} site - The site object.
+ * @returns {boolean} True if the CSP opportunity should be evaluated.
+ */
+export function isCspApplicable(site) {
+  if (site.getDeliveryType() === Site.DELIVERY_TYPES.AEM_EDGE) {
+    return true;
+  }
+
+  const hlxConfig = site.getHlxConfig?.();
+  return Boolean(hlxConfig?.rso?.owner && hlxConfig?.rso?.site);
+}
+
 function flattenCSP(csp) {
   return csp.flatMap((item) => {
     if (item.subItems?.items) {
@@ -129,8 +149,9 @@ export async function cspOpportunityAndSuggestions(auditUrl, auditData, context,
     return { ...auditData };
   }
 
-  // this opportunity is only relevant for aem_edge delivery type at the moment
-  if (site.getDeliveryType() !== Site.DELIVERY_TYPES.AEM_EDGE) {
+  // this opportunity is only relevant for edge-delivered (helix) sites: either
+  // the AEM_EDGE delivery type or a Crosswalk site (e.g. AEM_CS) with a valid hlxConfig
+  if (!isCspApplicable(site)) {
     log.debug(`[${AUDIT_TYPE}] [Site: ${site.getId()}] skipping CSP opportunity as it is of delivery type ${site.getDeliveryType()}`);
     return { ...auditData };
   }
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @bisswanger,

## Strengths

- `src/csp/csp.js:67-77`: Extracting `isCspApplicable` as a named, exported predicate is well done. It gives the gating logic a testable surface, a clear name, and a JSDoc that explains the "why" (Crosswalk sites labelled AEM_CS that are actually edge-delivered). This is how eligibility checks should be factored.
- `src/csp/csp.js:70`: The use of optional chaining on `site.getHlxConfig?.()` is appropriate here - `getHlxConfig` may not exist on all site objects, making this a legitimate boundary check rather than a silent skip.
- `test/audits/csp/csp.test.js:676-721`: Tests cover both the positive path (valid hlxConfig triggers the opportunity) and the negative path (empty rso skips), asserting on observable behavioral outcomes (`opportunityStub.create`, `cspOpportunity.addSuggestions`) rather than implementation internals.
- The change is additive and backward-compatible - previously eligible sites remain eligible, previously skipped sites get a second chance only when concrete evidence (hlxConfig.rso) confirms edge delivery.
- Change discipline is tight: one new function, one call-site swap, two tests. No unrelated changes mixed in.
- The gating logic operates on trusted internal data (site objects from the data access layer), not untrusted external input - no new trust boundary crossings.

## Issues

### Minor (Nice to Have)

1. **src/csp/csp.js:152** - The debug log message still reads "skipping CSP opportunity as it is of delivery type ..." which is slightly misleading for the Crosswalk case. When an AEM_CS site has an invalid hlxConfig, the skip reason is the missing hlxConfig, not strictly the delivery type. Consider adjusting to reflect the actual gate (e.g., "skipping CSP opportunity - site is not CSP-applicable (delivery type: ...)"). Low priority since it is a debug-level log, but when debugging why a Crosswalk site was skipped, the current message could point at the wrong thing.

2. **test/audits/csp/csp.test.js:699** - The negative test covers `rso: {}` (both owner and site missing) but does not cover the partial case where only one of `owner` or `site` is present (e.g., `rso: { owner: 'myorg' }` without `site`). The `&&` logic would also reject this, and a test for this boundary would document the contract explicitly. Consider adding a case with a partial rso asserting the opportunity is not created.

## Recommendations

- The concept "is this site edge-delivered regardless of its delivery-type label" is not specific to the CSP audit. If other audits hit the same Crosswalk labeling gap in the future, consider extracting the predicate to a shared utility (e.g., `src/common/site-utils.js`) rather than letting each audit re-derive the heuristic independently. Not blocking - the duplication does not yet exist.

## Assessment

Ready to merge? Yes
Reasoning: Clean, well-scoped change that correctly extends CSP audit eligibility to Crosswalk sites. The logic is sound, tests cover the meaningful behavioral paths, and no security, architectural, or convention concerns were identified. CI is pending - confirm it passes before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 36s | <b>Cost:</b> $2.24 | <b>Commit:</b> <code>c377fe2afc8192aaa2ff50b96694b2b5380f5c8b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2597, opened 2026-06-03
fix(rum-config-refresh): add www. prefix fallback for domain key lookup

sandsinh:
## Summary

- Expands the RUM domain candidate list in `rum-config-refresh` to include a `www.{domain}` fallback for any hostname that doesn't already start with `www.`
- Bare domain is always tried first; `www.` is only attempted if the bare domain returns 404
- Sites stored with a `www.` prefix are unaffected — no `www.www.` duplication

## Background

A bulk check across **306 sites** with `hasDomainKey: false` + a `lastCheckedAt` present revealed that **232 sites (76%)** have their RUM domain key registered under `www.{domain}` but not the bare domain. These sites were incorrectly persisted with `hasDomainKey: false`, blocking RUM data collection.

Fixes [SITES-46023](https://jira.corp.adobe.com/browse/SITES-46023)

## Changes

- `src/rum-config-refresh/handler.js` — add `withWwwFallback` helper; expand `domains` array
- `test/audits/rum-config-refresh.test.js` — 5 new test cases covering: www fallback success, no www.www. duplication, both bare+www fail, override www fallback

## Test plan

- [x] All 23 tests pass (`npm run test:spec -- test/audits/rum-config-refresh.test.js`)
- [x] 100% coverage on `src/rum-config-refresh/handler.js`
- [x] Existing tests unaffected

### DIFF
--- a/src/rum-config-refresh/handler.js
+++ b/src/rum-config-refresh/handler.js
@@ -70,8 +70,17 @@ export default async function rumConfigRefresh(message, context) {
     return ok({ skipped: true, reason: 'malformed baseURL' });
   }
 
-  // override-first: overrideBaseURL is the site's canonical RUM domain when set
-  const domains = [...new Set([overrideHostname, baseHostname].filter(Boolean))];
+  // override-first: overrideBaseURL is the site's canonical RUM domain when set.
+  // For each candidate, also try the www. variant as a fallback — domain keys are
+  // frequently registered under www.{domain} even when the site's base URL is bare.
+  const withWwwFallback = (d) => (d && !d.startsWith('www.') ? `www.${d}` : null);
+
+  const domains = [...new Set([
+    overrideHostname,
+    withWwwFallback(overrideHostname),
+    baseHostname,
+    withWwwFallback(baseHostname),
+  ].filter(Boolean))];
 
   let hasDomainKey = false;
   let timeoutId;
--- a/test/audits/rum-config-refresh.test.js
+++ b/test/audits/rum-config-refresh.test.js
@@ -317,6 +317,63 @@ describe('rum-config-refresh handler', () => {
       expect(mockConfig.updateRumConfig).to.have.been.calledWith(false);
       expect(context.log.warn).to.have.been.calledWithMatch('No domain key found');
     });
+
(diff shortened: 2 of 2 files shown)


MysticatBot (approved):
Hey @sandsinh,

## Strengths

- `src/rum-config-refresh/handler.js:73`: The `withWwwFallback` helper is well-named and its guard condition (`d && !d.startsWith('www.')`) correctly prevents the `www.www.` duplication case in a single, readable expression.
- `src/rum-config-refresh/handler.js:75-80`: Array ordering (bare, www-fallback, bare, www-fallback) preserves existing behavior for sites that already resolve on the bare domain. The `Set` dedup handles the edge case where override equals base without extra conditional logic.
- `test/audits/rum-config-refresh.test.js:320-379`: Tests cover the four key scenarios (fallback success, no double-www, both fail, override fallback) and verify call ordering with `sinon.assert.callOrder` - the right level of assertion for a feature whose correctness depends on try-order.
- Minimal blast radius: the change is purely additive. Existing sites that resolve on the bare domain hit the same code path as before; new lookups are appended to the candidate list.

## Assessment
Ready to merge? Yes
Reasoning: The change is focused, correct, and well-tested. It addresses a concrete data problem (232/306 sites with domain keys registered under www.) with the simplest possible approach - expanding the candidate list and letting the existing iteration logic do the work. No new abstractions, no new failure modes, and the existing shared timeout bounds total latency without adjustment.

## Recommendations

- Consider whether domain-variant expansion belongs in the audit handler or in the domain-key retrieval layer itself. If other consumers of `retrieveDomainkey` face the same 76% mismatch, centralizing the fallback would prevent multiple handlers from re-implementing the same expansion. Out of scope for this PR but worth tracking as a platform concern.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 31s | <b>Cost:</b> $2.27 | <b>Commit:</b> <code>da0471434d0865df3a3958babf6317085727af7d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2599, opened 2026-06-04
fix(common): honor auditContext.onDemand to bypass handler enabled-list gate

tathagat2241:
## Summary
The Slack `run audit` command was failing for any site that was not present in
the handler's enabled-list (e.g. `toc`, `meta-tags`, …). The audit worker's
centralized gate, `isAuditDisabledForSite`, always consulted
`configuration.isHandlerEnabledForSite`, so a one-off run from Slack was
treated identically to a scheduled run and got short-circuited with
`Audit <type> is disabled for site <id>, skipping`.
This PR introduces an explicit `onDemand` bypass that the api-service
`run-audit` Slack command sets in the SQS message. When present, the audit
worker skips the handler enabled-list check but still strictly enforces
entitlement (productCodes + `siteEnrollment`). Scheduled / API triggers do
not set `onDemand` and continue to honor the enabled-list gate unchanged.
> Companion change in api-service:
> adobe/spacecat-api-service#<api-service PR number>
## What changed
- `src/common/audit-utils.js`
  - `isAuditEnabledForSite(type, site, context, auditContext = {})` now
    accepts an `auditContext`. When `auditContext.onDemand === true` (or the
    string `"true"`), the handler enabled-list check is skipped. Entitlement
    is verified _before_ the bypass, so unentitled sites are still blocked.
  - `isAuditDisabledForSite` forwards `auditContext` to keep the strict
    boolean-inverse invariant.
  - JSDoc updated to document the bypass contract.
- `src/common/runner-audit.js`, `src/common/step-audit.js`,
  `src/common/async-job-runner.js`
  - Forward the incoming `auditContext` (built from the SQS message) into
    `isAuditDisabledForSite` so the flag propagates from the queue to the
    decision point.
- Tests: new cases in `test/common/{audit-utils,runner-audit,step-audit,async-job-runner}.test.js`
  covering:
  - bypass with `onDemand: true` (boolean) and `onDemand: "true"` (string),
  - entitlement still enforced under `onDemand` (negative path),
  - falsy / unrelated `onDemand` values do NOT bypass,
  - runners forward `auditContext` to the gate.
## Gating matrix after this PR
| Caller                          | Sets `onDemand` | Enabled-list check | Entitlement check |
|---------------------------------|:---------------:|:------------------:|:-----------------:|
| Slack `run audit`               | ✅              | bypassed           | enforced          |
| Scheduled trigger / API trigger | ❌              | enforced           | enforced          |
| `add-repo` / `add-site-modal`   | ❌              | enforced upstream  | enforced          |
## Test plan
- [x] `npm test` — 8417 passing, 0 failing.
- [x] 100% coverage on `src/common` (overall 100% statements, 100% branches,
      99.89% functions).

## Risk
- Low. Behavior change is gated behind an explicit, opt-in flag that only the
  `run-audit` Slack handler emits. All existing call sites (scheduled, API,
  `add-repo`, `add-site-modal`) keep their current behavior.
- The bypass cannot grant access to a site that lacks entitlement — that
  check runs first and is unchanged.
## Notes for reviewers
- The `preserveOnDemand` helper that already existed in `audit-utils.js` was
  scaffolding from an earlier change; the actual gate did not honor the flag
  yet. This PR closes that gap.
- The strict-inverse contract between `isAuditEnabledForSite` and
  `isAuditDisabledForSite` is now documented and reinforced by tests.

## Related Issues

Thanks for contributing!


src/common/audit-utils.js
```diff
@@ -67,6 +87,14 @@ export async function isAuditEnabledForSite(type, site, context) {
     return false;
   }
 
+  // On-demand bypass: skip the handler enabled-list check (entitlement already enforced above).
+  // Driven exclusively by an explicit `auditContext.onDemand` flag set by the Slack
+  // `run audit` command. Scheduled / API triggers do not set this flag.
+  if (auditContext?.onDemand === true || auditContext?.onDemand === 'true') {
```
rpapani: we shouldn't be using any additional flags, this will become another gate for us. Leave this method as is, introduce isAuditDisabledForSite, as everything is enabled by default, unless its disabled, so we'll check explicit disabled and guard that and remaining will execute. When the job comes from event-bridge from schedule, we'll honor the same logic and skip only if explicitly disabled, otherwise we'll just execute

src/common/audit-utils.js
```diff
@@ -89,10 +117,11 @@ export async function isAuditEnabledForSite(type, site, context) {
  * @param {string} type - Audit type
  * @param {Object} site - Site object
  * @param {Object} context - Context with dataAccess, log
+ * @param {Object} [auditContext] - Optional audit context; supports `onDemand` bypass flag
  * @returns {Promise<boolean>} True if audit should be skipped (disabled or not entitled)
  */
-export async function isAuditDisabledForSite(type, site, context) {
-  const enabled = await isAuditEnabledForSite(type, site, context);
+export async function isAuditDisabledForSite(type, site, context, auditContext = {}) {
+  const enabled = await isAuditEnabledForSite(type, site, context, auditContext);
```
rpapani: do an explicit disabled list check instead of negating the enabled list

rpapani (changes requested):

tkotthakota-adobe:
@tathagat2241 @rpapani 
So if audit is in disabled list then run audit from slack command won't run. Looks to me, this will be an issue as we won't be able to run the audit as there is no audit enable command any more (and audit in disabled list). 
 Should not we keep the disabled/enabled lists mainly for scheduled runs. Is there a way we can distinguish audit run from slack command vs scheduled run so that those list only applicable for schedule runs?

src/common/audit-utils.js
tathagat2241: Fixed: removed onDCemand flag.

src/common/audit-utils.js
tathagat2241: Fixed.

src/common/async-job-runner.js
```diff
@@ -90,13 +90,17 @@ export class AsyncJobRunner extends StepAudit {
 
       const site = await this.siteProvider(siteId, context);
 
-      if (await isAuditDisabledForSite(type, site, context)) {
-        log.info(`Audit ${type} is disabled for site ${site.getId()}, skipping`);
+      const { Configuration } = context.dataAccess;
+      const configuration = await Configuration.findLatest();
+      const handler = configuration.getHandlers()?.[type];
+      if (!isNonEmptyArray(handler?.productCodes)
+        || !(await checkProductCodeEntitlements(handler.productCodes, site, context))) {
```
rpapani: @tathagat2241 do we really need checkProductCodeEntitlements as part of this PR? Earlier we have the entitlement check, wouldn't that be sufficient, why to add one more check?

src/common/runner-audit.js
```diff
@@ -48,8 +49,12 @@ export class RunnerAudit extends BaseAudit {
     try {
       const site = await this.siteProvider(siteId, context);
 
-      if (await isAuditDisabledForSite(type, site, context)) {
-        log.info(`Audit ${type} is disabled for site ${site.getId()}, skipping`);
+      const { Configuration } = context.dataAccess;
+      const configuration = await Configuration.findLatest();
+      const handler = configuration.getHandlers()?.[type];
+      if (!isNonEmptyArray(handler?.productCodes)
+        || !(await checkProductCodeEntitlements(handler.productCodes, site, context))) {
```
rpapani: @tathagat2241 same here.. ideally we should check if the audit is disabled explicitly using your spacecat-shared newly introduced method? If you want to add entitlement check, please add to isHandlerDisabledForSite method similar to https://github.com/adobe/spacecat-audit-worker/blob/main/src/common/audit-utils.js#L50 rather than directly checking for product codes from here.. move this logic to isAuditDisabledForSite and also use your spacecat-shared method there

src/common/step-audit.js
```diff
@@ -116,8 +116,12 @@ export class StepAudit extends BaseAudit {
...
         site.requiresValidation = context.site.requiresValidation;
       }
-      if (await isAuditDisabledForSite(type, site, context)) {
-        log.info(`Audit ${type} is disabled for site ${site.getId()}, skipping`);
+      const { Configuration } = context.dataAccess;
+      const configuration = await Configuration.findLatest();
+      const handler = configuration.getHandlers()?.[type];
+      if (!isNonEmptyArray(handler?.productCodes)
+        || !(await checkProductCodeEntitlements(handler.productCodes, site, context))) {
+        log.info(`Audit ${type} skipped for site ${site.getId()}: missing product codes or site enrollment`);
```
rpapani: same here

rpapani (changes requested):

src/common/runner-audit.js
tathagat2241: Fixed.

src/common/async-job-runner.js
tathagat2241: Fixed.

src/common/step-audit.js
tathagat2241: fixed.

tkotthakota-adobe:
@tathagat2241 
Where are we checking onDemand to bypass handler enabled list?
How are we distinguishing between schedule run vs on demand slack command?

tkotthakota-adobe (commented):
Added few comments

tathagat2241:
@tkotthakota-adobe We have removed the onDemand flag. When it is a schedule run then job dispatcher sends the sqs message and only enabled sites are sent in the sqs message from job dispatcher so while running the audits we don't have to differentiate between if it is a scheduled run or a one off run. Job dispatcher also takes care of other checks like entitlement/ enrolement checks as well.

src/common/runner-audit.js
```diff
@@ -41,18 +40,12 @@ export class RunnerAudit extends BaseAudit {
...
 
   async run(message, context) {
-    const { log } = context;
     const { type, siteId } = message;
     const auditContext = buildRunnerAuditContext(message);
 
     try {
       const site = await this.siteProvider(siteId, context);
 
-      if (await isAuditDisabledForSite(type, site, context)) {
```
rpapani, line 51: if we remove this check altogether, the disabled audits also run right? we don't want that. We need to have this check with logic from your spacecat-shared util method

src/common/step-audit.js
```diff
@@ -125,10 +124,6 @@ export class StepAudit extends BaseAudit {
       if (context.site?.requiresValidation !== undefined) {
         site.requiresValidation = context.site.requiresValidation;
       }
-      if (await isAuditDisabledForSite(type, site, context)) {
```
rpapani, line 128: same here also

src/common/runner-audit.js
rpapani: response from @tathagat2241 
>We don't need to add that check in the audit worker.
**Scheduled audits** → gated by jobs-dispatcher via isHandlerEnabledForSite before the SQS message is sent
**Slack run audit** → gated by api-service/run-audit.js via entitlement check + isHandlerDisabledForSite before the SQS message is sent

rpapani (approved):

PR #2601, opened 2026-06-04
feat(prerender): skip HTML comparison for error/maintenance pages

anuj-adobe:
## Summary

- Reads `isErrorPage` flag from `scrape.json` metadata in `compareHtmlContent()`
- When `isErrorPage: true` (set by content-scraper when it detects a soft-403 / maintenance page), returns early with `needsPrerender: false` and `isErrorPage: true` — skipping `analyzeHtmlForPrerender()` entirely
- No RCV suggestion is created for pages that returned maintenance / WAF block content
- Logs a `warn`-level message to aid debugging

## Context

Companion PR to [spacecat-content-scraper#TBD] which adds:
- `src/support/error-page-detection.js` — generic error/maintenance page detector (title patterns + body patterns on small pages)
- Writes `isErrorPage: true/false` into `scrape.json` in `#store()`
- Skips `doMarkdownDiffAnalysis` in `processUrl()` for flagged pages (HTML files still uploaded for debugging)

This PR handles the audit-worker side: treat flagged pages as non-actionable so no irrelevant suggestion reaches customers.

## Test plan
- [ ] New test: `should skip HTML comparison and not create suggestion when isErrorPage is true in scrape.json`
  - Verifies `analyzeHtmlForPrerender` is NOT called
  - Verifies result has `isErrorPage: true` and `needsPrerender: false`
  - Verifies warn log is emitted
- [ ] All 249 prerender handler tests pass


MysticatBot (changes requested):
Hey @anuj-adobe,

## Strengths

- **src/prerender/handler.js:495-508** - The early-return pattern is the right choice. It keeps the error-page concern at the boundary of `compareHtmlContent` rather than threading conditional logic through `analyzeHtmlForPrerender` and downstream suggestion creation. Clean gate, easy to reason about.
- The return object preserves the full shape expected by downstream consumers (`hasScrapeMetadata`, `scrapeForbidden`, `isDeployedAtEdge`, etc.), so callers do not need conditional handling for the new code path. Stable contract.
- **src/prerender/handler.js:494** - The `!!metadata?.isErrorPage` pattern safely handles null metadata (optional chaining returns undefined, double-bang coerces to false). No null-safety risk introduced.
- **test/audits/prerender/handler.test.js:6762-6809** - The test exercises the full `processContentAndGenerateOpportunities` flow rather than unit-testing `compareHtmlContent` in isolation, verifying that the early return actually prevents suggestion creation end-to-end.
- The trust model is sound: `isErrorPage` is read from internal S3 metadata written by a companion service the team controls, not from user/external input. No new attack surface.
- Well-scoped change - two files, one behavior, one test. No unrelated cleanup mixed in.

## Issues

### Important (Should Fix)

1. **src/prerender/handler.js:498** - `error: true` is semantically misleading for this condition.

   In the existing catch block (~line 535), `error: true` means "analysis failed due to an exception." But an error page is not an analysis failure - it is an expected condition that was successfully detected and intentionally skipped. Downstream consumers that branch on `error: true` to identify failures will now also capture pages that were correctly classified as maintenance pages, conflating "something broke" with "nothing to do here."

   The `isErrorPage: true` flag already signals the condition distinctly. Setting `error: true` alongside it degrades signal-to-noise in any aggregation, dashboard, or alerting that uses the error field to surface problems.

   **How to fix:** Set `error: false` (or omit the field entirely) since the analysis did not fail. If downstream consumers need to distinguish "skipped intentionally" from "analyzed successfully," `isErrorPage: true` already provides that distinction.

### Minor (Nice to Have)

2. **src/prerender/handler.js:496** - `log.warn` for an expected condition. The PR description notes this is intentional for rollout visibility. For long-term operation, if error pages appear regularly across customer sites with maintenance pages or WAF blocks, warn-level logging will generate noise. The existing `scrapeForbidden` (403 detection) does not log at warn level. Consider downgrading to `info` once the feature is proven stable.

3. **test/audits/prerender/handler.test.js:6803-6804** - The test does not assert on `result.auditResult.results[0].error`. The production code sets this field in the return object, and its value affects downstream behavior. Adding `expect(result.auditResult.results[0].error).to.be.true` (or `.to.be.false` if the Important issue above is addressed) would catch regressions on this field.

## Recommendations

- The pattern of "scraper writes a classification flag, worker reads it to decide behavior" is solid and worth standardizing. As more classification signals emerge (bot-served content, geo-blocked responses, CDN edge pages), consider whether `compareHtmlContent` should have a single early-exit block that checks all skip-worthy metadata conditions rather than accumulating separate if-blocks.

## Assessment

Ready to merge? With fixes
Reasoning: The architecture is clean and the test coverage is solid. The one material concern is the `error: true` semantic collision - it conflates an intentional skip with an analysis failure, which can degrade downstream observability. Worth a quick fix before merge.

## Next Steps

1. Address the `error: true` semantics (decide whether this should be `false` given it is not a failure).
2. Minor items are optional but worth considering for long-term operational health.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 7s | <b>Cost:</b> $2.48 | <b>Commit:</b> <code>21cf82871890ce9012f1176b89450e1f605632f2</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anuj-adobe:
> **Issue 1**: Set \`error: false\` (or omit the field entirely) since the analysis did not fail.

Keeping `error: true` intentionally — the bot review raised a valid semantic point but missed a downstream consequence.

`scrapedUrlsSet` is built from `successfulComparisons` (i.e. `!result.error`). That set is passed to `syncSuggestions`, which uses it to decide which existing prerender suggestions to mark as **resolved** — implying content visibility is fine. If `error: false`, an error page URL enters that set and any existing suggestion for it gets closed with an implied 100% content visibility score. But we never saw real content; we saw a maintenance page.

`error: true` correctly quarantines the URL until a real scrape succeeds. `isErrorPage: true` is the distinguishing signal for anyone who needs to tell "intentional skip" apart from "actual failure." Added an inline comment to document this invariant.

MysticatBot (commented):
Hey @anuj-adobe,

## Strengths

- **src/prerender/handler.js:498** - The comment explaining why `error: true` is intentional directly addresses the prior review concern and documents the non-obvious coupling to `scrapedUrlsSet` and `syncSuggestions`. This is exactly the kind of comment that prevents someone from "fixing" it in 18 months and breaking suggestion resolution.
- **Previously flagged issues now addressed:** `log.warn` downgraded to `log.info` (correct level for an expected condition), and the test now asserts on the `error` field value.
- **src/prerender/handler.js:493-515** - The early-return gate preserves the full return shape expected by downstream consumers. No contract breakage, no conditional handling required by callers.
- **src/prerender/handler.js:1478** - Sparse spread `...(result.isErrorPage && { isErrorPage: true })` follows the identical pattern used for `scrapeError` on the next line. Consistent, readable.
- **test/audits/prerender/handler.test.js:6762-6809** - End-to-end test through `processContentAndGenerateOpportunities` verifies the early return actually prevents suggestion creation, not just that a function was skipped.
- **test/audits/prerender/handler.test.js:7059-7098** - S3 status persistence test covers both positive (flag present) and negative (flag absent on normal pages) paths.
- **Trust model** - `isErrorPage` is read from S3 metadata written by an internal companion service the team controls. No new attack surface, no user-controlled input.

**On the prior `error: true` semantics concern:** The author's counter-argument holds. The commit history shows they tried `error: false`, observed the real consequence (URL enters `scrapedUrlsSet`, incorrectly resolves existing suggestions), and reverted with clear rationale. The `error` field here is a control flag for downstream set membership. Prior concern resolved by author response.

## Issues

### Minor (Nice to Have)

1. **src/prerender/handler.js:498** - The `error` field now carries two distinct meanings in the same return-object contract: "analysis threw an exception" (from the catch block) and "page was intentionally skipped" (this new path). Consumers of the result object cannot distinguish the two cases without also checking `isErrorPage`. This is not blocking - both cases correctly exclude the URL from `scrapedUrlsSet` - but it is a fragility point for any future consumer that needs to differentiate "real failure" from "intentional skip." The PR already handles this correctly by adding `isErrorPage` as the discriminator; just noting the coupling for future design awareness.

2. **test/audits/prerender/handler.test.js** - No test covers the overlap scenario where a page has both `isErrorPage: true` AND `metadata.error.statusCode === 403` (scrapeForbidden). This is realistic since WAF blocks can return 403 and the content-scraper would also flag the page. The code handles it correctly (early return populates both flags), but an explicit test would confirm both flags coexist in the result. Low risk since the logic is straightforward.

## Recommendations

- The pattern of "upstream classifier writes a metadata flag, downstream worker gates behavior on it" is solid and worth standardizing. As more classification signals accumulate (bot-served content, geo-blocked responses, CDN edge pages), consider a single early-exit block in `compareHtmlContent` that checks all metadata-driven skip conditions in one place, rather than accumulating separate if-blocks that each duplicate the full return-object shape.

## Assessment

Reasoning: Well-scoped, well-tested feature gate. Two files, one behavior, one clear purpose. Prior review concerns are addressed or convincingly rebutted. Tests verify behavior at the right level of abstraction. No security concerns. Only minor observations for future awareness.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 49s | <b>Cost:</b> $2.41 | <b>Commit:</b> <code>231af2a54ada40a8f0d371ff97236f8287effd49</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @anuj-adobe,

## Strengths

- **src/prerender/handler.js (scrapeContext block)** - The DRY refactor is a clear win. Extracting `scrapeContext` into a single source of truth eliminates five-field duplication across three return paths. Adding a future skip condition (geoBlocked, botServed) is now one line in the early-exit block with no extra return-shape wiring.
- **src/prerender/handler.js (comment above isErrorPage check)** - The inline comment documenting why `error: true` is intentional (keeps URL out of `scrapedUrlsSet` so `syncSuggestions` won't resolve its suggestions) addresses the prior review concern directly in code. Future readers will find it without digging through PR history.
- **Previously flagged issues now addressed:** `log.warn` downgraded to `log.info` (prior review round 1), test now asserts on the `error` field value (prior review round 1), and author's counter-argument on `error: true` semantics is documented inline (prior review round 2).
- **test/audits/prerender/handler.test.js:6762-6809** - End-to-end test through `processContentAndGenerateOpportunities` verifies the early return actually prevents suggestion creation, not just that a function was skipped. The S3 status test covers both positive and negative paths for `isErrorPage` presence.
- **Trust model** - `isErrorPage` is read from S3 metadata written by an internal companion service. No new attack surface, no user-controlled input.
- Well-scoped change: two files, one behavior, one purpose. No unrelated cleanup mixed in.

## Issues

### Minor (Nice to Have)

1. **test/audits/prerender/handler.test.js:6803** - The "should skip HTML comparison" test asserts `isErrorPage`, `needsPrerender`, and `error`, but does not assert that `result.auditResult.results[0].url` equals the expected URL. If `url` were accidentally dropped from `scrapeContext`, this test would still pass. Adding `expect(result.auditResult.results[0].url).to.equal('https://example.com/page1')` would close the gap.

2. **No test for the overlap scenario** where metadata has both `isErrorPage: true` AND `error.statusCode === 403` (scrapeForbidden). The code handles it correctly (early return populates both `scrapeForbidden: true` and `isErrorPage: true` in the same object via `scrapeContext`), and the risk is low since the logic is straightforward. A single test case confirming the shape when both are present would document the intended contract for future readers.

## Assessment

Reasoning: Well-scoped, well-tested feature gate with a clean DRY refactor. Prior review concerns are resolved. No security, correctness, or architecture issues. Two minor test hardening suggestions are optional.

Note: CI build is still pending - confirm it passes before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 10s | <b>Cost:</b> $2.56 | <b>Commit:</b> <code>fe1c21598b88aebec89b9adc0901bc3251042541</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @anuj-adobe,

## Strengths

- **src/prerender/handler.js:492-501** - The `scrapeContext` extraction is a genuine DRY improvement. Three return paths previously hand-assembled the same 5-6 fields; now there is one source of truth. This eliminates drift between the success, error, and early-exit paths and makes the function's output contract explicit in one place.
- **src/prerender/handler.js:505-507** - The inline comment documenting why `error: true` is intentional answers the "why" without restating the "what", directly preventing a future maintainer from "fixing" it to `false`. Prior review concern resolved by author response.
- **Architecture fit** - The pattern "upstream scraper classifies, downstream worker gates on classification" is sound service decomposition. The scraper owns detection heuristics, the worker owns business logic gating. The `isErrorPage` flag in S3 metadata is a stable, versionless contract between the two services.
- **src/prerender/handler.js:1457** - The sparse spread `...(result.isErrorPage && { isErrorPage: true })` follows the identical pattern used for `scrapeError` on the next line. Consistent, additive, backward-compatible.
- **test/audits/prerender/handler.test.js:6762-6809** - End-to-end test through `processContentAndGenerateOpportunities` verifies the early return actually suppresses suggestion creation, not just that a function was skipped.
- **Trust model** - `isErrorPage` is read from internal S3 metadata written by a companion service under team control. No user-supplied or external input influences the skip decision.
- **Previously flagged issues addressed:** `log.warn` downgraded to `log.info`, test now asserts on `error` and `url` fields, overlap scenario test added, S3 status persistence test covers both positive and negative paths.

## Issues

### Minor (Nice to Have)

1. **src/prerender/handler.js:525** - Spread order precedence change in the success path. The old code placed metadata-derived fields after the analysis spread, so metadata values won on collisions. The new `{ ...scrapeContext, ...analysis }` reverses this: `analysis` now wins on key collisions. If the existing test suite passes this is not currently a problem, but it is a subtle contract change. Consider a brief comment (e.g., `// analysis fields intentionally override scrapeContext`) to document the intent for future readers.

2. **test/audits/prerender/handler.test.js** - The overlap test ("should set both isErrorPage and scrapeForbidden") asserts on `isErrorPage`, `needsPrerender`, `error`, and `scrapeError` but does not assert `scrapeForbidden: true`. Adding `expect(page.scrapeForbidden).to.be.true` would close the gap between what the test name promises and what it verifies.

## Recommendations

- The early-exit block comment ("Adding a new signal is one line here") is accurate today. If 2-3 more metadata-driven skip conditions accumulate, consider extracting the skip-condition evaluation into a named function (e.g., `shouldSkipAnalysis(metadata)`) so `compareHtmlContent` stays focused on the happy path and the skip logic becomes independently testable. Not worth doing now with one condition.

## Assessment

Reasoning: Well-scoped, well-tested feature gate with a clean DRY refactor. Prior review concerns are resolved or convincingly rebutted. No security, correctness, or architecture issues. All five review perspectives agree the change is merge-ready. Two minor hardening observations are optional.

Note: CI build is still pending - confirm it passes before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 52s | <b>Cost:</b> $2.35 | <b>Commit:</b> <code>62d3b16ab801d42637d3ec56b2a39df47b1ae96e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @anuj-adobe,

## Strengths

- **src/prerender/handler.js:493-502** - The `scrapeContext` extraction is a genuine DRY improvement. Three return paths previously hand-assembled the same 5-6 fields; now there is one source of truth. Eliminates drift and makes adding future skip conditions (geoBlocked, botServed) a one-liner with no return-shape wiring.
- **src/prerender/handler.js:507** - The inline comment documenting why `error: true` is intentional (`keeps URL out of scrapedUrlsSet so syncSuggestions won't resolve its suggestions`) captures the non-obvious business logic that was previously only explained in PR discussion. Future maintainers will find it without digging through threads.
- **src/prerender/handler.js:511** - The `// analysis fields intentionally override scrapeContext on key collision` comment directly addresses the prior review concern about spread precedence. Concise and correct.
- **test/audits/prerender/handler.test.js:6803** - The primary test correctly verifies `analyzeHtmlStub.called` is false, confirming the early-return path actually skips HTML comparison rather than just producing the right output shape.
- **test/audits/prerender/handler.test.js:6851-6856** - The overlap test asserting on `scrapeForbiddenCount` with an explanatory comment (scrapeForbidden is stripped from cleanResults but contributes to the aggregate count) documents the contract clearly.
- **Previously flagged issues now addressed:** spread-order comment added, overlap test assertion gap closed.
- **Trust model** - `isErrorPage` is read from internal S3 metadata written by a companion service under team control. No new attack surface, no user-controlled input.
- Well-scoped: two files, one behavior, one purpose.

## Issues

### Minor (Nice to Have)

1. **test/audits/prerender/handler.test.js** - No test verifies that `isErrorPage: false` (or `isErrorPage: 0`) in scrape.json does NOT trigger the early return. The production guard `if (metadata?.isErrorPage)` is a truthy check, and its behavior on falsy-but-defined values is straightforward JavaScript, but a single negative-case test would guard against someone later changing the condition to a presence check (e.g., `'isErrorPage' in metadata`). Cheap insurance, not blocking.

2. **test/audits/prerender/handler.test.js:7108** - The third new test (`should include isErrorPage in status...`) appears at a different indentation level than the two tests above it (within a different `describe` block for `uploadStatusSummaryToS3`). This is correct structurally, but worth a quick glance to confirm it lands in the intended suite scope and passes in the full test run.

## Assessment

Reasoning: Well-scoped, well-tested feature gate with a clean DRY refactor. All prior review concerns are resolved. No security, correctness, or architecture issues. Tests verify behavior at the right level of abstraction with full coverage. Two minor observations are optional and non-blocking.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 52s | <b>Cost:</b> $2.40 | <b>Commit:</b> <code>07ed2baaa50aafa1f20b97325f6bcf4d75d5e7c4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anuj-adobe:
## Review feedback addressed

**Spread order comment** (`src/prerender/handler.js`)
Added `// analysis fields intentionally override scrapeContext on key collision` above the `{ ...scrapeContext, ...analysis }` return to document the intent for future readers.

**Overlap test — `scrapeForbidden` assertion gap** (`test/audits/prerender/handler.test.js`)
`scrapeForbidden` is stripped from `cleanResults` at line 1724 (`map(({ hasScrapeMetadata, scrapeForbidden, ...result }) => result)`), so asserting `page.scrapeForbidden` would always be undefined. Instead added `expect(result.auditResult.scrapeForbiddenCount).to.equal(1)` — the observable effect that proves the flag was correctly processed into the aggregate.

**Negative-case test for `isErrorPage: false`** (`test/audits/prerender/handler.test.js`)
Added `'should NOT skip HTML comparison when isErrorPage is false in scrape.json'` which stubs `analyzeHtmlForPrerender` and asserts it IS called when the metadata has `isErrorPage: false`. Guards against any future change from the truthy guard `if (metadata?.isErrorPage)` to a presence check (`'isErrorPage' in metadata`).

**Third test scope** (`uploadStatusSummaryToS3` suite)
Confirmed `'should include isErrorPage in status...'` at line 7108 is correctly nested inside `describe('uploadStatusSummaryToS3')` — no structural issue.

MysticatBot (approved):
Hey @anuj-adobe,

## Strengths

- **src/prerender/handler.js:492-501** - The `scrapeContext` extraction is a genuine DRY improvement. Three return paths previously duplicated 5-6 fields; now there is one source of truth. Adding a future skip condition (geoBlocked, botServed) is a one-liner with no return-shape wiring.
- **src/prerender/handler.js:505-507** - The comment documenting why `error: true` is intentional captures a non-obvious business coupling (`scrapedUrlsSet` membership, `syncSuggestions` resolution). The author's counter-argument is convincing and now documented in code where it matters.
- **Architecture pattern** - "upstream classifier writes metadata, downstream worker gates on it" is the right service decomposition. The scraper owns detection heuristics, the worker owns business-logic gating. The S3 metadata flag is a stable, versionless contract between independently deployed services.
- **Trust model** - `isErrorPage` is read from internal S3 metadata written by a companion service under team control. No user-controlled input influences the skip decision. No new attack surface.
- **test/audits/prerender/handler.test.js:6808-6831** (new commit) - The negative-case test for `isErrorPage: false` guards against the truthy check being changed to a presence check. Cheap insurance for a subtle semantic difference.
- **Test coverage** - Four new tests cover positive path, negative path, overlap scenario (isErrorPage + scrapeForbidden coexisting), and S3 status persistence. The overlap test correctly asserts on `scrapeForbiddenCount` rather than the stripped `scrapeForbidden` field, showing understanding of the `cleanResults` contract.
- **Previously flagged issues all addressed** across five review rounds: `log.warn` downgraded to `log.info`, assertion on `error` field added, spread-order comment added, overlap test with `scrapeForbiddenCount`, and the negative-case test in this commit.
- **Change discipline** - Two files, one behavior, one purpose. No unrelated cleanup mixed in.

## Issues

No issues identified.

## Assessment

Ready to merge? Yes
Reasoning: Well-scoped feature gate with a clean DRY refactor, thorough test coverage, sound trust model, and all prior review concerns resolved across five iterative rounds. The single new commit since last review adds exactly the negative-case test that was suggested. No security, correctness, architecture, or convention concerns remain.

Note: CI build is still pending - confirm it passes before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 33s | <b>Cost:</b> $2.34 | <b>Commit:</b> <code>62b235f7829dd985fad7f229169fd8d6bccca900</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ssilare-adobe (approved):

PR #2602, opened 2026-06-04
fix(llm-error-pages): default new bucket-specific Opps to IGNORED

ymadobe:
## Summary
- The UI filters opportunities by `type=llm-error-pages` but PR #2461 introduced bucket-specific types (`llm-error-pages-{404,403,5xx}`) that the UI does not yet recognise. New Opps showed up in the UI with an empty Suggestions tab (suggestions are correctly written to DB but the UI's stricter type filter hides them).
- Temporary mitigation: after `convertToOpportunity` returns, if the Opp status is `NEW` (freshly created), set it to `IGNORED` so the UI does not surface it. Suggestions are still written, Mystique still runs, and the existing Excel/SharePoint flow is unaffected.
- Once the UI recognises the bucket types, drop this block and bulk-flip any accumulated IGNORED rows back to NEW.

## Follow-up tracked in Jira
This is a temporary workaround. Removal is tracked in **[LLMO-5328](https://jira.corp.adobe.com/browse/LLMO-5328)** (under epic [LLMO-3610](https://jira.corp.adobe.com/browse/LLMO-3610)):
> Once the LLMO UI ships a release that filters opportunities by `type IN ('llm-error-pages-404', 'llm-error-pages-403', 'llm-error-pages-5xx')` (or `type ILIKE 'llm-error-pages%'`), revert the `status=IGNORED` block introduced here and bulk-flip existing IGNORED rows back to NEW.

## Test plan
- [x] `npm run lint` — clean
- [x] `npm run test:spec -- test/audits/llm-error-pages/` — 203 passing
- [x] New unit test: `flips newly-created Opp status from NEW to IGNORED (pending UI support)`
- [x] New unit test: `does not flip status on Opps that are already non-NEW (e.g. APPROVED, IGNORED)`
- [ ] Post-deploy: confirm next Adobe weekly run creates Opps directly as IGNORED (no manual cleanup needed)


anagarwa:
Approving with condition that once [LLMO-5328](https://jira.corp.adobe.com/browse/LLMO-5328) is in place, the change will be reverted

anagarwa (approved):

PR #2606, opened 2026-06-05
feat(cdn-logs-report): stop uploading weekly reports to SharePoint [LLMO-5337]

akshaymagapu:
## Summary

The project-elmo-ui dashboards now read agentic & referral traffic from **PostgreSQL** (project-elmo-ui#1835), so the `cdn-logs-report` audit no longer needs SharePoint. This removes the SharePoint path and the dead code left behind. Jira: LLMO-5337.

## Changes

- **Stop using SharePoint.** Removed the weekly agentic/referral `.xlsx` generation, the SharePoint upload, and the `admin.hlx.page` bulk preview/publish from the audit.
- **Linear handler.** `runCdnLogsReport` now reads top-to-bottom: refresh agentic URL classification rules (DB) → agentic daily export → referral daily export → assemble result, via extracted helpers (`resolvePatternWeekOffset`, `generateAgenticPatterns`, `runReferralExport`). Pattern generation is contained — a transient Athena/DB failure can't block the daily DB exports.
- **Dead code removed.** `utils/report-runner.js`, `utils/excel-generator.js`, `constants/sheet-configs.js`; the weekly Athena queries (`createAgenticReportQuery` / `createReferralReportQuery`) and their `.sql` files; `saveExcelReportForBatch`; and the obsolete weekly-report config fields.
- **Removed the `cdn-reports-bulk-publish` audit** (handler + test + `index.js` registration) and the now-unused `bulkPublishToAdminHlx` / `runBulkJob` from the shared `utils/report-uploader.js`.
- **Left intact:** the daily agentic & referral DB exports (the live PG paths), DB-based classification rules, and the `report-uploader.js` functions still used by ~9 other audits (`uploadToSharePoint`, `saveExcelReport`, `publishToAdminHlx`, `createLLMOSharepointClient`).

## Behavior

`auditResult` now contains only the daily DB-export outcomes (`agentic-db-export` / `referral-db-export`); the audit performs no SharePoint upload or Helix publish.

## Testing

`npm run lint` clean; full suite passing; **100% coverage** across `src/cdn-logs-report/**`.

## Follow-up (out of scope)

Removing the orphaned SharePoint folders + `query-index` entries for these reports (ops cleanup).


src/cdn-logs-report/handler.js
```diff
@@ -22,15 +21,118 @@ import {
...
+  context,
+  s3Client,
+  athenaClient,
+  s3Config,
+  agenticReportConfig,
+  auditContext,
+}) {
+  const { log } = context;
+
+  if (!agenticReportConfig
```
MysticatBot: The guard combines `!agenticReportConfig` and `!pathHasData(...)` under a single log message ("No agentic report data found"). When the config is null, the message is misleading - there is no config, not no data. Split into two guards with distinct messages so production debugging can distinguish a broken `getConfigs` result from an empty S3 prefix.

src/cdn-logs-report/handler.js
```diff
@@ -22,15 +21,118 @@ import {
...
+}
+
+/**
+ * Weekly (non-daily) step: refresh the agentic URL classification rules and sync
+ * them to the database. The weekly SharePoint .xlsx reports were retired once the
+ * dashboards moved to PostgreSQL, so this is the only remaining weekly work.
+ *
+ * @returns {Promise<{ agenticReportHasData: boolean }>} whether agentic data was found
+ */
+async function generateAgenticPatterns({
```
MysticatBot, line 49: Unlike its sibling `runReferralExport` (which catches + returns a failure marker), this function lets exceptions from `loadSql`, `athenaClient.execute`, or `generatePatternsWorkbook` propagate. If Athena is temporarily unreachable, the entire audit fails and no daily DB exports reach PostgreSQL. Consider wrapping in try/catch and returning `{ agenticReportHasData: false }` on failure, matching the resilience intent documented in `runReferralExport`.

MysticatBot (changes requested):
Hey @akshaymagapu,

## Strengths

- **Clean linear decomposition** (handler.js:35-135): The handler now reads top-to-bottom as a clear sequence - patterns, agentic export, referral export, assemble result. The extracted `resolvePatternWeekOffset`, `generateAgenticPatterns`, and `runReferralExport` each have a single responsibility with well-chosen names. This is a material readability improvement over the nested loop-with-continue structure it replaced.
- **Error isolation in `runReferralExport`** (handler.js:97-135): Catching failures and returning a typed failure marker instead of throwing is the correct resilience pattern. The error is logged with full context (site ID, message, and the original error object).
- **Correct behavior preservation**: `resolvePatternWeekOffset` returns the same values the old code used for pattern generation (`weekOffsets[0]` was always -1 on Monday, 0 otherwise). The dual-offset loop only drove the now-deleted weekly report generation.
- **Self-contained deletion scope**: The removed chain (sheet-configs, excel-generator, report-runner, cdn-reports-bulk-publish, bulkPublishToAdminHlx/runBulkJob) has no remaining callers. Paths still used by other audits (uploadToSharePoint, saveExcelReport, createLLMOSharepointClient) are left intact.
- **Attack surface reduction**: Removing the outbound `admin.hlx.page` HTTP calls with `ADMIN_HLX_API_KEY` in an auth cookie and the SharePoint OAuth client instantiation is a security improvement.
- **Test rewrite matches the new surface**: Tests verify observable behavior (final auditResult shape, fullAuditRef) rather than internal call counts, and are organized by concern (happy path, patterns generation, daily referral export).

## Issues

### Important (Should Fix)

1. **handler.js:60-64 - Conflated guard combines two unrelated conditions into one log message**

   ```javascript
   if (!agenticReportConfig
     || !(await pathHasData(s3Client, agenticReportConfig.aggregatedLocation))) {
     log.info('No agentic report data found - skipping patterns generation');
     return { agenticReportHasData: false };
   }
   ```

   When `agenticReportConfig` is null/undefined, the log says "No agentic report data found" which is misleading - there is no config, not no data. A missing config might indicate a misconfiguration worth investigating; no data in S3 is normal for new sites.

   **Why it matters:** When debugging production issues, this log line cannot distinguish between a broken `getConfigs` result and an empty S3 prefix - two operationally different situations.

   **How to fix:** Split the guard:
   ```javascript
   if (!agenticReportConfig) {
     log.info('No agentic report config found - skipping patterns generation');
     return { agenticReportHasData: false };
   }
   if (!(await pathHasData(s3Client, agenticReportConfig.aggregatedLocation))) {
     log.info('No agentic report data found - skipping patterns generation');
     return { agenticReportHasData: false };
   }
   ```

2. **handler.js:49-95 - `generateAgenticPatterns` has no error containment, unlike its sibling `runReferralExport`**

   `runReferralExport` has explicit try/catch with the JSDoc comment "returning a failure marker instead of throwing so a referral problem never blocks the rest of the audit." But `generateAgenticPatterns` lets exceptions from `loadSql`, `athenaClient.execute`, or `generatePatternsWorkbook` propagate uncaught.

   This is pre-existing behavior (the old code propagated these too), but the new code makes the asymmetry more visible by placing both helpers side-by-side with one explicitly designed for isolation and the other not.

   **Why it matters:** If Athena is temporarily unreachable, the entire audit fails and no daily agentic/referral export data reaches PostgreSQL - even though the exports do not depend on pattern generation succeeding.

   **How to fix:** Wrap the body of `generateAgenticPatterns` in a try/catch that logs and returns `{ agenticReportHasData: false }`, mirroring `runReferralExport`. Add a test that verifies the containment (e.g., `mocks.loadSql` rejects, assert the handler still returns a result with the daily exports).

## Recommendations

- Since `process.env.ADMIN_HLX_API_KEY` is no longer consumed by this worker after merge, consider removing it from the deployed environment (Lambda env vars, Vault path) as a follow-up. A stale credential is a liability if it grants write access to `admin.hlx.page`.
- The `cdn-reports-bulk-publish` SQS trigger configuration (if any scheduler or dispatcher still emits these messages) should be cleaned up as a follow-up to avoid dead-letter queue noise.

## Assessment

Ready to merge? With fixes
Reasoning: This is a well-executed platform simplification with clean deletions and correct behavior preservation. The two Important findings are both about the `generateAgenticPatterns` helper - a misleading log message and missing error containment that could let a transient Athena failure block unrelated daily DB exports. Both are straightforward fixes.

## Next Steps

1. Split the conflated guard to distinguish missing config from missing data.
2. Add error containment to `generateAgenticPatterns` (with a corresponding test) to match the design intent expressed in `runReferralExport`.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 20s | <b>Cost:</b> $6.66 | <b>Commit:</b> <code>9a105d7250149864e8c616f99e5c41eb2071cce7</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

akshaymagapu:
Thanks for the review 🙏 — both **Important** items addressed in `1b3c587e`:

1. **Split the conflated guard** — `generateAgenticPatterns` now logs distinct messages: `No agentic report config found …` (missing/broken `getConfigs`) vs `No agentic report data found …` (empty S3 prefix).
2. **Error containment** — the pattern-generation body is wrapped in try/catch; a transient Athena/DB failure is logged and returns `{ agenticReportHasData: false }` instead of propagating, so it can no longer block the daily agentic/referral DB exports. Added a test (`contains patterns failures so the daily exports still run`) asserting the daily exports still run on an `athenaClient.execute` rejection. `handler.js` stays at 100% coverage.

On the recommendations:
- **`ADMIN_HLX_API_KEY`**: it's still consumed by this worker — `publishToAdminHlx` in `src/utils/report-uploader.js` reads it, and that function is still used by `llmo-onboarding-publish`, `frescopa-data-generation`, `related-urls`, and `llm-error-pages`. So it should **not** be removed from the environment. (Only `bulkPublishToAdminHlx`/`runBulkJob`, which also used it, were deleted.)
- **`cdn-reports-bulk-publish` trigger cleanup**: good call — tracking as a follow-up to confirm no scheduler/dispatcher still emits those SQS messages (would otherwise DLQ now that the handler is gone).

MysticatBot (approved):
Hey @akshaymagapu,

## Strengths

- **Previously flagged issues resolved cleanly** (handler.js:56-64, 96-100): Both prior Important findings are fully addressed. The guard is now split into two distinct checks with accurate, distinguishable log messages ("No agentic report config found" vs "No agentic report data found"). The error containment wraps the correct scope and logs with full context (site ID, message, original error object).
- **Error containment matches the stated resilience intent** (handler.js:96-100): The try/catch returns `{ agenticReportHasData: false }` which correctly signals downstream that patterns generation failed without blocking the daily DB exports - matching the design pattern established by `runReferralExport`.
- **Correct elimination of the second `fetchAgenticUrlClassificationRules` call**: The old code re-fetched after `generatePatternsWorkbook` to pass `remotePatterns` into `runWeeklyReport`. With weekly reports gone, that re-fetch had no consumer. Patterns are already synced to PG inside `generatePatternsWorkbook`.
- **Significant complexity reduction** (handler.js): The nested loop (`for reportConfig ... for weekOffset`) with conditional SharePoint client creation and mutable accumulators is replaced by a linear pipeline - patterns, agentic export, referral export, assemble. Each step returns a typed result and is independent.
- **Clean audit removal** (cdn-reports-bulk-publish): Handler, test, and HANDLERS registration removed together. The deleted `runBulkJob`/`bulkPublishToAdminHlx` had no remaining callers.
- **Behavior change from dual-week to single-week offset is intentional**: The old `[-1, 0]` on Mondays only drove per-week Excel reports. Pattern generation only needs a single representative week to sample CDN logs.
- **Test rewrite is proportional and behavior-focused** (handler.test.js): Rewritten from ~790 lines of deeply-coupled mock-wiring to ~210 lines of concise tests organized by concern, verifying observable outputs rather than internal stub call counts.
- **New test verifies the precise contract** ("contains patterns failures so the daily exports still run"): Rejects `athenaClient.execute`, asserts the error log, confirms `agenticReportHasData: false` propagation, and validates both daily exports still produce their batch IDs.

## Issues

No issues found.

## Assessment

Ready to merge? Yes
Reasoning: Both prior Important findings are cleanly addressed with appropriate error containment and a corresponding test. The overall PR is a well-scoped platform simplification that removes a dead integration path (SharePoint + admin.hlx.page), eliminates the `cdn-reports-bulk-publish` audit, and replaces a nested-loop handler with a clean linear pipeline while maintaining 100% coverage.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 30s | <b>Cost:</b> $3.46 | <b>Commit:</b> <code>1b3c587ef61c18d87a2f526646e80287c280c54a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2607, opened 2026-06-05
fix(accessibility): resolve stale a11y opportunities when no issues found (SITES-40450)

petcud:
Jira link: https://jira.corp.adobe.com/browse/SITES-40450

When the individual a11y opportunities flow finds no actionable issues, call resolveOpportunityIfNoIssues for each tracked a11y opportunity type so any existing NEW opportunity is transitioned to RESOLVED and its actionable suggestions are marked OUTDATED. Previously the early return left stale opportunities in NEW indefinitely, causing the ASO UI to show outdated results to customers who had already fixed all their accessibility issues.

### DIFF
--- a/src/accessibility/utils/generate-individual-opportunities.js
+++ b/src/accessibility/utils/generate-individual-opportunities.js
@@ -22,6 +22,7 @@ import {
   syncSuggestions,
   keepSameDataFunction,
   warnOnInvalidSuggestionData,
+  resolveOpportunityIfNoIssues,
 } from '../../utils/data-access.js';
 import {
   successCriteriaLinks, accessibilityOpportunitiesMap, URL_SOURCE_SEPARATOR, issueTypesForCodeFix,
@@ -818,7 +819,7 @@ export function calculateAccessibilityMetrics(aggregatedData) {
  */
 export async function createAccessibilityIndividualOpportunities(accessibilityData, context) {
   const {
-    site, log,
+    site, log, dataAccess,
   } = context;
 
   log.info(`[A11yIndividual] Creating accessibility opportunities for ${site.getBaseURL()}`);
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @petcud,

## Strengths

- Well-placed fix that closes a real state-machine gap - the audit is the authoritative source of truth, and the "no issues" path now correctly transitions downstream opportunity state to match (`src/accessibility/utils/generate-individual-opportunities.js:833-837`)
- Reuses the existing `resolveOpportunityIfNoIssues` contract from the shared data-access layer rather than inventing new resolution logic - keeps opportunity lifecycle management centralized
- Change discipline is excellent: two files, one concern, no unrelated cleanup
- Test verifies both call count and exact arguments for each opportunity type, catching regressions if the map changes (`test/audits/accessibility/generate-individual-opportunities.test.js:2051-2063`)
- The `Promise.all` over `Object.keys(accessibilityOpportunitiesMap)` is a fixed-size static map (2 entries), not an unbounded data collection - the CLAUDE.md N+1 warning does not apply here

## Issues

### Minor (Nice to Have)

**`Promise.all` without error isolation on cleanup path** - `src/accessibility/utils/generate-individual-opportunities.js:833-837`

If `resolveOpportunityIfNoIssues` rejects for one opportunity type (transient DB error), the entire early-return path throws and the caller never receives the `NO_OPPORTUNITIES` result. Since the audit already determined the site is clean, a transient DB failure during stale-opportunity cleanup would turn a healthy result into a hard error, potentially triggering unnecessary retries.

Consider `Promise.allSettled` or a per-item catch that logs a warning and continues, so cleanup is best-effort and the clean audit result always returns. Alternatively, if letting it propagate is intentional (so it retries on the next audit run), a test asserting that propagation behavior would document the choice.

## Assessment

Ready to merge? Yes

This is a focused, well-tested bug fix. It correctly resolves stale opportunities when no accessibility issues are found, uses established shared infrastructure, and introduces no new architectural surface or security concerns.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 28s | <b>Cost:</b> $2.02 | <b>Commit:</b> <code>0928854cb3ce306c83ac5c4ac661956ec1894d44</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

szergen (approved):

PR #2609, opened 2026-06-05
fix(cdn-logs-report): pin daily export S3 client to correct region (PermanentRedirect)

akshaymagapu:
## Problem

The daily **agentic** and **referral** exports were failing in production with:

> The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint. **PermanentRedirect** (HTTP 301)

Stack trace pointed at `uploadBundleToS3` → `runDailyAgenticExport`.

### Root cause

Both exports write their output (CSV bundle / referral CSV) to the **importer bucket** (`S3_IMPORTER_BUCKET_NAME`), which lives in **us-east-1**. But they reused the S3 client returned by `getCdnAwsRuntime`, which is regionalized to the **site's CDN bucket region** (`resolveSiteCdnRegion`):

```js
const s3Client = region === runtimeRegion ? context.s3Client : new S3Client({ region });
```

For any site whose configured CDN region is not `us-east-1`, the importer-bucket write hit the wrong regional endpoint and S3 returned `PermanentRedirect`, failing the export (e.g. site `69f318ca-bfd0-401a-a854-ca4619a7dacb`).

## Fix

Construct a dedicated `us-east-1` S3 client inside `runDailyAgenticExport` and `runDailyReferralExport` for the importer-bucket upload/cleanup, and drop the now-unused `s3Client` threading from the export call chain (`agentic-db-export.js`, `handler.js`). The Athena query path keeps using the site-region runtime — only the importer-bucket writes are pinned to `us-east-1`.

## Validation

- `agentic-daily-export.test.js`, `agentic-db-export.test.js`, `referral-daily-export.test.js`, `handler.test.js` — all green
- New assertions verify the S3 client is constructed with `{ region: 'us-east-1' }`
- 100% line/branch/statement coverage maintained on all changed source files

### DIFF
--- a/src/cdn-logs-report/agentic-daily-export.js
+++ b/src/cdn-logs-report/agentic-daily-export.js
@@ -16,11 +16,6 @@ import { loadSql, IMPORTER_BUCKET_REGION } from './utils/report-utils.js';
 import { weeklyBreakdownQueries } from './utils/query-builder.js';
 import { mapToAgenticTrafficBundle } from './utils/agentic-traffic-mapper.js';
 
-// The agentic bundle is always written to the importer bucket (S3_IMPORTER_BUCKET_NAME),
-// which lives in us-east-1 (IMPORTER_BUCKET_REGION). The injected CDN S3 client is
-// regionalized to the site's CDN region, so reusing it triggers S3 PermanentRedirect (301)
-// for non-us-east-1 sites. Always talk to the importer bucket through a us-east-1 client.
-
 export function getPreviousUtcDate(referenceDate = new Date()) {
   const previous = new Date(referenceDate);
   previous.setUTCDate(previous.getUTCDate() - 1);
@@ -264,6 +259,7 @@ export async function runDailyAgenticExport({
   const keyPrefix = getAgenticBundleKeyPrefix(site.getId(), trafficDate, bundleId);
   const bundleUri = `s3://${bundleBucket}/${keyPrefix}`;
   const uploadedFiles = buildBundleFileKeys(keyPrefix);
+  // Importer bucket is us-east-1, not the site's CDN region; reusing the CDN client 301s.
   const s3Client = new S3Client({ region: IMPORTER_BUCKET_REGION });
   let dispatch;
 
--- a/src/cdn-logs-report/agentic-daily-export.js
+++ b/src/cdn-logs-report/agentic-daily-export.js
@@ -12,15 +12,14 @@
 
 import { DeleteObjectsCommand, PutObjectCommand, S3Client } from '@aws-sdk/client-s3';
 import { v4 as uuidv4 } from 'uuid';
-import { loadSql } from './utils/report-utils.js';
+import { loadSql, IMPORTER_BUCKET_REGION } from './utils/report-utils.js';
 import { weeklyBreakdownQueries } from './utils/query-builder.js';
 import { mapToAgenticTrafficBundle } from './utils/agentic-traffic-mapper.js';
 
 // The agentic bundle is always written to the importer bucket (S3_IMPORTER_BUCKET_NAME),
-// which lives in us-east-1. The injected CDN S3 client is regionalized to the site's CDN
-// region, so reusing it triggers S3 PermanentRedirect (301) for non-us-east-1 sites.
-// Always talk to the importer bucket through a dedicated us-east-1 client.
-const AGENTIC_EXPORT_S3_REGION = 'us-east-1';
+// which lives in us-east-1 (IMPORTER_BUCKET_REGION). The injected CDN S3 client is
+// regionalized to the site's CDN region, so reusing it triggers S3 PermanentRedirect (301)
+// for non-us-east-1 sites. Always talk to the importer bucket through a us-east-1 client.
 
 export function getPreviousUtcDate(referenceDate = new Date()) {
   const previous = new Date(referenceDate);
@@ -265,7 +264,7 @@ export async function runDailyAgenticExport({
   const keyPrefix = getAgenticBundleKeyPrefix(site.getId(), trafficDate, bundleId);
   const bundleUri = `s3://${bundleBucket}/${keyPrefix}`;
   const uploadedFiles = buildBundleFileKeys(keyPrefix);
-  const s3Client = new S3Client({ region: AGENTIC_EXPORT_S3_REGION });
+  const s3Client = new S3Client({ region: IMPORTER_BUCKET_REGION });
   let dispatch;
 
   try {
--- a/src/cdn-logs-report/agentic-daily-export.js
+++ b/src/cdn-logs-report/agentic-daily-export.js
@@ -10,11 +10,17 @@
  * governing permissions and limitations under the License.
  */
 
-import { DeleteObjectsCommand, PutObjectCommand } from '@aws-sdk/client-s3';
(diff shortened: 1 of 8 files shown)


MysticatBot (commented):
Hey @akshaymagapu,

## Strengths

- Clear inline comments explaining the "why" at `agentic-daily-export.js:16-19` and `referral-daily-export.js:18-21` - future maintainers will not re-introduce the same bug by threading the CDN client back in.
- Correct root cause fix: the importer-bucket client was incorrectly conflated with the CDN-bucket client, and this PR properly decouples them.
- Clean parameter removal across the full call chain (`handler.js` -> `agentic-db-export.js` -> `runDailyAgenticExport`/`runDailyReferralExport`) leaves no dangling references.
- Tests assert the region contract directly (`S3ClientStub.calledOnceWith({ region: 'us-east-1' })`), making the fix regression-proof.
- Cleanup-on-failure paths (SQS dispatch failure, S3 upload failure) are verified to use the same correctly-regionalized client.

## Issues

### Minor (Nice to Have)

**1. Per-call `S3Client` instantiation - no connection reuse across Lambda warm starts**
- `src/cdn-logs-report/agentic-daily-export.js:268`, `src/cdn-logs-report/referral-daily-export.js:222`
- Each export call creates a fresh `S3Client` with its own HTTP connection pool. The previous pattern shared a single client across the handler lifecycle. For the current call patterns (one daily export per Lambda invocation, weekly path queues separate SQS messages), this is acceptable and not a correctness concern. Worth a follow-up if future work calls these functions in a tight loop within a single invocation - hoisting to module scope would be the natural fix.

**2. Removed negative assertion weakens early-return test path**
- `test/audits/cdn-logs-report/agentic-daily-export.test.js` (Athena unavailable test, ~line 701)
- The old test asserted `expect(s3Client.send).to.not.have.been.called` to confirm S3 was never touched on early exit. This was removed because the `s3Client` parameter no longer exists, but the replacement assertion (`S3ClientStub` was never called) is not present. The test still verifies the function throws, but doesn't prove it exited before any S3 work.
- Suggestion: Add `expect(S3ClientStub).to.not.have.been.called` in early-return tests to restore the "no S3 work" guarantee.

**3. Two identical region constants with different names**
- `AGENTIC_EXPORT_S3_REGION` (`agentic-daily-export.js:21`) and `REFERRAL_EXPORT_S3_REGION` (`referral-daily-export.js:23`) both hold `'us-east-1'` for the same reason (importer bucket location). If a third export type arrives, it will duplicate again. A shared constant in `utils/report-utils.js` would prevent drift, but is not worth blocking this fix for.

## Recommendations

- Consider hoisting the `S3Client` to module scope in a follow-up. The region is static, the bucket is static, and module-scope instantiation matches the existing architectural pattern. Tests already mock via `esmock` at the module level, so this would be a minimal refactor.

## Assessment

Reasoning: The fix is correct, well-scoped, and addresses a real production failure (PermanentRedirect for non-us-east-1 CDN sites). All findings are minor and none represent bugs or regressions. The change safely decouples the importer-bucket S3 client from the CDN-bucket client, which is the right architectural direction.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 11s | <b>Cost:</b> $3.88 | <b>Commit:</b> <code>94244de3480c08e01666d4188583b65473b19965</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

akshaymagapu:
Thanks for the review! Addressed in `113698aa` (+ comment trim in `2600abd7`):

- **3 (dup constants):** extracted a shared `IMPORTER_BUCKET_REGION` in `utils/report-utils.js`, used by both exports.
- **2 (early-exit assertion):** restored — the agentic early-return tests (Athena setup failure, missing analytics queue) now assert the importer `S3Client` is never even constructed, which is a stronger "no S3 work" guarantee than the old `send` check.
- **1 (per-call vs module-scope):** keeping per-call intentionally. It keeps the early-exit guarantee assertable (constructor-never-called = no S3 work), and the SDK v3 client is lazy so there's no warm-start connection cost until first `send`. Module scope would also construct the client at cold start for all ~100 audit types (handler graph is imported eagerly). Happy to revisit as a follow-up if needed.

MysticatBot (approved):
Hey @akshaymagapu,

## Strengths

- All three prior review findings addressed cleanly: shared `IMPORTER_BUCKET_REGION` constant eliminates duplication (`src/cdn-logs-report/utils/report-utils.js:16`), early-exit assertions restored with the stronger `S3ClientStub`-never-called guarantee (`test/audits/cdn-logs-report/agentic-daily-export.test.js:450`), and the per-call rationale is well-articulated.
- Clean parameter removal across the full call chain (`handler.js` -> `agentic-db-export.js` -> export functions) with no dangling references or dead parameters left behind.
- Comment placement at `agentic-daily-export.js:262` and `referral-daily-export.js:217` concisely explains the "why" - future maintainers will not re-thread the CDN client back in.
- Removing the caller-supplied `s3Client` parameter actually narrows the trust boundary - the S3 client is now internally governed using the Lambda's IAM role via the default credential chain, rather than caller-injectable.
- Region contract assertions (`S3ClientStub.calledOnceWith({ region: 'us-east-1' })`) verify the fix at the right abstraction level. Cleanup paths (SQS dispatch failure, S3 upload failure) also confirm the correctly-regionalized client is used for DeleteObjectsCommand.
- 100% coverage maintained per Codecov.

## Issues

### Minor (Nice to Have)

**Shared `s3ClientMock.send` stub state in referral tests**
- `test/audits/cdn-logs-report/referral-daily-export.test.js:27`
- The shared `s3ClientMock.send` stub is created in `beforeEach` with `.resolves({})`. The "cleans up uploaded CSV when S3 upload itself fails" test (line 486) overrides with `.onCall(0).rejects(...)`, mutating the shared stub. This works today because `sandbox.restore()` in `afterEach` resets everything, but a test added between the `onCall` setup and `afterEach` would inherit the mutated behavior.
- Not a correctness issue today, but a maintenance hazard for future test additions.
- Fix: Reset the stub at the start of tests that override call behavior: `s3ClientMock.send.reset(); s3ClientMock.send.onCall(0).rejects(...)`.

## Recommendations

- The importer bucket region is currently a code constant. If infrastructure ever moves to regional importer buckets (e.g., EU data residency), pulling this from environment config (`context.env`) would be the natural evolution. Not worth blocking this fix for, and the single constant makes such a future change straightforward.

## Assessment

Ready to merge? Yes
Reasoning: This is a well-scoped fix for a real production failure (PermanentRedirect for non-us-east-1 CDN sites). The root cause is correctly identified and fixed by decoupling the importer-bucket S3 client from the CDN-bucket client. All prior findings have been addressed, the code is clean, tests verify the region contract directly, and coverage remains at 100%. The single minor finding is a test-maintenance observation, not a correctness issue.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 38s | <b>Cost:</b> $2.49 | <b>Commit:</b> <code>2600abd71192fe6ec6b73a0c1bf76cdc5672ba2d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ctinp (approved):

PR #2613, opened 2026-06-08
fix: (prerender) enabled support of csv urls via slack command in case of ai-only mode

Danialkafeel:
### Problem

`ai-only` mode (`@spacecat run audit <site> audit:prerender mode:ai-only`) skips scraping and sends existing DB suggestions to Mystique for AI summary/prompt generation. Previously it always processed **all** eligible suggestions and sliced the first `MYSTIQUE_BATCH_SIZE` (320). Repeated runs kept hitting the **same first 320 suggestions** — there was no way to target a specific subset, making prompt backfills across thousands of suggestions impractical.

### Change

Adds support for a **CSV URL filter** in `ai-only` mode, supplied via a CSV file attached to the Slack `run audit` command (surfaced as `context.auditContext.urls`). Operators can now control exactly which URLs are processed in each run.

**`handleAiOnlyMode`**
- Now builds the `preBuiltCandidates` list from DB suggestions, applying the existing eligibility rules (skips domain-wide, `OUTDATED`, `SKIPPED`, `FIXED`, and edge-deployed suggestions; derives `scrapeJobId` from `originalHtmlKey` when missing).
- When `auditContext.urls` is present, restricts candidates to suggestions whose pathname matches the CSV list (domain-shift-safe via `toPathname`).
- Logs a `warn` listing any CSV URLs that had **no eligible DB suggestion** (not yet scraped, or in `OUTDATED`/`SKIPPED`/`FIXED` status), with a hint to run a normal prerender audit first for those URLs — so previously silent drops are now surfaced.
- When no CSV is attached, behavior is unchanged: all eligible suggestions are processed.

### Why unscraped URLs are safe

`ai-only` mode requires the S3 markdown diffs written by a prior scrape. CSV URLs that were never scraped have no matching DB suggestion, so they're correctly skipped — and now logged as a warning rather than dropped silently.

## Related Issues

Thanks for contributing!


src/prerender/handler.js
```diff
@@ -824,6 +754,87 @@ export async function handleAiOnlyMode(context) {
...
+    };
+  }
+
+  const csvUrls = context.auditContext?.urls;
+  const urlFilter = Array.isArray(csvUrls) && csvUrls.length > 0
+    ? new Set(csvUrls.map((u) => toPathname(u)))
+    : null;
+
+  if (urlFilter) {
+    log.info(`${LOG_PREFIX} ai-only: CSV URL filter active — restricting to ${urlFilter.size} URL(s). baseUrl=${baseUrl}, siteId=${siteId}`);
```
MysticatBot, line 900: issue (blocking): `s.getId?.()` is inconsistent with the rest of the codebase.

Every other `getId()` call on suggestion entities in this file uses the non-optional form (`s.getId()`). The removed code at the same logical location also used `s.getId()` without optional chaining.

If `getId` is always present on suggestion objects (which the rest of the file assumes), the optional chaining is misleading noise. If it genuinely could be undefined, then `suggestionId` silently becomes `undefined` and propagates into the candidate object and downstream SQS payload to Mystique - a latent bug.

**Fix:** Change `s.getId?.()` to `s.getId()` to match the existing entity contract.

src/prerender/handler.js
```diff
@@ -824,6 +754,87 @@ export async function handleAiOnlyMode(context) {
     };
   }
 
+  // Build candidates from DB suggestions, optionally filtered to a CSV URL list.
```
MysticatBot, line 880: issue (blocking): CI is failing. The repo CLAUDE.md documents: "Coverage enforcement is strict: 100% lines, branches, and statements across all src/**/*.js. Every code path must be tested."

The new code adds branches that likely need coverage, in particular: the early-return path here (no existing suggestions) when CSV URLs are active via `auditContext.urls`. In that scenario the helpful "run a normal prerender audit first" warning never fires because execution returns before reaching the URL filter logic. A test exercising this combination would both satisfy coverage and document the expected UX.

**Fix:** Add a test that sets `context.auditContext = { urls: [...] }` with the mock returning empty suggestions, verifying the early return fires with `suggestionCount: 0`.

MysticatBot (changes requested):
Hey @Danialkafeel,

_Changes: Moves candidate-building logic from `sendPrerenderGuidanceRequestToMystique` into `handleAiOnlyMode` and adds CSV URL filtering so ai-only mode Slack commands can target specific URLs (2 files)._

## Must fix before merge
1. [Important] `s.getId?.()` uses optional chaining inconsistently with the codebase pattern - `src/prerender/handler.js:777` (details inline)
2. [Important] CI failing against documented 100% coverage rule - new branches likely need additional test coverage - `src/prerender/handler.js:757` (details inline)

<details>

- nit: `preBuiltCandidates ?? []` in `sendPrerenderGuidanceRequestToMystique` is now dead code since both callers pass arrays - `src/prerender/handler.js:617`
- suggestion: Duplicate CSV URLs inflate the unmatched-URL warning count because `unmatchedUrls` is derived from the raw `csvUrls` array rather than the deduplicated `urlFilter` Set - consider deduplicating before computing unmatched - `src/prerender/handler.js:832`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 46s | <b>Cost:</b> $3.08 | <b>Commit:</b> <code>270660f26b3b79bedafd1d8db0c1e136e97d20d9</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/prerender/handler.js
Danialkafeel: It is used like this in multiple places, keeping it as a safeguard

src/prerender/handler.js
Danialkafeel: Done

MysticatBot (approved):
Hey @Danialkafeel,

**Verdict: Approve** - clean refactoring with good test coverage.
_Changes: Moves candidate-building logic into `handleAiOnlyMode` and adds CSV URL filtering so ai-only Slack commands can target specific URLs (2 files)._

<details>

- nit: Duplicate CSV URLs inflate the unmatched-URL warning count because `unmatchedUrls` is derived from the raw `csvUrls` array rather than the deduplicated `urlFilter` Set - consider filtering against the Set directly - `src/prerender/handler.js:832`
</details>

- CI coverage gap addressed - all modified and coverable lines now covered per Codecov.
- `getId?.()` optional chaining concern dropped - author's argument that this defensive pattern is used elsewhere in the codebase holds.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 33s | <b>Cost:</b> $2.81 | <b>Commit:</b> <code>00e397a92742408474b7e726283f35bd25773931</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/prerender/handler.js
```diff
@@ -829,6 +758,87 @@ export async function handleAiOnlyMode(context) {
...
+      mode: MODE_AI_ONLY,
+      opportunityId: opportunity.getId(),
+      fullAuditRef: `${MODE_AI_ONLY}/${opportunity.getId()}`,
+      auditResult: { message: 'No existing suggestions found', suggestionCount: 0 },
+    };
+  }
+
+  const csvUrls = context.auditContext?.urls;
+  const urlFilter = Array.isArray(csvUrls) && csvUrls.length > 0
+    ? new Set(csvUrls.map((u) => toPathname(u)))
```
ssilare-adobe: Please use `normalizePathnameWithQuery` instead of `toPathname(u)`

ssilare-adobe (changes requested):

ssilare-adobe:
@Danialkafeel Can you clarify if the change includes sending over all urls to mystique for AI generation? 

The suggestions were capped to 320 because of SQS limit of 256 Kb message size, if that is violated or there are too many suggestions which makes the message > 256 kb, then the AI generation fails, nice to have a tested flow on dev.

src/prerender/handler.js
Danialkafeel: Thanks Sahil, changed the CSV filter to `normalizePathnameWithQuery` for consistency with how suggestions are keyed in syncSuggestions; toPathname would have dropped the query string and mismatched query-distinct URLs. 
Also Added a test covering ?id=1 vs ?id=2.

Danialkafeel:
> @Danialkafeel Can you clarify if the change includes sending over all urls to mystique for AI generation?
> 
> The suggestions were capped to 320 because of SQS limit of 256 Kb message size, if that is violated or there are too many suggestions which makes the message > 256 kb, then the AI generation fails, nice to have a tested flow on dev.

No, the change does not send all URLs. The existing `slice(0, MYSTIQUE_BATCH_SIZE) (320)` cap in `sendPrerenderGuidanceRequestToMystique` is unchanged, so the SQS message stays under 256 KB. The CSV filter lets us scope a run to ≤320 URLs by design. 
I've also added a log.warn when the filtered set from csv urls exceeds 320 so it's clear the overflow won't be sent in that run

Danialkafeel:
Also validated this on dev: https://cq-dev.slack.com/archives/C060T2PPF8V/p1781008806363559?thread_ts=1781006189.441019&cid=C060T2PPF8V 

Verified the behaviour when some urls are not present in db (screenshot)
https://spacecat.coralogix.com/#/query-new/logs?id=mBhseVTxkaO7IEmFeLd07&page=0

Screenshot 2026-06-09 at 6 11 10 PM

src/prerender/handler.js
```diff
@@ -829,6 +758,101 @@ export async function handleAiOnlyMode(context) {
     };
   }
 
+  // Build candidates from DB suggestions, optionally filtered to a CSV URL list.
+  const existingSuggestions = await opportunity.getSuggestions();
+
+  if (!existingSuggestions || existingSuggestions.length === 0) {
```
anuj-adobe, line 883: can you pls refactor the code (changes added) to reduce the size of function and make it more readable?

Danialkafeel: Done. Please check now

Danialkafeel:
Not needed, already covered by another PR

PR #2616, opened 2026-06-08
fix(preflight): deduplicate empty heading opportunities

clotton:
## Summary
- **Root cause:** `getElementsFromCheck` for `heading-empty` was re-querying all empty headings of a given tag type (e.g. all empty H2s on the page), then attaching that full set to *every* opportunity card for that tag. With 2 empty H2s, both cards showed "1 of 2 instances found" pointing to the same 2 headings — appearing as duplicates in the preflight UI.
- **Fix:** Use `check.transformRules.selector` (already set per-element by the core heading handler) to target only the specific heading for each card — the same pattern already used by `heading-order-invalid`.
- **Customer impact:** Micron reported seeing duplicate empty-heading cards (SITES-46108). This resolves the visual duplication.

## Test plan
- [ ] Updated `should generate selectors for heading-empty check` → now uses `transformRules.selector`, verifies single-selector output even when multiple empty H2s are in the HTML
- [ ] Added `should generate no selectors for heading-empty check when transformRules is absent` → graceful no-op
- [ ] `npm run test:spec -- test/preflight/headings-selectors.test.js` passes (9/9, 1 pre-existing timeout failure unrelated to this change)

### DIFF
--- a/src/preflight/headings.js
+++ b/src/preflight/headings.js
@@ -88,15 +88,15 @@ function getElementsFromCheck(scrapeJsonObject, check) {
     }
 
     case 'heading-empty': {
-      // Find empty headings - extract tag name from check
-      const tagName = check.tagName?.toLowerCase();
-      if (tagName && /^h[1-6]$/.test(tagName)) {
-        const headingsArray = $(tagName).toArray();
-        // Find empty ones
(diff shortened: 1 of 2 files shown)


clotton:
Premature fix — investigation showed the page currently has 0 empty H2s, so this selector bug wasn't the cause of SITES-46108. Closing to re-investigate from scratch.

PR #2618, opened 2026-06-08
fix(preflight): heading-empty duplicate cards when multiple empty headings exist

clotton:
## Summary

The Micron `markets-industries/ai` page had two carousel components with blank heading fields. Each rendered an empty `<h2 class="cmp-carousel-v2__heading"></h2>` in the preflight scrape. This triggered two separate `heading-empty` checks — but a bug in `getElementsFromCheck` caused both cards to receive selectors for *all* empty H2s on the page, making them look identical.

**Root cause:** In `src/preflight/headings.js`, the `heading-empty` case re-queried every empty heading and attached all their selectors to every card. With N empty H2s, N cards were each labelled "1 of N instances found" and appeared as duplicates.

**Fix:** Use `check.transformRules.selector` — already set by the headings handler for the specific element that triggered each check — instead of re-querying. Each card now resolves exactly its own element.

**Confirmed via S3:** The prod scrape for this site (`spacecat-prod-scraper`, site `7b8f919a`, path `markets-industries/ai/scrape.json`, dated 2026-03-21) contains exactly two empty `cmp-carousel-v2__heading` H2s at positions #14 and #15 — matching the two duplicate cards Micron reported. The Micron page has since been updated (carousel headings filled in), but the fix is needed for any page with multiple empty headings.

## Changes

- `src/preflight/headings.js`: `heading-empty` case now uses `check.transformRules.selector` to target the specific element rather than re-querying all empty headings

## Test plan

- [ ] Updated existing `heading-empty` test to include `transformRules.selector` (matches actual handler behavior)
- [ ] Added regression test: 2 empty H2s → 2 checks → each card gets exactly 1 selector (no duplicates)
- [ ] Added test: `heading-empty` without `transformRules.selector` produces no selectors (no fallback to re-query all)
- [ ] `npm run test:spec -- test/preflight/headings-selectors.test.js` → 10 passing, 1 pre-existing timeout unrelated to this change

Closes SITES-46108

### DIFF
--- a/src/preflight/headings.js
+++ b/src/preflight/headings.js
@@ -45,7 +45,7 @@ function getSeoImpact(checkType) {
  * @param {Object} check - The check result from validatePageHeadingFromScrapeJson
  * @returns {Object} Element targets with selectors
  */
-function getElementsFromCheck(scrapeJsonObject, check) {
+function getElementsFromCheck(scrapeJsonObject, check, log) {
   if (!scrapeJsonObject?.scrapeResult?.rawBody) {
     return {};
   }
@@ -100,6 +100,8 @@ function getElementsFromCheck(scrapeJsonObject, check) {
             selectors.push(resolvedSelector);
           }
         }
+      } else {
+        log.debug(`[preflight-headings] heading-empty check has no transformRules.selector — element target will be omitted (check: ${JSON.stringify(check.transformRules)})`);
       }
       break;
     }
@@ -284,7 +286,7 @@ export default async function headings(context, auditContext) {
 
           // Generate selectors from the scraped HTML
           if (scrapeJsonObject) {
-            const elementData = getElementsFromCheck(scrapeJsonObject, check);
+            const elementData = getElementsFromCheck(scrapeJsonObject, check, log);
             if (elementData?.elements?.length > 0) {
               Object.assign(opportunity, elementData);
             }
--- a/src/preflight/headings.js
+++ b/src/preflight/headings.js
@@ -88,15 +88,18 @@ function getElementsFromCheck(scrapeJsonObject, check) {
     }
 
     case 'heading-empty': {
-      // Find empty headings - extract tag name from check
-      const tagName = check.tagName?.toLowerCase();
-      if (tagName && /^h[1-6]$/.test(tagName)) {
-        const headingsArray = $(tagName).toArray();
-        // Find empty ones
-        const emptyHeadings = headingsArray.filter((h) => $(h).text().trim().length === 0);
-        selectors = emptyHeadings
-          .map((h) => getDomElementSelector(h))
-          .filter(Boolean);
+      // Use the specific element selector from the check rather than re-querying all empty
+      // headings — re-querying attaches every empty heading's selector to every card, making
+      // N empty headings produce N cards each showing "1 of N instances found".
+      const selector = check.transformRules?.selector;
+      if (selector) {
+        const element = $(selector).get(0);
+        if (element) {
+          const resolvedSelector = getDomElementSelector(element);
+          if (resolvedSelector) {
+            selectors.push(resolvedSelector);
+          }
+        }
       }
       break;
     }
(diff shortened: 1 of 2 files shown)


MysticatBot (commented):
Hey @clotton,

**Verdict: Comment** - clean, well-scoped bug fix with solid regression coverage.
_Changes: fixes duplicate heading-empty cards by using `check.transformRules.selector` to target each specific empty heading instead of re-querying all empty headings on the page (2 files)._

<details>

- nit: Silent no-op when `transformRules.selector` is absent produces no log or metric - if the upstream headings handler regresses and stops populating this field, cards will silently lack element targets with no signal in Coralogix. A `log.debug` when `selector` is falsy would make this failure mode observable. - `src/preflight/headings.js:91`
- suggestion: Assert `toElementTargets` call count in the regression test (e.g., `.calledTwice`) to catch spurious extra invocations beyond the per-check-no-duplicate guarantee. - `test/preflight/headings-selectors.test.js:345`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 46s | <b>Cost:</b> $1.99 | <b>Commit:</b> <code>70fb8453fe314c3c0c7bf7ffbe6417a749d7a7e1</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @clotton,

**Verdict: Approve** - clean fix with both prior findings addressed.
_Changes: replaces the re-query-all pattern in the heading-empty case with per-check `transformRules.selector` targeting to eliminate duplicate cards (2 files)._

- Debug log now surfaces when `transformRules.selector` is absent, with stringified `transformRules` context for debugging upstream regressions.
- `toElementTargets` call count assertion (`.calledTwice`) guards against the duplicate-card regression in the new multi-heading test.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 28s | <b>Cost:</b> $2.10 | <b>Commit:</b> <code>12852ca1e6a3944335f646d79fdbbdd34f69a0ab</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

GeezerPelletier:
## Code Review

The fix is correct and the root cause diagnosis is accurate. The core approach — using `check.transformRules.selector` instead of re-querying all empty headings — directly resolves the duplicate-card bug. A few observations:

### Production change (`src/preflight/headings.js`)

**Silent DOM miss** — when `$(selector).get(0)` returns `null` (the selector from `transformRules` doesn't match anything in the scraped HTML), the element is silently dropped with no log. The `else` branch only fires when `transformRules.selector` is absent entirely. A debug log on the inner miss path would help diagnose future issues:
```js
const element = $(selector).get(0);
if (!element) {
  log.debug(`[preflight-headings] heading-empty selector "${selector}" matched no element in scraped HTML`);
}
```

**Double-resolution coupling** — the code takes `check.transformRules.selector`, finds the element in the DOM, then calls `getDomElementSelector(element)` to produce a normalized selector. This implicitly assumes the scrape HTML parsed by `getElementsFromCheck` is identical to what the handler saw when it generated `transformRules.selector`. That should hold in practice (same scrape source), but it's worth keeping in mind as an implicit dependency.

**Behavior change for checks without `transformRules.selector`** — the old code would still find empty headings via DOM query even without `transformRules`. The new code intentionally produces no selectors in that case (to prevent the duplicate-card bug), and this is explicitly tested. Just flagging it as a meaningful contract change for any code path that might produce `heading-empty` checks without `transformRules.selector` set.

**Minor**: in the `else` debug log, `JSON.stringify(check.transformRules)` prints `undefined` when `transformRules` is absent. `JSON.stringify(check)` or a relevant subset would be more actionable.

### Tests (`test/preflight/headings-selectors.test.js`)

The three new test cases directly target the bug scenario — good coverage. One gap: the anti-duplication test verifies `toElementTargets` call args and call count, but doesn't assert on what the resulting `auditsResult` entries actually contain. If the bug regressed such that both calls used the same selector, the mock stubs would still satisfy these assertions. Asserting on the output data structure would catch a broader class of regressions.

The `mockSharedUtils` extraction correctly reflects the actual module dependency (`../../src/headings/shared-utils.js`) and is applied consistently across all `esmock` call sites.

The 10 s timeout bump — a one-line comment explaining why (esmock loading overhead, slow CI, etc.) would help future contributors.

### Summary

Clean, well-scoped fix with solid test coverage. The suggestions above are all nits — nothing blocking. LGTM.

GeezerPelletier (approved):

PR #2621, opened 2026-06-09
fix: preserve suggestion updatedBy when re-audit keeps the status

tarunsinghdev:
## Problem

`syncSuggestions` (the shared sync used by ~30 audit types, incl. **cwv** and **broken-backlinks**) stamps `updatedBy = 'system'` **unconditionally** on every matched suggestion on **every re-audit** — regardless of whether the status changed or who last acted.

Consequence: for any still-failing suggestion whose status is *preserved* (`ERROR` / `APPROVED` / `IN_PROGRESS` / `REJECTED` / …), the real actor is overwritten to `system` within a week. So `updatedBy` is unreliable for triage/attribution of failed PLG auto-fixes — a failure genuinely caused by a **user action** shows up as `system`, masking who actually did it.

(The **alt-text** audit doesn't have this problem — it never re-stamps existing suggestions, only the opportunity. This PR brings the shared sync in line with that behavior.)

## Fix

Only attribute the update to `system` when this re-audit **actually transitions the status**. When the status is preserved (`mergeStatusFunction` returns `null`), leave `updatedBy` untouched. Suggestion `data` (metrics, etc.) is still merged and saved either way — only the audit-trail actor is no longer clobbered.

```diff
     const newStatus = mergeStatusFunction(existing, newDataItem, { ...context, isTBYB });
-    // null indicates to keep existing status
     if (newStatus !== null) {
       existing.setStatus(newStatus);
+      existing.setUpdatedBy('system');
     }
-    existing.setUpdatedBy('system');
```

## Behavior

| Case | Before | After |
|------|--------|-------|
| Status preserved (`ERROR`/`APPROVED`/`IN_PROGRESS`/`REJECTED`…) | `updatedBy` → `system` (real actor lost) | `updatedBy` **preserved** |
| Status changed by audit (`ERROR→NEW` re-dispatch, `OUTDATED→NEW` resurrect) | `updatedBy` → `system` | `updatedBy` → `system` (unchanged — correct, the audit did it) |
| New suggestions | unaffected | unaffected |
| Suggestion `data` merge + save | always | always (unchanged) |

## Testing

- Added a regression test: `updatedBy` is **not** re-stamped when `mergeStatusFunction` returns `null` (status preserved), and `data` is still merged + saved.
- Extended the `ERROR → NEW` test to assert `setUpdatedBy('system')` **is** called when the audit transitions the status.
- Tightened the "preserve REJECTED" test to assert `updatedBy` is not touched.
- `test/utils/data-access.test.js` `syncSuggestions` suite: no new failures vs. `main`.

## Risk / scope

Shared helper used by ~30 audit types, but the change is conservative — it only **keeps more information** (a real actor) and never drops a status transition. No consumer should depend on `updatedBy` always being `system` post-audit (that was the bug).

## Related Issues

Thanks for contributing!


MysticatBot (commented):
Hey @tarunsinghdev,

**Verdict: Comment** - clean fix, no blocking issues.
_Changes: moves `setUpdatedBy('system')` inside the status-transition conditional so the real actor is preserved when a re-audit keeps the existing status (2 files)._

<details>

- nit: the 6-line inline comment (lines 505-511) restates the PR description at length - consider trimming to 1-2 lines that state the contract (`null = keep status + preserve current updatedBy; non-null = transition + attribute to system`) and let the git history carry the backstory - `src/utils/data-access.js:505`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 26s | <b>Cost:</b> $2.11 | <b>Commit:</b> <code>b56fe008329fb34e6d1d90076d621d55a66b1e25</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2624, opened 2026-06-09
feat: read brandalf feature-flag from db instead of making API call

rarescheseli:
Read brandalf feature flag via dataAccess instead of the SpaceCat API

**Summary**
* replace the HTTP feature-flags API call in isBrandalfEnabled with a direct read of the feature_flags table through the mysticat PostgREST client (dataAccess.services.postgrestClient), matching the pattern already used for brand-presence data.
* update both call sites (llmo-customer-analysis/handler.js, offsite-brand-presence-enrichment.js) to pass the PostgREST client; drops the `SPACECAT_API_BASE_URL`/`SPACECAT_API_KEY` dependency for this check.
Preserve the existing three-state contract (`true`/`false`/`null`), where a missing row resolves to `false`.

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/llmo-customer-analysis/handler.js
+++ b/src/llmo-customer-analysis/handler.js
@@ -296,7 +296,8 @@ export async function runLlmoCustomerAnalysis(finalUrl, context, site, auditCont
       log,
     });
     if (orgId) {
-      const isV2 = onboardingMode !== 'v1' && await isBrandalfEnabled(orgId, env, log);
+      const isV2 = onboardingMode !== 'v1'
+        && await isBrandalfEnabled(orgId, context.dataAccess?.services?.postgrestClient, log);
       if (isV2) {
         organizationId = orgId;
         const brand = await findActiveBrandForSite(context, { orgId, siteId });
--- a/src/utils/brandalf-utils.js
+++ b/src/utils/brandalf-utils.js
@@ -10,47 +10,45 @@
  * governing permissions and limitations under the License.
  */
 
-import { tracingFetch as fetch } from '@adobe/spacecat-shared-utils';
+const BRANDALF_FLAG_NAME = 'brandalf';
+const FEATURE_FLAG_PRODUCT = 'LLMO';
 
 /**
- * Checks whether the brandalf feature flag is enabled for an organization
- * by calling the SpaceCat API feature-flags endpoint.
+ * Checks whether the brandalf feature flag is enabled for an organization by
+ * reading the `feature_flags` table directly via the mysticat PostgREST client
+ * (available at `context.dataAccess.services.postgrestClient`).
  *
  * @param {string} organizationId - SpaceCat org UUID
- * @param {object} env - Environment variables (needs SPACECAT_API_BASE_URL, SPACECAT_API_KEY)
+ * @param {object} postgrestClient - mysticat PostgREST client (dataAccess.services.postgrestClient)
  * @param {object} log - Logger
  * @returns {Promise<boolean|null>} true/false when the flag state is known, null when unknown
  */
-export async function isBrandalfEnabled(organizationId, env, log) {
-  const { SPACECAT_API_BASE_URL: apiBase, SPACECAT_API_KEY: apiKey } = env || {};
+export async function isBrandalfEnabled(organizationId, postgrestClient, log) {
   if (!organizationId) {
     return false;
   }
-  if (!apiBase || !apiKey) {
-    log?.warn('SPACECAT_API_BASE_URL or SPACECAT_API_KEY not configured; cannot check brandalf flag');
+  if (!postgrestClient?.from) {
+    log?.warn('PostgREST client not available; cannot check brandalf flag');
     return null;
   }
 
   try {
-    const url = `${apiBase}/organizations/${encodeURIComponent(organizationId)}/feature-flags?product=LLMO`;
-    const response = await fetch(url, {
-      headers: { 'x-api-key': apiKey },
-    });
+    const { data, error } = await postgrestClient
+      .from('feature_flags')
+      .select('flag_value')
+      .eq('organization_id', organizationId)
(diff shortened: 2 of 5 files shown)


MysticatBot (commented):
Hey @rarescheseli,

**Verdict: Comment** - clean refactor, one minor test-coupling observation.
_Changes: replaces the HTTP feature-flags API call with a direct PostgREST DB query for the brandalf flag, preserving the three-state contract (5 files)._

<details>

- nit: the happy-path test (`'returns true when the brandalf flag is enabled'`) pins 5 interaction assertions on the query-builder chain (`.from`, `.select`, `.eq` x3). These will break on any query refactor (reordering `.eq()`, adding `.limit(1)`) even when observable behavior is unchanged. Consider isolating the query-shape contract in a dedicated test and keeping the behavioral test focused on the return value alone. - `test/utils/brandalf-utils.test.js:53`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 42s | <b>Cost:</b> $2.46 | <b>Commit:</b> <code>7c8d509fb9709c6f6d76d7e4d28ade645ef2c9e4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

AndreiAlexandruParaschiv (approved):

dzehnder (approved):
Hey @rarescheseli,

## Strengths

- Clean, behavior-preserving transport swap. The `true`/`false`/`null` three-state contract is preserved exactly: missing client -> `null`, query `error` -> `null`, thrown exception -> `null`, absent row -> `false`, explicit row -> `flag_value === true`. `src/utils/brandalf-utils.js:51-53`
- Good architectural consistency. The flag check now shares the exact PostgREST client, connection, and failure mode as the brand-presence data read it gates (`offsite-brand-presence-enrichment.js:379`), instead of reaching its data over a separate HTTP transport with its own auth and failure domain. Net reduction in moving parts.
- Security-positive. Drops the `SPACECAT_API_KEY`/`SPACECAT_API_BASE_URL` dependency for this check (one fewer transmitted secret), and has no injection surface - `organizationId`, `'LLMO'`, and `'brandalf'` are passed as values to PostgREST `.eq()` filters (two of three are hardcoded constants), not interpolated into SQL.
- Fails closed/unknown, not open: DB error, exception, and missing client all return `null` rather than a truthy value. Warn logs include the org ID and `error.message` only - nothing sensitive. `src/utils/brandalf-utils.js:45,54`
- Constants extracted (`BRANDALF_FLAG_NAME`, `FEATURE_FLAG_PRODUCT`), JSDoc updated to the new signature, and the new unit suite covers all three states including every error/throw path with log assertions.

## Issues

### Minor (Nice to Have)

- The deleted consumer tests pinned handler behavior under the helper's error condition; that coverage is now partially lost. `test/audits/llmo-customer-analysis.test.js` - the two removed tests ("should skip brand resolution when brandalf helper warns and returns false" / "should continue when brandalf helper warns on error") asserted the BP schedule is still created with `body.brand_id` undefined when the helper warned. Under this PR those same conditions now return `null` instead of `false`. The pure `false` -> no-`brand_id` path is still covered by the retained test (around line 1770, `resolves(false)`), but no retained consumer test stubs `resolves(null)`. Since the consumer relies on `null` being falsy in the `&&` at `src/llmo-customer-analysis/handler.js:299`, a future refactor of that line to `=== true`/`=== false` would not be caught. Fix: add one consumer test stubbing `mockIsBrandalfEnabled.resolves(null)` and asserting the schedule is still created without `brand_id` - the `null`-equivalent of the retained `false` test.
- `data?.flag_value === true` assumes a strictly boolean column. `src/utils/brandalf-utils.js:54` - strict equality maps any non-boolean truthy value (`1`, `"true"`) to `false`. This is correct if `feature_flags.flag_value` is a Postgres `boolean` column (which deserializes to a JS boolean), and the comment implies it is. Worth a quick schema confirmation; if you want to pin the mapping against type drift, a test feeding `flag_value: 'true'` would do it. No code change needed if the column is boolean.

Note: the happy-path test's interaction assertions on the query-builder chain (`from`/`select`/`eq` x3) are somewhat over-specified and brittle to query refactors. This was already raised in the existing automated review on this PR, so I am not re-filing it - just confirming two reviewers independently agree.

## Recommendations

- Optional symmetry: `src/llmo-customer-analysis/handler.js:299` collapses a `null` return (client absent or query error) into the non-v2 path via the `&&` short-circuit, with no log distinguishing "flag is genuinely false" from "could not determine the flag." The enrichment caller (`offsite-brand-presence-enrichment.js`) does distinguish `=== null` and logs it. This is pre-existing behavior (the helper returned `null` before this PR too), so it is out of this PR's net scope - but since you are in this file, a one-line `=== null` warn would close the gap.

- This is now the second consumer in the worker to read mysticat tables directly via PostgREST, and the first to use one as a feature gate. That trades an API contract for a schema contract (a rename of `feature_flags`/`flag_value` would surface as a silent `null`/warn, not a loud error). Acceptable given the brand-presence precedent and not fixable in this repo, but a third or fourth `.from('table')` consumer is the signal to invest in a shared typed data-access layer, and whether this read access should be least-privilege-scoped (read-only role/RLS) is a mysticat-data-service/platform concern.

## Assessment

Ready to merge? Yes - the two findings are Minor and optional. Clean, well-scoped, behavior-preserving refactor that aligns the flag check with the data path it gates, removes a credential and an HTTP hop, preserves the three-state contract, and updates both call sites, the JSDoc, and test coverage. CI and codecov are green.

## Next Steps

1. Consider adding the one `resolves(null)` consumer test to restore the downstream coverage the deleted tests provided under the new contract.
2. Confirm `feature_flags.flag_value` is a boolean column (likely already true).
3. Everything else is optional.

PR #2627, opened 2026-06-09
fix: Llmo 5495 Auto-trigger FAQs After Related URLs Are Published

baranovskyalexandr:
## Related Issues

Fixes LLMO-5495 — Auto-trigger FAQs After Related URLs Are Published
Relates to LLMO-5035 — FAQs Related URLs 4-week lookback

---

## What

After `guidance:related-urls` successfully writes and publishes Related URLs to the brand presence sheet, enqueue a best-effort follow-up `faqs` audit for the same site.

Also ships the LLMO-5035 fix: `faqs/guidance-handler` now looks back up to 4 weekly brand presence sheets to find the latest available Related URLs instead of only checking the current week.

## Why

Two independent orchestration gaps left FAQ suggestions stale:

1. **LLMO-5495 (orchestration):** `related-urls` and `faqs` run on separate schedules with separate enablement lists. `related-urls` can successfully populate the brand presence sheet while `faqs` does not run afterward — leaving customers with generic or empty-URL FAQ suggestions until the next scheduled `faqs` run. Observed for `careinsurance.com` and `hdfc.bank.in`.

2. **LLMO-5035 (lookback):** When `faqs` did run, it only read Related URLs from the current week's sheet. If the current week's sheet was missing or had no Related URLs yet, it fell back to empty — ignoring perfectly good data from prior weeks.

### `src/related-urls/guidance-handler.js`

- Added `enqueueFaqsAudit(site, context)`:
  - Checks `isAuditEnabledForSite('faqs', ...)` before enqueueing
  - Emits `log.warn` when `faqs` is not enabled — observable signal for ops that suggestions will not refresh despite a successful Related URLs write
  - Sends `{ type: 'faqs', siteId, auditContext: { triggeredBy: 'related-urls' } }` to the audit queue
  - Errors are caught and logged; never propagate to the `ok()` response
- Enqueue call is placed **outside** the upload/publish `try/catch` so a failure there cannot return a misleading `badRequest`

**Dedupe note:** week-boundary dedupe (`auditedAt >= weekStart`) was considered and explicitly rejected — it would skip the enqueue exactly when `faqs` ran earlier in the same week before `related-urls` wrote the sheet, which is the main bug scenario. Duplicate runs (weekly scheduled + chained) are intentionally allowed; a proper dedupe would require SQS queue inspection or a `triggeredBy` audit record and is left as a follow-up.

**Disabled `faqs` note:** if `related-urls` is enabled but `faqs` is not, the new warning is observable but suggestions will not self-heal. Paired enablement or config validation is a separate product decision (see LLMO-5495).

### `src/faqs/guidance-handler.js` (LLMO-5035)

- `loadLatestRelatedUrlsByGroup()` now iterates up to 4 previous weekly brand presence sheets (`brandpresence-all-w{N}-{YYYY}.xlsx`) and returns the first week that has non-empty Related URLs data
- Stops at the first week with data (no unnecessary SharePoint reads)
- Returns an empty `Map` only if all 4 weeks are empty or missing

### Tests

- `test/audits/related-urls/guidance-handler.test.js` — 9 new tests covering: successful enqueue, disabled faqs with warning, zero rows written (no enqueue), SQS unavailable, `sendMessage` failure, missing audit queue, `Configuration.findLatest` rejection, queue URL assertion, and the core scenario (faqs ran earlier same week — enqueue still fires)
- `test/audits/faqs/guidance-handler.test.js` — 8 new tests covering: 4-week lookback, early exit on first week with data, `itemNotFound` treated as missing, customers without Related URLs unaffected, real ExcelJS integration

---

- [ ] make sure to link the related issues in this description ✅

Thanks for contributing!

### DIFF
--- a/src/related-urls/guidance-handler.js
+++ b/src/related-urls/guidance-handler.js
@@ -26,6 +26,7 @@ import {
   RELATED_URLS_COLUMN_HEADER, RELATED_URLS_DELIMITER,
   buildColumnMap, getColumn,
 } from '../faqs/utils.js';
+import { isAuditEnabledForSite } from '../common/index.js';
 
 const WEEKS_TO_LOOK_BACK = 4;
 const MAX_URLS_TO_WRITE = 5;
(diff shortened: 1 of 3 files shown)


duynguyen:
I think there’s still one issue here: awaiting enqueueFaqsAudit means the new FAQ trigger still sits on the request success path even though the PR describes it as best-effort and says it must not affect the ok() response. Errors are caught, but slow calls to Configuration.findLatest() or sqs.sendMessage() can still delay this handler long enough to time out after the sheet has already been written and published. If this follow-up is truly best-effort, I’d suggest bounding it with a short timeout or otherwise decoupling it from the response path.

duynguyen (approved):

PR #2628, opened 2026-06-09
feat: replace duplicate wwwUrlResolver with shared lib delegation and persist resolved domain as overrideBaseURL

tkotthakota-adobe:
https://jira.corp.adobe.com/browse/SITES-46095

Problem and Fix:                                                                                                            
                                                         
The audit worker had its own copy of the URL resolver that didn't include the ghost key fix, so audits would still query the wrong domain. It also never saved the resolved domain, so every audit repeated the same RUM probe. The fix removes the duplicate code, delegates to the shared library (which now has the ghost key check), and saves the correct domain to the site config so future audits use it directly. 

Related PR: https://github.com/adobe/spacecat-shared/pull/1656/


ramboz (changes requested):
## Summary
This PR removes the audit-worker's fork of `wwwUrlResolver`, delegates to the shared-lib implementation (which carries the ghost-key fix), and auto-persists the resolved domain to `overrideBaseURL` so later audits skip the RUM probe. The refactor direction is good and the implementation matches the stated intent. One merge-gating issue remains (coverage), plus a few reliability/correctness items worth addressing before merge.

> Note: the `@adobe/spacecat-shared-utils` gist-tarball dependency is understood to be the project's standard mechanism for deploying the unpublished cross-repo change ([spacecat-shared#1656](https://github.com/adobe/spacecat-shared/pull/1656)) to the dev environment, and is expected to resolve to a published version once that ships. Not flagged.

## Strengths
- **Genuine de-duplication.** Collapsing ~40 lines of forked resolution logic into a single `sharedWwwUrlResolver(site, rumApiClient, log)` call removes a real maintenance hazard — the forked copy drifting from the canonical one is exactly what caused the ghost-key bug this fixes.
- **The persist pattern matches repo convention.** `site.setConfig(Config.toDynamoItem(siteConfig))` mirrors `rum-config-refresh`, `paid-traffic-analysis`, and `llmo-customer-analysis`. The deep `Config` import path is the established idiom here, not a smell.
- **The new tests are well-targeted.** The ghost-key fallback test and the persist test both assert on observable outcomes (`spacecat.com` fallback, `savedConfig.fetchConfig` shape) rather than just exercising code paths.

## Issues

### Blockers (Must Fix Before Merge)

**B1. The change orphans `toggleWWWHostname`, which will fail the 100% coverage gate** — `src/support/utils.js:125`
`base-audit.js` was the only caller of `toggleWWWHostname` (`toggleWWW` at line 111 is a *separate* function and does not call it). After this PR removes the call site, `toggleWWWHostname` has zero callers. Its line 130 is **not** under the `/* c8 ignore next 3 */` (which only covers lines 127–129), so it was previously covered transitively through the old resolver's tests and now becomes unreachable. The repo enforces 100% lines/branches/statements, so `npm test` should fail here. (Codecov's green check does not contradict this — it only scores lines modified in the diff, and `utils.js` isn't in the diff.) Clean fix: **delete the now-dead `toggleWWWHostname`** in this PR.

### Should Fix

**S1. `site.save()` fires on every audit even when the override is already correct** — `src/common/base-audit.js` (`wwwUrlResolver`)
The guard is `resolvedHostname !== baseHostname`. Once `overrideBaseURL` is persisted, the shared resolver short-circuits and returns the override-derived hostname (e.g. `www.spacecat.com`), but `baseHostname` is still the apex (`spacecat.com`) — so the two differ on *every* subsequent run and the code re-persists an identical value each time. `existingFetchConfig` is already read but only spread, never compared. The stated goal ("future audits skip the RUM probe entirely") is half-met: the probe is skipped, but a redundant DB write is added on every audit, for every override-needed site. Add a comparison and a test asserting `save` is not called when the override already matches:
```js
const desiredOverride = `https://${resolvedHostname}`;
if (resolvedHostname !== baseHostname
    && existingFetchConfig.overrideBaseURL !== desiredOverride
    && typeof site.save === 'function') { … }
```

**S2. Fire-and-forget `site.save()` diverges from convention and can silently lose the write** — `src/common/base-audit.js` (`wwwUrlResolver`)
All three existing handlers `await site.save()`. This PR uses `site.save().catch((e) => log.error(...))` and returns immediately. Consequences: (a) in Lambda, the runtime can freeze the event loop after the handler chain unwinds, so a not-yet-settled background save may never flush — defeating the optimization non-deterministically; (b) the next line logs `persisted overrideBaseURL=…` *before* the save resolves, claiming success that may not have happened; (c) errors are swallowed to a log line. Prefer `await` (a single write on the resolution path, not a hot loop), or document the non-blocking tradeoff explicitly.

**S3. Auto-persisted override is sticky and never re-validated** — `src/common/base-audit.js` (`wwwUrlResolver`)
Because the shared resolver short-circuits on `overrideBaseURL`, once the audit sets it the domain is **never re-probed**. If a resolution is ever wrong — a transient RUM blip, or a future edge case the ghost-key fix doesn't cover — the bad domain becomes permanently sticky and self-reinforcing, with no self-healing path short of manual config surgery. At minimum document the one-way-door behavior; ideally consider re-validation or only persisting on a confirmed bundle-data hit.

**S4. New persistence branch likely leaves uncovered branches under the 100% gate** — `test/common/audit.test.js`
The persist test only exercises the happy path: `save` resolves, and `getFetchConfig`/`updateFetchConfig`/`setConfig` all exist. That leaves the `.catch((e) => …)` rejection callback and the `?.` short-circuit branches unexercised — both count against c8 branch/function coverage. Add a `save`-rejects test (and drop the unnecessary optional chaining per N2), or this compounds B1 at the gate.

### Nice to Have

**N1. Lockfile churn unrelated to this change** — `package-lock.json`
The diff sprinkles `"peer": true` onto ~20 unrelated packages (`helix-universal`, `client-dynamodb`, `babel`, `octokit`, `acorn`, …) — the signature of regenerating the lockfile with a different npm/Node version than the committed baseline. If this isn't an expected artifact of the gist-install flow, regenerate with the repo's canonical Node 24 / npm version so only the intended dependency line changes.

**N2. Over-defensive optional chaining inconsistent with repo style** — `src/common/base-audit.js`
`getFetchConfig?.()`, `updateFetchConfig?.()`, and `setConfig?.()` imply the methods might be absent — but the persist test runs against a real `Config`/site where they exist, and existing handlers call `site.setConfig(...)` directly. If `updateFetchConfig` were ever actually undefined, the `?.` would silently persist a config *without* the override (a no-op write that looks successful). Drop the `?.` so a genuine API mismatch fails loudly, and to recover the branch coverage from S4.

**N3. Tests hardcode the shared lib's internal bundle URL** — `test/common/audit.test.js`
The `nock` stubs pin `https://bundles.aem.page/bundles/www.spacecat.com/2023/03/12`. That date/path is the shared lib's internal probe shape; if it changes, these audit-worker tests break for reasons unrelated to audit-worker code. Acceptable for now, just a coupling to be aware of.

## Recommendations
- Consider whether auto-persisting `overrideBaseURL` belongs in `wwwUrlResolver` at all, versus the dedicated `rum-config-refresh` handler that already owns site-config mutation. Mutating customer config as a side effect of every audit's URL resolution is a broad behavioral change for a function whose name implies a pure read.

## Verdict
**Ready to merge:** No
**Reasoning:** B1 (orphaned `toggleWWWHostname`) will fail the 100% coverage gate; resolve it and ideally S1/S2 before merge.

---
*Note: reviewed from the PR diff plus full repo context at `main`, cross-referenced against shared-lib PR #1656. I could not run the test suite or coverage gate locally (`node_modules` not installed; worktree sits at `main`, not the PR branch), so B1/S4 are reasoned statically — please confirm with `npm install && npm test`.*

tkotthakota-adobe (commented):
## Review

### Blockers (Must Fix Before Merge)

**B1. Dead `toggleWWWHostname` will fail the 100% coverage gate — `src/support/utils.js:125`**

`base-audit.js` was the only caller of `toggleWWWHostname` (`toggleWWW` at line 111 is a separate function and does not call it). After this PR removes the call site, `toggleWWWHostname` has zero callers. Its line 130 is not under the `/* c8 ignore next 3 */` (which only covers lines 127–129), so it was previously covered transitively through the old resolver's tests and now becomes unreachable. The repo enforces 100% lines/branches/statements, so `npm test` should fail here. (Codecov's green check does not contradict this — it only scores lines modified in the diff, and `utils.js` is not in the diff.) Clean fix: delete the now-dead `toggleWWWHostname` in this PR.

---

### Should Fix Before Merge

**S1. `site.save()` fires on every audit even when the override is already correct — `src/common/base-audit.js` (`wwwUrlResolver`)**

The guard is `resolvedHostname !== baseHostname`. Once `overrideBaseURL` is persisted, the shared resolver short-circuits and returns the override-derived hostname (e.g. `www.spacecat.com`), but `baseHostname` is still the apex (`spacecat.com`) — so the two differ on every subsequent run and the code re-persists an identical value each time. `existingFetchConfig` is already read but only spread, never compared. The stated goal ("future audits skip the RUM probe entirely") is half-met: the probe is skipped, but a redundant DB write is added on every audit, for every override-needed site. Add a comparison and a test asserting save is not called when the override already matches:

```js
const desiredOverride = `https://${resolvedHostname}`;
if (resolvedHostname !== baseHostname
    && existingFetchConfig.overrideBaseURL !== desiredOverride
    && typeof site.save === 'function') { … }
```

**S2. Fire-and-forget `site.save()` diverges from convention and can silently lose the write — `src/common/base-audit.js` (`wwwUrlResolver`)**

All three existing handlers `await site.save()`. This PR uses `site.save().catch((e) => log.error(...))` and returns immediately. Consequences: (a) in Lambda, the runtime can freeze the event loop after the handler chain unwinds, so a not-yet-settled background save may never flush — defeating the optimisation non-deterministically; (b) the next line logs `persisted overrideBaseURL=…` before the save resolves, claiming success that may not have happened; (c) errors are swallowed to a log line. Prefer `await` (a single write on the resolution path, not a hot loop), or document the non-blocking tradeoff explicitly.

**S3. Auto-persisted override is sticky and never re-validated — `src/common/base-audit.js` (`wwwUrlResolver`)**

Because the shared resolver short-circuits on `overrideBaseURL`, once the audit sets it the domain is never re-probed. If a resolution is ever wrong — a transient RUM blip, or a future edge case the ghost-key fix does not cover — the bad domain becomes permanently sticky and self-reinforcing, with no self-healing path short of manual config surgery. At minimum document the one-way-door behaviour; ideally consider re-validation or only persisting on a confirmed bundle-data hit.

**S4. New persistence branch likely leaves uncovered branches under the 100% gate — `test/common/audit.test.js`**

The persist test only exercises the happy path: save resolves, and `getFetchConfig`/`updateFetchConfig`/`setConfig` all exist. That leaves the `.catch((e) => …)` rejection callback and the `?.` short-circuit branches unexercised — both count against c8 branch/function coverage. Add a save-rejects test (and drop the unnecessary optional chaining per N2), or this compounds B1 at the gate.

---

### Nice to Have

**N1. Lockfile churn unrelated to this change — `package-lock.json`**

The diff sprinkles `"peer": true` onto ~20 unrelated packages (helix-universal, client-dynamodb, babel, octokit, acorn, …) — the signature of regenerating the lockfile with a different npm/Node version than the committed baseline. If this is not an expected artefact of the gist-install flow, regenerate with the repo's canonical Node 24 / npm version so only the intended dependency line changes.

**N2. Over-defensive optional chaining inconsistent with repo style — `src/common/base-audit.js`**

`getFetchConfig?.()`, `updateFetchConfig?.()`, and `setConfig?.()` imply the methods might be absent — but the persist test runs against a real `Config`/`site` where they exist, and existing handlers call `site.setConfig(...)` directly. If `updateFetchConfig` were ever actually `undefined`, the `?.` would silently persist a config without the override (a no-op write that looks successful). Drop the `?.` so a genuine API mismatch fails loudly, and to recover coverage from S4.

**N3. Tests hardcode the shared lib's internal bundle URL — `test/common/audit.test.js`**

The nock stubs pin `https://bundles.aem.page/bundles/www.spacecat.com/2023/03/12`. That date/path is the shared lib's internal probe shape; if it changes, these audit-worker tests break for reasons unrelated to audit-worker code. Acceptable for now, just a coupling to be aware of.

---

### Recommendations

Consider whether auto-persisting `overrideBaseURL` belongs in `wwwUrlResolver` at all, versus the dedicated `rum-config-refresh` handler that already owns site-config mutation. Mutating customer config as a side effect of every audit's URL resolution is a broad behavioural change for a function whose name implies a pure read.

### Verdict

**Ready to merge: No**

B1 (`toggleWWWHostname` orphan) will fail the 100% coverage gate; resolve it and ideally S1/S2 before merge.

tkotthakota-adobe:
Hi @ramboz Thanks for the review. Comments are addressed.
image

ramboz (approved):

PR #2631, opened 2026-06-10
fix(cdn-logs-report): trigger weekly agentic rollup on empty Sunday export

akshaymagapu:
## Problem

The weekly agentic-traffic rollup (`wrpc_refresh_agentic_traffic_weekly`) is only ever triggered as a **side-effect of the Sunday daily import completing** in `mysticat-projector-service` — there is no independent/scheduled trigger.

For sites that have agentic traffic **Mon–Fri but no data on Sunday**, the chain breaks:

1. On Monday, `runDailyAgenticExport` exports yesterday = **Sunday**.
2. Sunday has 0 rows → the export returns `skipped: true` and **never sends a `batch.completed` event** ([agentic-daily-export.js:244](https://github.com/adobe/spacecat-audit-worker/blob/main/src/cdn-logs-report/agentic-daily-export.js#L244)).
3. No batch → no projector import → no weekly-refresh follow-up message.
4. The just-closed ISO week's weekly rollup is **never refreshed**, even though Mon–Fri data was already imported on their own days. (Mid-week imports don't trigger the closed-week rollup because the week isn't closed yet — the projector's `getClosedWeeklyRefreshDateRange` only returns a range once the week's Sunday is in the imported range.)

Until now the only remedy was running the manual `backfill-llmo mode=weekly-db` Slack command, which calls the weekly rollup RPC directly.

## Fix

When the **week-closing Sunday** daily export is skipped (0 rows), call `wrpc_refresh_agentic_traffic_weekly` directly for the ISO week ending on that Sunday — mirroring the api-service `backfill-llmo mode=weekly-db` path. This rolls up whatever days were already imported into the weekly view.

- Scoped narrowly: only fires when the export is **skipped** *and* the traffic date is a **Sunday** (week close). Days with data still flow through the normal batch → projector path; non-Sunday empty days are ignored.
- Best-effort: failures (missing PostgREST client, RPC error, thrown exception) are logged but **never block the rest of the audit**.
- Reuses the existing `context.dataAccess.services.postgrestClient` RPC pattern (as in `replaceAgenticUrlClassificationRules`) and `generateReportingPeriods` for the Mon–Sun week range.

## Testing

- 6 new unit tests covering: Sunday-skip triggers the RPC with the correct week range; non-Sunday skip and Sunday-with-data do **not** trigger; missing PostgREST client, RPC error, and thrown RPC error are all handled without failing the caller.
- `agentic-db-export.js` retains **100% line/branch/statement** coverage.
- Full `cdn-logs-report` suite (handler + daily-export + report-utils) passes (53 tests).

### DIFF
--- a/src/cdn-logs-report/handler.js
+++ b/src/cdn-logs-report/handler.js
@@ -26,17 +26,31 @@ import { getConfigs } from './constants/report-configs.js';
 import { generatePatternsWorkbook } from './patterns/patterns-uploader.js';
 import { runAgenticDbExports } from './utils/agentic-db-export.js';
 import { runDailyReferralExport } from './referral-daily-export.js';
+import { getPreviousUtcDate } from './agentic-daily-export.js';
 
 /**
- * Picks the single week offset used to sample CDN logs for pattern generation.
- * An explicit weekOffset wins; otherwise the previous full week on Mondays,
- * else the current week.
+ * Resolves the reporting period used to sample CDN logs for pattern generation.
+ *
+ * A date-based (backfill) run carries an explicit `auditContext.date`; the period is
+ * derived from that run's traffic date (`date - 1`, matching the daily export) so
+ * patterns are sampled from the week that actually has uploaded data. This matters for
+ * sites whose logs only cover an earlier period — e.g. backfilling May data in June
+ * must not sample the (empty) current week, which would yield no URLs.
+ *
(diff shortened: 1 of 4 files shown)


akshaymagapu:
Superseded by #2632, which carries the same work on a fresh branch cut from main (plus the date-based pattern-sampling fix and weekOffset cleanup).

PR #2633, opened 2026-06-10
fix(cdn-logs-report): stop emitting citability score + deployed-at-edge (LLMO-5646)

akshaymagapu:
## Problem

The PG agentic-traffic dashboard now computes the per-URL citability score client-side (project-elmo-ui `useRcvEnrichedUrls`, merged to main), and the URL-wise score as well. So the daily export no longer needs to produce `citability_score` / `deployed_at_edge` — the backend value is ignored by the UI.

This supersedes the original intent of this PR (re-sourcing citability from prerender suggestions + `status.json`). Instead of replacing the citability source, we remove it from the audit worker entirely.

## Change

Removes all citability derivation from `agentic-traffic-mapper.js`:

- Drop `getCitabilityMap`, the prerender `status.json` S3 read (`readPrerenderStatusJson`), and the active-opportunity lookup.
- Delete the `prerender-citability.js` util and its test.
- The `dimensions` JSONB is now always emitted as an empty object `{}`.

## Why this is safe to ship alone

Verified end to end:

- **Import**: `dimensions` lands as `{}`; PostgreSQL JSONB tolerates missing keys.
- **Refresh RPCs**: `agentic_traffic_parse_citability_score` → NULL, `agentic_traffic_is_deployed_at_edge` → false (no error).
- **API**: the agentic-traffic controller is null-safe.
- **UI**: computes citability client-side and ignores the backend value.

The only stale artifact is the server-side URL export RPC (still has the two columns); that is handled separately in **LLMO-5648**. Broader data-service RPC simplification is **LLMO-5647**.

## Validation

- `prerender-citability.js` removed; no remaining references to it or `getCitabilityMap` in the repo.
- `agentic-traffic-mapper.js` and `agentic-daily-export.js` at **100%** lines/branches/statements/functions.
- Full `test/audits/cdn-logs-report/**` suite passing (179 tests); `eslint` clean.

## Notes

- The branch name predates this scope change; the net diff is a pure removal.

LLMO-5646


dhavkuma:
PageCitability has all the URLs even those which does not have suggestions generated at Prerender Audit ( we do skip some where content gain is not at par)

Are we okay missing out on those urls while creating this agentic traffic map?

src/cdn-logs-report/utils/agentic-traffic-mapper.js
```diff
@@ -79,39 +80,87 @@ function inferContentType(urlPath = '') {
...
 
+// Reads the NEW suggestions from the site's prerender opportunity (replaces PageCitability).
+// Fetching by status keeps the query scoped to fresh, per-URL rows.
 async function getCitabilityScores(site, context) {
-  const pageCitability = context?.dataAccess?.PageCitability;
-  if (!pageCitability?.allBySiteId) {
+  const { Opportunity, Suggestion: SuggestionDA } = context?.dataAccess ?? {};
+  if (!Opportunity?.allBySiteIdAndStatus || !SuggestionDA?.allByOpportunityIdAndStatus) {
     return [];
   }
```
MysticatBot: issue (blocking): @dhavkuma raised that PageCitability covered all URLs (including those the prerender audit skipped due to low content gain), while this new implementation only sources from NEW prerender suggestions. URLs that previously had a citability score will now export with no `citability_score` dimension at all.

This is a material data-completeness regression for downstream consumers of the agentic-traffic bundle. The PR cannot merge without an explicit answer to that question: is the narrower coverage intentional (and acceptable to the dashboard consumers), or should the query be widened to recapture those URLs?

If intentional, a brief code comment documenting the scope reduction would help future maintainers.

src/cdn-logs-report/utils/agentic-traffic-mapper.js
```diff
@@ -79,39 +80,87 @@ function inferContentType(urlPath = '') {
...
-    const existingScore = acc[pathname];
 
-    if (!existingScore || new Date(score.getUpdatedAt()) > new Date(existingScore.updatedAt)) {
+    const isDeployed = !!data.edgeDeployed;
+    const coveredByDomainWide = !!data.coveredByDomainWide;
+    const coveredByPattern = !!data.coveredByPattern;
+    const isDeployedAtEdge = isDeployed || coveredByDomainWide || coveredByPattern;
+
+    const { wordCountBefore, wordCountAfter } = data;
+    const hasUsableWordCounts = Number.isFinite(wordCountBefore)
```
MysticatBot: issue (blocking): When `suggestion.getUpdatedAt?.()` returns `undefined`, `new Date(undefined)` produces `Invalid Date`. The comparison `new Date(updatedAt) > new Date(existing.updatedAt)` is then always `false`.

Consequence: if the first suggestion processed for a pathname has no `updatedAt`, it gets stored with `updatedAt: undefined`. Any subsequent suggestion for the same pathname (even with a valid timestamp) can never replace it, because the comparison always fails.

Fix: normalize missing timestamps so comparisons remain deterministic:
```js
const updatedAt = suggestion.getUpdatedAt?.() || new Date(0).toISOString();
```
Or guard before storing: skip the entry if `updatedAt` is falsy.

MysticatBot (changes requested):
Hey @akshaymagapu,

_Changes: Replaces PageCitability as the citability-score source with prerender suggestions, mirroring the UI scoring logic (2 files)._

## Must fix before merge

1. [Important] Unanswered data-completeness question from @dhavkuma - `src/cdn-logs-report/utils/agentic-traffic-mapper.js:149` (details inline)
2. [Important] `new Date(undefined)` comparison bug when `updatedAt` is missing - `src/cdn-logs-report/utils/agentic-traffic-mapper.js:120` (details inline)

Note: CI checks are currently failing (branch-deploy, it-postgres) - resolve before merge.

<details>

- nit: `getLlmVisibilityScore` can return a negative score when `wordCountBefore < 0` since `Math.min(100, negative)` passes through. Consider `Math.max(0, Math.min(100, ...))` - `src/cdn-logs-report/utils/agentic-traffic-mapper.js:88`
- suggestion: Rename `getCitabilityScores` to `getPrerenderSuggestions` since it now returns suggestions, not scores - `src/cdn-logs-report/utils/agentic-traffic-mapper.js:142`
- nit: Opportunity status uses a string literal `'NEW'` while the Suggestion status uses `Suggestion.STATUSES.NEW` in the same function. Use the constant for both for consistency - `src/cdn-logs-report/utils/agentic-traffic-mapper.js:149`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 57s | <b>Cost:</b> $4.03 | <b>Commit:</b> <code>3976f570901d0fb51a7c252c1400922fc3045c90</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

akshaymagapu:
@dhavkuma yes, that's intended. We're matching the dashboard 1:1 now. Citability comes from the same NEW prerender suggestions the UI reads via `getLlmVisibilityScore`, so URLs the prerender audit skips show no score in the UI today and they'll carry no `citability_score` in the bundle either. The old PageCitability path could show scores for URLs the UI doesn't, which is the mismatch we're dropping with the table.

@anuj-adobe does matching the UI coverage exactly work for you, or should the export keep the wider URL set?

MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - all prior findings addressed correctly, no new blocking issues.
_Changes: Replaces the PageCitability table with prerender suggestions as the citability-score source, mirroring the UI scoring logic 1:1 (2 files)._

- `new Date(undefined)` comparison bug - now normalized with epoch-0 fallback
- Data-completeness question from @dhavkuma - confirmed intentional, scope-note comment added in code
- Negative score floor added with `Math.max(0, ...)`
- Function renamed from `getCitabilityScores` to `getPrerenderSuggestions`
- Opportunity status now uses `Opportunity.STATUSES.NEW` constant

Note: CI checks (branch-deploy, build) were still pending at review time - verify they pass before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 53s | <b>Cost:</b> $2.60 | <b>Commit:</b> <code>4ceeb7873d9983fd3339bf8cc9b747942ad64625</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - clean refactor, no blocking issues found on re-review.
_Changes: Replaces PageCitability with prerender suggestions + S3 status.json as the citability-score source, extracting scoring logic into a pure utility (4 files)._

Note: CI checks (build) were still pending at review time - verify they pass before merge.

<details>

- nit: `readPrerenderStatusJson` catch produces a generic "Unexpected end of JSON input" message when status.json is empty (0 bytes) - including the S3 Key in the warning would aid debugging - `src/cdn-logs-report/utils/agentic-traffic-mapper.js:96`
- suggestion: `normalizePath` returns the original falsy value (`undefined`/`''`) rather than a canonical sentinel - consider documenting or returning `null` to make the contract explicit for callers - `src/cdn-logs-report/utils/prerender-citability.js:25`
- nit: `hasDomainWideFix` does a literal `.includes('/*')` on a field named `allowedRegexPatterns` - a one-line comment noting this is a convention string (not a regex match) would prevent a future "fix" - `src/cdn-logs-report/utils/prerender-citability.js:57`
- suggestion: The test at line 242 bundles 5 independent scenarios (wrong type, no DA, missing Suggestion DAO, null opps, null suggestions) into a single `it()` block - splitting would improve failure diagnosis - `test/audits/cdn-logs-report/agentic-traffic-mapper.test.js:242`

</details>

- `new Date(undefined)` comparison bug - now uses epoch-0 fallback
- Data-completeness question from @dhavkuma - confirmed intentional, matching dashboard 1:1
- Negative score floor added with `Math.max(0, ...)`
- Function renamed from `getCitabilityScores` to `getCitabilityMap`
- Opportunity status uses `Audit.AUDIT_TYPES.PRERENDER` constant

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 55s | <b>Cost:</b> $3.23 | <b>Commit:</b> <code>3988e54e1a4c90ba1f01c4f6fcead8a3301f58e4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anuj-adobe:
I've restored the experience on UI by PR: https://github.com/adobe/project-elmo-ui/pull/2085 after discussing with @vivesing from ELMO UI team.
Do we still need this change on backend?

MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - clean removal, no blocking issues.
**Complexity: HIGH** - large diff (mostly test refactoring).
_Changes: Removes all citability-score and deployed-at-edge computation from the agentic traffic mapper, emitting empty dimensions (3 files)._

<details>

- question: @anuj-adobe asked (June 16) whether this backend removal is still needed after they "restored the experience on UI" via project-elmo-ui#2085. Could you confirm the UI computes citability client-side and does not consume the backend `dimensions` value? - PR thread
- nit: `const dimensions = {};` is allocated per-row but never populated. Consider a module-level `Object.freeze({})` constant, or note in a follow-up whether the field itself can be dropped from the schema (LLMO-5647/5648 may cover this) - `src/cdn-logs-report/utils/agentic-traffic-mapper.js:127`
</details>

- All prior findings were against a different implementation (prerender-based citability source) - PR scope changed to pure removal, rendering them moot

Note: CI checks (build) were still pending at review time - verify they pass before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 48s | <b>Cost:</b> $5.38 | <b>Commit:</b> <code>c032a295752e7dd2b6b2b263957a8025c6ed067e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2634, opened 2026-06-10
feat(llm-error-pages): dual-publish observation:llm-broken-urls behin…

ymadobe:
<p><strong>Summary</strong></p>
<p>Adds a second SQS publish from the LLM Error Pages audit handler that emits the new Mysticat blackboard message <code>observation:llm-broken-urls</code>. The legacy <code>guidance:llm-error-pages</code> publish is unchanged — both run side-by-side during shadow validation. Gated by env var <code>OBSERVATION_LLM_BROKEN_URLS_ENABLED</code>, default off, zero production risk at merge time.</p>
<p><strong>Why</strong></p>
<p>LLM Error Pages is migrating from the legacy Mystique-crew architecture to the Mysticat blackboard architecture. The new pipeline ingests via SQS → <code>o_llm_broken_urls</code> fact → verifier → alternatives → projector → opportunity. This PR is the audit-worker-side producer for that pipeline. Feature-flagging lets this land before the mystique-side ingestion task and projector merge, decoupling the rollout sequence.</p>
<p><strong>Changes</strong></p>
<ul>
<li>New conditional SQS publish inside <code>runAuditAndSendToMystique</code>, gated on <code>env.OBSERVATION_LLM_BROKEN_URLS_ENABLED === 'true'</code>.</li>
<li>Re-groups the existing consolidated 404 set by URL only (the legacy path keys by url+provider) to produce one observation entry per URL, with user-agent providers unioned.</li>
<li>Wire format matches the Pydantic validator on the mystique ingestion task: camelCase <code>siteId</code>, <code>auditId</code>, <code>baseURL</code>, <code>deliveryType</code>, <code>userAgents</code>, <code>lastSeen</code>, with <code>period</code> sub-object and <code>type: "observation:llm-broken-urls"</code> discriminator.</li>
<li>Defensive caps on user-agent count (50) and string length (512) to mirror mystique's max bounds.</li>
<li><code>messageBaseUrl</code> truthiness guard — skips publish with warn log if absent rather than emitting a malformed <code>baseURL</code>.</li>
<li>New test coverage exercising the publish path, flag-off path, missing-baseURL path, multi-provider dedup, oversized-UA cap, and missing-optional-field fallbacks. 100% line/branch/func/stmt coverage on handler.js.</li>
</ul>
<p><strong>Rollout</strong></p>
<ol>
<li>Merge with flag off — no behavior change.</li>
<li>After dependent mystique + projector PRs deploy, flip <code>OBSERVATION_LLM_BROKEN_URLS_ENABLED=true</code> in dev.</li>
<li>Validate observation shape against mystique's parsed-task logs.</li>
<li>Promote flag stage → prod.</li>
<li>Phase 4: remove legacy publish + the flag itself.</li>
</ol>
<hr>


anagarwa (commented):
Hey @ymadobe,

Overall this is a well-engineered PR. The failure isolation design, the security regression tests, and the inline documentation all show careful thinking. A few minor points below.

## Strengths

- **Failure isolation** (`handler.js:73-186`): The broad try/catch on the shadow publish is exactly the right pattern here - swallowing errors is justified by the explicit contract that the new path must never affect the legacy path or trigger SQS re-delivery. The rationale is documented in the function's JSDoc, which makes the intent clear to future maintainers.
- **Security regression tests** (`test/audits/llm-error-pages/utils.test.js:573-591`): The three SECURITY-prefixed tests pinning the `toPathOnly` contract are excellent. They explicitly document the origin-defense dependency and would catch a future refactor that accidentally preserved absolute URLs.
- **Test suite breadth** (`test/audits/llm-error-pages/handler.test.js:240-551`): Flag-off path, strict `'true'` enforcement, wire schema validation, multi-provider dedup, UA caps, size budget, and failure isolation are all explicitly verified. Assertion density of 3.29 per test is healthy.
- **Inline documentation quality** (`handler.js:34-186`): Phase 4 cutover instructions, cross-repo coordination notes (`observedThrough` vs `lastSeen`), and defense layer explanations are unusually thorough and will save the next engineer significant archaeology time.
- **Strict flag semantics** (`handler.js:210`): `=== 'true'` strict comparison is the right posture for a feature gate controlling production traffic.

## Issues

### Minor (Nice to Have)

**`OBSERVATION_MAX_URLS` cap is dead code on the effective input**

`handler.js:122` - `sorted404` is already sliced to 50 entries at the call site, so `byUrl` can have at most 50 entries. The `.slice(0, OBSERVATION_MAX_URLS)` call is a no-op under normal conditions. Either remove the slice or consolidate `TOP_404_LIMIT` and `OBSERVATION_MAX_URLS` into a single constant to prevent them drifting out of sync if one is ever changed.

**Serialization size check uses `.length` (UTF-16 code units) not byte count**

`handler.js:174` - `.length` counts UTF-16 code units; the SQS 256 KB limit is in bytes (UTF-8 encoded). For non-ASCII characters in user-agent strings, this will undercount the byte size. LLM bot UA strings are ASCII-only in practice today, but consider `Buffer.byteLength(serialized, 'utf8')` for a byte-accurate check.

**Two nearly identical tests for missing baseURL**

`handler.test.js:366 and 491` both verify "observation publish skipped with a warn log when site has no baseURL configured" with nearly identical fixtures and assertions. Remove one, or if the distinction is intentional, document it in the test name.

**Missing spec in `/docs/specs/`**

CLAUDE.md states significant new functionality must be developed spec-first. No spec exists in `docs/specs/` for this migration.

## Recommendations

Document `OBSERVATION_LLM_BROKEN_URLS_ENABLED` in an ops runbook or env-var reference before the flag is flipped in any environment. The PR body's rollout steps are a good start but disappear after merge.

## Assessment

Ready to merge? Yes
Reasoning: The code is correct, well-tested, and feature-flagged for zero production impact at merge time. The minor issues above are worth addressing but do not block merge. Ensure the mystique-side field-name alignment (`observedThrough` vs `lastSeen`) lands before the flag is enabled anywhere.

anagarwa (approved):

PR #2635, opened 2026-06-10
fix(llm-error-pages): HEAD-check Mystique suggestions and write empty rows to DB

ymadobe:
Two coordinated fixes for the LLM error-pages spreadsheets flow:

1. **HEAD-check defence-in-depth.** New `src/llm-error-pages/url-health-check.js` filters Mystique-suggested URLs that return a confident 4xx/5xx on the live site. Timeouts / network errors / 405 Method-Not-Allowed are kept (inconclusive — never penalise a slow customer site). When the filter empties a link's list, also blanks `aiRationale` so the row stays internally consistent.
2. **Remove Excel↔DB drift.** The `!suggestedUrls?.length` gate at the DB-write path silently dropped empty-suggestion rows while Excel still wrote them. Removed — what lands in DB now matches what lands in Excel.

## Test plan

- [ ] CI green
- [ ] 100% line/branch/statement coverage on `guidance-handler.js` + `url-health-check.js` (verified locally)
- [ ] Post-deploy: confirm CloudWatch shows `[LLM-ERROR-PAGES] Dropped N suggested URL(s) that failed HEAD check` for sites with hallucinated picks
- [ ] Spot-check next weekly motilaloswal Excel — no 404 URLs in Suggested URLs column
- [ ] Confirm empty-suggestion rows appear in both Excel and the Suggestion DB table (no more drift)


anagarwa (dismissed):
# PR Review — fix(llm-error-pages): HEAD-check Mystique suggestions and write empty rows to DB

5 reviewers (staff-engineer, architect, project-conventions-reviewer, security, qa-engineer) reviewed this change in parallel. The HEAD-check defence-in-depth is a sound idea and the test diet is genuinely good, but several material issues need to land before merge.

## Blockers (Major)

### 1. SSRF: new module fetches arbitrary URLs from an untrusted upstream with no host/IP/scheme guard
**`src/llm-error-pages/url-health-check.js` + `guidance-handler.js`**

URLs arrive via SQS (`data.brokenLinks[].suggestedUrls`) from Mystique — the same upstream this PR explicitly distrusts — and flow straight into `fetch()` with no scheme allowlist, no hostname validation, and no IP-range check. A message containing `http://169.254.169.254/latest/meta-data/iam/...`, `http://10.x.x.x/...`, or `http://localhost:port/...` causes the Lambda to probe AWS IMDS, internal VPC services, and loopback. Even though only `res.status` is consumed, the side effects are real: blind port/service discovery and (if IMDSv2 hop-limit isn't 1) credential reachability — and this Lambda's middleware chain holds S3, SQS, Secrets Manager, and dataAccess permissions.

**This repo already ships the correct mitigation** at `src/site-detection/handler.js:105-200` (`isPrivateIP` + `resolvesToPublicAddress` + `redirect: 'manual'`). Please reuse it.

Fix:
- Lift `isPrivateIP` / `resolvesToPublicAddress` from `src/site-detection/handler.js` into a shared util (e.g. `src/support/url-safety.js`).
- Restrict scheme to `http:`/`https:`.
- Reject IP literals and any hostname that resolves to RFC1918 / loopback / link-local / CGNAT / multicast / IPv6 ULA.
- Consider scoping to the customer's `baseUrl` registrable domain — that matches the business logic ("alternative URL for a broken link on this site").

### 2. SSRF via redirect: `redirect: 'follow'` defeats any future host guard
**`src/llm-error-pages/url-health-check.js:46`**

An attacker-controlled URL can 302 to `http://169.254.169.254/...` and native fetch will follow without re-validating the new host. The peer module at `src/site-detection/handler.js:200` already uses `redirect: 'manual'` for this exact reason.

Fix: change to `redirect: 'manual'` and treat 3xx as inconclusive (keep the URL — consistent with the existing "only drop confident 4xx/5xx" policy).

### 3. Cross-codebase drift: this is the fourth HEAD-reachability implementation in the repo
**Architecture-level**

`src/internal-links/helpers.js:218` (`checkLinkWithHead`), `src/redirect-chains/handler.js:95-230`, and `src/sitemap/common.js:174` already implement HEAD-based probes with their own concurrency, UA, timeout, and 405 handling. Additionally, almost every audit imports `tracingFetch` from `@adobe/spacecat-shared-utils` so calls appear in Honeycomb/Coralogix traces — this new module uses bare `fetch` and will be invisible.

Fix: at minimum swap to `tracingFetch`. Better: file a follow-up to consolidate into a shared util and reuse `checkLinkWithHead` semantics.

### 4. Operational risk: 10-global concurrency hammers a single origin
**`src/llm-error-pages/url-health-check.js:23,74-82`**

The docstring notes "typical batch is 50 broken links × 3 suggestions = 150 URLs, almost all hitting the same domain". With 10 global concurrency and 10 ECS workers, a single audit can put 50–100 concurrent HEADs against one customer origin. Cloudflare/Akamai bot challenges trigger; the inconclusive-keeps rule then turns the filter into a silent no-op precisely on the sites it most needs to protect.

Fix: cap per-host concurrency to 2–3 (group by `new URL(u).origin`, p-limit per host). At minimum, log the outcome distribution (`kept_2xx, kept_3xx, kept_405, kept_error, dropped_4xx, dropped_5xx`) so "filter silently disabled" is detectable.

### 5. Telemetry gap: a silent filter between Mystique and the user-facing report needs a metric
**`src/llm-error-pages/guidance-handler.js:71`**

One `log.info("Dropped N...")` per invocation gives CloudWatch rows but no aggregable signal. If Mystique regresses and starts sending 90% bad URLs, no one will know without manual per-siteId log inspection. "Dropped 0" is also ambiguous: Mystique healthy vs. customer rate-limiting our probes.

Fix: emit a structured metric with `{siteId, total, dropped_4xx, dropped_5xx, kept_inconclusive}`. The premise of the PR is "we no longer trust the upstream" — that requires a dashboard.

### 6. Missing spec & ADR (CLAUDE.md explicit rule violations)
**`docs/specs/`, `docs/decisions/`**

CLAUDE.md lines 179-185 require:
- *"Significant new functionality must be developed spec-first. Add a spec to `/docs/specs/`."* A new 89-LOC module + behavior change at the audit-worker/Mystique boundary qualifies (precedent: `docs/specs/alt-text-url-fallback-chain.md` for an equivalent pattern in `image-alt-text`).
- *"Technical or architectural decisions with long-term impact must be recorded as ADRs in `/docs/decisions/`."* Two decisions here meet this bar: (a) consumer-side defence-in-depth vs. tightening upstream Mystique, (b) the Excel↔DB contract change persisting empty-suggestion rows.

Fix: add `docs/specs/llm-error-pages-head-check-and-db-consistency.md` and `docs/decisions/001-llm-error-pages-consumer-side-url-validation.md`. Both should reference a Mystique follow-up ticket so this workaround has an expiry condition.

### 7. Test gap: AbortController timeout path is never exercised
**`test/audits/llm-error-pages/url-health-check.test.js`**

`HEAD_TIMEOUT_MS = 5000` + `controller.abort()` produce an `AbortError` (`DOMException('aborted', 'AbortError')`), distinct from the generic `Error('network unreachable')` the test uses. The timeout path is the highest-risk behavior in this module (protects the handler from hanging) and is untested.

Fix:
```js
it('keeps URL when HEAD times out (AbortError)', async () => {
  fetchStub.rejects(Object.assign(new Error('aborted'), { name: 'AbortError' }));
  const result = await filterReachableUrls(['https://slow.example/x'], log);
  expect(result).to.deep.equal(['https://slow.example/x']);
});
```

### 8. Test gap: batch serialization (the point of HEAD_CONCURRENCY) is not asserted
**`test/audits/llm-error-pages/url-health-check.test.js` (>10 URLs test)**

The 25-URL test only asserts `callCount === 25`. A `Promise.all(urls.map(...))` refactor would still pass. Per-host rate-limiting is the docstring's stated motivation — add an inflight-counter probe:
```js
let inflight = 0, maxInflight = 0;
fetchStub.callsFake(async () => {
  inflight += 1; maxInflight = Math.max(maxInflight, inflight);
  await new Promise((r) => setImmediate(r));
  inflight -= 1; return { status: 200 };
});
// ... assert expect(maxInflight).to.be.at.most(10);
```

### 9. Test correctness: "brokenLinks absent" test comment contradicts actual behavior
**`test/audits/llm-error-pages/guidance-handler.test.js`**

The test comment says "The Excel block throws on `brokenLinks.length` but that's caught." Post-normalization this is wrong — `brokenLinks` is always an array, so the Excel block no-ops rather than throws, and `uploadToSharePoint`/`publishToAdminHlx` **are** called (writing an empty workbook to SharePoint). Either fix the comment and assert what really happens, or add an early-return for missing brokenLinks. Current state is undefined behavior with a misleading comment.

## Minor

**10. `clearTimeout(timer)` belongs in `finally`** (`url-health-check.js:39-62`): if anything between `setTimeout` and `clearTimeout` throws, the timer leaks and aborts a settled request. Three-line fix.

**11. `filterReachableUrls` name overstates the guarantee** — the function keeps URLs that are "not confidently broken" (405s, timeouts, and network errors all pass). Rename to `filterOutConfirmedBrokenUrls` so future callers aren't misled into trusting it as a liveness check.

**12. User-Agent reads as evasion** (`url-health-check.js:25`): mimicking Chrome with `Spacecat/1.0` buried after the UA may trip JA3 vs. UA heuristics at customer WAFs and obscures Adobe identity in forensics. Use `AdobeSpacecat-LLMErrorPages/1.0 (+https://www.adobe.com/)` instead.

**13. `log.warn("No suggested URLs...")` will flood post-PR** (`guidance-handler.js:117-119`): fires WARN per-row for both Mystique-supplied empty rows and HEAD-emptied rows. On a 50-link batch with 15 empties that's 15 WARNs per audit. Demote to `log.debug` or aggregate.

**14. Contract change for empty-suggestion rows is undocumented for downstream consumers** (`guidance-handler.js:184-204`): previously empty-suggestion rows were not persisted; now they are. Confirm with UI/AEM owners or add a release note. The spec in #6 should capture this.

**15. Branch hole: surviving URLs + missing aiRationale** (`guidance-handler.js:79`): `aiRationale = filtered.length === 0 ? '' : (link.aiRationale ?? '')` — the `??` fallback when URLs survive but Mystique sent no rationale is not exercised. One-line test covers it.

## Strengths

- Comments at the intentional behavior-change sites are excellent — the multi-line block at `guidance-handler.js:54-58` and `184-191` explains *why* the gate is removed and what invariant the HEAD-check guarantees. Exactly the kind of comment that earns its keep two years on.
- The 405-keep / network-error-keep / timeout-keep semantics are deliberate and well-documented. Fail-open on inconclusive is the right business choice.
- Renaming `brokenLinks` → `incomingBrokenLinks` and shadowing with the cleaned value is principled change discipline — the rest of the handler reads unchanged.
- Test coverage is genuinely good: optional-chaining branches, batch processing, order preservation, and the defensive `Array.isArray` guard are all exercised. The renamed test (`persists empty-suggestion rows to DB (Excel↔DB consistency — gate removed)`) explicitly acknowledges the behavior reversal — exemplary for future reviewers.
- No new dependencies; uses native `fetch`.
- N+1 rule (CLAUDE.md lines 215-217) preserved — `Suggestion.saveMany` still used in the guidance handler. The PR does not fall into the pattern CLAUDE.md explicitly warns against for this exact area.

---
*Reviewed with [Claude Code](https://claude.ai/code)*

MysticatBot (changes requested):
Hey @ymadobe,

**Verdict: Request changes** - solid revision that addresses all prior SSRF/concurrency/telemetry concerns, but the scheduler's error contract, documentation requirements, and one residual security gap need attention.
_Changes: Adds HEAD-based reachability filtering for Mystique-suggested URLs with SSRF protection and fixes Excel-DB drift by persisting empty-suggestion rows (6 files)._

## Must fix before merge

1. [Important] Scheduler silently drops URLs on unexpected rejection (violates fail-open contract) - `src/llm-error-pages/url-health-check.js:170` (details inline)
2. [Important] "Sync by hand" note needs canonicality declaration and follow-up reference - `src/support/url-safety.js:21` (details inline)
3. [Important] DNS rebinding TOCTOU residual risk undocumented in shared security module - `src/support/url-safety.js:153` (details inline)
4. [Important] Missing spec per CLAUDE.md rule: "Significant new functionality must be developed spec-first. Add a spec to /docs/specs/" - no spec file added for 366 LOC of new non-trivial functionality (custom scheduler, SSRF boundary, fail-open semantics). Precedent: `docs/specs/alt-text-url-fallback-chain.md`.
5. [Important] Missing ADR per CLAUDE.md rule: "Technical or architectural decisions with long-term impact must be recorded as ADRs in /docs/decisions/" - two decisions qualify: (a) consumer-side defence-in-depth vs fixing upstream Mystique, (b) reversing the empty-suggestion-row DB contract.
6. [Important] Per-host concurrency cap not actually asserted in test - `test/audits/llm-error-pages/url-health-check.test.js:157` (details inline)

<details>
<summary>Non-blocking (2): minor issues</summary>

- nit: Structured log embeds JSON inside a string template rather than passing as structured metadata - `src/llm-error-pages/guidance-handler.js:83`. If the logging framework supports a second arg for structured fields, prefer that for native field access in Coralogix/CloudWatch.
- nit: Inconsistent optional chaining on `log.warn` - `isUrlSafeToFetch` uses `log.warn?.()` but `resolvesToPublicAddress` uses `log.warn()` without the guard - `src/support/url-safety.js:60`. Latent paper cut if a minimal log object is ever passed.
</details>

- SSRF guard added (url-safety.js with scheme allowlist, IP literal check, DNS resolution)
- redirect:'manual' prevents redirect-based SSRF bypass
- tracingFetch used for observability
- Per-host concurrency cap implemented (HEAD_PER_HOST_CONCURRENCY=3 + runWithHostCap)
- Structured head-check-summary telemetry emitted on every invocation
- AbortError timeout path tested
- Test comment/behavior mismatch corrected
- clearTimeout positioned correctly (unconditional after try/catch)
- Function renamed to filterOutConfirmedBrokenUrls
- User-Agent set to AdobeSpacecat-LLMErrorPages/1.0
- Per-link empty log demoted from warn to debug
- aiRationale ?? '' fallback tested

src/llm-error-pages/url-health-check.js
```diff
@@ -0,0 +1,208 @@
...
+        state.inflightGlobal += 1;
+        inflightPerHost.set(host, (inflightPerHost.get(host) ?? 0) + 1);
+        Promise.resolve()
+          .then(() => taskFn(urls[idx], idx))
+          .then((value) => {
+            results[idx] = value;
+          })
+          .finally(() => onTaskDone(host));
+      }
+    };
```
MysticatBot, line 175: issue (blocking): **Scheduler silently drops URLs on unexpected rejection - violates fail-open contract.**

When `taskFn` rejects (e.g., an unexpected `TypeError` from a `tracingFetch` instrumentation bug, or a system-level error that escapes the inner catches), the `.then((value) => { results[idx] = value; })` path is skipped. The `results[idx]` slot stays `undefined`, which is falsy in `urls.filter((_, idx) => reachable[idx])`, so the URL is silently dropped.

The module's own JSDoc states: "Inconclusive results (timeouts, network errors, 405 Method Not Allowed) are kept." A rejection is the ultimate inconclusive result, yet the current code drops rather than keeps.

**Fix:** Add a `.catch` that defaults to keep:
```js
Promise.resolve()
  .then(() => taskFn(urls[idx], idx))
  .then((value) => { results[idx] = value; })
  .catch(() => { results[idx] = true; })  // fail-open: keep on unexpected error
  .finally(() => onTaskDone(host));
```

src/support/url-safety.js
```diff
@@ -0,0 +1,158 @@
...
+
+import dns from 'dns/promises';
+
+/**
+ * SSRF-related URL safety helpers. Lifted from src/site-detection/handler.js so
+ * additional audits (e.g. llm-error-pages HEAD probes) can apply the same
+ * private/loopback/link-local block list before fetching outbound URLs.
+ *
+ * NOTE: the site-detection module currently still uses its own private copies
+ * of these helpers; both versions are kept in sync by hand for now to avoid
```
MysticatBot: issue (blocking): **"Sync by hand" is a documented drift hazard for security-critical code.**

This module is in `src/support/` (shared), yet `src/site-detection/handler.js` maintains a private fork of the same helpers. The site-detection copy documents a known DNS-rebinding limitation with a tracked follow-up. When that fix lands, this copy will not get it.

**Fix (within this PR's scope):**
1. Add a comment here declaring this file as the canonical copy going forward.
2. Add a corresponding comment in `src/site-detection/handler.js` referencing this file as the target migration destination.
3. Reference a follow-up ticket (in the PR description) to migrate site-detection to import from `src/support/url-safety.js`.

src/support/url-safety.js
```diff
@@ -0,0 +1,158 @@
...
+    : rawHost;
+
+  // If the host is itself an IP literal, skip DNS and check directly. Otherwise
+  // a private IP in URL form would still pass resolvesToPublicAddress (because
+  // dns.lookup happily returns the literal back).
+  if (IP_LITERAL_HOSTNAME_REGEX.test(host)) {
+    if (isPrivateIP(host)) {
+      log.warn?.(`[url-safety] rejecting URL with private IP literal host: ${url}`);
+      return false;
+    }
```
MysticatBot, line 182: issue (blocking): **DNS rebinding TOCTOU gap is undocumented in a module designed for reuse.**

The DNS check resolves the hostname and confirms public addresses. Then `fetch()` performs its own resolution. An attacker controlling DNS with TTL=0 can serve a public IP on the first lookup (passes guard) and `169.254.169.254` on the second (when fetch connects).

**Mitigations in this environment:** IMDSv2 hop-limit=1, VPC egress rules, attacker needs Mystique data compromise. These are strong, but environmental - not inherent to the guard.

**Why it matters:** This module lives in `src/support/` with generic naming, clearly designed for reuse. A future caller in a less-constrained context (different Lambda, local dev, non-VPC deployment) inherits the gap without knowing.

**Fix:** Add a block comment above `isUrlSafeToFetch` documenting:
- The known TOCTOU window between DNS validation and fetch connection
- The environmental controls that mitigate it (IMDSv2, VPC)
- The ideal fix (pin resolved IP via undici lookup option) as a tracked follow-up

test/audits/llm-error-pages/url-health-check.test.js
```diff
@@ -0,0 +1,189 @@
...
+    const result = await filterOutConfirmedBrokenUrls(urls, log);
+    expect(result).to.have.length(12);
+    expect(fetchStub).to.have.callCount(12);
+  });
+
+  it('does not crash when log.debug is missing on the 405 path (optional chaining branch)', async () => {
+    fetchStub.resolves({ status: 405 });
+    const minimalLog = {}; // no .debug
+    const result = await filterOutConfirmedBrokenUrls(['https://a.example/x'], minimalLog);
+    expect(result).to.deep.equal(['https://a.example/x']);
```
MysticatBot, line 202: issue (blocking): **Per-host concurrency cap is not asserted - a `Promise.all` refactor would still pass this test.**

The test creates 12 same-host URLs, asserts they all complete and fetch is called 12 times. But the stated invariant (at most 3 in-flight per host) is never observed. This was flagged in the prior review and the test was added, but the assertion that proves the cap is still missing.

**Fix:**
```js
it('enforces per-host concurrency cap of 3', async () => {
  let inflight = 0, maxInflight = 0;
  fetchStub.callsFake(async () => {
    inflight += 1;
    maxInflight = Math.max(maxInflight, inflight);
    await new Promise((r) => setImmediate(r));
    inflight -= 1;
    return { status: 200 };
  });
  const urls = Array.from({ length: 12 }, (_, i) => `https://same.example/${i}`);
  await filterOutConfirmedBrokenUrls(urls, log);
  expect(maxInflight).to.be.at.most(3);
});
```

MysticatBot (commented):
Hey @ymadobe,

**Verdict: Comment** - all prior findings addressed, no blocking issues remain.
_Changes: Adds HEAD-based reachability filtering for Mystique-suggested URLs with SSRF protection, per-host concurrency caps, and fixes Excel-DB drift by persisting empty-suggestion rows (9 files)._

<details>

- nit: Inconsistent optional chaining on `log.warn` - `isUrlSafeToFetch` uses `log.warn?.()` but `resolvesToPublicAddress` uses `log.warn()` without the guard. Either both should guard or neither should (caller contract says pass a real logger). - `src/support/url-safety.js:60`
- nit: Structured log embeds JSON inside a template string (`head-check-summary ${JSON.stringify(...)}`) rather than passing as structured metadata. If the logging framework supports a second arg for structured fields, prefer that for native field access in Coralogix. - `src/llm-error-pages/guidance-handler.js:88`
- nit: Global concurrency cap (HEAD_CONCURRENCY=10) is not observationally asserted in tests. The ">10 URLs" test uses 25 distinct hosts and asserts `callCount === 25`, but a `Promise.all` refactor with no concurrency limit would also pass. The per-host cap test demonstrates the correct pattern (inflight counter + setImmediate yield + `at.most` assertion) - apply the same with 25 hosts and assert `maxInflight <= 10`. - `test/audits/llm-error-pages/url-health-check.test.js:157`
- suggestion: Deduplication behavior across multiple broken links is untested. The handler deduplicates suggested URLs into a Set before HEAD-checking, then maps results back per-link. A test with two broken links sharing the same suggested URL would document this behavior. - `src/llm-error-pages/guidance-handler.js:68`
- suggestion: `runWithHostCap` is a generic per-host + global concurrency scheduler with no domain logic. When the follow-up consolidation PR happens, consider moving it to `src/support/` for reuse by other audits that need throttled outbound probes. - `src/llm-error-pages/url-health-check.js:108`

</details>

- Scheduler fail-open on unexpected rejection (`.catch` belt-and-suspenders added)
- "Sync by hand" note replaced with canonicality declaration and follow-up reference
- DNS rebinding TOCTOU gap documented inline with mitigations
- Spec added at `docs/specs/llm-error-pages-head-check-and-db-consistency.md`
- ADR added at `docs/decisions/001-llm-error-pages-consumer-side-url-validation.md`
- Per-host concurrency cap now observationally asserted with `maxInflight` counter

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 39s | <b>Cost:</b> $4.55 | <b>Commit:</b> <code>8188d93d3c9dde4f497ef36fc87594751edf4f3c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @ymadobe,

**Verdict: Approve** - all prior nits addressed, no new concerns.
_Changes: Adds HEAD-based reachability filtering for Mystique-suggested URLs with SSRF protection, per-host concurrency caps, and fixes Excel-DB drift by persisting empty-suggestion rows (9 files)._

- Structured log now passes metadata as second arg (Coralogix native field indexing)
- Optional chaining removed from `log.warn` in url-safety.js (caller contract tightened - fail-closed on broken logger is the correct security posture)
- Global concurrency cap observationally asserted with `maxInflight` counter
- URL deduplication across broken links tested

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 30s | <b>Cost:</b> $2.47 | <b>Commit:</b> <code>fb10c9875a8950009585238d0599c09a1ce41661</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2642, opened 2026-06-11
feat(cdn-analysis): retry transient Athena failures via delayed re-queue

akshaymagapu:
## Problem

`cdn-logs-analysis` intermittently fails with **transient** Athena errors and aborts the whole audit:

```
cdn-logs-analysis audit failed for site <id>. Reason: Query exhausted resources at this scale factor.
```

This error (and throttling / internal / worker-node errors) is usually driven by momentary Athena cluster load and **succeeds on a later run**. Today a single blip fails the audit and fires an alert.

## Change

The runner now catches a **retryable** Athena failure and **re-enqueues the same `auditContext` as a delayed sub-audit**, instead of failing outright:

- **Delay 850s** (under the SQS 900s cap) so the Athena cluster has time to recover before the retry runs in a fresh Lambda.
- **Preserves the original `auditContext`**, only adding/bumping `retryCount` — no invented fields.
- **Soft result** on the current invocation so SQS does not *also* redeliver it (no double-processing).
- **Bounded to 2 retries** (`MAX_ANALYSIS_RETRIES`). After that it logs and rethrows — so the existing `("cdn-logs-analysis" "failed")` alert still fires on a genuinely stuck site, just after retries are exhausted.

### Retryable vs not

Matched against AWS-documented **System-category** error signatures (`exhausted resources at this scale factor`, `internal error`, `throttl`, `too many requests`, `rate exceeded`, `slow down`, `service unavailable`, `worker node`, …). **User errors** (syntax, missing table/column, access denied, bad data) are deliberately excluded — a retry just fails identically.

> Note: Athena exposes a structured `Retryable` boolean on `QueryExecution.Status.AthenaError`. Keying off that is more robust than string matching, but it requires surfacing the field from `spacecat-shared-athena-client` (separate package + release). Left as a follow-up.

### Kill switch

Gated by env **`CDN_ANALYSIS_RETRY_ENABLED`** (default `true`). Set to `false` to instantly revert to the prior fail-fast behavior — the runner skips the retry branch and goes straight to log + throw.

## Behavior / monitoring impact

- Transient blips that recover are **no longer** logged as failures (intentional noise reduction). Only the final post-retry failure throws and hits the dashboard.
- A hard failure now surfaces ~28 min later worst-case (2 × 850s) instead of immediately — the tradeoff for giving Athena recovery room.
- No new alert string needed: the existing query still catches the final failure via the framework's "audit failed" log.

## Safety

- Bounded — `retryCount` propagated in the re-queued message; no infinite loop.
- No downstream impact — cdn-logs-analysis uses the no-op default message sender; the persisted soft result (`{ retryScheduled: true }`) mirrors existing `{ skipped }` / `{ scanAndTriggerOnly }` shapes.
- Graceful degradation — if the re-enqueue itself fails, it throws out of the runner → identical to today.

## Tests

6 new tests (retry scheduling, retryCount increment, give-up-and-rethrow, non-retryable pass-through, singular/plural log wording, kill switch). `src/cdn-analysis/handler.js` remains at **100%** coverage; lint clean.

### DIFF
--- a/src/cdn-analysis/handler.js
+++ b/src/cdn-analysis/handler.js
@@ -41,7 +41,31 @@ const SQS_MAX_DELAY_SECONDS = 900;
 const CDN_LOGS_REPORT_DELAY_SECONDS = 800;
 const CDN_LOGS_REPORT_STAGGER_SECONDS = 30;
 
+// Re-enqueue the window as a delayed sub-audit on transient Athena failures.
+// Disable with env CDN_ANALYSIS_RETRY_ENABLED=false. Delay kept under the 900s SQS cap.
+const MAX_ANALYSIS_RETRIES = 2;
+const ANALYSIS_RETRY_DELAY_SECONDS = 850;
+// Transient (System-category) Athena failures worth retrying; user errors excluded.
+const RETRYABLE_ATHENA_ERROR_PATTERNS = [
+  'exhausted resources at this scale factor',
+  'internal error',
+  'internalerror',
+  'resource limit exceeded',
+  'throttl',
+  'too many requests',
+  'rate exceeded',
+  'slow down',
+  'slowdown',
+  'service unavailable',
+  'worker node',
+];
+
 const pad2 = (n) => String(n).padStart(2, '0');
+
+function isRetryableAthenaError(message = '') {
+  const normalized = message.toLowerCase();
+  return RETRYABLE_ATHENA_ERROR_PATTERNS.some((pattern) => normalized.includes(pattern));
+}
 
 function isValidAuditContext(auditContext) {
   if (!isNonEmptyObject(auditContext)) {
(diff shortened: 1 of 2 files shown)


src/cdn-analysis/handler.js
```diff
@@ -528,8 +552,53 @@ export async function processCdnLogs(auditUrl, context, site, auditContext) {
...
-  return processCdnLogs(auditUrl, context, site, auditContext);
+  const { log, env } = context;
+  try {
+    return await processCdnLogs(auditUrl, context, site, auditContext);
+  } catch (e) {
+    const retryCount = Number(auditContext?.retryCount) || 0;
+    const retriesEnabled = String(env?.CDN_ANALYSIS_RETRY_ENABLED ?? 'true').toLowerCase() === 'true';
+
+    if (retriesEnabled && isRetryableAthenaError(e.message) && retryCount < MAX_ANALYSIS_RETRIES) {
+      await requeueAnalysisRetry(context, site, auditContext, retryCount + 1);
```
MysticatBot: issue (blocking): If `requeueAnalysisRetry` throws (e.g., `Configuration.findLatest()` fails or `sqs.sendMessage` rejects), the SQS/DB error propagates instead of the original Athena error `e`. The operator sees a misleading failure reason in logs/alerts - the Athena error that triggered the retry is lost.

The PR description states "if the re-enqueue itself fails, it throws out of the runner - identical to today" but the error identity changes: today the Athena error propagates; after this change, a *different* error propagates, which will not match the existing `("cdn-logs-analysis" "failed")` alert pattern expecting Athena error text.

Fix: wrap the requeue call so the original error is preserved on failure:

```js
try {
  await requeueAnalysisRetry(context, site, auditContext, retryCount + 1);
} catch (requeueErr) {
  log.error(`cdn-logs-analysis requeue failed for siteId=${site.getId()}: ${requeueErr.message}`);
  throw e; // preserve the original Athena error
}
```

MysticatBot (changes requested):
Hey @akshaymagapu,

**Verdict: Request changes** - one blocking concern around error propagation when the requeue mechanism itself fails.
_Changes: Adds a bounded retry-via-SQS-requeue for transient Athena failures in cdn-logs-analysis, with kill switch and soft-result to avoid double-processing (2 files)._

## Must fix before merge
1. [Important] Requeue failure masks the original Athena error - `src/cdn-analysis/handler.js:584` (details inline)

<details>

- nit: String-matching for retryability is acknowledged as transitional (PR description notes the structured `Retryable` boolean follow-up) but the code has no pointer to that follow-up - a one-line comment or TODO marking this list as a bridge pattern would prevent the approach being cargo-culted into other audit handlers - `src/cdn-analysis/handler.js:47`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 50s | <b>Cost:</b> $2.57 | <b>Commit:</b> <code>848d5c4f78694de0f9150fd1808016bc5274a352</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (commented):
Hey @akshaymagapu,

**Verdict: Comment** - both prior findings addressed; two edge-case nits noted, neither blocking.
_Changes: Adds bounded retry-via-SQS-requeue for transient Athena failures in cdn-logs-analysis, with structured retryable flag, kill switch, and soft-result to prevent double-processing (2 files)._

<details>

- nit: `isRetryableAthenaError` accesses `error.message.toLowerCase()` without guarding against errors that lack a `message` property - if a non-Error value is thrown from `processCdnLogs`, this produces a TypeError that masks the original failure. In practice AWS SDK errors always carry `.message`, so low risk. Guard with `const msg = error.message || '';` - `src/cdn-analysis/handler.js:39`
- nit: `Number(auditContext?.retryCount) || 0` passes negative values through (e.g. `-5 || 0` yields `-5`), allowing more retries than intended. Requires a malformed SQS message on an IAM-gated queue, so negligible practical risk. `Math.max(0, Number(auditContext?.retryCount) || 0)` would clamp - `src/cdn-analysis/handler.js:76`
</details>

- Requeue failure now preserves original Athena error (logs SQS failure, rethrows original)
- String-matching for retryability now documented as fallback-only; structured `error.retryable` flag checked first

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 35s | <b>Cost:</b> $2.26 | <b>Commit:</b> <code>6bff4ff7bba7f54bef80ed4fea237c6a1976009e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - all prior concerns addressed cleanly; retry mechanism is well-bounded and properly tested.
_Changes: Adds bounded retry-via-SQS-requeue for transient Athena failures in cdn-logs-analysis, with structured retryable flag, kill switch, negative-retryCount clamping, and original-error preservation on requeue failure (2 files)._

- Requeue failure now preserves original Athena error (try/catch around requeueAnalysisRetry, rethrows original)
- String-matching documented as fallback; structured `error.retryable` boolean checked first
- `String(error.message)` guards against undefined message property
- `Math.max(0, ...)` clamps negative retryCount values

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 31s | <b>Cost:</b> $2.89 | <b>Commit:</b> <code>a307a149ac424273663527b81d2ecc38acf4751b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2644, opened 2026-06-11
fix(prerender): use null for valuable when AI summary is absent

anuj-adobe:
## Summary

- When Mystique sends no valid AI summary, the audit worker now stores \`valuable: null\` instead of defaulting to \`true\`
- \`null\` accurately signals "undetermined" — the suggestion exists but valuability has not been determined (AI analysis could not be completed)
- Also fixes \`isValuable\` fallback from \`true\` to \`null\` for non-boolean inputs
- Companion Mystique PR: https://git.corp.adobe.com/experience-platform/mystique/pull/2656

## Context

Previously, \`guidance-handler.js\` had two places that defaulted to \`true\` when there was no signal:

1. \`const isValuable = typeof valuable === 'boolean' ? valuable : true\` — when \`valuable\` is \`null\` (non-boolean), this asserted \`true\`
2. \`currentData.valuable ?? true\` — when a new suggestion had no existing value and no valid AI summary, this stored \`true\`

Both now use \`null\` as the fallback, meaning "we don't know yet" rather than asserting the suggestion is valuable without evidence.

## Backward compatibility

- **Existing suggestions in DB** — unaffected; \`??\` preserves the existing stored value
- **Audit worker before Mystique** — old Mystique sends \`valuable: true\` (boolean); \`isValuable = true\`, no change for successful summaries. New suggestions with failed summaries get \`null\` instead of \`true\` (desired)
- **Mystique before Audit Worker** — old worker's \`?? true\` fallback kicks in; degrades to old behavior harmlessly
- **Downstream consumers** — \`if (valuable)\` and \`if (valuable === true)\` correctly exclude \`null\`; \`if (!valuable)\` now matches both \`false\` and \`null\`, which is semantically correct ("not confirmed valuable")

## Changed file

- \`src/prerender/guidance-handler.js\`: \`isValuable\` fallback \`true\` → \`null\`; \`currentData.valuable ?? true\` → \`?? null\`

## Test plan

- [ ] Confirm existing tests pass: \`npm test\`
- [ ] Verify suggestions with no AI summary store \`valuable: null\` in the DB
- [ ] Verify suggestions with a valid AI summary are unaffected (\`valuable: true/false\` set by Mystique)


src/prerender/guidance-handler.js
```diff
@@ -166,7 +166,7 @@ export default async function handler(message, context) {
 
       // Track if AI summary is meaningful
       const hasValidAiSummary = aiSummary && aiSummary.toLowerCase() !== 'not available';
-      const isValuable = typeof valuable === 'boolean' ? valuable : true;
+      const isValuable = typeof valuable === 'boolean' ? valuable : null;
```
MysticatBot, line 169: issue (blocking): The test at `test/audits/prerender/guidance-handler.test.js:208` ("should handle valuable flag correctly (defaults to true if not provided)") asserts `expect(updatedData.valuable).to.equal(true)` at line 239. With this change, the assertion will fail since the fallback is now `null`.

**Why it matters:** CI enforces 100% coverage with strict assertions. The test documents the intended behavior - it should reflect the new semantics so future maintainers understand the `null` contract.

**Fix:** Update the test name to reflect the new default (e.g. "defaults to null if not provided") and change the assertion to `.to.equal(null)`.

MysticatBot (changes requested):
Hey @anuj-adobe,

**Verdict: Request changes** - one test asserts the old default and needs updating.
_Changes: Changes the `valuable` field default from `true` to `null` when AI summary is absent, correctly signaling "undetermined" instead of falsely asserting valuable (1 file)._

## Must fix before merge

1. [Important] Existing test asserts old `true` default, will fail with new `null` behavior - `test/audits/prerender/guidance-handler.test.js:208-239` (details inline)


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 56s | <b>Cost:</b> $2.29 | <b>Commit:</b> <code>d7ad8fa4d3aa9c5933da3dd5d1913124567eff08</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2646, opened 2026-06-11
fix: remove unreachable `payload.checks` narrowing branch in preflight audit

radhikagpt1208:
## Summary

Removes the unreachable `payload.checks` narrowing branch from the preflight audit handler. The handler now resolves the enabled checks purely from per-site configuration and persists the resolved set into the job metadata for the MFE to consume.

## Background

This is a follow-up cleanup to the preflight changes in [spacecat-audit-worker#1820](https://github.com/adobe/spacecat-audit-worker/pull/1820/changes#diff-56d6bf732d419d5caa8e25f33ec139ba9c1d6ab7e1dcdab550db87e0adfe9203), which introduced per-site `${audit}-preflight` resolution plus an optional narrowing of the resolved set against any `payload.checks` already on the job:

```js
let enabledChecks = siteEnabled;
const currentPayload = jobEntity.getMetadata()?.payload;
if (currentPayload?.checks?.length) {
  enabledChecks = currentPayload.checks.filter((c) => siteEnabled.includes(c));
}
```
## Why the narrowing branch is dead code
The intersection branch only fires when an inbound job already carries payload.checks. In the real flow that never happens:
- The MFE never sends checks. It submits preflight via POST /preflight/jobs with only { urls, step, siteId }. The API controller (createPreflightJob) creates the AsyncJob with payload = { siteId, urls, step, enableAuthentication } — no checks.
- The worker is the first writer of checks. The handler computes the enabled set from global config and writes it into the job metadata, which the MFE then reads back from GET /preflight/jobs/{jobId} (metadata.payload.checks).
- IDENTIFY and SUGGEST are two separate jobs. The MFE issues 2 independent submitPreflightJob calls (one per step), each creating its own AsyncJob with its own jobId and fresh payload. So there is no cross-step metadata inheritance — the SUGGEST job does not see the IDENTIFY job's persisted checks.
- Because no inbound job ever has payload.checks, `currentPayload?.checks?.length` is always falsy and the narrowing branch is never executed.

## Impact
No behavior change in production — the removed branch was unreachable.
The MFE continues to receive the resolved check list via the job-status response, exactly as before.

Thanks for contributing!


GeezerPelletier (approved):
LGTM — the unreachability argument holds. `createPreflightJob` is the only creator and never sets `checks`; this handler is the first writer of `payload.checks`; IDENTIFY and SUGGEST are independent AsyncJobs so SUGGEST never inherits IDENTIFY-time `checks`. The original "SUGGEST inherits IDENTIFY-time decision" comment described an intent the MFE-driven flow never realized.

On worker retry of the same `jobId`, the new overwrite-with-fresh behavior is arguably more correct than the old intersection — an intersection would silently drop a check newly enabled between attempts. Good catch.

PR #2650, opened 2026-06-12
fix(prerender): prevent duplicate domain-wide suggestions via buildKey symmetry

ssilare-adobe:
## Summary

Fixes a bug where the prerender audit created a new domain-wide aggregate suggestion on every audit run when the existing one was in a non-preservable state (OUTDATED, REJECTED, etc.), resulting in an ever-growing list of duplicate domain-wide suggestions that could never be cleaned up.

## Root Cause

`buildKey` inside `processOpportunityAndSuggestions` was asymmetric:

- **Incoming** domain-wide item (from `allSuggestions`) has a `.key` field → `buildKey` returned `'domain-wide-aggregate|prerender'`
- **Stored** domain-wide suggestion (from `existing.getData()`) has no `.key` — only `isDomainWide: true` → `buildKey` fell through to `normalizePathnameWithQuery(data.url)` → returned `/*%20(All%20Domain%20URLs)`

These two keys never matched, so `syncSuggestions` always treated the incoming domain-wide as a brand-new suggestion and inserted a duplicate. The `isDomainWide` guard in `handleOutdatedSuggestions` then prevented the old copies from ever being cleaned up.

## Fix

Extended `buildKey` to recognise `isDomainWide: true` in stored data and return the same constant key as the incoming item:

```js
// Before
const buildKey = (data) => {
  if (data.key) return data.key;
  return normalizePathnameWithQuery(data.url);
};

// After
const buildKey = (data) => {
  if (data.key === DOMAIN_WIDE_SUGGESTION_KEY || data.isDomainWide) {
    return DOMAIN_WIDE_SUGGESTION_KEY;
  }
  return normalizePathnameWithQuery(data.url);
};
```

Now `syncSuggestions` correctly matches the existing domain-wide record and updates it in place instead of inserting a new one.

## Changes

- **`src/prerender/handler.js`** — fix `buildKey` to return `DOMAIN_WIDE_SUGGESTION_KEY` for both incoming (`.key`) and stored (`isDomainWide:true`) domain-wide data
- **`test/audits/prerender/sync-suggestions.test.js`** — update `buildKey` mirror constant; add Case 6 that proves no duplicate is created when existing stored domain-wide has `isDomainWide:true`
- **`test/audits/prerender/handler.test.js`** — update assertions in "OUTDATED" and "not preservable" tests to verify `buildKey` symmetry; fix hardcoded `'domain-wide-key'` probe to use the real constant

## Testing

- All 386 prerender tests pass (`handler`, `sync-suggestions`, `ai-only-mode`, `guidance-handler`, `utils`)
- New test Case 6 in `sync-suggestions.test.js` directly reproduces the bug scenario (existing `isDomainWide:true` in OUTDATED state) and asserts the suggestion is updated in place with no `addSuggestions` call

## Related Issues

None

### DIFF
--- a/test/audits/prerender/handler.test.js
+++ b/test/audits/prerender/handler.test.js
@@ -3443,8 +3443,9 @@ describe('Prerender Audit', () => {
         expect(buildKey({ url: 'https://example.com/some/page?filter=iphone' })).to.equal('/some/page?filter=iphone');
         expect(buildKey({ url: 'https://www.example.com/some/page?filter=mac' })).to.equal('/some/page?filter=mac');
 
-        // Domain-wide suggestions (with a `key` field) pass through unchanged
-        expect(buildKey({ key: 'domain-wide-key' })).to.equal('domain-wide-key');
+        // Domain-wide suggestions return the constant key regardless of how the data is stored
+        expect(buildKey({ key: 'domain-wide-aggregate|prerender' })).to.equal('domain-wide-aggregate|prerender');
+        expect(buildKey({ isDomainWide: true, url: 'https://example.com/* (All Domain URLs)' })).to.equal('domain-wide-aggregate|prerender');
       });
 
       it('should normalize scrapedUrlsSet to pathname so domain-shifted suggestions are correctly outdated', async () => {
@@ -3713,7 +3714,11 @@ describe('Prerender Audit', () => {
         expect(domainWideSuggestion).to.be.undefined;
       });
 
-      it('should create new domain-wide suggestion when existing one is OUTDATED without deployed flags', async () => {
+      it('should pass domain-wide suggestion to syncSuggestions when existing one is OUTDATED — buildKey symmetry prevents duplicate', async () => {
+        // When the existing domain-wide is not preservable (OUTDATED, no edgeDeployed),
+        // the new domain-wide aggregate is always included in newData. The fixed buildKey
+        // returns DOMAIN_WIDE_SUGGESTION_KEY for stored data with isDomainWide:true, so
+        // syncSuggestions matches the existing record and updates it instead of creating a duplicate.
         const existingDomainWideSuggestion = {
           getId: () => 'existing-domain-wide-id',
           getStatus: () => 'OUTDATED',
@@ -3764,6 +3769,12 @@ describe('Prerender Audit', () => {
         const domainWideSuggestion = syncArgs.newData.find((s) => s.key);
         expect(domainWideSuggestion).to.exist;
         expect(domainWideSuggestion.key).to.equal('domain-wide-aggregate|prerender');
+
+        // Verify buildKey treats both new (data.key) and stored (isDomainWide) domain-wide
+        // records as the same key — the core fix preventing duplicates
+        const { buildKey } = syncArgs;
+        expect(buildKey({ key: 'domain-wide-aggregate|prerender' })).to.equal('domain-wide-aggregate|prerender');
+        expect(buildKey({ isDomainWide: true, url: 'https://example.com/* (All Domain URLs)' })).to.equal('domain-wide-aggregate|prerender');
       });
 
       it('should not detect domain-wide suggestion by pathPattern alone (requires isDomainWide flag)', async () => {
@@ -8051,7 +8062,10 @@ describe('Prerender Audit', () => {
       );
     });
 
-    it('should create new domain-wide suggestion when existing ones are not preservable', async () => {
+    it('should pass domain-wide aggregate to syncSuggestions when existing one is not preservable — buildKey match prevents duplicate', async () => {
+      // The domain-wide item is always present in newData when no preservable one exists.
+      // The fixed buildKey returns DOMAIN_WIDE_SUGGESTION_KEY for stored isDomainWide:true data,
+      // so syncSuggestions matches and updates the existing record rather than inserting a new one.
       const existingDomainWideSuggestion = {
         getStatus: () => 'OUTDATED',
         getId: () => 'outdated-domain-wide-id',
@@ -8102,9 +8116,13 @@ describe('Prerender Audit', () => {
       await mockHandler.processOpportunityAndSuggestions('https://example.com', auditData, context);
 
       expect(syncSuggestionsStub).to.have.been.calledOnce;
-      const { newData } = syncSuggestionsStub.getCall(0).args[0];
+      const { newData, buildKey } = syncSuggestionsStub.getCall(0).args[0];
       const domainWide = newData.find((item) => item.key);
       expect(domainWide).to.exist;
+
+      // buildKey must return the same constant for both new and stored domain-wide data
+      expect(buildKey({ key: 'domain-wide-aggregate|prerender' })).to.equal('domain-wide-aggregate|prerender');
+      expect(buildKey({ isDomainWide: true })).to.equal('domain-wide-aggregate|prerender');
     });
 
     it('should include suggestionId in auditRunCandidates from saved suggestions', async () => {
@@ -8990,7 +9008,10 @@ describe('Prerender Audit', () => {
       );
     });
 
-    it('should create new domain-wide suggestion when existing ones are not preservable', async () => {
+    it('should pass domain-wide aggregate to syncSuggestions when existing one is not preservable — buildKey match prevents duplicate', async () => {
+      // The domain-wide item is always present in newData when no preservable one exists.
+      // The fixed buildKey returns DOMAIN_WIDE_SUGGESTION_KEY for stored isDomainWide:true data,
+      // so syncSuggestions matches and updates the existing record rather than inserting a new one.
       const existingDomainWideSuggestion = {
         getStatus: () => 'OUTDATED',
         getId: () => 'outdated-domain-wide-id',
@@ -9041,9 +9062,13 @@ describe('Prerender Audit', () => {
       await mockHandler.processOpportunityAndSuggestions('https://example.com', auditData, context);
 
       expect(syncSuggestionsStub).to.have.been.calledOnce;
-      const { newData } = syncSuggestionsStub.getCall(0).args[0];
+      const { newData, buildKey } = syncSuggestionsStub.getCall(0).args[0];
       const domainWide = newData.find((item) => item.key);
       expect(domainWide).to.exist;
+
+      // buildKey must return the same constant for both new and stored domain-wide data
+      expect(buildKey({ key: 'domain-wide-aggregate|prerender' })).to.equal('domain-wide-aggregate|prerender');
+      expect(buildKey({ isDomainWide: true })).to.equal('domain-wide-aggregate|prerender');
     });
   });
 
--- a/test/audits/prerender/sync-suggestions.test.js
+++ b/test/audits/prerender/sync-suggestions.test.js
@@ -19,9 +19,12 @@ import { toPathname } from '../../../src/prerender/utils/utils.js';
 
 use(sinonChai);
 
-// Mirrors the buildKey in processOpportunityAndSuggestions
+// Mirrors the buildKey in processOpportunityAndSuggestions.
+// Domain-wide suggestions must return the same constant regardless of whether data comes
+// from allSuggestions (has .key) or from an existing DB record (has isDomainWide:true).
+const DOMAIN_WIDE_KEY = 'domain-wide-aggregate|prerender';
 const buildKey = (data) => {
-  if (data.key) return data.key;
+  if (data.key === DOMAIN_WIDE_KEY || data.isDomainWide) return DOMAIN_WIDE_KEY;
   return toPathname(data.url);
 };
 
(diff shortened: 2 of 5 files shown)


anuj-adobe (commented):
## Review: test coverage for  symmetry fix

The fix itself is correct. Three test gaps to address before merge.

test/audits/prerender/sync-suggestions.test.js
```diff
@@ -19,9 +19,12 @@ import { toPathname } from '../../../src/prerender/utils/utils.js';
...
-// Mirrors the buildKey in processOpportunityAndSuggestions
+// Mirrors the buildKey in processOpportunityAndSuggestions.
+// Domain-wide suggestions must return the same constant regardless of whether data comes
+// from allSuggestions (has .key) or from an existing DB record (has isDomainWide:true).
+const DOMAIN_WIDE_KEY = 'domain-wide-aggregate|prerender';
 const buildKey = (data) => {
-  if (data.key) return data.key;
+  if (data.key === DOMAIN_WIDE_KEY || data.isDomainWide) return DOMAIN_WIDE_KEY;
   return toPathname(data.url);
 };
```
anuj-adobe: **Drift risk: this is a hand-copied mirror of the production `buildKey`, not the real function.**

If someone changes `buildKey` in `handler.js` (adds a branch, renames the constant, etc.) this mirror stays silently stale and Case 6 keeps passing against the wrong logic.

Suggested fix — export `buildKey` from `handler.js` and import it here:

```js
// handler.js — add named export
export const buildDomainWideSuggestionKey = (data) => {
  if (data.key === DOMAIN_WIDE_SUGGESTION_KEY || data.isDomainWide) return DOMAIN_WIDE_SUGGESTION_KEY;
  return normalizePathnameWithQuery(data.url);
};
```

```js
// sync-suggestions.test.js
import { buildDomainWideSuggestionKey as buildKey } from '../../../src/prerender/handler.js';
```

Alternatively, extract `buildKey` to `src/prerender/utils/utils.js` alongside `toPathname` — it already lives there for other URL helpers.

test/audits/prerender/sync-suggestions.test.js
```diff
@@ -219,6 +222,52 @@ describe('Prerender syncSuggestions integration', () => {
...
+      mapNewSuggestion: (d) => ({ data: d.data ?? d }),
+      existingSuggestions: [existingDomainWide],
+    });
+
+    // Existing was updated in place — no duplicate created
+    expect(saveMany).to.have.been.calledOnce;
+    expect(addSuggestions).to.not.have.been.called;
+    // Fresh aggregated data was applied
+    expect(existingDomainWide.getData().contentGainRatio).to.equal(1.2);
+    expect(existingDomainWide.getData().wordCountBefore).to.equal(15);
```
anuj-adobe, line 260: **Missing: test for the exact production scenario — multiple concurrent NEW domain-wide duplicates from a first-run race.**

The production incident for `dickssportinggoods.com` had **5 NEW** suggestions created simultaneously by a race condition: 5 concurrent Lambda invocations all called `findPreservableDomainWideSuggestion` before any had committed, saw 0 existing domain-wides (brand-new opportunity), and all reached `addSuggestions` simultaneously.

After PR #2650, for an **OUTDATED** existing domain-wide the race is also fixed — `buildKey` now matches so all threads take the `toUpdate` (idempotent UPDATE) path instead of `toCreate` (INSERT). The only remaining gap is a truly-zero-existing state (first ever audit run for a new opportunity), which is a narrow window. A follow-up DB unique partial index on `(opportunity_id) WHERE is_domain_wide = true` in mysticat-data-service would close that completely.

What's missing is a test verifying that once N NEW duplicates already exist from such a race, subsequent single-threaded runs don't create a (N+1)th:

```js
it('Case 6b — multiple NEW domain-wide duplicates from a prior race do not accumulate further on subsequent runs', async () => {
  const makeNewDomainWide = () => makeSuggestion({
    isDomainWide: true,
    url: 'https://example.com/* (All Domain URLs)',
    contentGainRatio: 1.0,
  }, 'NEW');

  const dup1 = makeNewDomainWide();
  const dup2 = makeNewDomainWide();
  const dup3 = makeNewDomainWide();

  // Handler found dup1 via findPreservableDomainWideSuggestion and excluded
  // domain-wide from newData — so syncSuggestions receives no domain-wide item.
  await syncSuggestions({
    context, opportunity,
    newData: [],
    buildKey, mergeDataFunction,
    mapNewSuggestion: (d) => ({ data: d.data ?? d }),
    existingSuggestions: [dup1, dup2, dup3],
  });

  // Nothing created, nothing outdated — all three stay NEW
  expect(addSuggestions).to.not.have.been.called;
  expect(bulkUpdateStatus).to.not.have.been.called;
  expect(saveMany).to.not.have.been.called;
});
```

test/audits/prerender/handler.test.js
```diff
@@ -3764,6 +3769,12 @@ describe('Prerender Audit', () => {
         const domainWideSuggestion = syncArgs.newData.find((s) => s.key);
         expect(domainWideSuggestion).to.exist;
         expect(domainWideSuggestion.key).to.equal('domain-wide-aggregate|prerender');
+
+        // Verify buildKey treats both new (data.key) and stored (isDomainWide) domain-wide
+        // records as the same key — the core fix preventing duplicates
+        const { buildKey } = syncArgs;
+        expect(buildKey({ key: 'domain-wide-aggregate|prerender' })).to.equal('domain-wide-aggregate|prerender');
+        expect(buildKey({ isDomainWide: true, url: 'https://example.com/* (All Domain URLs)' })).to.equal('domain-wide-aggregate|prerender');
       });
```
anuj-adobe, line 3778: **This test verifies the contract passed to `syncSuggestions` but mocks it out entirely.**

That means it can't catch a bug inside `syncSuggestions` itself. The full chain (real `handler.js` `buildKey` + real `syncSuggestions`) is only tested via `sync-suggestions.test.js` Case 6, which uses a local mirror of `buildKey` (see comment there).

The two tests complement each other, but the split creates a seam. If the `buildKey` mirror drifts, `sync-suggestions.test.js` stays green against wrong logic while this test verifies the right function never reaches real `syncSuggestions`.

This is acceptable as-is if the mirror drift issue is fixed (exporting the real `buildKey`). If the export approach is taken, Case 6 then exercises the actual production function end-to-end and this comment can be removed.

test/audits/prerender/sync-suggestions.test.js
ssilare-adobe: Fixed in c58f447d. Extracted the key-building logic to shared modules:

- `DOMAIN_WIDE_SUGGESTION_KEY` → `src/prerender/utils/constants.js`
- `normalizePathnameWithQuery` + `buildSuggestionKey` → `src/prerender/utils/utils.js`

The test now imports the real production function:
```js
import { buildSuggestionKey as buildKey } from '../../../src/prerender/utils/utils.js';
```

Went with `utils.js` over `handler.js` to avoid pulling in heavy `handler.js` deps (`@aws-sdk/client-s3`, etc.) in the unit test.

test/audits/prerender/sync-suggestions.test.js
ssilare-adobe: Added Case 6b in c58f447d — creates 3 OUTDATED domain-wide duplicates (mirroring the production incident) and asserts:

1. All three are matched and data-refreshed via `saveMany` (same `buildSuggestionKey` for all)
2. No new suggestion created (`addSuggestions` not called)
3. `bulkUpdateStatus` is never called with `OUTDATED` — the `isDomainWide` guard in `handleOutdatedSuggestions` protects them

If the guard is accidentally removed in a future PR, assertion 3 will catch it.

test/audits/prerender/handler.test.js
ssilare-adobe: Resolved by the export approach in c58f447d. Both test files now use the same real `buildSuggestionKey` from `src/prerender/utils/utils.js`:

- **`sync-suggestions.test.js`** (Cases 6 & 6b): imports `buildSuggestionKey` and exercises the full chain with real `syncSuggestions` — end-to-end coverage of the dedup logic.
- **`handler.test.js`**: verifies the correct function reference is passed to `syncSuggestions` via the mocked call's `buildKey` arg.

No mirror to drift anymore — a change to `buildSuggestionKey` in `utils.js` is immediately tested by both suites.

test/audits/prerender/sync-suggestions.test.js
anuj-adobe: Case 6b still doesn't reproduce the production incident. The production data shows all 5 suggestions had status **NEW**, not OUTDATED:

```json
{ "status": "NEW", "updatedAt": "2026-06-11T12:33:50.316Z" }  // × 4
{ "status": "NEW", "updatedAt": "2026-06-11T12:36:45.846Z" }  // × 1
```

OUTDATED and NEW follow completely different code paths in `syncSuggestions`:

**Author's Case 6b (OUTDATED):**
- `findPreservableDomainWideSuggestion` returns null (OUTDATED is not preservable) → domain-wide **is** in `newData`
- All 3 OUTDATED entries match via `buildKey` → `toUpdate` path → `saveMany` called
- The guard exercised is the status filter (`OUTDATED` already excluded from `handleOutdatedSuggestions`)

**Actual production scenario (NEW):**
- `findPreservableDomainWideSuggestion` finds a NEW → returns it → domain-wide **not** in `newData`
- All 5 NEW entries are not in `newDataKeys` → `handleOutdatedSuggestions` sees them as candidates
- The **`isDomainWide` guard** (line 279 of `data-access.js`) is what protects them — this is the guard that must not regress
- `saveMany` and `addSuggestions` are never called

The test that covers the real scenario and guards the `isDomainWide` protection:

```js
it('Case 6b — multiple NEW domain-wide duplicates from a race do not get outdated or duplicated further', async () => {
  // Production state: 5 concurrent Lambdas created 5 NEW domain-wides.
  // On the next single-threaded run, findPreservableDomainWideSuggestion found one of them
  // and excluded domain-wide from newData entirely.
  const dup1 = makeSuggestion({ isDomainWide: true, url: 'https://example.com/* (All Domain URLs)' }, 'NEW');
  const dup2 = makeSuggestion({ isDomainWide: true, url: 'https://example.com/* (All Domain URLs)' }, 'NEW');
  const dup3 = makeSuggestion({ isDomainWide: true, url: 'https://example.com/* (All Domain URLs)' }, 'NEW');

  await syncSuggestions({
    context, opportunity,
    newData: [],  // domain-wide excluded because findPreservableDomainWideSuggestion found a NEW one
    buildKey, mergeDataFunction,
    mapNewSuggestion: (d) => ({ data: d }),
    existingSuggestions: [dup1, dup2, dup3],
  });

  // isDomainWide guard in handleOutdatedSuggestions must protect all three NEW entries
  const outdatedCall = bulkUpdateStatus.getCalls().find((c) => c.args[1] === 'OUTDATED');
  expect(outdatedCall).to.be.undefined;
  // No new domain-wide created
  expect(addSuggestions).to.not.have.been.called;
  // No spurious data updates (nothing to match against in newData)
  expect(saveMany).to.not.have.been.called;
});
```

The current Case 6b is a valid test for the `buildKey` fix on its own, but it should be renamed to reflect what it actually tests (OUTDATED duplicates via `buildKey` match) and this NEW-status test added alongside it.

ssilare-adobe: Good catch — the production data had all NEW-status duplicates, not OUTDATED. Fixed in 9ae11fb2:

- **Case 6b** renamed to clarify it tests the OUTDATED-via-buildKey-match path (the fix's primary mechanism)
- **Case 6c** added for the exact production scenario: 3 NEW domain-wide duplicates, `newData: []` (domain-wide excluded because `findPreservableDomainWideSuggestion` found one), and asserts:
  1. `bulkUpdateStatus` never called with `OUTDATED` — the `isDomainWide` guard protects all three
  2. `addSuggestions` not called — no new domain-wide created
  3. `saveMany` not called — nothing in `newData` to match against, so no spurious updates

This directly guards the `isDomainWide` check in `handleOutdatedSuggestions` line 279.

anuj-adobe (approved):
All three review concerns are addressed:

1. **Drift risk** — `buildSuggestionKey` is now a named export from `src/prerender/utils/utils.js`. Both `sync-suggestions.test.js` and `handler.js` use the same real function — no copy-paste mirror to drift.

2. **Production scenario coverage** — Case 6c correctly reproduces the actual incident: 3 NEW domain-wide duplicates, domain-wide excluded from `newData` (as `findPreservableDomainWideSuggestion` would do), and asserts the `isDomainWide` guard in `handleOutdatedSuggestions` protects them from being marked OUTDATED. Case 6b covers the complementary OUTDATED path that the `buildKey` fix addresses directly.

3. **Full chain** — `sync-suggestions.test.js` now imports and exercises the real `buildSuggestionKey` + real `syncSuggestions` end-to-end.

All 8 `sync-suggestions` tests pass. LGTM.

anuj-adobe (approved):
All three review concerns are addressed:

1. **Drift risk** — `buildSuggestionKey` is now a named export from `src/prerender/utils/utils.js`. Both `sync-suggestions.test.js` and `handler.js` use the same real function — no copy-paste mirror to drift.

2. **Production scenario coverage** — Case 6c correctly reproduces the actual incident: 3 NEW domain-wide duplicates, domain-wide excluded from `newData` (as `findPreservableDomainWideSuggestion` would do), and asserts the `isDomainWide` guard in `handleOutdatedSuggestions` protects them from being marked OUTDATED. Case 6b covers the complementary OUTDATED path that the `buildKey` fix addresses directly.

3. **Full chain** — `sync-suggestions.test.js` now imports and exercises the real `buildSuggestionKey` + real `syncSuggestions` end-to-end.

All 8 `sync-suggestions` tests pass. LGTM.

PR #2652, opened 2026-06-12
feat(prerender): support CSV URL scoping for ai-only mode batches

ssilare-adobe:
## Summary

Adds CSV URL scoping to the ai-only prerender flow so batched audits only process their subset of URLs.

### What changed

- **`sendPrerenderGuidanceRequestToMystique`** — accepts optional `urlScope` (a Set of URLs) to filter candidates before sending to Mystique
- **`handleAiOnlyMode`** — when `auditContext.urls` is present (CSV batch from run-audit command), passes the URL set as a scope filter

Mode detection is unchanged — the worker reads `mode: 'ai-only'` from the SQS `data` field via its existing `getModeFromData(data)`.

### How batch support works

The API service sends one SQS message per batch:
```json
{
  "type": "prerender",
  "data": "{\"mode\":\"ai-only\"}",
  "auditContext": { "urls": ["url1", ..., "url320"] }
}
```

The worker:
1. Detects `ai-only` from `data.mode` (existing logic)
2. Finds the opportunity and loads all its suggestions from DB
3. Filters suggestions to only those matching the batch's `auditContext.urls`
4. Sends the scoped set to Mystique

Without `auditContext.urls`, all eligible suggestions are sent (existing behavior preserved).

## Test plan (56 tests passing)

- [x] Scopes suggestions to `auditContext.urls` when provided
- [x] Filters out suggestions not in `auditContext.urls`
- [x] Sends all suggestions when `auditContext.urls` is not set
- [x] Returns 0 when no suggestions match the URL scope
- [x] All existing tests pass unchanged

### DIFF
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -527,11 +527,19 @@ async function compareHtmlContent(url, context) {
 }
 
 /**
- * Parses the mode from the data field
+ * Parses the mode from the data field or auditContext.
+ * The mode may arrive in either location:
+ *   - data.mode: set via Slack command keyword arguments (legacy)
+ *   - auditContext.mode: set via run-audit command's mode:ai-only flag
  * @param {string|Object} data - The data field from the message
+ * @param {Object} [auditCtx] - The auditContext from the message
  * @returns {string|null} - The mode value or null
  */
-function getModeFromData(data) {
+function getModeFromData(data, auditCtx) {
+  if (auditCtx?.mode) {
+    return auditCtx.mode;
+  }
+
   if (!data) {
     return null;
   }
@@ -582,6 +590,7 @@ async function sendPrerenderGuidanceRequestToMystique(
   context,
   preBuiltCandidates,
   generatePrompts = false,
+  urlScope = null,
 ) {
   const {
     log, sqs, env, site,
@@ -680,6 +689,11 @@ async function sendPrerenderGuidanceRequestToMystique(
       suggestionsPayload = candidates;
     }
 
+    // When a URL scope is provided (CSV batch), filter to only matching URLs
+    if (urlScope && suggestionsPayload.length > 0) {
+      suggestionsPayload = suggestionsPayload.filter((s) => urlScope.has(s.url));
+    }
+
     if (suggestionsPayload.length === 0) {
       log.info(`${LOG_PREFIX} No eligible suggestions to send to Mystique for opportunityId=${opportunityId}. baseUrl=${baseUrl}, siteId=${siteId}`);
       return 0;
@@ -730,7 +744,7 @@ async function sendPrerenderGuidanceRequestToMystique(
  */
 export async function handleAiOnlyMode(context) {
   const {
-    site, log, dataAccess, data,
+    site, log, dataAccess, data, auditContext,
   } = context;
   const { Opportunity } = dataAccess;
   const siteId = site.getId();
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @ssilare-adobe,

**Verdict: Approve** - clean, well-scoped feature addition with solid test coverage.
_Changes: adds optional CSV URL scoping to the ai-only prerender flow so batched audits only process their subset of URLs (2 files)._

<details>

- nit: URL comparison via `Set.has(s.url)` is exact-match only - if upstream CSVs ever contain URLs with different casing, trailing slashes, or query params vs DB-stored suggestions, matches will silently fail. Consider documenting this as the explicit contract or adding normalization. - `src/prerender/handler.js:683`
- suggestion: log how many suggestions were excluded by the scope filter (not just the scope size at entry) to make batch debugging easier - `src/prerender/handler.js:683`
- nit: no test for `auditContext = { urls: [] }` boundary (should behave identically to absent urls) - `test/audits/prerender/ai-only-mode.test.js`
- nit: positional parameter list on `sendPrerenderGuidanceRequestToMystique` is now 5 params with null-passing at call sites - worth watching if a 6th appears - `src/prerender/handler.js:585`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 4s | <b>Cost:</b> $3.62 | <b>Commit:</b> <code>b3fd2dd9184f80e2a7fd95c96d67dd290338419e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/prerender/handler.js
```diff
@@ -680,6 +681,11 @@ async function sendPrerenderGuidanceRequestToMystique(
       suggestionsPayload = candidates;
     }
 
+    // When a URL scope is provided (CSV batch), filter to only matching URLs
+    if (urlScope && suggestionsPayload.length > 0) {
```
anuj-adobe, line 685: The `suggestionsPayload.length > 0` guard is redundant. `[].filter(...)` already returns `[]` — the condition adds no protection and obscures the intent.

```js
// Before
if (urlScope && suggestionsPayload.length > 0) {
  suggestionsPayload = suggestionsPayload.filter((s) => urlScope.has(s.url));
}

// After
if (urlScope) {
  suggestionsPayload = suggestionsPayload.filter((s) => urlScope.has(s.url));
}
```

test/audits/prerender/ai-only-mode.test.js
```diff
@@ -738,6 +738,52 @@ describe('Prerender AI-Only Mode', () => {
...
+    });
+
+    it('should return 0 when no suggestions match auditContext.urls', async () => {
+      context.auditContext = {
+        urls: ['https://example.com/no-match'],
+      };
+
+      const result = await importTopPages(context);
+
+      expect(result.auditResult.suggestionCount).to.equal(0);
```
anuj-adobe, line 783: Missing assertion that SQS was **not** called. The behaviour this test is supposed to guard is that Mystique is never invoked when the scope filters everything out. Checking `suggestionCount` alone doesn't catch a bug where the empty-array early-return is accidentally removed — the count would still be 0 but a message would be sent.

```js
expect(result.auditResult.suggestionCount).to.equal(0);
expect(mockSqs.sendMessage).to.not.have.been.called; // ← add this
```

anuj-adobe (changes requested):
Two issues — one in source, one in tests.

src/prerender/handler.js
ssilare-adobe: Discussed to merge as `ai-approved`

test/audits/prerender/ai-only-mode.test.js
ssilare-adobe: Discussed to merge as `ai-approved`

PR #2655, opened 2026-06-12
feat: llm error pages suggestion history

ymadobe:

## Related Issues

Thanks for contributing!


anagarwa (changes requested):
**5-reviewer consolidated review** (staff-engineer ×2, architect, project-conventions, security ×2, qa ×2)

---

## Strengths

- `upsertWeekHistory` (handler.js:100–113) is pure, idempotent (Map-keyed by `periodIdentifier`), oldest-first, bounded by `RETENTION_WEEKS`. Correctly imported without esmock in tests so real `parsePeriodIdentifier` ordering is exercised.
- Retention classifier (handler.js:459–481) is a clean single-pass `forEach` with two guards up front (`scrapedUrls.has()` and `!lastSeen`) before any tier decision.
- Bulk operations (`removeByIds`, `bulkUpdateStatus`) follow CLAUDE.md's N+1 prevention patterns; no pool-exhaustion risk.
- `Promise.allSettled` + per-bucket try/catch prevents one bucket failure from poisoning others.
- 15 new test cases with 100% codecov on modified lines, covering all four sweep outcomes.
- The IGNORED-flip trade-off is explicitly documented in the commit message rather than left as a silent surprise.

---

## Critical — Must Fix

### C1. Malformed `periodIdentifier` triggers silent hard-delete via epoch fallback
`handler.js:466–472` + `utils.js:651–654`

`parsePeriodIdentifier` returns `new Date(0)` (epoch 1970) for any string that doesn't match `/^w(\d{2})-(\d{4})$/i`. The only sweep guard is `if (!lastSeen) return;` — catches falsy values, not truthy-but-unparseable ones. Epoch is always < `deleteCutoff`, so a suggestion with `periodIdentifier: '2026-W15'`, `'w5-2026'` (no leading zero), or any future format change is silently routed to `toDelete` and hard-purged by `removeByIds` — a true DELETE with no recovery.

The `!lastSeen` guard was added for exactly this "age unknowable" class; the same treatment must apply to parse failures. Confirmed independently by security, qa, staff-engineer ×2.

**Fix:**
```js
const seenAt = parsePeriodIdentifier(lastSeen);
if (seenAt.getTime() === 0) {
  log.warn(`[LLM-ERROR-PAGES] Skipping suggestion ${s.getId()} with unparseable periodIdentifier: "${lastSeen}"`);
  return;
}
```
Add a test asserting `removeByIds` is NOT called when `periodIdentifier` is a truthy non-matching string.

---

### C2. OUTDATED skip-list contains a non-existent status and is missing four real ones
`handler.js:475–477`

The list `['OUTDATED', 'FIXED', 'RESOLVED', 'REJECTED', 'APPROVED']` has two bugs:

- **`RESOLVED` does not exist** in the canonical `Suggestion.STATUSES` enum (`NEW, APPROVED, IN_PROGRESS, SKIPPED, FIXED, ERROR, OUTDATED, PENDING_VALIDATION, REJECTED`). Dead branch — the check never fires.
- **`SKIPPED`, `IN_PROGRESS`, `PENDING_VALIDATION`, `ERROR` are all missing.** A user who explicitly skipped a suggestion, or has one `IN_PROGRESS`, will have it silently flipped to OUTDATED ~5 weeks later. The canonical skip list at `src/utils/data-access.js:262–267` correctly includes `SKIPPED` and `ERROR`; this sweep is inconsistent with the existing convention.

Only the `FIXED` case is tested at handler.test.js:1819–1851; all other terminal states are uncovered.

**Fix:** Remove `RESOLVED`; add `SKIPPED`, `ERROR`, `IN_PROGRESS`, `PENDING_VALIDATION`. Consider importing the enum from `SuggestionDataAccess.STATUSES` rather than maintaining a parallel string list. Add test cases for `SKIPPED` and `IN_PROGRESS`.

---

## Important — Should Fix

### I1. No spec or ADR — CLAUDE.md compliance gap
`docs/specs/`, `docs/decisions/` (absent from PR)

CLAUDE.md uses "must" language for both: specs for "significant new functionality" and ADRs for "non-obvious behavior choices with long-term impact." This PR introduces: the `data.history[]` per-week schema (a new DB contract for UI consumers), the 3-tier retention policy, the hard-purge-includes-terminal-states decision, the IGNORED-flip trade-off. The existing `docs/specs/llm-error-pages-head-check-and-db-consistency.md` covers unrelated work. Confirmed by all reviewers.

**Fix:** Add `docs/decisions/002-llm-error-pages-retention-and-history.md` covering the 3-tier thresholds with rationale, the hard-purge-ignores-status decision, the `data.history[]` schema as a cross-system contract, and the IGNORED-flip temporary trade-off. Extend the existing spec or add a sibling covering the new schema.

---

### I2. Hard-delete tier purges customer-actioned terminal states with no recovery path
`handler.js:471–472, 457`; test at handler.test.js:1789–1812

The middle tier explicitly skips FIXED/APPROVED with the comment "avoids silently regressing a customer action." The delete tier destroys those same rows unconditionally at week 7 — a FIXED record is an audit trail of the customer's remediation. If the upstream Athena query returns zero rows for several weeks due to a transient pipeline issue, URLs missing from `scrapedUrls` over multiple runs will age into the >6-week tier and be purged regardless of customer-actioned status.

**Fix (one of):** either extend the terminal-state skip to the delete tier; **or** document the explicit trade-off in the ADR required by I1, including any compliance review, and enrich the delete log to include IDs and statuses of purged terminal-state rows.

---

### I3. `RETENTION_WEEKS` controls three independent product decisions
`handler.js:52–55, 112`

The same constant drives: (a) DB delete cutoff, (b) OUTDATED window ceiling, (c) history array length in `upsertWeekHistory`. Bumping to 12 weeks silently doubles the per-suggestion JSON payload. These are independent decisions that happen to share a value today. Confirmed by architect, staff-engineer ×2.

**Fix:** Split into `HISTORY_RETENTION_WEEKS = 6` and `DB_RETENTION_WEEKS = 6`. Reference the history constant at handler.js:112; reference the DB constant for cutoffs. Export `HISTORY_RETENTION_WEEKS` so handler.test.js:2333–2342 can use it instead of hardcoding `6`.

---

### I4. `userAgents` and `agentTypes` stored in `data.history[]` without a length cap
`handler.js:72–81` (`buildWeekHistoryEntry`)

`groupErrorsByUrl` aggregates all unique user-agents seen for a URL in a week. A bot rotating UAs produces an unbounded array. This was stored once on the top-level `data` field (overwritten weekly), but `data.history[]` retains it for 6 weeks, multiplying potential storage by up to 6×. On a heavily-crawled error URL, this can significantly bloat the suggestion `data` blob. Additionally, raw UA strings may contain PII or arbitrary text; storing them for 6 weeks in a UI-exposed field increases the window for inadvertent exposure.

**Fix:** cap in `buildWeekHistoryEntry`:
```js
const MAX_AGENT_ENTRIES = 10;
return {
  ...
  agentTypes: (item.agentTypes || []).slice(0, MAX_AGENT_ENTRIES),
  userAgents: (item.userAgents || []).slice(0, MAX_AGENT_ENTRIES),
};
```

---

### I5. Backfill path runs retention sweep with wall-clock cutoffs against historical data
`handler.js:341–342`; `backfillAudit` (handler.js:629–657)

`deleteCutoff` and `outdatedCutoff` are computed from `Date.now()`. When `backfillAudit` calls `runAuditAndSendToMystique` for a historical week (e.g. `weekOffset=3`), the Athena query returns URLs from 3 weeks ago, so `scrapedUrls` only contains that historical week's error URLs. Any suggestion whose URL wasn't in that historical week but exists in the current week is not protected by the `scrapedUrls.has(data.url)` guard, and will be classified by age relative to today's cutoffs. A suggestion last seen 5 weeks ago in a non-backfilled week would be marked OUTDATED by a historical backfill run.

**Fix:** skip the retention sweep when running in backfill mode: `if (auditContext.weekOffset !== undefined) return;` before the sweep block. Add a regression test asserting no `removeByIds`/`bulkUpdateStatus` calls when `weekOffset` is set.

---

### I6. IGNORED-flip compounding effect is worse than documented
`handler.js:345, 377–387`

The commit message documents "new opp created weekly, history doesn't accumulate." What's not captured: 3 buckets × weekly run = 3 new IGNORED opps/site/week. Over the 6-week window: 18 stale IGNORED opportunities with full suggestion sets that the sweep structurally cannot touch (pre-fetch at line 345 filters to `NEW` only). This is unbounded growth even after the UI cutover — there is no backfill cleanup path.

**Fix:** file a tracking issue, add its number to the TODO comment at handler.js:377–381, and add a `log.warn` once per run when the IGNORED-flip path is hit so operators can see in Coralogix that history accumulation is currently dark for that site.

---

### I7. No test for the first-rollout scenario: existing suggestions have no `history` field
`handler.js:412–415`

`mergeDataFunction` calls `upsertWeekHistory(existingData.history, ...)` with `existingData.history === undefined` for every suggestion created before this PR ships. The pure unit test for `upsertWeekHistory(undefined, ...)` exercises the helper in isolation, but no integration-level test asserts that the handler's merge path produces a correct one-entry history when the stored suggestion has no prior history. This is the code path that runs on every suggestion in the fleet on the first post-deploy audit run.

**Fix:** add a test: call `mergeDataFunction({ url: '/p1' /*, no history */ }, newDataItem)` and assert `merged.history.length === 1` with the correct `periodIdentifier`.

---

### I8. Deletion log records count only — forensic gap for a hard-delete operation
`handler.js:498–500`

`removeByIds` is irrecoverable. The log line emits a count but not the deleted IDs, making it impossible to reconstruct which suggestions were destroyed if a bug or disputed deletion needs investigation.

**Fix:**
```js
if (toDelete.length > 0) {
  const deletedIds = toDelete.map((s) => s.getId());
  await Suggestion.removeByIds(deletedIds);
  log.info(`[LLM-ERROR-PAGES] Deleted ${deletedIds.length} suggestions older than ${RETENTION_WEEKS} weeks for ${auditType}`, {
    suggestionIds: deletedIds,
    urls: toDelete.map((s) => s.getData()?.url),
  });
}
```

---

### I9. `existingSuggestions` status may be stale after `syncSuggestions` mutates in-memory objects
`handler.js:389, 391, 474`

`defaultMergeStatusFunction` can call `existing.setStatus(NEW)` on in-memory suggestion objects when a URL resurfaces. The sweep reads `s.getStatus()` on those same instances. Today safe because `scrapedUrls.has(data.url)` early-return covers the only URLs where `syncSuggestions` changes status, but the invariant lives in `src/utils/data-access.js`, not here.

**Fix:** add a comment at handler.js:461–463: *"The scrapedUrls guard covers all URLs syncSuggestions may have mutated in-memory — status reads below are safe while syncSuggestions only flips status for matching URLs."*

---

## Minor — Nice to Have

**M1. Retention sweep tests are real-clock-coupled** — `handler.test.js:1854–2002`  
Use `sandbox.useFakeTimers(new Date('2026-04-13T00:00:00Z').getTime())` to pin the clock and enable exact boundary tests at `weeksAgo(4)` and `weeksAgo(6)`.

**M2. `RETENTION_WEEKS` not exported; test hardcodes `6`** — `handler.test.js:2333–2342`  
Export the constant (or `HISTORY_RETENTION_WEEKS` once split per I3) and use it in the test.

**M3. IGNORED TODO comment lacks a tracking handle** — `handler.js:377–381`  
File the follow-up issue and add its number to the comment.

**M4. Stale test comment** — `handler.test.js:2160`  
"no lastSeen → return false" → "no lastSeen → no push to toDelete/toOutdate".

**M5. Sweep-skip tests assert only one of two bulk paths** — `handler.test.js:1758–2002`  
The `no periodIdentifier`, `current-run guard`, and `fresh tier` tests each assert `removeByIds` was not called, but not that `bulkUpdateStatus` was also not called. Add the symmetric assertion to each.

---

## Assessment

**Request changes.** Two Critical data-mutation bugs must be fixed before merge: the epoch-fallback hard-delete path (C1, one-line guard + one test) and the incomplete skip-list that silently flips `SKIPPED`/`IN_PROGRESS` suggestions to OUTDATED (C2, corrected list + tests). The Important findings include two additional behavioral gaps: unbounded UA storage in history entries (I4) and the backfill-vs-wall-clock interaction (I5). The CLAUDE.md-required spec + ADR (I1) must also be addressed. The code itself is well-structured, follows the repo's bulk-operation patterns, and the tier semantics are clearly commented at the call site.

MysticatBot (approved):
Hey @ymadobe,

**Verdict: Approve** - all Critical and Important code-level findings from the prior review are correctly addressed.
_Changes: hardens the 3-tier retention sweep with epoch-sentinel guard, protected-status enum, split retention constants, agent-array cap, and backfill-mode bypass (2 files)._

<details>

- nit: `Suggestion.bulkUpdateStatus(toOutdate, 'OUTDATED')` uses a string literal while the rest of the sweep references `SuggestionModel.STATUSES.*` - `src/llm-error-pages/handler.js:533`
- suggestion: Add a test asserting a stale OUTDATED suggestion IS deleted past DB_RETENTION_WEEKS (documents the intended behavior and prevents a regression if someone later adds OUTDATED to PROTECTED_SWEEP_STATUSES) - `test/audits/llm-error-pages/handler.test.js`
- suggestion: Add an integration test for `mergeDataFunction` when `existingData` has no `history` field (the first-rollout path every suggestion hits post-deploy; unit test covers the helper but not the wiring) - `test/audits/llm-error-pages/handler.test.js`
- suggestion: Log deleted suggestion IDs in the `removeByIds` call for forensic recovery on irrecoverable hard-deletes - `src/llm-error-pages/handler.js:527`
</details>

- C1: epoch-fallback hard-delete now guarded with `seenAt.getTime() === 0` check + `log.warn`
- C2: `PROTECTED_SWEEP_STATUSES` sourced from canonical `SuggestionModel.STATUSES` enum
- I2: Protected-status guard applies to both delete and outdate tiers
- I3: Split into `DB_RETENTION_WEEKS`, `HISTORY_RETENTION_WEEKS`, `NEW_WINDOW_WEEKS`
- I4: `MAX_AGENT_ENTRIES = 10` caps `agentTypes`/`userAgents` per history entry
- I5: `isBackfill` guard skips sweep when `weekOffset` is set

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 45s | <b>Cost:</b> $4.36 | <b>Commit:</b> <code>14d00154c948bbac240e4d2f3b5dd0127b932efb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anagarwa (approved):

PR #2659, opened 2026-06-15
feat(toc): wire TOC opportunity into Impact Engine prompt generation (LLMO-5611)

anagarwa:
## Summary

- Adds `sendTocGuidanceRequestToMystique()` which fires a `guidance:table-of-contents` SQS message to Mystique after `syncSuggestions` runs
- Each suggestion payload carries `url`, `title`, `headings` (extracted from HAST via new `extractHeadingTextsFromHast()`), and `hasPrompts` flag so Mystique can generate LLMO impact-engine prompts
- New suggestions are initialised with `prompts: []` and `hasPrompts: false`; already-prompted and FIXED/OUTDATED/SKIPPED suggestions are skipped to avoid redundant work

Relates to LLMO-5611 / Epic LLMO-28 — TOC in Impact Engine.

## Test plan

- [ ] Run existing TOC handler tests (`npm run test:spec -- test/audits/toc/handler.test.js`) — all green
- [ ] Verify SQS message shape matches `TocOpportunityRequestMessageData` in Mystique
- [ ] Deploy to stage; trigger a TOC audit; confirm a `guidance:table-of-contents` message lands on the Mystique queue with correct suggestion payloads
- [ ] Confirm Mystique task processes the message and writes `prompts` back to suggestion data

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -91,6 +91,7 @@ import llmoCustomerAnalysis from './llmo-customer-analysis/handler.js';
 import llmoOnboardingPublish from './llmo-onboarding-publish/handler.js';
 import headings from './headings/handler.js';
 import toc from './toc/handler.js';
+import tocGuidance from './toc/guidance-handler.js';
 import vulnerabilities from './vulnerabilities/handler.js';
 import vulnerabilitiesCodeFix from './vulnerabilities-code-fix/handler.js';
 import prerender from './prerender/handler.js';
@@ -198,6 +199,7 @@ const HANDLERS = {
   hreflang,
   headings,
   toc,
+  'guidance:table-of-contents': tocGuidance,
   prerender,
   'guidance:prerender': prerenderGuidance,
   'product-metatags': productMetatags,
--- a/src/toc/eligibility-utils.js
+++ b/src/toc/eligibility-utils.js
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/**
+ * True when a TOC suggestion is still open (not FIXED/OUTDATED/SKIPPED) and therefore
+ * still eligible to send/receive Impact Engine prompts. Shared by the outbound
+ * (sendTocGuidanceRequestToMystique) and inbound (guidance-handler) eligibility checks
+ * so the two stay in sync.
+ * @param {Object} suggestion - Suggestion entity
+ * @param {Object} Suggestion - The Suggestion data-access collection (for STATUSES)
+ * @returns {boolean} True if the suggestion is non-terminal
+ */
+export function isEligibleTocSuggestion(suggestion, Suggestion) {
+  const status = suggestion.getStatus();
+  return status !== Suggestion.STATUSES.FIXED
+    && status !== Suggestion.STATUSES.OUTDATED
+    && status !== Suggestion.STATUSES.SKIPPED;
+}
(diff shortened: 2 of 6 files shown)


anuj-adobe:
## Code review — TOC guidance handler + Mystique wiring

Findings below, most severe first. Paired with the Mystique-side review on experience-platform/mystique#3162.

### 🟠 Correctness

**1. `src/toc/handler.js:~586` (mapNewSuggestion) — re-auditing wipes Mystique-persisted prompts.**
`mapNewSuggestion` now injects `prompts: []` and `hasPrompts: false` into `data`, and the existing `mergeDataFunction` returns `{ ...existingSuggestion, ...converted }` (`handler.js:442-443`). On every subsequent TOC audit, the new empty `prompts` / `hasPrompts:false` **overwrite** the prompts Mystique previously persisted (except for `edgeDeployed` suggestions) — and because `hasPrompts` is reset to `false`, `sendTocGuidanceRequestToMystique` then re-requests a full LLM regeneration each audit cycle. Preserve the existing `prompts`/`hasPrompts` in `mergeDataFunction` when the new item's prompts are empty (e.g. carry them over unless the new item actually supplies prompts).

### 🟡 Cleanup (DRY / code smell)

**2. `src/toc/handler.js:236` & `src/toc/guidance-handler.js:52` — duplicated terminal-status filter (DRY).**
The `status === FIXED || OUTDATED || SKIPPED` skip block is copy-pasted in both `sendTocGuidanceRequestToMystique` (outbound eligibility) and `groupEligibleSuggestionsByUrl` (write-back eligibility). They must be kept in sync by hand; drift silently changes which suggestions get prompts. Extract one shared helper, e.g. `isEligibleTocSuggestion(suggestion, Suggestion)`.

**3. `src/toc/handler.js:180` — `extractHeadingTextsFromHast` is over-broad and misnamed.**
It collects *every* text node in the HAST subtree (including inter-element whitespace/newline nodes and any non-heading text), with no tag filtering and no dedup. The `headings` array sent to Mystique for prompt grounding can contain `'\n'` / `'  '` noise and duplicate fragments, degrading prompt quality. Filter to heading nodes and trim/drop blank strings.

**4. Minor — `suggestionId` is dead payload.** Each outbound candidate includes `suggestionId`, but the reply is matched purely by URL (as `guidance-handler.js`'s own docstring notes — the id doesn't round-trip through Mystique). It reads as a correlation key that isn't one; drop it or wire the return path to it.

<sub>🤖 Review assisted by Claude Code</sub>

anagarwa:
Addressed in 880e3658:

1. **Re-audit wipes Mystique-persisted prompts** — Fixed. `mergeDataFunction` now preserves the existing `prompts`/`hasPrompts` unless the new item actually supplies prompts, instead of letting `mapNewSuggestion`'s always-empty seed values clobber them on every subsequent audit.
2. **Duplicated terminal-status filter (DRY)** — Fixed. Extracted `isEligibleTocSuggestion()` into a new `src/toc/eligibility-utils.js`, used by both `handler.js` and `guidance-handler.js` so the outbound/inbound eligibility checks can't drift apart.
3. **`extractHeadingTextsFromHast` over-broad/misnamed** — Fixed. Now walks only heading anchor nodes (`a[data-selector]`, matching the shape `tocArrayToHast` produces) instead of every text node in the tree, and results are deduped.
4. **`suggestionId` dead payload** — left as-is (out of scope for this pass, per explicit direction).

Also found and fixed while verifying: `test/audits/toc.test.js`'s log mocks never stubbed `.warn`, so this PR's own `sendTocGuidanceRequestToMystique()` call (which `opportunityAndSuggestions` now invokes unconditionally) was crashing 21 existing tests — confirmed this was already failing on `ci / build` before this fix, independent of the review findings above.

All 182 TOC tests pass locally; lint clean.

🤖 Fixes applied with Claude Code

anuj-adobe (approved):
Re-reviewed at `f94e510` — the main points from the earlier review are addressed:

1. ✅ Re-audit no longer wipes Mystique-persisted prompts — `mergeDataFunction` now preserves existing `prompts`/`hasPrompts` when the new item is empty (with a clear explanatory comment).
2. ✅ Duplicated FIXED/OUTDATED/SKIPPED filter extracted to a shared `eligibility-utils.js` (`isEligibleTocSuggestion`) used by both `handler.js` and `guidance-handler.js`.
3. ✅ `extractHeadingTextsFromHast` now targets `a[data-selector]` heading links specifically, trims, and the caller dedups via `new Set` — with a test pinning the clean output.

Minor (non-blocking): `suggestionId` is still sent in the outbound candidates even though the reply is matched by URL — harmless dead payload, fine to leave or drop later.

No new concerns. LGTM 👍

ymadobe (approved):

PR #2662, opened 2026-06-15
fix(llm-error-pages): un-hide bucket Opps now that ELMO UI supports them

ymadobe:
## Summary

Reverts the temporary #2602 workaround that flipped newly-created
`llm-error-pages-{404,403,5xx}` opportunities to `IGNORED` to hide them from a
UI that didn't yet recognise the bucket types. The ELMO UI now supports these
types (project-elmo-ui #1990), so new Opps should be visible (`NEW`) and the
rows the workaround accumulated should be un-hidden.

Closes LLMO-5328 (under LLMO-3610).

## Changes

- `handler.js`: replace the `NEW → IGNORED` flip with a system-managed
  `IGNORED → NEW` un-hide. New Opps are left `NEW`; previously hidden rows are
  flipped back as their bucket recurs in subsequent audit runs.
- Guarded on `getUpdatedBy() === 'system'` so a customer's explicit IGNORE is
  never overridden.
- Tests updated: "leaves a NEW Opp as NEW", "restores a system-IGNORED Opp to
  NEW", "does not un-hide a customer-IGNORED Opp".

## Notes / rollout

- The un-hide happens per-site as each bucket recurs (errors found again). For
  an immediate one-shot un-hide of all accumulated rows, a separate DB migration
  could bulk-update `status=IGNORED AND updatedBy='system'` → `NEW`, but this
  handler change resolves it within a weekly cycle without a migration.

## Test plan

- [x] `npx eslint src/llm-error-pages/handler.js` — clean
- [x] llm-error-pages handler suite — 87 passing
- [ ] CI green
- [ ] Post-deploy: confirm next weekly run logs `Restored opportunity … to NEW`
      and the Opps surface in ELMO


MysticatBot (approved):
Hey @ymadobe,

**Verdict: Approve** - clean, well-scoped revert of a temporary workaround with correct guard logic.
_Changes: Replaces the temporary NEW-to-IGNORED workaround with an IGNORED-to-NEW un-hide guarded by `updatedBy === 'system'`, now that the ELMO UI supports bucket-specific opportunity types (2 files)._

Note: CI checks are still pending - confirm build passes before merge.

<details>

- nit: `setUpdatedBy('system')` is a no-op on the un-hide path since the guard already requires `getUpdatedBy() === 'system'` - the value is unchanged. Harmless, but a future reader may wonder if it is intentional (e.g., to bump `updatedAt`). Consider removing or adding a brief comment. - `src/llm-error-pages/handler.js:543`
- suggestion: The old test suite covered status=APPROVED falling through the guard; the new suite only covers status=NEW and status=IGNORED. If 100% branch coverage requires it, a test with a non-IGNORED, non-NEW status (e.g., APPROVED) would confirm the condition is truly `=== IGNORED`-only. - `test/audits/llm-error-pages/handler.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 2s | <b>Cost:</b> $2.18 | <b>Commit:</b> <code>25fab048da5e3391082d8d119e566fd4fa9ae1ef</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anagarwa (approved):

PR #2665, opened 2026-06-15
fix(prerender): normalize URL pathnames to lowercase to prevent case-variant duplicate suggestions

anuj-adobe:
## Problem

Two prerender utility functions — `normalizePathnameWithQuery` and `toPathname` — were not lowercasing the pathname before building suggestion dedup keys.

This caused URLs that differ only in case (e.g. `/Page1` vs `/page1`) to produce different keys. When `syncSuggestions` ran, it could not find the existing stored suggestion by key, so it created a new undeployed duplicate instead of updating the existing (possibly already-deployed) one.

The same miss existed in both PRs that introduced these functions (#2397 for `toPathname`, #2650 for `normalizePathnameWithQuery`) — neither PR's description mentions case-sensitivity as an in-scope concern.

## Fix

Added `.toLowerCase()` to the **pathname portion only** in both functions:

- **`normalizePathnameWithQuery`** — used as the dedup key for individual URL suggestions in `buildSuggestionKey`, `scrapedUrlsSet` lookups, and page record maps
- **`toPathname`** — used in edge-deployed URL matching (`markDeployedUrlSuggestionsAsCovered`) and deployed pathnames set

Query params are intentionally **not** lowercased — they can contain signed or case-sensitive tokens where lowercasing would break matching.

## Behavior test layer

Added `test/audits/prerender/behavior/` — a new test layer that runs real handler logic with only I/O boundaries mocked (no stubs on business logic). This gives end-to-end confidence that key constraints hold across the full flow, complementing the heavily-mocked `handler.test.js`.

### `suggestion-deduplication.behavior.test.js`
Enforces that an audit never creates duplicate suggestions:
- An existing NEW domain-wide is preserved — no new domain-wide is created
- A case-variant URL (`/Page1` existing, `/page1` incoming) resolves to the same key — no duplicate added

### `edge-deployed-protection.behavior.test.js`
Enforces that `edgeDeployed` suggestions are never marked OUTDATED:
- A suggestion with `edgeDeployed` set is excluded from `bulkUpdateStatus` even when absent from audit results and no `scrapedUrlsSet` is provided (worst-case scenario)
- A counterpart test confirms non-deployed stale suggestions ARE correctly passed to `bulkUpdateStatus`, so the guard is verified from both sides

## Test plan
- [ ] `test/audits/prerender/behavior/suggestion-deduplication.behavior.test.js` — 4 tests pass
- [ ] `test/audits/prerender/behavior/edge-deployed-protection.behavior.test.js` — 3 tests pass
- [ ] `test/audits/prerender/utils.test.js` — 9 new unit tests for `toPathname` (case normalization) and `normalizePathnameWithQuery` (full coverage including query param casing preserved)
- [ ] `test/audits/prerender/handler.test.js` — dedup test updated to reflect current behavior


MysticatBot (changes requested):
Hey @anuj-adobe,

**Verdict: Request changes** - one inconsistency in the error fallback path undermines the stated fix.
_Changes: adds `.toLowerCase()` to URL pathname normalization in two utility functions to prevent case-variant duplicate suggestions, plus a new behavior test layer (5 files)._

## Must fix before merge
1. [Important] `toPathname` catch block returns raw URL without lowercasing, inconsistent with the fix applied to `normalizePathnameWithQuery` - `src/prerender/utils/utils.js:59` (details inline)

<details>

- suggestion: In `normalizePathnameWithQuery`'s catch block, `url.toLowerCase()` lowercases the entire raw string including any query-like portion. Since the catch path fires for unparseable strings this is low-risk, but if consistency with the happy-path intent matters, consider splitting on `?` before lowercasing only the left side - `src/prerender/utils/utils.js:78`
- suggestion: Add a unit test for percent-encoded paths (e.g. `/Page%2F1`) to document whether `%2F` becoming `%2f` is acceptable for downstream comparisons - `test/audits/prerender/utils.test.js`
</details>

src/prerender/utils/utils.js
```diff
@@ -56,7 +56,7 @@ function hasNonHtmlExtension(pathname) {
 export function toPathname(url) {
   try {
     const { pathname } = new URL(url);
-    return pathname === '/' ? pathname : pathname.replace(/\/$/, '');
+    return pathname === '/' ? pathname : pathname.replace(/\/$/, '').toLowerCase();
```
MysticatBot, line 59: issue (blocking): The catch block returns the raw `url` without `.toLowerCase()`, while `normalizePathnameWithQuery` (line 78) now lowercases its catch-block return. Since `toPathname` is used for dedup key building in edge-deployed pathname matching (`deployedAtEdgePathnames` set), a malformed URL that hits this catch path will not be case-normalized, preserving the original bug for that code path.

The PR description states the goal is to normalize case for dedup keys. This catch block is part of that contract.

**Fix:**
```js
} catch {
  return url.toLowerCase();
}
```

Then update the existing test assertion `toPathname('not-a-valid-url')` to expect the lowercased value.

MysticatBot (approved):
Hey @anuj-adobe,

**Verdict: Approve** - prior blocking finding is resolved; remaining items are non-blocking nits.
_Changes: adds `.toLowerCase()` to URL pathname normalization in two utility functions to prevent case-variant duplicate suggestions, plus a new behavior test layer (5 files)._

<details>

- nit: `normalizePathnameWithQuery` catch block lowercases the entire raw string including any query-like portion, inconsistent with the happy path that preserves query casing. Low-risk since unparseable URLs rarely carry meaningful query params, but a `url.indexOf('?')` split before lowercasing would make the contract consistent across both paths - `src/prerender/utils/utils.js:80`
- suggestion: Add a unit test for percent-encoded paths (e.g. `https://example.com/%41BC/Page` expecting `/abc/page`) to document the decode-then-lowercase ordering - `test/audits/prerender/utils.test.js`
</details>

- `toPathname` catch block now lowercases the raw URL, consistent with `normalizePathnameWithQuery`

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 37s | <b>Cost:</b> $2.52 | <b>Commit:</b> <code>578dbb463c8debc0589542b442110c831e772de2</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2667, opened 2026-06-15
feat(offsite-brand-presence): auto-trigger analysis audits after DRS completion

AndreiAlexandruParaschiv:
## Summary

After a manual `offsite-brand-presence` run, the reddit / youtube / cited analysis audits currently have to be triggered by hand. This automates that: once the DRS scrape jobs reach a terminal state, the existing DRS status poll handler triggers the downstream analysis audit for every domain whose jobs **succeeded**.

- Trigger point: `src/offsite-brand-presence/drs-status-handler.js`, right after it posts the completion summary.
- Audit types are **derived from the jobs that actually ran** (no hardcoded list):
  | Job domain | Audit triggered |
  |---|---|
  | `reddit.com` | `reddit-analysis` |
  | `youtube.com` | `youtube-analysis` |
  | `wikipedia.org` | `wikipedia-analysis` |
  | `top-cited` | `cited-analysis` |
- Only domains with a `COMPLETED`/`COMPLETED_WITH_ERRORS` job are triggered. `FAILED`, `CANCELLED`, and still-running-at-deadline domains are skipped (no fresh data). Each audit type triggers at most once even across multiple datasets.
- The originating Slack context is forwarded so each analysis audit posts its results to the same thread (these audits already support `auditContext.slackContext`).
- Triggering is **best-effort and isolated** in a try-catch, so a transient SQS error cannot fail the handler or cause the poll message to be redelivered (which would re-post the summary and re-trigger).

Reuses the same SQS chaining pattern the prompt-generation handler uses for `llmo-customer-analysis`. No changes to the analysis audits themselves.

### Scope
Fires for runs that go through the completion poll — i.e. **manual Slack-triggered `offsite-brand-presence` runs**, which is the manual flow being automated. Scheduled runs don't poll today; extending to those is a separate change.

## Test Plan

- [x] `npm test` — full suite passes, 100% lines/branches/statements coverage
- [x] `npm run lint` — clean
- [ ] Manual: trigger `offsite-brand-presence` from Slack; confirm that after the DRS completion summary, reddit/youtube/cited (and wikipedia when present) analysis audits run automatically and post to the same thread.
- [ ] Manual: confirm a domain whose scrape failed does not get its analysis audit triggered.


MysticatBot (approved):

**Verdict: Approve** - clean, well-scoped automation that follows established SQS chaining patterns.
_Changes: auto-triggers downstream analysis audits (reddit, youtube, cited, wikipedia) after DRS scrape jobs reach a terminal success state, with best-effort isolation (3 files)._

<details>

- nit: `triggerAnalysisAudits` fetches `Configuration.findLatest()` independently, though the handler already resolves it on the re-enqueue path; hoisting once and passing it in would avoid a redundant DB round-trip on the terminal path - `src/offsite-brand-presence/drs-status-handler.js:81`
- suggestion: the sequential `await sqs.sendMessage` loop could be `Promise.all` since the messages are independent and the set is bounded (max 4 types), matching how the handler already uses `Promise.all` for `drsClient.getJob` calls above - `src/offsite-brand-presence/drs-status-handler.js:85`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 41s | <b>Cost:</b> $3.01 | <b>Commit:</b> <code>a531537f9c4f0deebd277312d9857c33de40699a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/offsite-brand-presence/drs-status-handler.js
```diff
@@ -13,10 +13,75 @@
...
+ * @param {object} context - Universal context (sqs, dataAccess, log)
+ */
+async function triggerAnalysisAudits(statuses, siteId, slackContext, context) {
+  const { sqs, dataAccess, log } = context;
+
+  const auditTypes = new Set();
+  for (const s of statuses) {
+    if (DRS_SUCCESS_STATUSES.has(s.status)) {
+      const auditType = resolveAnalysisAuditType(s.domain);
+      if (auditType) {
```
rarescheseli, line 63: nit: we could trigger the audits from here, no need for the `for` loop below.

rarescheseli: never mind, we use the set to not trigger the audit twice for domains that have multiple scrape jobs (like `reddit_posts` and `reddit_comments`).

rarescheseli (approved):

PR #2670, opened 2026-06-15
fix(preflight): wait for meta tags so slow-rendering SPAs aren't captured as empty shells

clotton:
## Problem

Preflight reports a false **"h1 not found"** on slow-rendering AEM author SPAs. The content scraper's default profile captures the page after a fixed ~5s wait; on pages like Cox `business-internet` the `<h1>` paints later (~6–7s), so the scraper captures the ~492-byte empty React shell (`h1=0`) instead of the hydrated page.

Jira: [SITES-46324](https://jira.corp.adobe.com/browse/SITES-46324)

## Fix

Preflight scrapes now pass `waitTimeoutForMetaTags: 5000` in the content-scraper `options`. The scraper ([`abstract-handler.js`](https://github.com/adobe/spacecat-content-scraper/blob/main/src/handlers/abstract-handler.js), shipped in v1.72.7) waits for `title`/`h1`/`description` via a **non-throwing** `Promise.allSettled` before capturing, and resolves the instant they land:

- **Fast pages are unaffected** — resolves immediately once the tags are present.
- **Slow SPA pages are captured after hydration** instead of as an empty shell.
- **Non-throwing** — pages legitimately missing an `<h1>` are not failed; they just hit the cap.

One file of behavior change: [`src/preflight/handler.js`](src/preflight/handler.js).

## Validation

**Unit tests** — green at 100% coverage (lines/branches/statements). New focused test asserts the option is set; two existing `scrapePages` expectations updated.

**Live empirical proof** — ran the *real* merged scraper handler against the live Cox author page. Same page, only the wait strategy changed:

| Case | h1 | bodyLen |
|------|----|---------|
| Control (default 5s) | **0** | 492 (empty shell) |
| **`waitTimeoutForMetaTags: 5000`** (this PR's path) | **1** | 318,693 |

The false "h1 not found" is resolved.

## Known trade-off

The wait short-circuits on the first `<h1>`, so a very-slow page is captured mid-hydration (318KB at first-h1 vs ~421KB fully-settled). This fixes the h1 false-positive; fuller render-readiness detection is a scraping-layer concern (DRS roadmap), not preflight.

### DIFF
--- a/test/audits/preflight.test.js
+++ b/test/audits/preflight.test.js
@@ -23,6 +23,7 @@ import { TierClient } from '@adobe/spacecat-shared-tier-client';
 import {
   scrapePages, PREFLIGHT_STEP_SUGGEST, PREFLIGHT_STEP_IDENTIFY,
   AUDIT_BODY_SIZE, AUDIT_LOREM_IPSUM, AUDIT_H1_COUNT,
+  PREFLIGHT_META_TAGS_WAIT_MS,
 } from '../../src/preflight/handler.js';
 import { runLinksChecks } from '../../src/preflight/links-checks.js';
 import { MockContextBuilder } from '../shared.js';
@@ -536,6 +537,7 @@ describe('Preflight Audit', () => {
         options: {
           enableAuthentication: true,
           screenshotTypes: [],
+          waitTimeoutForMetaTags: PREFLIGHT_META_TAGS_WAIT_MS,
         },
       });
     });
@@ -568,8 +570,26 @@ describe('Preflight Audit', () => {
         options: {
           enableAuthentication,
           screenshotTypes: [],
+          waitTimeoutForMetaTags: PREFLIGHT_META_TAGS_WAIT_MS,
         },
       });
+    });
+
+    it('sets waitTimeoutForMetaTags so slow-rendering SPA pages are captured after hydration', async () => {
+      const context = {
+        site: { getId: () => 'site-123', getDeliveryType: () => Site.DELIVERY_TYPES.AEM_EDGE },
+        job: {
+          getMetadata: () => ({
+            payload: {
+              step: PREFLIGHT_STEP_IDENTIFY,
+              urls: ['https://main--example--page.aem.page'],
+            },
+          }),
+        },
+      };
+      const result = await scrapePages(context);
+      expect(result.options.waitTimeoutForMetaTags).to.equal(PREFLIGHT_META_TAGS_WAIT_MS);
+      expect(PREFLIGHT_META_TAGS_WAIT_MS).to.equal(5000);
     });
 
     it('includes promiseToken in options if context.promiseToken exists', async () => {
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @clotton,

**Verdict: Approve** - clean, focused fix with appropriate test coverage.
_Changes: Adds a `waitTimeoutForMetaTags` option to preflight scrape calls so slow-rendering SPA pages are captured after meta-tag hydration instead of as empty shells (2 files)._

<details>

- nit: The new test's first assertion (`expect(result.options.waitTimeoutForMetaTags).to.equal(PREFLIGHT_META_TAGS_WAIT_MS)`) is borderline tautological since it checks the option equals the same constant used to set it. The second line (`equal(5000)`) is the one that actually pins behavior. Consider collapsing to a single `expect(result.options.waitTimeoutForMetaTags).to.equal(5000)` for clarity. - `test/audits/preflight.test.js:593`
- nit: `PREFLIGHT_META_TAGS_WAIT_MS` is exported but used in exactly one place. A local (unexported) const would avoid accidental coupling from other modules importing it. The export exists so the test can reference it, but that couples the test to the implementation detail rather than the observable value. - `src/preflight/handler.js:46`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 44s | <b>Cost:</b> $2.03 | <b>Commit:</b> <code>139333b401db9478fa484b948258e58042bfbc03</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

GeezerPelletier:
**Verdict: Approve pending one question**

Clean, minimal fix — the approach is sound, the PR body's empirical proof (492B shell → 318KB hydrated) is exactly right, and the non-throwing `Promise.allSettled` path means legitimately-missing `<h1>` pages won't regress.

**One thing to confirm before merge:** does `package.json` enforce `spacecat-content-scraper >= 1.72.7`? The fix silently becomes a no-op if the scraper is pinned to an older version (no error, no test failure — just a regression back to the empty-shell capture). If the scraper is a separately-deployed service rather than an npm dependency, this doesn't apply, but worth a quick note either way.

Everything else looks good — constant is well-named and exported, options spread is idiomatic, test coverage is green.

GeezerPelletier (approved):

test/audits/preflight.test.js
```diff
@@ -568,10 +570,28 @@ describe('Preflight Audit', () => {
...
+            payload: {
+              step: PREFLIGHT_STEP_IDENTIFY,
+              urls: ['https://main--example--page.aem.page'],
+            },
+          }),
+        },
+      };
+      const result = await scrapePages(context);
+      expect(result.options.waitTimeoutForMetaTags).to.equal(PREFLIGHT_META_TAGS_WAIT_MS);
+      expect(PREFLIGHT_META_TAGS_WAIT_MS).to.equal(5000);
```
bhellema, line 592: I'm not sure of the value of testing the constant here.

PR #2673, opened 2026-06-16
fix(broken-backlinks): sync suggestion rank with traffic_domain on re-audit

jsanchez07:
## Problem

  The `rank` field on broken-backlinks suggestions diverges from
  `data.traffic_domain` over time. At creation both are set from the same
  Semrush value, but on every re-audit `syncSuggestions` calls
  `setData(mergedData)` — updating `data.traffic_domain` — while `rank`
  is never touched and stays frozen at the original creation value.

  This caused 129/168 suggestions on a representative staging opportunity
  to have `rank ≠ data.traffic_domain`, with many stuck at `2147483647`
  (INT32_MAX overflow from old raw Semrush visit counts that predated the
  0–100 authority score migration).

  ## Fix

  In the `toUpdate` loop in `syncSuggestions`, reuse the existing
  `mapNewSuggestion` function (already used at creation time) to extract
  and apply the updated rank on every re-audit:

  ```js
  const { rank } = mapNewSuggestion(newDataItem);
  if (rank !== undefined) {
    existing.setRank(rank);
  }
  
  For all other audit types mapNewSuggestion returns rank: 0, so the
  change is a no-op outside of broken-backlinks.

  Testing

  - Updated 19 mock suggestion objects in data-access.test.js to
  include setRank: sinon.stub()
  - Added assertion that setRank is called with the correct value in the
  OUTDATED→PENDING_VALIDATION update path
  - Verified pre-fix baseline in staging: 129/168 mismatched suggestions
  on opportunity 1a0899c7-17b3-460e-823f-89f50c79700c


src/utils/data-access.js
```diff
@@ -519,6 +519,11 @@ export async function syncSuggestions({
 
     warnOnInvalidSuggestionData(mergedData, opportunityType, log);
 
+    const { rank } = mapNewSuggestion(newDataItem);
+    if (rank !== undefined) {
```
mikeweis: Also guard against `rank` being `null`

mikeweis (approved):
Approved with needed update. 👍

PR #2675, opened 2026-06-16
fix: downgrade product-metatags empty-extraction log to warn + re-pin data-access 3.75.2 (pageCitability url-upsert)

solaris007:
Two recoverable-noise fixes - the last items from the prod ERROR-noise review.

## Change A - product-metatags empty-extraction -> warn
`src/product-metatags/handler.js:621` logged at error when `extractedTagsCount === 0`. That is a benign summary for non-product sites / empty scrapes (every non-product site hits it). The genuine per-object failures stay at error (handler.js:407 "No Scraped tags found in S3", :432 "Scrape result is empty"). Downgraded only the count==0 summary to warn (~143/6h of prod ERROR noise).

## Change B - re-pin data-access 3.75.2 (pageCitability url-upsert)
data-access 3.75.2 (adobe/spacecat-shared#1683) makes `PageCitability.create` upsert on the globally-unique `url` column, eliminating the read-then-insert duplicate-key race (Postgres 23505, ~100/hour in prod) between page-citability/handler.js:202 and prerender/handler.js:1401. No call-site change needed - the existing creates become race-safe `ON CONFLICT (url) DO UPDATE`.

## Testing
- `npm test`: 8603 passing, 0 failing, 100% coverage.
- lint clean.
- 3 product-metatags test assertions updated from error -> warn to match Change A (same message/args, no weakened coverage).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@adobe/mysticat-shared-seo-client": "1.6.0",
         "@adobe/spacecat-helix-content-sdk": "1.4.33",
         "@adobe/spacecat-shared-athena-client": "1.11.0",
-        "@adobe/spacecat-shared-data-access": "3.75.1",
+        "@adobe/spacecat-shared-data-access": "3.75.2",
         "@adobe/spacecat-shared-data-access-v2": "npm:@adobe/spacecat-shared-data-access@2.109.0",
         "@adobe/spacecat-shared-drs-client": "1.12.1",
         "@adobe/spacecat-shared-google-client": "1.6.3",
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @solaris007,

**Verdict: Approve** - clean, well-scoped noise reduction with correct test updates.
_Changes: Downgrades a benign zero-extraction-count log from error to warn and bumps data-access to 3.75.2 for race-safe PageCitability upsert (4 files)._

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 21s | <b>Cost:</b> $1.96 | <b>Commit:</b> <code>81f4b214a6401e49a266d77cecb412c84c505562</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2663, opened 2026-06-15
feat(offsite-brand-presence): report DRS job completion status in Slack

AndreiAlexandruParaschiv:
## Summary

Enhances the `offsite-brand-presence` audit so that, after it submits DRS scrape jobs and posts their `job_id`s to a Slack thread, it posts **one follow-up summary** to the same thread once all jobs reach a terminal DRS status (`COMPLETED`, `COMPLETED_WITH_ERRORS`, `FAILED`, `CANCELLED`) — or after a ~20-minute max-wait budget.

- New message type `offsite-brand-presence-drs-status`: a poll handler (`src/offsite-brand-presence/drs-status-handler.js`) that calls `drsClient.getJob()` per job and re-enqueues itself with an SQS delay until all jobs are terminal or the deadline passes, then posts the summary.
- The runner enqueues the poll only for **manual Slack runs** (channelId + threadTs present) with at least one successfully-submitted job; scheduled runs and all-failed submissions enqueue nothing.
- State (job list, Slack context, absolute deadline) rides inside the re-enqueued SQS message, so each poll is self-describing — no DB, no DRS metadata round-trip.
- **No changes to DRS, infrastructure, or the shared DRS client** — `getJob()` and the delayed-re-enqueue pattern already exist.

## Test Plan

- [x] `npm test` — full suite passes, 100% lines/branches/statements coverage
- [x] `npm run lint` — clean
- [ ] Manual: trigger `offsite-brand-presence` from Slack; confirm the job-id message posts immediately and a single completion summary follows in the same thread once jobs finish (or after ~20 min) with correct per-job statuses and error details for failures.
- [ ] Manual: confirm a scheduled (non-Slack) run posts no completion summary and incurs no polling.

PR #2682, opened 2026-06-16
feat(offsite-brand-presence): granular per-audit DRS scrape + analyze

AndreiAlexandruParaschiv:
## Summary

Today, `run audit reddit-analysis` / `youtube-analysis` / `cited-analysis` only *analyze* content that a prior **offsite-brand-presence** run already scraped — `filterUrlsByDrsStatus` is a DRS status lookup, not a scrape. Run standalone with nothing scraped, they fail with "no DRS content available."

This makes each analysis audit self-contained: when it finds no DRS content, it **requests a domain-scoped scrape** and analyzes once it completes — without triggering the other two audits.

## How it works

- **Self-healing, no new flags.** Each runner already catches `DrsNoContentAvailableError`. Now, instead of just failing, it enqueues a **domain-scoped `offsite-brand-presence` run** (`messageData.domainScope`) and returns `pending_scrape`.
- **`offsite-brand-presence` scoping:** `messageData.domainScope` restricts collection + scraping to a single bucket — an `OFFSITE_DOMAINS` key (`reddit.com`, `youtube.com`) or `top-cited` (→ cited-analysis).
- **Re-trigger via the existing poll:** the scoped run scrapes that domain, polls to completion (#2663), then re-triggers the analysis (#2667). By then its scraped content is available, so it analyzes.
- **Loop guard:** poll-triggered analyses now carry `drsScrapeRequested: true`. If a post-scrape analysis still finds no data, it fails gracefully instead of requesting another scrape. This also makes the normal offsite→analyze path never re-scrape.
- **Slack context forwarded**, so the scoped run's job-id/completion messages and the analysis results post to the same thread.

| Audit | Scoped scrape |
|---|---|
| reddit-analysis | `reddit.com` |
| youtube-analysis | `youtube.com` |
| cited-analysis | `top-cited` |

Reuses the existing collect→scrape→poll→trigger pipeline (no duplication); the shared `requestOffsiteScrape` helper lives in `offsite-audit-utils.js`.

## Test Plan

- [x] `npm test` — full suite passes (8644), 100% coverage
- [x] `npm run lint` — clean
- [ ] Manual: with nothing scraped, `run audit reddit-analysis` → posts a scoped DRS scrape, and after completion reddit-analysis runs on the scraped URLs — youtube/cited are not triggered.
- [ ] Manual: full `offsite-brand-presence` run still triggers all three analyses once (no re-scrape loop).

> Scope: the re-trigger relies on the completion poll, which runs for Slack-triggered runs (those carry a thread). Scheduled-run coverage remains the separate follow-up noted earlier.


src/utils/offsite-audit-utils.js
```diff
@@ -140,3 +140,28 @@ export function resolveMystiqueUrlLimit(auditContext, log, logPrefix) {
...
+ * analysis audit — by which point its scraped content is available.
+ *
+ * @param {object} context - Universal context (sqs, dataAccess, log)
+ * @param {string} siteId - The site ID
+ * @param {string} domainScope - An OFFSITE_DOMAINS key (e.g. 'reddit.com') or 'top-cited'
+ * @param {object} [slackContext] - Forwarded so notifications/results post to the thread
+ */
+export async function requestOffsiteScrape(context, siteId, domainScope, slackContext) {
+  const { sqs, dataAccess, log } = context;
+  const configuration = await dataAccess.Configuration.findLatest();
```
MysticatBot: issue (blocking): `requestOffsiteScrape` has no error handling around `Configuration.findLatest()` or `sqs.sendMessage()`. If either throws (transient DynamoDB timeout, SQS throttle), the exception propagates to the outer handler and the audit fails with an opaque infra error rather than the intentional `DrsNoContentAvailableError` message.

The existing `scheduleDrsStatusPoll` in the offsite-brand-presence handler wraps its equivalent SQS call in try/catch for exactly this reason.

**Fix:** Wrap the body in try/catch. On failure, log a warning and fall through to the `pending_scrape` return (the next scheduled run will attempt again). The audit should degrade gracefully rather than crash on a transient infra blip.

src/offsite-brand-presence/handler.js
```diff
@@ -810,9 +832,12 @@ export async function offsiteBrandPresenceRunner(finalUrl, context, site, auditC
 
   // Sort once, partition into per-domain + top-cited buckets
   const excludedFromTopCited = [...Object.keys(OFFSITE_DOMAINS), ...TOP_CITED_EXCLUDED_DOMAINS];
-  const {
-    topByDomain, topCited,
-  } = selectTopUrls(allUrls, DRS_URLS_LIMIT, excludedFromTopCited);
+  let { topByDomain, topCited } = selectTopUrls(allUrls, DRS_URLS_LIMIT, excludedFromTopCited);
+
+  if (domainScope) {
```
MysticatBot, line 853: issue (blocking): If `domainScope` does not match any `OFFSITE_DOMAINS` key or `'top-cited'`, `scopeBucketsToDomain` empties every bucket silently. The runner then triggers DRS scraping with zero URLs, schedules a status poll with zero jobs, and the downstream analysis audit re-triggers with `drsScrapeRequested: true` only to find no content and fail. This is a multi-hop silent no-op.

**Fix:** Validate early, fail fast. Check `domainScope` against `Object.keys(OFFSITE_DOMAINS)` plus `'top-cited'` and return an explicit error for unrecognized values.

MysticatBot (changes requested):

**Verdict: Request changes** - two robustness gaps that would cause confusing failures under transient infra issues or invalid input.
_Changes: adds self-healing DRS scrape requests to reddit/youtube/cited analysis audits with a loop guard and domain-scoped offsite-brand-presence runs (11 files)._

## Must fix before merge

1. [Important] Unhandled async failure in `requestOffsiteScrape` - callers sit inside a catch block but the function itself can throw on SQS/DynamoDB transients - `src/utils/offsite-audit-utils.js:157` (details inline)
2. [Important] `domainScope` not validated against known values - an unrecognized scope silently empties all buckets, producing a no-op scrape chain with no error signal - `src/offsite-brand-presence/handler.js:837` (details inline)

<details>

- nit: `log?.info` optional chaining at `src/utils/offsite-audit-utils.js:167` is misleading - `sqs` and `dataAccess` are used without guards on the same context object, so `log` is equally guaranteed present. Drop the `?.` for consistency.
- suggestion: Add a distinct `status` value (e.g. `no_content_after_scrape`) to the post-scrape failure return so downstream consumers can distinguish "scrape completed but yielded nothing" from a generic failure - `src/cited-analysis/handler.js:262`.
- suggestion: Verify whether `spacecatOrgId` (read by `offsiteBrandPresenceRunner` from `messageData`) is needed for DRS routing. If so, `requestOffsiteScrape` should forward it in `messageData` alongside `domainScope`.
- suggestion: Add a dedicated unit test block for `requestOffsiteScrape` in `test/utils/offsite-audit-utils.test.js` covering the `slackContext` absent branch and the `Configuration.findLatest()` rejection path.
</details>

Note: CI checks are still pending.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 46s | <b>Cost:</b> $6.70 | <b>Commit:</b> <code>c72ba188ab700845783de600a1b79aa755db894a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/utils/offsite-audit-utils.js
AndreiAlexandruParaschiv: Fixed in 74c32affa. `requestOffsiteScrape` now wraps the `Configuration.findLatest()` + `sqs.sendMessage()` calls in try/catch — a transient failure logs a warning and the analysis audit degrades to its `pending_scrape` result instead of throwing an opaque infra error. Added a test (`still returns pending_scrape when the scrape request fails`) covering the catch path.

src/offsite-brand-presence/handler.js
AndreiAlexandruParaschiv: Fixed in 74c32affa. The runner now validates `messageData.domainScope` against `VALID_DOMAIN_SCOPES` (`Object.keys(OFFSITE_DOMAINS)` + `'top-cited'`) and fails fast with an explicit `Unknown domainScope: <value>` error before any collection/scraping, instead of silently emptying every bucket. Added a test (`aborts with an explicit error for an unrecognized domainScope`).

MysticatBot (approved):

**Verdict: Approve** - both prior findings addressed cleanly, no new blocking issues.
_Changes: adds self-healing DRS scrape requests to reddit/youtube/cited analysis audits with a loop guard and domain-scoped offsite-brand-presence runs (11 files)._

<details>
<summary>Non-blocking (2): suggestions</summary>

- suggestion: Consider centralizing the domain-to-scope mapping (`'reddit.com'`, `'youtube.com'`, `'top-cited'`) in a shared constant in `offsite-audit-utils.js` rather than hardcoding in each handler. If a new domain is added, each handler must be updated independently today. - `src/reddit-analysis/handler.js:182`, `src/youtube-analysis/handler.js:182`, `src/cited-analysis/handler.js:265`
- suggestion: Document the loop-guard invariant on `triggerAnalysisAudits` - any future caller that triggers analysis audits after scraping MUST set `drsScrapeRequested: true`. The invariant is implicit today; a JSDoc note would protect against regression. - `src/offsite-brand-presence/drs-status-handler.js:77`
</details>

- `requestOffsiteScrape` now wraps Configuration/SQS calls in try/catch, degrading gracefully on transient failures (test confirms)
- `domainScope` validated against `VALID_DOMAIN_SCOPES` with fail-fast on unrecognized values (test confirms)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 26s | <b>Cost:</b> $8.79 | <b>Commit:</b> <code>0b4c82dafbbcab2f8dbfe40ad8e00423ca461d84</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2690, opened 2026-06-17
fix(cited-analysis): prevent SQS 256 KB message size limit failures

AndreiAlexandruParaschiv:
## Summary

- The `sendMystiqueMessagePostProcessor` was sending full URL Store objects enriched with unbounded `prompts` arrays, causing SQS to reject the `guidance:cited-analysis` message with _"Message must be shorter than 262144 bytes"_ (seen on lovesac.com with 50 URLs)
- Root cause: each URL accumulates one `prompts` entry per subPrompt across every topic that references it — 50 URLs × many topics × many subPrompts easily blows past 256 KB
- Added three layers of defence (matching the pattern in `llm-error-pages/handler.js`):
  1. **Field projection** — strip URL Store metadata (`siteId`, `byCustomer`, `audits`, timestamps) that Mystique never reads; only `url`, `categories`, `timesCited`, `prompts` are sent
  2. **Prompt cap** — `MAX_PROMPTS_PER_URL = 20` prevents runaway growth; Mystique uses prompts only for suggestion synthesis so beyond-20 entries have negligible impact
  3. **Size guard** — if the message still exceeds the 200 KB safety budget after projection and capping, URLs are dropped from the tail until it fits (audit proceeds with fewer URLs rather than failing entirely)

## Test plan

- [ ] All 43 existing cited-analysis handler tests pass
- [ ] New test: `should strip URL Store metadata and keep only url/categories/timesCited/prompts` — verifies field projection
- [ ] New test: `should reduce URL count when serialised message exceeds size budget` — verifies size guard triggers and reduces URL count with a warning log

### DIFF
--- a/src/cited-analysis/handler.js
+++ b/src/cited-analysis/handler.js
@@ -29,6 +29,15 @@ import { resolveBrandForSite, applyBrandScope } from '../utils/brand-resolver.js
 
 const LOG_PREFIX = '[Cited]';
 
+// SQS standard-queue maximum payload is 256 KB (262144 bytes). Stay under a
+// safety budget so worst-case serialisation doesn't hit the hard reject.
+const SQS_MAX_SAFE_BYTES = 200 * 1024;
+
+// Cap subPrompts carried per URL. Each URL can accumulate a subPrompt entry
+// for every topic it appears in; without a cap the prompts array alone can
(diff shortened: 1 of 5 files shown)


AndreiAlexandruParaschiv:
Superseded by #2691 — rebased cleanly onto main.

PR #2691, opened 2026-06-17
fix(cited-analysis): prevent SQS 256 KB message size limit failures

AndreiAlexandruParaschiv:
## Summary

- The `sendMystiqueMessagePostProcessor` was sending full URL Store objects enriched with unbounded `prompts` arrays, causing SQS to reject the `guidance:cited-analysis` message with _"Message must be shorter than 262144 bytes"_ (seen on lovesac.com with 50 URLs)
- Root cause: each URL accumulates one `prompts` entry per subPrompt across every topic that references it — 50 URLs × many topics × many subPrompts easily blows past 256 KB

**Fix (three commits):**

1. **Field projection + size guard** — strip URL Store metadata (`siteId`, `byCustomer`, `audits`, timestamps) Mystique never reads; only send `url`, `categories`, `timesCited`, `prompts`. Size guard drops URLs from the tail if the payload still exceeds the 200 KB safety budget.

2. **Cap at 40 URLs, keep full prompts** — introduce `CITED_ANALYSIS_URLS_LIMIT = 40` (cited-specific, doesn't touch the global 50 used by reddit/youtube). Each URL requires a full DRS scrape; 50 URLs was taking 30-40 min. With field projection, 40 URLs + full prompts fits within the SQS budget, so the `MAX_PROMPTS_PER_URL = 20` cap is not needed.

3. **Slack failure notification** — when the post-processor throws (SQS send fails, any reason), post a `:x:` message to the Slack thread that triggered the audit. Previously, Mystique was never reached so the guidance-handler never ran, leaving the operator with no feedback.

## Test plan

- [ ] All 46 cited-analysis handler tests pass
- [ ] `should strip URL Store metadata and keep only url/categories/timesCited/prompts`
- [ ] `should reduce URL count when serialised message exceeds size budget`
- [ ] `should limit URLs to CITED_ANALYSIS_URLS_LIMIT when many URLs exist`
- [ ] `should fall back to CITED_ANALYSIS_URLS_LIMIT when urlLimit is absent`
- [ ] `should post a Slack failure message when SQS send fails and slackContext is present`
- [ ] `should not post to Slack when SQS send fails without slackContext`

### DIFF
--- a/src/cited-analysis/handler.js
+++ b/src/cited-analysis/handler.js
@@ -30,6 +30,15 @@ import { resolveBrandForSite, applyBrandScope } from '../utils/brand-resolver.js
 
 const LOG_PREFIX = '[Cited]';
 
+// SQS standard-queue maximum payload is 256 KB (262144 bytes). Stay under a
+// safety budget so worst-case serialisation doesn't hit the hard reject.
+const SQS_MAX_SAFE_BYTES = 200 * 1024;
+
+// Cap subPrompts carried per URL. Each URL can accumulate a subPrompt entry
+// for every topic it appears in; without a cap the prompts array alone can
+// push a 50-URL payload well past 256 KB.
+const MAX_PROMPTS_PER_URL = 20;
+
 /**
  * Cited Analysis Audit Handler
  *
@@ -321,8 +330,20 @@ async function sendMystiqueMessagePostProcessor(auditUrl, auditData, context) {
     log.info(`${LOG_PREFIX} urlLimit=${urlLimit} (URLs sent to Mystique)`);
 
     const { urls, sentimentConfig } = storeData;
+    // Project only the fields Mystique reads (url, categories, prompts,
+    // timesCited) and cap prompts per URL so the payload stays bounded.
+    // URL Store metadata (siteId, byCustomer, audits, timestamps) is not
+    // needed downstream and contributes significant per-URL bloat.
     const enrichedUrls = enrichUrlsWithTopicData(urls, sentimentConfig.topics)
-      .slice(0, urlLimit);
+      .slice(0, urlLimit)
+      .map(({
+        url: urlStr, categories, timesCited, prompts,
+      }) => ({
+        url: urlStr,
+        ...(categories?.length > 0 && { categories }),
+        ...(timesCited > 0 && { timesCited }),
+        ...(prompts?.length > 0 && { prompts: prompts.slice(0, MAX_PROMPTS_PER_URL) }),
+      }));
 
     const baseMessage = {
       type: 'guidance:cited-analysis',
@@ -350,6 +371,22 @@ async function sendMystiqueMessagePostProcessor(auditUrl, auditData, context) {
     }
     const message = applyBrandScope(baseMessage, brand);
 
+    // Safety guard: if the serialised message still exceeds the budget after
+    // per-URL projection and prompt capping, drop URLs from the tail until
+    // it fits rather than letting SQS reject the send entirely.
+    let sentUrlCount = message.data.urls.length;
+    while (sentUrlCount > 1) {
+      const bytes = Buffer.byteLength(JSON.stringify(message), 'utf8');
+      if (bytes <= SQS_MAX_SAFE_BYTES) {
+        break;
+      }
+      sentUrlCount -= 1;
+      message.data.urls = enrichedUrls.slice(0, sentUrlCount);
+      log.warn(
+        `${LOG_PREFIX} Message size ${bytes} bytes exceeds budget; reducing to ${sentUrlCount} URLs`,
+      );
+    }
+
     log.debug(`${LOG_PREFIX} Built Mystique message type ${message.type}`);
     await sqs.sendMessage(env.QUEUE_SPACECAT_TO_MYSTIQUE, message);
     const scopeForLog = brand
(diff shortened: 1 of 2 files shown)


src/cited-analysis/handler.js
```diff
@@ -350,14 +375,30 @@ async function sendMystiqueMessagePostProcessor(auditUrl, auditData, context) {
     }
     const message = applyBrandScope(baseMessage, brand);
 
+    // Safety guard: if the serialised message still exceeds the budget after
```
MysticatBot, line 379: issue (blocking): The `while (sentUrlCount > 1)` condition guarantees at least one URL is always sent, but if that single remaining URL (with up to 20 capped prompts of arbitrary length) still exceeds the 200 KB budget, the message is sent anyway and SQS will reject it - the exact failure this PR aims to prevent.

**Why it matters:** The PR description notes URLs can have "many topics x many subPrompts" - a single URL with 20 long prompts and extensive category data could exceed 200 KB on its own.

**Suggested fix:** After the loop exits, add a final size check when `sentUrlCount === 1`. Either truncate prompts further on that last URL, or log an error and skip the send:

```js
if (sentUrlCount === 1) {
  const finalBytes = Buffer.byteLength(JSON.stringify(message), 'utf8');
  if (finalBytes > SQS_MAX_SAFE_BYTES) {
    log.error(
      `${LOG_PREFIX} Single-URL payload (${finalBytes} bytes) still exceeds budget; truncating prompts`,
    );
    message.data.urls[0].prompts = message.data.urls[0].prompts?.slice(0, 5);
  }
}
```

test/audits/cited-analysis/handler.test.js
```diff
@@ -695,12 +701,92 @@ describe('Cited Analysis Handler', () => {
...
+        },
+      };
+
+      const postProcessor = citedAnalysisHandler.default.postProcessors[0];
+      await postProcessor(baseURL, auditData, context);
+
+      const sentMessage = context.sqs.sendMessage.firstCall.args[1];
+      const sentUrl = sentMessage.data.urls[0];
+      expect(sentUrl).to.have.property('url', 'https://techreview.io/review-of-example');
+      expect(sentUrl).to.have.property('categories');
```
MysticatBot, line 759: issue (blocking): The `MAX_PROMPTS_PER_URL = 20` cap logic is never tested in isolation. The size-budget test creates URLs with exactly 20 prompts (the cap value itself), and the metadata-stripping test uses only 1 subPrompt. Neither exercises the path where a URL has >20 prompts and the cap trims it.

**Suggested fix:** Add a test case with a URL containing, say, 30 prompts and assert `sentMessage.data.urls[0].prompts.length === 20`.

MysticatBot (changes requested):

**Verdict: Request changes** - two gaps in the safety guard need addressing before this is production-ready.
_Changes: Adds three layers of SQS message size defense (field projection, prompt cap, size guard) to the cited-analysis handler to prevent 256 KB limit rejections (2 files)._

## Must fix before merge
1. [Important] Size guard has no protection when a single URL still exceeds budget - `src/cited-analysis/handler.js:378` (details inline)
2. [Important] No test exercises the `MAX_PROMPTS_PER_URL` cap - `test/audits/cited-analysis/handler.test.js:754` (details inline)

<details>

- nit: The trim loop re-serializes the full message on every iteration; a binary search or byte-per-URL estimate would converge faster for large payloads - `src/cited-analysis/handler.js:379`
- suggestion: Assert `Buffer.byteLength(JSON.stringify(sentMessage), 'utf8') <= 200 * 1024` in the size-guard test for a stronger proof the fix works - `test/audits/cited-analysis/handler.test.js:778`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 12s | <b>Cost:</b> $6.01 | <b>Commit:</b> <code>bb507d816da811aa1f693df69bea65881a3ed599</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):

**Verdict: Approve** - prior findings addressed, no blocking issues remain.
_Changes: Adds layered SQS message size defense (field projection, 40-URL cap, iterative trimming, last-resort prompt stripping) and Slack failure notification to the cited-analysis handler (2 files)._

<details>

- nit: The size-guard while-loop calls `JSON.stringify(message)` on each iteration, giving O(n) serialization passes bounded by the 40-URL cap. Not a problem today but worth a comment if the cap ever increases - `src/cited-analysis/handler.js:383`
- suggestion: When the size guard drops URLs, consider adding a `meta.truncatedFrom` integer to the message so Mystique can distinguish "site only has N URLs" from "payload was trimmed" - `src/cited-analysis/handler.js:376`
</details>

- Size guard now handles the single-URL edge case with last-resort prompt stripping
- MAX_PROMPTS_PER_URL cap removed in favor of the size guard approach (simpler, handles all cases)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 4s | <b>Cost:</b> $6.71 | <b>Commit:</b> <code>489b662fc8b87d264c5cb34574a2bfbebfd88919</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2692, opened 2026-06-17
feat: apply URL limit during DRS filtering instead of post-processor slicing

rarescheseli:
### Summary
Moves the `urlLimit` enforcement from the post-processor (`.slice()` after enrichment) into `filterUrlsByDrsStatus`, so URLs are capped earlier, before topic enrichment and the SQS payload is built.
* `filterUrlsByDrsStatus` accepts an optional limit parameter and stops collecting once reached
* `fetchStoreData` in cited, reddit, and youtube handlers now receives `urlLimit` and forwards it to the DRS filter
* `resolveMystiqueUrlLimit` is called before `fetchStoreData` instead of after
* removes redundant `.slice(0, urlLimit)` from all three post-processors

## Related Issues

Thanks for contributing!


src/utils/offsite-audit-utils.js
```diff
@@ -112,7 +121,15 @@ export async function filterUrlsByDrsStatus(urls, datasetIds, siteId, drsClient,
     );
   }
 
-  const filtered = urls.filter((item) => availableUrls.has(item.url));
+  const filtered = [];
+  for (const item of urls) {
+    if (availableUrls.has(item.url)) {
+      filtered.push(item);
+      if (limit && filtered.length >= limit) {
```
MysticatBot: issue (blocking): After the early-exit `break` here, the downstream log says `DRS availability filter: removed ${removed} URL(s) not yet scraped` - but `removed = urls.length - filtered.length` now conflates two causes: (1) URLs DRS marked unavailable, and (2) URLs that were available but skipped because the limit was reached.

An operator seeing "removed 460 URL(s) not yet scraped" would misattribute the cause when many of those URLs were actually available but past the cap.

Fix: note when the limit was the binding constraint, e.g.:
```js
if (limit && filtered.length === limit) {
  log?.info(`${prefix} URL limit reached (${limit}), additional available URLs not collected`);
}
```
Or compute `removed` only against URLs that were checked and rejected by DRS (not those skipped by the cap).

MysticatBot (changes requested):
Hey @rarescheseli,

**Verdict: Request changes** - one operational observability issue to address.
_Changes: moves URL limit enforcement from post-processor slicing into `filterUrlsByDrsStatus` to cap URLs earlier, before topic enrichment (8 files)._

## Must fix before merge
1. [Important] Log conflates DRS-filtered and limit-capped URLs - `src/utils/offsite-audit-utils.js:128` (details inline)

<details>

- nit: stale comment in cited post-processor references removed `.slice()` enforcement but the guarantee now depends on upstream DRS filtering - `src/cited-analysis/handler.js:342`
- nit: `limit` truthiness check (`if (limit && ...)`) treats `0` as "no limit"; a `limit > 0` guard would make the contract unambiguous - `src/utils/offsite-audit-utils.js:128`
- suggestion: post-processor logs `urlLimit=X (URLs sent to Mystique)` but no longer enforces that count; consider logging actual `enrichedUrls.length` instead - `src/reddit-analysis/handler.js:248`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 20s | <b>Cost:</b> $8.12 | <b>Commit:</b> <code>c02b84c3da0c4397135069a1e3fa3507a91b4a4d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @rarescheseli,

**Verdict: Approve** - prior blocking finding resolved, remaining items are non-blocking nits.
_Changes: moves URL limit enforcement from post-processor `.slice()` into `filterUrlsByDrsStatus` to cap URLs earlier, before topic enrichment (8 files)._

<details>

- nit: `limit` truthiness check (`if (limit && ...)`) treats `0` as "no limit" on the main path but `slice(0, 0)` would return empty on the fallback paths - inconsistent contract. Consider `limit != null && filtered.length >= limit` or documenting that `limit` must be a positive integer - `src/utils/offsite-audit-utils.js:253`
- nit: post-processor logs `urlLimit=X (URLs sent to Mystique)` but no longer enforces that count; the parenthetical suggests enforcement happens here when it now happens upstream - `src/reddit-analysis/handler.js:247`, `src/youtube-analysis/handler.js:247`
- nit: stale comment "CITED_ANALYSIS_URLS_LIMIT=40 keeps the payload within the SQS budget without capping per-URL prompts" references removed `.slice()` enforcement; the limit is now applied during DRS filtering, not at the post-processor stage - `src/cited-analysis/handler.js:370`
</details>

- Log conflation between DRS-filtered and limit-capped URLs now uses separate `drsRejected` counter and distinct log messages

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 52s | <b>Cost:</b> $3.15 | <b>Commit:</b> <code>caa1b531fd34e639ccad362e6fd6e2b1f6dceebc</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2693, opened 2026-06-17
feat(sqs): propagate organizationId to Mystique-bound messages

askayastha22:
**Jira:** [SITES-46272](https://jira.corp.adobe.com/browse/SITES-46272)

## What

Auto-attaches the organization ID (`organizationId`) to every message sent to the Mystique queue (`QUEUE_SPACECAT_TO_MYSTIQUE`), at the SQS transport layer in `src/support/sqs.js`.

Previously, none of the ~32 Mystique-bound `sqs.sendMessage` call sites included an organization ID — payloads only carried `siteId` and `deliveryType`. This change makes the org identity available to Mystique across the SpaceCat → Mystique boundary.

## How

Mirrors the existing `traceId` propagation pattern already in `sendMessage`:

- When the target queue is `QUEUE_SPACECAT_TO_MYSTIQUE`, the message carries a `siteId`, and no `organizationId` is already present, the org ID is resolved (`Site.findById(siteId)` → `site.getOrganizationId()`, a plain getter — no extra `Organization` lookup) and attached as top-level `organizationId` (camelCase, consistent with `siteId`/`deliveryType` and the inbound message convention).
- Resolution is **cached per invocation** (keyed by `siteId`) so a handler emitting many messages for the same site pays the `Site.findById` lookup only once — avoids the N+1 pattern.
- Resolution **never blocks the send**: a failed lookup logs a warning and the message is sent without `organizationId`.
- No-op for non-Mystique queues; generic `sendMessage` behavior is unchanged elsewhere.

### Why the transport layer (vs. per-handler)

There is no single chokepoint at the payload level — the ~32 Mystique payloads are built inline across many handlers. `SQS.sendMessage` is the single transport chokepoint, and the file already establishes this exact pattern for `traceId`. This covers **all current and future** Mystique sends with no per-handler changes and a single, concentrated test surface.

**Tradeoff:** the SQS support client now reaches into `context.dataAccess.Site` at runtime to resolve the org ID (gated to the Mystique queue, best-effort). It does so via the injected `context` — no new module import / no circular-dependency concern. The alternative — resolving once onto `context.organizationId` via middleware — keeps the support client fully decoupled but needs a reliable universal set-point across audit and guidance handlers.

## Testing

- `sqs.js`: **100% lines / branches / statements**
- 10 tests covering: happy path, unresolvable site (no org id), site-not-found, non-Mystique queue, missing queue env, pre-set `organizationId` not overwritten, missing `siteId`, cache reuse, resolution throwing, missing `dataAccess`.
- `npm run lint`: clean

## Related PRs

- **Consumer:** [experience-platform/mystique#2813](https://git.corp.adobe.com/experience-platform/mystique/pull/2813) — consumes this `organizationId` and lands it (plus `siteId`) on the Langfuse trace metadata of the SQS-driven guidance/detection/autofix crews + tasks.

### DIFF
--- a/src/support/sqs.js
+++ b/src/support/sqs.js
@@ -34,8 +34,10 @@ class SQS {
   }
 
   /**
-   * Resolves the organization ID for a site, caching the result (including null) per siteId.
-   * Never throws — a failed lookup must not block the message send.
+   * Resolves the organization ID for a site, caching only successful resolutions
+   * (including a genuine null) per siteId. Never throws — a failed lookup must not
+   * block the message send, and a transient failure is left uncached so later
+   * sends for the same site can retry.
    *
    * @param {string} siteId - The site ID carried on the message.
    * @returns {Promise<string|null>} The organization ID, or null if it can't be resolved.
@@ -45,17 +47,19 @@ class SQS {
       return this.#organizationIdCache.get(siteId);
     }
 
-    let organizationId = null;
     try {
       const { Site } = this.context?.dataAccess ?? {};
       const site = await Site?.findById(siteId);
-      organizationId = site?.getOrganizationId?.() ?? null;
+      const organizationId = site?.getOrganizationId?.() ?? null;
+      // Cache only on success. A transient Site.findById failure (timeout/throttle)
+      // must NOT be cached as a permanent null — that would suppress the org ID for
+      // every subsequent message for this site within the invocation.
+      this.#organizationIdCache.set(siteId, organizationId);
+      return organizationId;
     } catch (e) {
       this.log.warn(`Failed to resolve organizationId for Mystique message (siteId: ${siteId}): ${e.message}`);
+      return null;
     }
-
-    this.#organizationIdCache.set(siteId, organizationId);
-    return organizationId;
   }
 
   async sendMessage(queueUrl, message, msgGroupId, delaySeconds = 0, msgDedupId = undefined) {
--- a/src/support/sqs.js
+++ b/src/support/sqs.js
@@ -11,7 +11,6 @@
  */
 import { SendMessageCommand, SQSClient } from '@aws-sdk/client-sqs';
 import { hasText } from '@adobe/spacecat-shared-utils';
-import { getImsOrgId } from '../utils/data-access.js';
 
 /**
  * @class SQS utility to send messages to SQS
@@ -20,9 +19,9 @@ import { getImsOrgId } from '../utils/data-access.js';
  * @param {object} context - context object for trace ID propagation
  */
 class SQS {
-  // Per-invocation cache of resolved IMS org IDs, keyed by siteId, so a handler
+  // Per-invocation cache of resolved organization IDs, keyed by siteId, so a handler
   // that emits many Mystique messages for the same site only pays the DB lookup once.
-  #imsOrgIdCache = new Map();
+  #organizationIdCache = new Map();
 
   constructor(region, log, context) {
     this.sqsClient = new SQSClient({ region });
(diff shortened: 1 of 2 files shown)


src/support/sqs.js
```diff
@@ -29,6 +33,31 @@ class SQS {
...
+    let organizationId = null;
+    try {
+      const { Site } = this.context?.dataAccess ?? {};
+      const site = await Site?.findById(siteId);
+      organizationId = site?.getOrganizationId?.() ?? null;
+    } catch (e) {
+      this.log.warn(`Failed to resolve organizationId for Mystique message (siteId: ${siteId}): ${e.message}`);
+    }
+
+    this.#organizationIdCache.set(siteId, organizationId);
```
MysticatBot: issue (blocking): Transient DB failures get cached as a permanent `null` for the siteId.

When `Site.findById` throws (timeout, throttle), the catch block leaves `organizationId = null`, and then this line caches that null. All subsequent Mystique messages for the same siteId within this invocation will hit the cache and skip the lookup - even if the DB recovers immediately.

This conflates "site genuinely has no organizationId" (a correct, cacheable answer) with "the lookup failed" (a transient condition that should allow retry).

**Fix:** Move the `this.#organizationIdCache.set(siteId, organizationId)` call inside the try block (after the assignment from `site.getOrganizationId()`). Leave the catch path uncached so subsequent sends for the same site can retry.

```js
try {
  const { Site } = this.context?.dataAccess ?? {};
  const site = await Site?.findById(siteId);
  organizationId = site?.getOrganizationId?.() ?? null;
  this.#organizationIdCache.set(siteId, organizationId);  // cache only on success
} catch (e) {
  this.log.warn(`Failed to resolve organizationId for Mystique message (siteId: ${siteId}): ${e.message}`);
}

return organizationId;
```

This preserves the "never throws" contract while ensuring transient failures do not poison the cache for the rest of the invocation.

MysticatBot (changes requested):
Hey @askayastha22,

**Verdict: Request changes** - one correctness issue in the cache semantics.
_Changes: Adds automatic organizationId propagation to Mystique-bound SQS messages by resolving the org via Site.findById at the transport layer, with per-invocation caching (2 files)._

## Must fix before merge
1. [Important] Transient failures cached as permanent null, suppressing org ID for subsequent messages - `src/support/sqs.js:57` (details inline)

<details>

- nit: `siteId` is interpolated directly into the log string without sanitization; if structured logging is not guaranteed, control characters in a malformed siteId could forge log entries - `src/support/sqs.js:54`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 55s | <b>Cost:</b> $6.25 | <b>Commit:</b> <code>bcce66a4aa68ffbed725c54fbdbe1154b606ed8c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/support/sqs.js
askayastha22: Fixed in f39f0999 — moved the cache write inside the `try` so only successful resolutions (including a genuine `null`) are cached; the catch path now returns `null` uncached so later sends for the same site retry. Added a regression test asserting a transient first-call failure is not cached and the second send re-resolves. `sqs.js` stays at 100% coverage.

On the non-blocking nit (siteId in the warn string): leaving as-is — `siteId` is a system-generated UUID from the data layer (not external free-text) and direct interpolation of ids into logs is the established convention across this codebase. Happy to sanitize if preferred.

MysticatBot (approved):
Hey @askayastha22,

**Verdict: Approve** - cache-poisoning fix is solid, no new issues found.
_Changes: Adds automatic organizationId propagation to Mystique-bound SQS messages with per-invocation caching at the transport layer, fixing the previously flagged transient-failure cache poisoning (2 files)._

- Transient DB failures no longer cached as permanent null - cache write moved inside try block, regression test confirms retry semantics (f39f0999)

**Out of scope, worth tracking:** The SpaceCat-to-Mystique message schema in `mysticat-architecture/context/external-contracts.md` does not yet include the new `organizationId` field - consider updating that contract doc in a follow-up to mysticat-architecture.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 9m 59s | <b>Cost:</b> $4.98 | <b>Commit:</b> <code>f39f0999a6ed66c8861ea1a271d74bf369a1a184</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

papineni05 (approved):

PR #2697, opened 2026-06-18
fix(cited-analysis): cap prompts at 5 per URL to bound SQS payload

AndreiAlexandruParaschiv:
## What

Caps the per-URL `prompts` array sent to Mystique for cited-analysis at **5 prompts per URL** (`MAX_PROMPTS_PER_URL`), keeping the first 5 in stored order.

## Why

Investigating a real Lovesac payload showed the SQS size pressure comes from the **number of prompts per URL**, not the number of URLs. Individual URLs carried 100+ prompts (e.g. `wayfair`/`dreamsofa` ~150 each), totalling ~180–200 KB of prompt text across ~33 URLs — enough to blow the 200 KB SQS budget at the full URL count. Previously the code kept full prompts and compensated by silently dropping whole URLs from the tail (the size guard).

Capping prompts fixes the root cause: **all 50 URLs keep their data** and the payload drops to ~18 KB (50 URLs × 5 prompts × ~70 chars).

## Changes

- `src/cited-analysis/handler.js`: add `MAX_PROMPTS_PER_URL = 5` and slice each URL's prompts in the payload projection; refresh stale comments.
- Because the prompt cap removes the SQS-size pressure that motivated lowering the URL caps, `MYSTIQUE_URLS_LIMIT` returns to **50** and the cited-analysis cap aligns to **50**. The earlier 50→25 reduction is reverted.
- The existing size guard stays as a last-resort safety net for pathologically long prompt *strings*.

## Testing

- Added a test asserting prompts are capped to 5 in stored order.
- Reworked the size-guard test so it still exercises the guard after the prompt cap (long prompt strings).
- Full suite: **8668 passing**, **100% coverage**, lint clean.


MysticatBot (approved):

**Verdict: Approve** - clean, well-motivated fix that addresses the root cause of SQS payload bloat.
_Changes: caps per-URL prompts at 5 in the Mystique payload and reverts the cited-analysis URL limit to the global MYSTIQUE_URLS_LIMIT (50), keeping the size guard as a last-resort safety net (2 files)._

<details>

- nit: `MAX_PROMPTS_PER_URL` is handler-private but the test re-declares it as a mirror constant with a comment to keep in sync manually. If someone changes the handler value and forgets the test, the test silently passes with a stale value. Consider exporting it from a shared constants file. - `test/audits/cited-analysis/handler.test.js:27`
- suggestion: the old `CITED_ANALYSIS_URLS_LIMIT = 40` comment cited DRS scrape timing (30-40 min at 50 URLs) as a secondary motivation alongside SQS size. The prompt cap addresses the SQS concern but the DRS timing concern is orthogonal. Worth a one-line note in the PR description or code comment confirming that DRS timing at 50 URLs is acceptable in practice. - `src/cited-analysis/handler.js:270`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 13s | <b>Cost:</b> $4.35 | <b>Commit:</b> <code>a1a6eef0c37d90bfba2c649060ef6b7db47637db</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2699, opened 2026-06-18
fix(broken-internal-links): cache inconclusive URLs

oiatsiuk:
**Problem**
The Avalara broken-internal-links audit was getting hard-killed by AWS Lambda mid-batch. Inconclusive (ETIMEOUT) URLs were being re-validated on every request, causing cascading slowdowns on rate-limited sites. Combined with a timeout check that only fired between batches (every 10 pages), a single slow page could exhaust the remaining Lambda time before the next batch boundary.

**Solution**
Inconclusive URLs (e.g. ETIMEOUT) are now cached as working so they are not re-validated within the same audit run
Added a timeout check inside the per-link-batch loop (every 10 links within a page). When timeout is detected: cache is saved to S3, batch is released without being marked complete, continuation SQS message is sent, and the next Lambda re-processes the same batch using the warm cache

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/internal-links/crawl-batch-orchestration.js
+++ b/src/internal-links/crawl-batch-orchestration.js
@@ -182,7 +182,7 @@ export function createCrawlBatchOrchestration({
   }
 
   async function processOneBatch({
-    auditId, batchNum, batchStartIndex, batchSize, scrapeResultPaths, context, log,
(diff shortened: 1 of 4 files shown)


mikeweis (approved):
LGTM 👍

PR #2701, opened 2026-06-18
fix(preflight): throttle link-probe concurrency to prevent timeout false positives (SITES-46696)

clotton:
## Problem (SITES-46696)

Preflight's broken-links check probes **every link on a page at once** via an unbounded `Promise.all`. On a slow author environment this overwhelms the author: renders queue past the probe's 10s timeout, `tracingFetch` aborts, and `checkLinkStatus` records the timeout as **Status 0 / broken** — a flood of false "broken link" positives.

Micron hit this: **119 of 150** internal author-host links were reported broken, all working pages, all Status 0 from `Request timeout after 10000ms`.

Status 0 is a race: `(per-probe render time) × (probes hitting the author at once)` exceeding the 10s timeout. Firing all links simultaneously maximizes the left side.

## Fix

Route every probe through a shared concurrency limiter instead of unbounded `Promise.all`.

- `createConcurrencyLimiter(max)` — small queue-based limiter (no new dependency).
- Default **5**, overridable via `options.linkCheckConcurrency` or env `PREFLIGHT_LINK_CHECK_CONCURRENCY`.
- Capping concurrency keeps each probe's effective wait under the 10s window, so working links are no longer falsely flagged.
- Behaviour for genuinely broken links (404/410/5xx) is **unchanged** — only the all-at-once firing changes.

The default is deliberately conservative (correctness first); it's configurable so throughput can be tuned later.

## Validation

- **Unit:** 7 new tests + existing suite — limiter runs all tasks, never exceeds the cap, propagates rejections, falls back to serial on invalid input, and reads the option/env/default. `links-checks.js` at **100%** coverage (lines/branches/funcs/stmts); lint clean.
- **Mechanism (real pipeline):** the timeout→Status 0→false-broken path was reproduced end-to-end via an actual preflight run (95/150 on a slow AEM CS author), independent of this change.
- End-to-end before/after on the patched worker will be confirmed post-deploy by re-running preflight on the slow-author repro page (expect 0 broken where it was 95).

## Note on scope

This is the **interim fix in spacecat-audit-worker** to stop Micron's false positives now. The durable home is the links-audit migration (SITES-44289) — the same throttle should land in the new probe utilities so it isn't lost.

### DIFF
--- a/src/preflight/links-checks.js
+++ b/src/preflight/links-checks.js
@@ -16,6 +16,48 @@ import { getDomElementSelector, toElementTargets } from './utils/dom-selector.js
 import { DEFAULT_USER_AGENT } from '../internal-links/helpers.js';
 
 /**
+ * Default number of link probes performed concurrently. Kept low so the audit
+ * does not overwhelm a slow author environment: firing every link at once makes
+ * authenticated author renders queue past the per-request timeout, which the
+ * audit then records as Status 0 / broken (false positives — see SITES-46696).
+ * Override per run via options.linkCheckConcurrency or env
+ * PREFLIGHT_LINK_CHECK_CONCURRENCY.
+ */
+export const DEFAULT_LINK_CHECK_CONCURRENCY = 5;
+
+/**
+ * Minimal promise concurrency limiter. Returns a scheduler that runs at most
+ * `max` tasks at a time and resolves with each task's result. Falls back to
+ * serial (1) for invalid input.
+ *
+ * @param {number} max - maximum number of concurrently running tasks
+ * @returns {(task: () => Promise<*>) => Promise<*>} scheduler function
+ */
+export function createConcurrencyLimiter(max) {
+  const limit = Number.isFinite(max) && max >= 1 ? Math.floor(max) : 1;
+  let active = 0;
+  const queue = [];
+  const next = () => {
+    if (active >= limit || queue.length === 0) {
+      return;
+    }
+    active += 1;
+    const { task, resolve, reject } = queue.shift();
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @clotton,

**Verdict: Approve** - clean, well-tested interim fix with no blocking concerns.
_Changes: Adds a shared promise concurrency limiter to throttle link-probe requests, preventing timeout false positives on slow author environments (2 files)._

<details>

- nit: Concurrency resolution uses `||` so explicit `0` silently falls through to the default (5) rather than the limiter's serial fallback (1) - a `log.info` when the resolved value differs from what was passed would help operators debug config issues - `src/preflight/links-checks.js:243`
- suggestion: Add a multi-page integration test confirming the limiter is shared across all pages in a single `runLinksChecks` call (guards against a regression where limiter creation accidentally moves inside the page-processing loop) - `test/preflight/links-checks.test.js`
- suggestion: Consider capping the concurrency ceiling in `createConcurrencyLimiter` (e.g., `Math.min(Math.floor(max), 50)`) to protect against misconfiguration via env var with very large finite values - `src/preflight/links-checks.js:28`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 30s | <b>Cost:</b> $3.01 | <b>Commit:</b> <code>6e80cbbca9cdd2023f4858072ab889b6fd8a3e45</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2703, opened 2026-06-19
fix(prerender): prevent duplicate suggestion keys in guidance saveMany

ssilare-adobe:
## Summary

- Replace `toPathname` with `normalizePathnameWithQuery` in `guidance-handler.js` to preserve query parameters when keying suggestions, preventing key collisions for filter/faceted URLs
- Add defensive dedup guard (Set-based) before `saveMany` to prevent duplicate `suggestion_id`s in the batch

## Problem

The prerender guidance handler used `toPathname(url)` to build a lookup map of existing suggestions and to match incoming Mystique results. `toPathname` strips query parameters and hostname, so URLs like:

- `casio.com/us/watches/casio/standard.filter.metals-K84v...`
- `casio.com/us/watches/casio.filter.IN_STOCK-MTP-K84v...`

...collapsed to the same key. This caused the same suggestion object to be pushed to the `saveMany` batch twice, triggering PostgreSQL error:

```
ON CONFLICT DO UPDATE command cannot affect row a second time
(code: 21000)
```

**Affected sites (prod):** `casio.com` (320 suggestions), `aia.com.hk` (175 suggestions), and likely many of the 1,101 messages in the `spacecat-dead-letter-queue`.

## Root Cause

Keying mismatch between suggestion **creation** (uses `normalizePathnameWithQuery` — preserves query params) and suggestion **update** in the guidance handler (used `toPathname` — drops query params). Two distinct suggestions with different query params resolved to the same map key, so the second overwrote the first. Both Mystique results then matched the same suggestion object → duplicate IDs in the batch.

## Investigation

Full E2E trace via CloudWatch (audit-worker) + Splunk (Mystique prod):

| Time (IST) | System | Event |
|---|---|---|
| 13:03:07 | audit-worker | Received prerender trigger, queued to `spacecat-to-mystique` |
| 13:03:08 | Mystique | Picked up message, started processing |
| 13:11:14 | Mystique | `batch_complete` — 175/175 succeeded |
| 13:11:46 | Mystique | Sent result to `mystique-to-spacecat` ✅ |
| 13:11:46 | audit-worker | Received response, fetched presigned URL from S3 |
| 13:11:46 | audit-worker | ❌ `DataAccessError: Failed to save many` |

Mystique processed and delivered successfully. The crash was on the audit-worker side when persisting.

## Test plan

- [x] Add test: casio.com filter URLs with same pathname but different query-encoded suffixes are treated as distinct suggestions
- [x] Add test: plain pathname vs pathname+query are keyed distinctly
- [x] Add test: aia.com.hk domain variant + locale path URLs match correctly
- [x] Add test: dedup guard prevents duplicate suggestion IDs when URLs normalize to same key
- [x] Add test: large batch (50 filter URLs) processes without ID collisions
- [x] All 41 tests pass (36 existing + 5 new)


MysticatBot (approved):
Hey @ssilare-adobe,

**Verdict: Approve** - well-scoped fix with correct root cause analysis and solid test coverage.
_Changes: Replaces `toPathname` with `normalizePathnameWithQuery` in the guidance handler to preserve query parameters in suggestion map keys, and adds a Set-based dedup guard before `saveMany` to prevent duplicate ID batch errors (2 files)._

<details>

- nit: The dedup guard discards duplicates silently with no log output. Adding a `log.debug` inside the `if (seenIds.has(id))` branch (with the duplicate URL) would make this defensive path observable in production without adding noise. - `src/prerender/guidance-handler.js:202`
- nit: Quality metric counters (`valuableCount`, `validAiSummaryCount`, `suggestionsWithPrompts`) are incremented pre-dedup while `totalSuggestions` uses the post-dedup count, so logs could show `totalSuggestions=1, valuableSuggestions=2` when the guard fires. Consider adding a `dedupDroppedCount` to the log line, or noting this is by-design. - `src/prerender/guidance-handler.js:223`
</details>

Note: CI checks are still pending - verify build passes before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 16s | <b>Cost:</b> $4.20 | <b>Commit:</b> <code>ea98cd398dc23d87d854c02e4e42f10a1f63c033</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2706, opened 2026-06-20
fix: Sitemap audit now saves both "error" and "url" suggestions

mikeweis:
* enabled co-existence of both "error" and "url" suggestions

## Related Issues
 * Jira ticket -> https://jira.corp.adobe.com/browse/SITES-45985

Thanks for contributing!


src/sitemap/common.js
```diff
@@ -1113,6 +1115,18 @@ export async function getBaseUrlPagesFromSitemaps(inputUrl, initialUrls, log) {
           if (pages.length > 0) {
             pagesBySitemap[sitemapUrl] = pages;
           }
+        } else {
+          log?.warn(
+            `Sitemap: off-domain sitemap — skipping valid sitemap URL outside audit scope for ${inputUrl}: ${sitemapUrl}`,
+          );
+        }
+      } else {
+        const reason = sitemapData.reasons?.find((r) => r?.error);
```
oiatsiuk, line 1124: `find` will only capture the first error reason per sitemap — if `reasons` can ever contain multiple errors for a single sitemap, the rest would be silently dropped. Consider using `filter`. Is this a case worth guarding against?

mikeweis:
@oiatsiuk -- regarding the concern for line 1124 of **common.js** use of `find`.  Your concern is that if there were 2+ errors per sitemap.xml file, then using `find` would only return the first one.  Using `filter` would find 2+, if they existed.
 * Answer: These new / refined error messages can, at most, be on `1` sitemap.xml file.  Therefore using  `find` is correct.
 
Thanks for the concern! 👍

src/sitemap/handler.js
```diff
@@ -393,23 +404,43 @@ export function getPagesWithIssues(auditData) {
...
+  const { reasons, details } = auditData.auditResult;
 
-  const response = success
-    ? []
-    : reasons
-      .filter((reason) => reason?.error)
-      .map((reason) => buildErrorSuggestionFromReason(reason));
+  const errorSuggestions = buildErrorSuggestionsFromReasons([
+    ...(reasons ?? []),
+    ...(details?.sitemapErrors ?? []),
```
jsanchez07, line 437: In findSitemap, the no-valid-paths-extracted failure path now does this:

 ```
 return toPersistedAuditResult({
    reasons: [{
      value: filteredSitemapUrls[0],
      error: ERROR_CODES.NO_VALID_PATHS_EXTRACTED,
    }],
    url: inputUrl,
    details: { issues: notOkPagesFromSitemap, sitemapErrors },
  });
```

  In this failure case, sitemapErrors will contain whatever errors were collected during processing. But reasons also carries an error (NO_VALID_PATHS_EXTRACTED). When generateSuggestions merges [...reasons, ...details.sitemapErrors], you could end up with duplicate suggestions if the same broken sitemap is the cause of both — e.g., the one broken sitemap causes NO_VALID_PATHS_EXTRACTED in reasons AND also appears in sitemapErrors.
  
  The deduplication in buildErrorSuggestionsFromReasons handles exact key matches, so this is likely fine in practice — but worth confirming the test "should dedupe identical error suggestions" covers this end-to-end path rather than just the utility function in isolation.

jsanchez07 (approved):
Approve after the answer being posed here is answered

oiatsiuk (approved):

src/sitemap/handler.js
mikeweis: 👍  will fix & resolve

src/sitemap/common.js
mikeweis: 👍  (_see my above comment_)

PR #2719, opened 2026-06-23
fix(prerender): scope scraping URLs to site.baseUrl subpath (LLMO-5748)

dhavkuma:
## Summary

Fixes LLMO-5748 — top organic pages, scrape jobs, and suggestions for the prerender audit must be scoped to the site's configured `baseUrl`.

### Background

Two sites can share the same domain but be distinct entries (e.g. `nba.com` and `nba.com/kings` are separate sites). `SiteTopPage.allBySiteIdAndSourceAndGeo` fetches top pages by `siteId` only — no path filtering. For a site configured at `nba.com/kings`, top pages from `nba.com/lakers` stored under the same `siteId` would be scraped and generate suggestions, polluting results across sites.

The same gap existed for explicit/CSV URLs passed via `auditContext.urls`.

> **Note:** The original ticket description claimed these were already "innately scoped" — code review found they were not. Agentic/Athena URLs were the only source already correctly scoped (via `buildSiteFilters` SQL regex in `cdn-utils.js`).

### Changes

**`src/prerender/utils/site-scope.js`** — new utility (not imported from other audits):
- `isUrlWithinSite(url, site)` — checks whether a URL belongs to the site identified by `site.getBaseURL()`
- `filterUrlsBySite(urls, site)` — filters a URL array to the site's scope
- Internal `isUrlWithinBaseUrl` checks: hostname match (via `stripWWW`, www-agnostic), port match, and path prefix — consistent with `internal-links/subpath-filter.js`
- Accepts the full `site` object (not a raw baseUrl string) so future scoping rules can be added without changing call sites

**`src/prerender/handler.js`** — two guards added in `submitForScraping`:
- Top organic pages (`SiteTopPage`) — filtered after `rebaseUrl`, before dedup
- CSV/explicit `auditContext.urls` — filtered after `rebaseUrl`, before dedup

**No change for root-domain sites** — `filterUrlsBySite` is a no-op when `baseUrl` has no path, so existing behaviour is fully preserved.

**Updated test** — the LLMO-5145 test (`should include all top pages regardless of subpath`) shipped in #2696 was updated to reflect the corrected behaviour.

### Test plan

- [ ] `src/prerender/utils/site-scope.test.js` — 18 tests, 100% line/branch/statement coverage
  - null url, null site, missing `getBaseURL`
  - unparseable URL or baseUrl
  - hostname mismatch, port mismatch
  - www / non-www treated as same host
  - root-domain site — all URLs pass through
  - URL within site path, URL equal to site path exactly
  - `/kingsley` does not match `/kings` site (prefix ≠ path child)
- [ ] `test/audits/prerender/handler.test.js` — existing suite (261 tests) fully passing
  - `filters CSV auditContext.urls to site baseUrl`
  - `does not filter CSV auditContext.urls for root-domain sites`
  - `filters top organic page URLs to site baseUrl`
  - `does not filter top organic page URLs for root-domain sites`


dhavkuma:
@MysticatBot review

src/prerender/handler.js
```diff
@@ -986,8 +987,9 @@ export async function submitForScraping(context) {
   if (Array.isArray(auditContext?.urls) && auditContext.urls.length > 0) {
     const preferredBase = getPreferredBaseUrl(site, context);
     const rebasedCsvUrls = auditContext.urls.map((url) => rebaseUrl(url, preferredBase, log));
+    const scopedCsvUrls = filterUrlsBySite(rebasedCsvUrls, site);
```
MysticatBot: issue (blocking): `filterUrlsBySite` uses `site.getBaseURL()` for hostname+path matching, but the URLs have already been rebased to `preferredBase` (via `getPreferredBaseUrl`). When `overrideBaseURL` is configured in site config, or `finalUrl` resolves to a different domain, the rebased URLs have a different hostname than `site.getBaseURL()` - so the filter rejects them all.

CI confirms this: three existing tests fail (`uses overrideBaseURL from site config as domain for csvUrls rebasing`, `should fall back to top pages when baseUrl is empty`, `uses overrideBaseURL from site config as domain for organic and included URL rebasing`).

The fix needs to account for the fact that after `rebaseUrl`, URLs live on the `preferredBase` domain, not on `site.getBaseURL()` domain. Options:
- Filter BEFORE rebasing (against the original URL domain+path)
- Filter using `preferredBase` as the scope anchor instead of `site.getBaseURL()`
- Extract only the path-prefix check and skip hostname comparison when rebasing is involved

The path-prefix component of the check is the part that fixes LLMO-5748; the hostname comparison is what breaks the `overrideBaseURL` scenario.

MysticatBot (changes requested):
Hey @dhavkuma,

**Verdict: Request changes** - CI confirms the new filter breaks when `overrideBaseURL` or `finalUrl` differs from `site.getBaseURL()`.
_Changes: adds subpath-scope filtering to two URL sources in the prerender audit so locale-specific sites only scrape their own pages (4 files)._

## Must fix before merge
1. [Critical] `filterUrlsBySite` rejects all URLs when `preferredBase` differs from `site.getBaseURL()` - `src/prerender/handler.js:990,1037` (details inline)

<details>

- suggestion: the core algorithm in `site-scope.js` is near-identical to the existing `isWithinAuditScope` in `src/internal-links/subpath-filter.js` (same imports, same hostname+port+path logic). Consider reusing or promoting to a shared utility to avoid divergence - `src/prerender/utils/site-scope.js`
- suggestion: add a `log.info` or `log.debug` showing how many URLs were removed by site-scope filtering at each call site, so operators can diagnose silent empty-result scenarios - `src/prerender/handler.js:990`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 9m 55s | <b>Cost:</b> $6.16 | <b>Commit:</b> <code>87acdb1cb340f6c1b70a8fe90149ae1e0412ef7b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dhavkuma:
moved to https://github.com/adobe/spacecat-audit-worker/pull/2724

PR #2716, opened 2026-06-22
fix(prerender): revert to inline suggestions in Mystique SQS message

ssilare-adobe:
## Summary

- Reverts S3-based suggestion dispatch back to inline suggestions in the SQS message payload
- Caps inline suggestions at `MYSTIQUE_BATCH_SIZE` (320), matching the previous behavior
- S3 upload of suggestions to Mystique is currently broken; this restores the working inline path

## Changes

- `src/prerender/handler.js`: Removed S3 `PutObjectCommand` for suggestions upload; sends `data.suggestions` array inline with `batchIndex`/`totalBatches` fields instead of `suggestionsS3Key`/`suggestionsS3Bucket`
- `src/prerender/utils/constants.js`: Restored `MYSTIQUE_BATCH_SIZE` export (= `DAILY_BATCH_SIZE` = 320)
- `test/audits/prerender/ai-only-mode.test.js`: Updated assertions from S3 body parsing to SQS inline payload; added test for 320-cap behavior

## Test plan

- [x] `ai-only-mode.test.js` — 65 passing
- [x] `handler.test.js` — 257 passing
- [x] Lint clean on changed files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2709, opened 2026-06-22
feat(prerender): upload suggestions to S3 and send S3 link to Mystique

ssilare-adobe:
## Summary

- Upload all prerender suggestions to S3 (`prerender/mystique-suggestions/{opportunityId}.json`) and send only the S3 key via SQS to Mystique
- Removes the 256 KB SQS message size limit that previously capped batches at 320 suggestions
- Eliminates multi-batch sequential chaining — all suggestions are dispatched in a single message
- Mystique downloads suggestions from S3, processes them, and the guidance-handler cleans up the S3 file on completion
- Slack completion notifications preserved for Slack-triggered runs

## Flow

```mermaid
flowchart TD
    A[Trigger: Slack / Scheduled / Full prerender] --> B[sendPrerenderGuidanceRequestToMystique]
    B --> C[Upload all suggestions to S3]
    C --> D[Send SQS message with S3 key]

    D --> E[Mystique downloads from S3 and processes]
    E --> F[guidance-handler: save AI summaries]
    F --> G{mystiqueSession?}

    G -- No --> H[Done]
    G -- Yes --> I[Slack: AI summaries complete]
    I --> J[Delete S3 suggestions file]
    J --> K[Clear mystiqueSession]
    K --> H
```

## Test plan

- [x] All 430 prerender tests pass
- [x] 100% line/branch/statement coverage on changed files
- [x] Verified S3 upload always happens with correct key format
- [x] Verified SQS message contains `suggestionsS3Key` + `suggestionsS3Bucket` instead of inline `suggestions`
- [x] Verified `mystiqueSession` only saved when Slack context present
- [x] Verified S3 suggestions file deleted after Mystique completes
- [x] Verified cleanup errors are isolated (don't invalidate saved suggestions)

PR #2720, opened 2026-06-23
revert(prerender): revert inline suggestions revert (#2716)

ssilare-adobe:
## Summary

Reverts #2716 (fix(prerender): revert to inline suggestions in Mystique SQS message).

This restores the S3-based suggestion dispatch path that was introduced in #2709 — uploading suggestions to S3 and sending the S3 key/bucket to Mystique instead of inlining the full suggestions array in the SQS message.

## Changes reverted

- `src/prerender/handler.js`: Re-adds `PutObjectCommand` for S3 suggestions upload; sends `suggestionsS3Key`/`suggestionsS3Bucket` instead of inline `data.suggestions`
- `src/prerender/utils/constants.js`: Removes the `MYSTIQUE_BATCH_SIZE` export that was restored in #2716
- `test/audits/prerender/ai-only-mode.test.js`: Reverts assertions back to S3 body parsing

## Test plan

- [ ] Run `npm run test:spec -- test/audits/prerender/ai-only-mode.test.js`
- [ ] Run `npm run test:spec -- test/audits/prerender/handler.test.js`
- [ ] Lint clean on changed files

### DIFF
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -36,7 +36,7 @@ import {
   TOP_ORGANIC_URLS_LIMIT,
   PRERENDER_RECENT_PROCESSING_TIME_DAYS,
   MODE_AI_ONLY,
-  MYSTIQUE_BATCH_SIZE,
+  MYSTIQUE_SUGGESTIONS_S3_PREFIX,
 } from './utils/constants.js';
 import { isAiOnlyMode, buildUrlScopeForMode } from './mode-selector.js';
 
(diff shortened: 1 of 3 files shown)


MysticatBot (approved):
Hey @ssilare-adobe,

**Verdict: Approve** - clean revert restoring S3-based suggestion dispatch with consistent tests.
_Changes: Reverts #2716 to restore S3-based suggestion upload + SQS key-reference dispatch to Mystique, removing the inline batching path (3 files)._

<details>

- nit: Test assertions that parse `mockS3Client.send.firstCall.args[0].input.Body` without first asserting the command type are fragile if a future change adds another S3 call earlier in the flow - `test/audits/prerender/ai-only-mode.test.js:393`
- suggestion: If two audit runs for the same opportunityId overlap, the S3 key `prerender/mystique-suggestions/${opportunityId}.json` overwrites silently - consider appending a timestamp or run ID in a follow-up if concurrency is a concern - `src/prerender/handler.js:676`
</details>

Note: CI checks are still pending - confirm green before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 14s | <b>Cost:</b> $3.09 | <b>Commit:</b> <code>1106ae2a35c3fc3ecd855c82360599e081ba75b0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2723, opened 2026-06-23
fix: enable Bright Data URL validation by default and fix CDN-block false negatives

jsanchez07:
## Problem
  
  `BRIGHT_DATA_VALIDATE_URLS` defaulted to `false`, meaning Bright Data suggested
  URLs were never validated before being shown to customers. This allowed 404 URLs
  to reach the dashboard as actionable suggestions.

  Two additional bugs existed in `filterBrokenSuggestedUrls`:
  - The filter used `response.ok` (2xx only), which treated CDN blocks (403, 429)
    as broken URLs. Sites like ups.com that block bot HTTP requests had all their
    Bright Data suggestions incorrectly filtered out.
  - `fetch()` had no timeout, leaving requests open indefinitely if a site hung.

  ## Changes

  - **`src/backlinks/handler.js`**: Changed `BRIGHT_DATA_VALIDATE_URLS` default
    from `false` to `true`. The env var can still override this to `false` if needed.

  - **`src/utils/url-utils.js`**:
    - Filter now only removes confirmed 404s (`response.status !== 404`). CDN blocks
      (403, 429, 5xx) are treated as potentially valid pages and kept.
    - Added 5-second fetch timeout via `AbortController`. Timed-out URLs are
      filtered out (treated as unverifiable).
    - `fetchFn` is now injectable as an optional parameter to support unit testing
      of timeout behavior without ESM global stubbing.

  - **Tests**: Added coverage for CDN-block pass-through and fetch timeout behavior.
    Fixed three existing backlinks tests that relied on the old `false` default.

  ## Validation

  The `BRIGHT_DATA_VALIDATE_URLS` path only runs when both `BRIGHT_DATA_API_KEY`
  and `BRIGHT_DATA_ZONE` are present, so sites without Bright Data credentials are
  unaffected. A before/after comparison on a production site with Bright Data
  suggestions will be run after deploy to confirm 404 suggestions are filtered.

### DIFF
--- a/src/utils/url-utils.js
+++ b/src/utils/url-utils.js
@@ -173,7 +173,12 @@ export function isPreviewPage(url) {
   return urlObj.hostname.endsWith('.page');
 }
 
-export async function filterBrokenSuggestedUrls(suggestedUrls, baseURL) {
+export async function filterBrokenSuggestedUrls(
+  suggestedUrls,
+  baseURL,
+  timeoutMs = 5000,
+  fetchFn = fetch,
+) {
   // Strip www from both sides for consistent domain comparison
   const baseDomain = stripWWW(new URL(baseURL).hostname);
   const checks = suggestedUrls.map(async (suggestedUrl) => {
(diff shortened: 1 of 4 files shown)


src/utils/url-utils.js
oiatsiuk, line 206: if fetch throws a non-timeout error (e.g. DNS failure, connection refused), the outer catch silently returns null and filters the URL out. This was pre-existing behavior, but since validation is now enabled by default it will happen more often. Worth adding a log warning so these cases are visible in production logs

test/utils/url-utils.test.js
```diff
@@ -252,6 +252,43 @@ describe('filterBrokenSuggestedUrls', () => {
     const result = await utils.filterBrokenSuggestedUrls(suggestedUrls, baseURL);
     expect(result).to.deep.equal(['https://www.example.com/page2']);
   });
+
+  it('should keep URLs blocked by CDN (403, 429) since they may be valid pages', async () => {
```
oiatsiuk: tests cover 403 and 429, but the PR description also mentions 5xx. Could you add a case for 500 or 503 to confirm it's kept as well?

oiatsiuk (approved):

src/utils/url-utils.js
jsanchez07: I added logs for this

test/utils/url-utils.test.js
jsanchez07: I added those checks as well.

PR #2724, opened 2026-06-23
feat(prerender): filter URLs by site scope before scraping

dhavkuma:
## Summary

Adds site-scope filtering to the prerender audit's `submitForScraping` step so that only URLs within the site's configured base path are submitted for scraping. This prevents processing irrelevant pages when a site is scoped to a subpath (e.g. `bulk.com/uk`).

> **Reconciled with #2605 (LLMO-5145):** that PR scopes the **domain-wide suggestion** for subpath sites; this PR scopes the **URLs submitted for scraping**. Both behaviors are kept — `main` is merged in and the combined logic coexists.

### What changed

- **`src/prerender/utils/site-scope.js`** (new) — pure utility module:
  - `isWithinSiteScope(url, siteBaseUrl)` — `true` if the URL is within the site's base-path scope. Domain-only base URLs (no subpath) pass everything through. Handles relative URLs, www stripping, port comparison, and a trailing-slash guard against false-positive prefix matches (`/uk` vs `/ukraine`).
  - `filterBySiteScope(urls, siteBaseUrl)` — array filter over `isWithinSiteScope` (guards non-array input).

- **`src/prerender/handler.js`** — `submitForScraping` rebases all URL sources (CSV, included, organic top pages, agentic) onto the preferred base, then applies a **single** `filterBySiteScope` per merged candidate set:
  - **CSV branch** — filters the merged explicit URLs.
  - **Slack branch** — filters the merged organic + included set.
  - **Daily-batch branch** — filters the merged organic + included + agentic set **before** the `DAILY_BATCH_SIZE` slice, so out-of-scope URLs don't consume batch slots and starve in-scope ones.
  - `processContentAndGenerateOpportunities` keeps a defensive scope re-filter on scrape-result paths.

### Design decisions

- `isWithinSiteScope(url, null/'')` returns `true` — no `siteBaseUrl` means no restriction, so sites without a configured base URL are unaffected.
- Scope is applied to the **merged** candidate set (one filter per branch) rather than per-source. Scope is a per-URL predicate that commutes with dedup/recent/edge filtering, so the submitted-URL set is identical while the code is simpler.
- The scope filter runs **before** the daily-batch slice — placing it after would let out-of-scope URLs occupy slots and shrink the in-scope batch on subpath sites.
- Scoping uses `site.getBaseURL()`. **Known limitation:** if a subpath-scoped site's `preferredBase` (`overrideBaseURL`/`finalUrl`) resolves to a different host, scope filtering could over-drop; not handled here.

## Test plan

- [x] `test/audits/prerender/site-scope.test.js` — unit tests for `isWithinSiteScope`/`filterBySiteScope`: guard conditions, domain-only scope, relative/absolute URLs, www normalization, port mismatch, trailing slash, query strings, multi-segment subpaths, prefix false-positives, error handling.
- [x] `submitForScraping` integration tests in `handler.test.js` for scope filtering on CSV, included, and agentic URLs, plus a regression test that the scope filter runs **before** the daily-batch slice (in-scope URLs survive a full batch of out-of-scope ones).
- [x] Reconciled with #2605: combined the duplicate `Subpath URL scoping (LLMO-5145)` test blocks into one; both URL-filtering and subpath-suggestion behaviors covered.
- [x] 484 prerender tests passing, 0 failing.
- [x] 100% line/branch/statement/function coverage on `src/prerender/`.
- [x] ESLint clean.


src/prerender/utils/site-scope.js
```diff
@@ -0,0 +1,78 @@
+/*
```
dhavkuma: WIll move this utility to https://github.com/adobe/spacecat-shared/pull/1714

dhavkuma (commented):
utility to be moved to spacecat-shared via https://github.com/adobe/spacecat-shared/pull/1714

MysticatBot (changes requested):
Hey @dhavkuma,

**Verdict: Request changes** - three issues need resolution before merge.
**Complexity: HIGH** - large diff (600+ lines across production code and tests).
_Changes: Adds site-scope URL filtering to the prerender audit so only URLs within the site's configured base path are submitted for scraping (4 files)._
Note: Recommend a human read before merge - possible divergence from the documented design (heuristic, flagged by the conventions reviewer). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge

1. [Important] PR description contradicts implementation on organic top pages - `src/prerender/handler.js:585` - The description says "Organic top pages - intentionally not filtered" but the code applies `filterBySiteScope` to `topPagesUrls` and the test confirms filtering IS applied. Either update the description or remove the filter from this path.

2. [Important] Potential hostname mismatch between rebase target and scope filter - `src/prerender/handler.js:536` - URLs are rebased to `preferredBase` (from `getPreferredBaseUrl`, which may return an `overrideBaseURL`) then scope-checked against `site.getBaseURL()`. If these diverge beyond www normalization, the host comparison in `isWithinSiteScope` will reject all rebased URLs, producing a silently empty list. Fix: verify/document that these always share the same hostname, or apply scope filtering before rebasing.

3. [Important] No observability when scope filtering drops URLs - `src/prerender/utils/site-scope.js:75` - `filterBySiteScope` silently removes URLs across 5 call sites with no log output. A misconfigured `siteBaseUrl` would drop all URLs with the only symptom being `submittedUrls=0` in downstream metrics. Fix: log the count of scope-filtered URLs at each call site.

<details>

- nit: `filterBySiteScope` at `src/prerender/handler.js:1222` re-filters URLs in `processContentAndGenerateOpportunities` that already passed scope check in `submitForScraping` - add a comment explaining the defensive intent, or remove if purely redundant.
- suggestion: Add a guard `if (!Array.isArray(urls)) return [];` in `filterBySiteScope` before this moves to spacecat-shared - callers outside this repo won't have the same input guarantees. - `src/prerender/utils/site-scope.js:75`
- nit: The bare `catch` at `src/prerender/utils/site-scope.js:68` swallows all exceptions (not just URL parse errors). Consider catching only `TypeError` or logging a warning so upstream regressions in `prependSchema`/`stripWWW` don't silently reject all URLs.
- suggestion: Add test cases for URLs with query strings (`?foo=bar`) and multi-segment base paths (`/eu/uk/en`) to confirm pathname-only matching works as expected. - `test/audits/prerender/site-scope.test.js`

</details>

Note: CI checks are currently pending - verify before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 52s | <b>Cost:</b> $5.82 | <b>Commit:</b> <code>98e85725f5a1fb15a26ffe46a6ab065cd369673f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anuj-adobe:
do we really need to filter each type differently? or we can simply filter out urls that are not in the scope? I think that way it would add only one line more in `submitForScrapping`

dhavkuma:
Update on this PR.

**Reconciled with #2605 (LLMO-5145).** Since #2605 (subpath-scoped domain-wide suggestion) merged to `main` and takes a different approach to the same ticket, I merged `main` in and kept **both** behaviors: URL scope-filtering (this PR, in `submitForScraping`) stays active, and #2605's subpath-scoped suggestion logic is preserved. The two duplicate `Subpath URL scoping (LLMO-5145)` test blocks were consolidated into one.

**Refactor:** the four per-source `filterBySiteScope` calls are now a single scope filter on each merged candidate set, applied **before** the `DAILY_BATCH_SIZE` slice so out-of-scope URLs can't consume batch slots and starve in-scope ones. Added a regression test that fails if that ordering is reversed.

**Status of the review items:**

- **#1 (description vs. code on organic top pages)** — resolved; the description matches the code (organic top pages *are* scope-filtered).
- **#2 (host divergence between rebase target and scope base)** — not changed. The case is narrow: it only triggers when a site is subpath-scoped **and** `preferredBase` (`overrideBaseURL`/`finalUrl`) resolves to a different host than `site.getBaseURL()`. Scoping is intentionally kept against `site.getBaseURL()`; left as a known limitation rather than expanding this PR.
- **#3 (no logging when scope filtering drops URLs)** — not added in this PR.
- **Non-blocking:** array guard, defensive re-filter comment, and query-string/multi-segment tests are all in. The bare `catch` in `site-scope.js` is intentionally a catch-all (returns `false` on any parse failure).

484 prerender tests passing, 100% line/branch/statement/function coverage on `src/prerender/**`.

Note: the scope utility is slated to move to spacecat-shared (adobe/spacecat-shared#1714).

src/prerender/handler.js
```diff
@@ -538,14 +539,15 @@ export async function submitForScraping(context) {
...
     const rebasedCsvUrls = auditContext.urls.map((url) => rebaseUrl(url, preferredBase, log));
-    const { urls: explicitUrls, filteredCount } = mergeAndGetUniqueHtmlUrls(
+    const { urls: mergedCsvUrls, filteredCount } = mergeAndGetUniqueHtmlUrls(
       rebasedCsvUrls,
       { includeQueryParams: true },
     );
+    const explicitUrls = filterBySiteScope(mergedCsvUrls, site.getBaseURL());
 
     log.info(`
     ${LOG_PREFIX} prerender_submit_scraping_metrics:
```
MysticatBot, line 554: issue (blocking): **Hostname mismatch between preferredBase and scope-check target can silently drop all URLs.**

URLs are rebased to `preferredBase` (from `getPreferredBaseUrl` - returns `overrideBaseURL` > `context.finalUrl` > `site.getBaseURL()`), then scope-filtered against `site.getBaseURL()`. When these differ in hostname, `isWithinSiteScope` rejects every rebased URL because `stripWWW(parsedUrl.hostname) !== stripWWW(parsedBaseURL.hostname)` at line 63 of site-scope.js.

Engaging with the prior pushback: the argument that "the case is narrow" (requires subpath-scoped site AND hostname divergence) is valid on probability, but the consequence is severe - ALL submitted URLs are silently dropped, producing `submittedUrls=0` with no error. Combined with zero observability (finding 2), this creates a production debugging nightmare: the audit silently produces nothing for affected sites, and no log or metric reveals why.

Fix options (pick one):
- Scope-check against `preferredBase` instead of `site.getBaseURL()` (since that is the hostname URLs actually carry post-rebase)
- Add a guard: if scope filter removes ALL URLs, log a warning with both hostnames so the mismatch is diagnosable
- Assert hostname compatibility at the top of `submitForScraping` and skip scope filtering when they diverge

src/prerender/handler.js
```diff
@@ -662,7 +664,10 @@ export async function submitForScraping(context) {
       [...organicDeduped, ...includedDeduped, ...agenticDeduped],
       { includeQueryParams: true },
     );
-    const batchedUrls = crossDeduped.slice(0, DAILY_BATCH_SIZE);
+    // Single site-scope filter on the merged candidate set, applied before the daily-batch
```
MysticatBot, line 671: issue (blocking): **No observability when scope filtering removes URLs.**

The `filteredCount` metric at lines 683+ captures dedup removals but not scope removals. An operator investigating `submittedUrls=0` would see `filteredOutUrls=0` and have no way to know scope filtering dropped everything.

Acknowledging the logging commit was added then reverted - but the underlying concern remains: silent total data loss with no diagnostic signal. This compounds finding 1: the hostname mismatch scenario is already narrow and hard to reproduce, and the absence of any metric makes it invisible in production.

Minimal fix: compute `scopeFilteredCount = preFilterLength - postFilterLength` and add it to the existing metrics log line. This is a single arithmetic operation per call site, not a per-URL log.

src/prerender/utils/site-scope.js
```diff
@@ -0,0 +1,81 @@
...
+    const normalizedBaseHost = stripWWW(parsedBaseURL.hostname);
+    if (normalizedUrlHost !== normalizedBaseHost || parsedUrl.port !== parsedBaseURL.port) {
+      return false;
+    }
+
+    return parsedUrl.pathname.startsWith(basePathWithSlash) || parsedUrl.pathname === basePath;
+  } catch {
+    return false;
+  }
+}
```
MysticatBot: issue (blocking): **Bare catch swallows all exceptions including unexpected ones.**

The try block contains calls to `prependSchema`, `stripWWW`, `new URL()`, property access on the results, and string operations. The catch returns `false` for all of them indiscriminately. If `prependSchema` or `stripWWW` regresses (e.g., throws TypeError on a previously-valid input after a shared-utils upgrade), every call to `isWithinSiteScope` would return `false`, and `filterBySiteScope` would produce an empty array - silently zeroing out all audit traffic.

This is fail-closed for a single URL (acceptable), but the same catch covers the `siteBaseUrl` parse (lines 37-38) which is constant across all URLs in one invocation. A parse failure there drops EVERY URL in the batch.

Fix: catch only TypeError (the expected exception class from `new URL()`) or at minimum log a warning on the first unexpected exception so the failure is observable.

MysticatBot (changes requested):
Hey @dhavkuma,

**Verdict: Request changes** - three issues need resolution before merge.
**Complexity: HIGH** - large diff (500+ changed lines across production and test code).
_Changes: Adds site-scope URL filtering to the prerender audit's submitForScraping step so only URLs within the site's configured base path are submitted for scraping (4 files)._
Note: Recommend a human read before merge - possible divergence from the documented design (heuristic, flagged by the conventions reviewer). The bot review is a complement to, not a replacement for, a human read here.
Note: CI checks are currently pending - verify before merge.

## Must fix before merge

1. [Important] Scope filter can silently drop ALL URLs for subpath-scoped sites when preferredBase hostname differs from site.getBaseURL() - `src/prerender/handler.js:553` (details inline)
2. [Important] No observability when scope filtering removes URLs - operators cannot diagnose silent empty lists - `src/prerender/handler.js:667` (details inline)
3. [Important] Bare catch block swallows all exceptions and returns false - an upstream regression in prependSchema/stripWWW would silently zero-out all submitted URLs - `src/prerender/utils/site-scope.js:68` (details inline)

<details>

- nit: Relative URL branch does not normalize path-traversal segments - `/uk/../fr/page` passes raw `startsWith('/uk/')` check. Not reachable in current handler usage (all URLs are absolute after rebase), but will matter when this moves to spacecat-shared. - `src/prerender/utils/site-scope.js:54`
- nit: Relative URL exactly at basePath with a query string (`/uk?foo=bar`) is incorrectly rejected by the `url === basePath` check. Same reachability caveat applies. - `src/prerender/utils/site-scope.js:55`

</details>

- PR description now correctly states organic top pages ARE filtered by scope (matches implementation).
- Consolidation from four per-source filterBySiteScope calls to single filter on merged candidate set (addresses anuj-adobe's simplification question).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 16m 49s | <b>Cost:</b> $7.72 | <b>Commit:</b> <code>c4b2cc2ac029be87887ecd08173005408246eb63</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/prerender/handler.js
dhavkuma: Added a warning when scope filtering reduces a non-empty candidate set to empty, logging both `preferredBase` and `siteBaseUrl` so the hostname mismatch is diagnosable (`applySiteScopeFilter` in handler.js). Filtering behavior is unchanged — the narrow mismatch case is now observable instead of silent. (bacfa126)

src/prerender/handler.js
dhavkuma: Added `scopeFilteredUrls` (preFilter − postFilter) to all three `prerender_submit_scraping_metrics` log lines (CSV/Slack/non-Slack), so scope removals are visible separately from the dedup `filteredOutUrls`. (bacfa126)

src/prerender/utils/site-scope.js
dhavkuma: Resolved by removing the local util: `site-scope.js` (and its test) is deleted and the prerender handler now consumes `filterBySiteScope` from `@adobe/spacecat-shared-utils` (bumped to 1.123.1, where the util landed via adobe/spacecat-shared#1714). The bare-catch concern is tracked as a follow-up in spacecat-shared. (c8326896)

dhavkuma:
@anuj-adobe re: [your question](https://github.com/adobe/spacecat-audit-worker/pull/2724#issuecomment-4799637580) on filtering each type differently — agreed, and that's how it works now. The per-source `filterBySiteScope` calls were consolidated into a single scope filter on the *merged* candidate set in `submitForScraping` (applied before the `DAILY_BATCH_SIZE` slice so out-of-scope URLs don't consume batch slots). The util itself has since moved to `@adobe/spacecat-shared-utils` (1.123.1, via adobe/spacecat-shared#1714), so the handler just imports `filterBySiteScope` from there.

src/prerender/handler.js
```diff
@@ -1217,7 +1269,12 @@ export async function processContentAndGenerateOpportunities(context) {
     // Skip expensive URL fetching and comparison when domain is known to be bot-blocked
     if (!isDomainBlocked) {
       if (scrapeResultPaths?.size > 0) {
-        urlsToCheck = Array.from(context.scrapeResultPaths.keys());
+        // Defensive re-filter: scrape result paths should already be in-scope (submitForScraping
+        // scoped them before submission), but a stale scrape job could carry out-of-scope paths.
+        urlsToCheck = filterBySiteScope(
+          Array.from(context.scrapeResultPaths.keys()),
```
MysticatBot: issue (blocking): **Defensive re-filter bypasses the observability wrapper this PR establishes.**

The three `submitForScraping` branches all route through `applySiteScopeFilter`, which counts removals and warns on complete elimination. This defensive re-filter calls `filterBySiteScope` directly:

```js
urlsToCheck = filterBySiteScope(
  Array.from(context.scrapeResultPaths.keys()),
  site.getBaseURL(),
);
```

If a stale scrape job carries out-of-scope paths that get dropped here, the removal is silent. The subsequent log reports `urlsToCheck.length` post-filter, so operators cannot distinguish "5 URLs scraped successfully" from "8 URLs scraped, 3 silently dropped by scope guard." This defeats the observability investment made in the rest of the PR.

**Fix:** Use `applySiteScopeFilter` here too (pass `preferredBase` from context if available, or a placeholder like `"n/a (defensive re-filter)"`), or add a local `log.info` when the delta is nonzero. The contract this PR established is: any time the scope filter removes URLs, there is a log signal.

MysticatBot (changes requested):
Hey @dhavkuma,

**Verdict: Request changes** - one inconsistency in the observability contract this PR establishes.
**Complexity: HIGH** - large diff (500+ changed lines across production and test code; dependency update).
_Changes: Adds site-scope URL filtering to the prerender audit so only URLs within the site's configured base path are submitted for scraping, with observability via a wrapper function and metrics (3 files)._
Note: CI checks are currently pending - verify before merge.

## Must fix before merge

1. [Important] Defensive re-filter bypasses the observability wrapper - `src/prerender/handler.js:1272` (details inline)

<details>

- nit: `hasRecentOrganic` comparison changed from `topPagesUrls.length` to `rebasedTopPagesUrls.length` but `rebaseUrl` is a 1:1 map so the values are always identical - a one-line comment noting this invariant would prevent future misreads. - `src/prerender/handler.js:681`

</details>

- Hostname mismatch between preferredBase and siteBaseUrl now emits a warning via `applySiteScopeFilter` (bacfa126).
- Scope filtering count now reported as `scopeFilteredUrls` in all three metrics log lines (bacfa126).
- Bare catch block removed - utility moved to `@adobe/spacecat-shared-utils` with follow-up tracked in spacecat-shared (c8326896).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 46s | <b>Cost:</b> $5.65 | <b>Commit:</b> <code>902c66a4b7cc5fbbd1df3e43f00c0a55722a2b6c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/prerender/handler.js
dhavkuma: Good catch — routed the defensive re-filter through `applySiteScopeFilter` too, so a stale scrape job dropping out-of-scope paths is no longer silent. It now warns on complete elimination and logs the dropped count when non-zero (`Defensive scope re-filter dropped N out-of-scope path(s)…`). `preferredBase` isn't meaningful here, so it's passed as `"n/a (defensive re-filter)"`. (164f0cd6)

MysticatBot (approved):
Hey @dhavkuma,

**Verdict: Approve** - prior finding resolved cleanly, no new issues.
**Complexity: MEDIUM** - medium diff (690 lines across production and test code; dependency update).
_Changes: Routes the defensive scope re-filter through `applySiteScopeFilter` for consistent observability, plus a main merge (3 files)._

<details>

- nit: Spurious blank line introduced in `package-lock.json` between `"version"` and `"resolved"` for `@adobe/spacecat-shared-utils` - likely an artifact of the registry URL change. Harmless but unusual; the next clean `npm install` should normalize it. - `package-lock.json:5485`

</details>

- Defensive re-filter now routes through `applySiteScopeFilter` with appropriate context (`'n/a (defensive re-filter)'`), giving the same observability guarantees as submission-time filtering (164f0cd).
- Hostname mismatch between preferredBase and siteBaseUrl now emits a warning via `applySiteScopeFilter` (bacfa126).
- Scope filtering count now reported as `scopeFilteredUrls` in all three metrics log lines (bacfa126).
- Bare catch block removed - utility moved to `@adobe/spacecat-shared-utils` with follow-up tracked in spacecat-shared (c8326896).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 14s | <b>Cost:</b> $3.92 | <b>Commit:</b> <code>9453a3b8a137d72149fbf34d44c8bd95397f6ebb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/prerender/handler.js
```diff
@@ -538,14 +562,21 @@ export async function submitForScraping(context) {
 
   const siteId = site.getId();
   const isSlackTriggered = !!(auditContext?.slackContext?.channelId);
+  const preferredBase = getPreferredBaseUrl(site, context);
```
dhavkuma, line 541: @anuj-adobe refactor: this change is to move up the var init - this is used in multple places.
old L:543 & L:590

src/prerender/handler.js
```diff
@@ -538,14 +562,21 @@ export async function submitForScraping(context) {
...
 
   if (Array.isArray(auditContext?.urls) && auditContext.urls.length > 0) {
-    const preferredBase = getPreferredBaseUrl(site, context);
     const rebasedCsvUrls = auditContext.urls.map((url) => rebaseUrl(url, preferredBase, log));
-    const { urls: explicitUrls, filteredCount } = mergeAndGetUniqueHtmlUrls(
+    const { urls: mergedCsvUrls, filteredCount } = mergeAndGetUniqueHtmlUrls(
       rebasedCsvUrls,
       { includeQueryParams: true },
     );
+    const { scoped: explicitUrls, scopeFilteredCount } = applySiteScopeFilter(
```
dhavkuma: @anuj-adobe Filtering should be done before we end function execution at old L:561.

src/prerender/handler.js
```diff
@@ -12,7 +12,7 @@
 
 import { PutObjectCommand } from '@aws-sdk/client-s3';
 import { Audit, Suggestion } from '@adobe/spacecat-shared-data-access';
-import { detectBotBlocker } from '@adobe/spacecat-shared-utils';
+import { detectBotBlocker, filterBySiteScope } from '@adobe/spacecat-shared-utils';
```
dhavkuma, line 15: for quick reference https://github.com/adobe/spacecat-shared/blob/7a7a1730a77cfb3a52968765549d5f74e648ae0b/packages/spacecat-shared-utils/src/url-helpers.js#L541

src/prerender/handler.js
```diff
@@ -662,7 +703,17 @@ export async function submitForScraping(context) {
       [...organicDeduped, ...includedDeduped, ...agenticDeduped],
       { includeQueryParams: true },
     );
-    const batchedUrls = crossDeduped.slice(0, DAILY_BATCH_SIZE);
+    // Single site-scope filter on the merged candidate set, applied before the daily-batch
+    // slice so out-of-scope URLs don't consume batch slots and starve in-scope ones.
+    const nonSlackScoped = applySiteScopeFilter(
```
dhavkuma: Filtering here is required to batch only scoped urls below in line
    const batchedUrls = nonSlackScoped.scoped.slice(0, DAILY_BATCH_SIZE);

src/prerender/handler.js
```diff
@@ -616,7 +648,16 @@ export async function submitForScraping(context) {
       [...organicSlackDeduped, ...includedSlackDeduped],
       { includeQueryParams: true },
     );
-    finalUrls = crossSlackDeduped;
+    // Single site-scope filter on the merged candidate set (scoped here, not per-source).
```
dhavkuma, line 623: Can move this to later in the execution after resolving https://github.com/adobe/spacecat-audit-worker/pull/2724/changes#r3511436472

src/prerender/handler.js
anuj-adobe: can we rename it to siteScopedUrls or something similar?

src/prerender/handler.js
```diff
@@ -57,6 +57,30 @@ const AUDIT_TYPE = Audit.AUDIT_TYPES.PRERENDER;
...
+ * but scope-checked against siteBaseUrl's host, so divergent hosts silently drop every URL and
+ * produce submittedUrls=0 with no other signal.
+ *
+ * @param {string[]} urls - candidate URLs (already rebased onto preferredBase)
+ * @param {string} siteBaseUrl - scope base (site.getBaseURL())
+ * @param {string} preferredBase - rebase target the URLs carry
+ * @param {string} siteId - for log correlation
+ * @param {Object} log - logger
+ * @returns {{ scoped: string[], scopeFilteredCount: number }}
+ */
```
anuj-adobe: It seems claude verbose comment, can we make it concise as it unnecessarily increases the file size?

src/prerender/handler.js
```diff
@@ -57,6 +57,30 @@ const AUDIT_TYPE = Audit.AUDIT_TYPES.PRERENDER;
...
+function applySiteScopeFilter(urls, siteBaseUrl, preferredBase, siteId, log) {
```
anuj-adobe: we can remove this method as discussed

anuj-adobe (commented):
pls check inline comments

src/prerender/handler.js
dhavkuma: removed this fucntion

src/prerender/handler.js
dhavkuma: removed this function

src/prerender/handler.js
```diff
@@ -538,14 +538,17 @@ export async function submitForScraping(context) {
 
   const siteId = site.getId();
   const isSlackTriggered = !!(auditContext?.slackContext?.channelId);
+  const preferredBase = getPreferredBaseUrl(site, context);
+  const siteBaseUrl = site.getBaseURL();
```
dhavkuma, line 542: `siteBaseUrl` (instead of preferredBase) would be used to base site scoping. Refer to https://wiki.corp.adobe.com/pages/viewpage.action?pageId=3825916773&spaceKey=AEMSites&title=LLMO%2B-%2BSubpath%2BURL%2BSupport

dhavkuma (commented):
added inline comments for adding links or reasons

src/prerender/handler.js
dhavkuma: renamed to `scopedUrls`

test/audits/prerender/handler.test.js
```diff
@@ -1604,6 +1627,213 @@ describe('Prerender Audit', () => {
...
+                'https://bulk.com/fr/page-2',
+                'https://bulk.com/uk/page-3',
+              ],
+            },
+            finalUrl: 'https://bulk.com/uk',
+            log: { info: sinon.stub(), warn: sinon.stub(), debug: sinon.stub() },
+            env: {},
+          };
+
+          const result = await mockHandler.submitForScraping(context);
```
MysticatBot, line 1656: issue (blocking): **Test expects defensive scope-filter + logging in `processContentAndGenerateOpportunities` that is no longer implemented.**

This test asserts:
- `totalUrlsChecked` equals 2 (only in-scope URLs from scrapeResultPaths)
- `log.info` is called with `/Defensive scope re-filter dropped 1 out-of-scope path/`

But the production diff shows NO changes to `processContentAndGenerateOpportunities` - the defensive re-filter that was added via `applySiteScopeFilter` in commit 164f0cd was lost when the wrapper was removed in commit 5a837d1f. The final `handler.js` has no scope filtering in that function, so `totalUrlsChecked` will be 3 (all scrapeResultPaths) and no log message is emitted. This is likely the CI build failure.

Fix: re-add the inline scope filter + logging in `processContentAndGenerateOpportunities`:
```js
const allPaths = Array.from(context.scrapeResultPaths.keys());
const urlsToCheck = filterBySiteScope(allPaths, site.getBaseURL());
const defensiveDropped = allPaths.length - urlsToCheck.length;
if (defensiveDropped > 0) {
  log.info(`${LOG_PREFIX} Defensive scope re-filter dropped ${defensiveDropped} out-of-scope path(s) for siteId=${siteId}`);
}
```
Or, if the defensive re-filter was intentionally dropped, remove this test.

MysticatBot (changes requested):
Hey @dhavkuma,

**Verdict: Request changes** - test asserts behavior that was lost during the wrapper removal refactor.
**Complexity: MEDIUM** - medium diff (599 lines across production and test code).
_Changes: Addresses anuj-adobe's review feedback by removing the `applySiteScopeFilter` wrapper and calling `filterBySiteScope` directly with inline scope-count metrics (2 files)._
Note: CI checks are currently failing - likely due to the finding below.

## Must fix before merge

1. [Important] Test asserts defensive scope-filter behavior lost during refactor - `test/audits/prerender/handler.test.js:~1656` (details inline)

- `applySiteScopeFilter` wrapper removed and replaced with direct `filterBySiteScope` calls per anuj-adobe's feedback.
- Verbose JSDoc comment removed per anuj-adobe's "claude verbose comment" note.
- `nonSlackScoped` renamed to `scopedUrls` per anuj-adobe's naming suggestion.
- `rebasedTopPagesUrls` usage order fixed (init before use).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 23s | <b>Cost:</b> $2.04 | <b>Commit:</b> <code>aa08f64fe0fe11c4238e752f4e552753e6eda182</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @dhavkuma,

**Verdict: Approve** - prior finding resolved cleanly, no new issues.
**Complexity: MEDIUM** - medium diff (569 lines across production and test code).
_Changes: Removes the stale test for defensive scope re-filter behavior that was lost during the wrapper removal refactor (1 file)._
Note: CI checks are currently pending - verify before merge.

<details>

- nit: The defensive scope re-filter in `processContentAndGenerateOpportunities` was intentionally dropped (the wrapper and its observability contract are gone), but nothing in the PR description documents that decision. A one-line note in the PR body would help future readers understand why the defense-in-depth layer was removed rather than preserved with inline logging. - `src/prerender/handler.js`

</details>

- Stale test for defensive scope-filter behavior removed (b24b4a8) - the test asserted `processContentAndGenerateOpportunities` filtering and logging that no longer exists after the `applySiteScopeFilter` wrapper was dropped per anuj-adobe's review feedback.
- `applySiteScopeFilter` wrapper removed and replaced with direct `filterBySiteScope` calls per anuj-adobe's feedback.
- Verbose JSDoc comment removed per anuj-adobe's "claude verbose comment" note.
- `nonSlackScoped` renamed to `scopedUrls` per anuj-adobe's naming suggestion.
- `rebasedTopPagesUrls` usage order fixed (init before use).
- Hostname mismatch between preferredBase and siteBaseUrl now visible via `scopeFilteredUrls` metric in all three log lines.
- Bare catch block removed - utility moved to `@adobe/spacecat-shared-utils` with follow-up tracked in spacecat-shared.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 38s | <b>Cost:</b> $1.41 | <b>Commit:</b> <code>b24b4a8808a87c9eacbc6b62aec90472bf228ea4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anuj-adobe (approved):
LGTM

PR #2727, opened 2026-06-24
fix(preflight): only report DNS failures as broken external links (SITES-47125)

johobot:
## Problem

Valid external links on bot-protected sites were being false-flagged as **"Status 0 / Broken External Links"** in the preflight links audit. Confirmed against a customer (ups.com): both `https://www.ups.com/us/en/support/shipping-support/legal-terms-conditions/california-privacy-rights` and `https://www.ups.com/ppwa/doWork?loc=en_US` load fine in a browser but were reported as broken.

Jira: [SITES-47125](https://jira.corp.adobe.com/browse/SITES-47125)

## Root cause

The fix for [SITES-40919](https://jira.corp.adobe.com/browse/SITES-40919) (#2476, commit `d83ffb19`) changed the GET-fallback catch block in `src/preflight/links-checks.js` to return `{ status: 0 }` for **any** thrown fetch error — correct for genuinely unreachable hosts (DNS `ENOTFOUND`).

ups.com resolves DNS fine but **resets the HTTP/2 stream** (`ERR_HTTP2_STREAM_ERROR` / "NGHTTP2_INTERNAL_ERROR") to block bots. That throws, bypasses `isBrokenStatus()` (where 403/429 are already excluded), and lands in the catch block that reported it as broken. The existing bot-block protection only covered HTTP responses, not connection-level refusals.

## Fix

Discriminate by error type. Only a **DNS-resolution failure (`ENOTFOUND`)** — the domain does not exist — is a definitive broken link (`status: 0`). All other network-level failures (connection reset, HTTP/2 stream errors, TLS, timeout) mean the host is reachable but returned no usable response — indistinguishable from a valid page that blocks bots — so they are logged as "probe inconclusive" and not reported.

This **preserves the SITES-40919 case** (nonexistent domains / intranet hosts that fail DNS still surface as `status: 0`) while fixing the SITES-47125 false positives. Both behaviors verified end-to-end against the real network.

## Tests

- `test/preflight/links-checks.test.js` — added the HTTP/2 bot-block case (not broken) and the explicit `ENOTFOUND`-with-`code` case (broken); updated the old "any error → status 0" tests.
- `test/audits/preflight.test.js` — added integration coverage (via nock) for both the DNS-failure → broken and connection-level → not-broken paths, internal and external.

## Verification

- ✅ Full suite: 8752 passing
- ✅ 100% line/branch/statement coverage (100% on `links-checks.js`)
- ✅ Lint clean
- ✅ Real-network e2e: ups.com URLs skipped ("probe inconclusive"); dead domain still reported (status 0)

### DIFF
--- a/src/preflight/links-checks.js
+++ b/src/preflight/links-checks.js
@@ -131,11 +131,17 @@ function isBrokenStatus(status) {
  * connection level (e.g. ups.com resets the HTTP/2 stream with NGHTTP2_INTERNAL_ERROR), so these
  * must NOT be reported as broken (SITES-47125).
  *
+ * Node's DNS resolver reliably sets `err.code === 'ENOTFOUND'`, which is the primary signal. The
+ * message check is a narrow fallback for wrappers that strip `.code`: it only fires when no code
+ * is present and matches the canonical `getaddrinfo ENOTFOUND` format — not any message that
+ * merely contains the substring, which would re-introduce false positives (PR #2727 review).
+ *
  * @param {Error} err
  * @returns {boolean}
  */
 function isDnsResolutionFailure(err) {
-  return err.code === 'ENOTFOUND' || /ENOTFOUND/.test(err.message);
+  return err.code === 'ENOTFOUND'
+    || (err.code == null && /getaddrinfo ENOTFOUND/.test(err.message));
 }
 
 /**
--- a/src/preflight/links-checks.js
+++ b/src/preflight/links-checks.js
@@ -118,6 +118,27 @@ function isBrokenStatus(status) {
 }
 
 /**
+ * Returns true only for a DNS-resolution failure — the one network-level error that
+ * definitively indicates a broken link.
+ *
+ * A DNS-resolution failure (`ENOTFOUND` — getaddrinfo cannot resolve the hostname) means the
+ * domain does not exist, so the link is unambiguously broken (SITES-40919: nonexistent domains
+ * and intranet hosts that fail DNS must surface as status 0).
+ *
+ * Every other thrown error — connection reset, HTTP/2 stream errors, TLS failures, connection
+ * refused, request timeouts — means the host resolves and is reachable but did not return a
+ * usable HTTP response. That is indistinguishable from a valid page that blocks crawlers at the
+ * connection level (e.g. ups.com resets the HTTP/2 stream with NGHTTP2_INTERNAL_ERROR), so these
+ * must NOT be reported as broken (SITES-47125).
+ *
+ * @param {Error} err
+ * @returns {boolean}
+ */
+function isDnsResolutionFailure(err) {
+  return err.code === 'ENOTFOUND' || /ENOTFOUND/.test(err.message);
+}
+
+/**
  * Helper function to check if a link is broken
  * @param {string} href - The URL to check
  * @param {string} pageUrl - The source page URL
@@ -203,16 +224,24 @@ async function checkLinkStatus(href, pageUrl, context, options = {
 
     return null;
   } catch (finalErr) {
-    // Network-level failure (DNS, timeout, unsupported protocol) — the link is
-    // unreachable, which is a broken-link condition. status: 0 is the sentinel
-    // for "no HTTP response received".
-    log.info(`[preflight-audit] ${linkType} link ${href} unreachable (${finalErr.message}) — reporting as broken`);
-    return {
-      urlTo: href,
-      href: pageUrl,
-      status: 0,
-      ...toElementTargets(selectors),
-    };
+    // Only a DNS-resolution failure (ENOTFOUND) is a definitive broken-link condition: the
+    // domain does not exist. status: 0 is the sentinel for "no HTTP response received".
+    if (isDnsResolutionFailure(finalErr)) {
+      log.info(`[preflight-audit] ${linkType} link ${href} unreachable — DNS does not resolve (${finalErr.message}) — reporting as broken`);
+      return {
+        urlTo: href,
+        href: pageUrl,
+        status: 0,
+        ...toElementTargets(selectors),
+      };
+    }
+
+    // Any other network-level failure (connection reset, HTTP/2 stream error, TLS, timeout)
+    // means the host is reachable but did not return a usable response — indistinguishable
+    // from a valid page that blocks bots at the connection level. Do NOT report as broken
+    // (SITES-47125: ups.com and similar bot-protected sites were false-flagged as Status 0).
+    log.info(`[preflight-audit] ${linkType} link ${href} probe inconclusive (${finalErr.message}) — not reporting as broken`);
+    return null;
   }
 }
 
--- a/src/preflight/links-checks.js
+++ b/src/preflight/links-checks.js
@@ -199,7 +199,11 @@ async function checkLinkStatus(href, pageUrl, context, options = {
       return null;
     }
   } catch (err) {
-    log.warn(`[preflight-audit] HEAD request failed (${err.message}), retrying with GET: ${href}`);
+    // DEBUG, not WARN: a thrown HEAD is expected and self-correcting here — we always retry with
+    // GET, and on bot-protected hosts (which reset the connection) every probe throws HEAD first
+    // yet ends up correctly not-broken. Logging that at WARN produced high-volume, non-actionable
+    // noise that also looked like a failure for links we then skip (SITES-47125 review).
+    log.debug(`[preflight-audit] HEAD request failed (${err.message}), retrying with GET: ${href}`);
   }
 
   // GET fallback — HEAD was inconclusive (broken status, fallback status, or network error).
@@ -246,7 +250,10 @@ async function checkLinkStatus(href, pageUrl, context, options = {
     // means the host is reachable but did not return a usable response — indistinguishable
     // from a valid page that blocks bots at the connection level. Do NOT report as broken
     // (SITES-47125: ups.com and similar bot-protected sites were false-flagged as Status 0).
-    log.info(`[preflight-audit] ${linkType} link ${href} probe inconclusive (${finalErr.message}) — not reporting as broken`);
+    // DEBUG, not INFO: on a page with many bot-blocked external links this is the common,
+    // non-actionable path and at INFO it floods the logs. The DNS-failure branch above stays at
+    // INFO because that is the actionable "reporting as broken" case (SITES-47125 review).
+    log.debug(`[preflight-audit] ${linkType} link ${href} probe inconclusive (${finalErr.message}) — not reporting as broken`);
     return null;
   }
 }
--- a/test/preflight/links-checks.test.js
+++ b/test/preflight/links-checks.test.js
@@ -356,7 +356,27 @@ describe('preflight/links-checks - runLinksChecks', () => {
     expect(context.log.error).to.not.have.been.called;
   });
 
-  it('flags internal link as broken (status 0) on a DNS-resolution error (ENOTFOUND)', async () => {
+  it('does NOT flag as broken when a non-DNS error message merely contains the substring "ENOTFOUND"', async () => {
+    // The message-fallback must only match the canonical Node format ("getaddrinfo ENOTFOUND"),
+    // not any message that happens to contain the substring — otherwise a wrapper error like this
+    // would re-introduce the false positives this fix removes (MysticatBot review on PR #2727).
+    const wrappedError = new Error('Retry after ENOTFOUND was cached'); // no .code, non-canonical
+    fetchStub.onFirstCall().rejects(wrappedError);
+    fetchStub.onSecondCall().rejects(wrappedError);
+
+    const result = await runLinksChecks(
+      [pageUrl],
+      makeScrapedObjects('<a href="https://other.com/page">link</a>'),
+      context,
+    );
+
+    expect(result.auditResult.brokenExternalLinks).to.have.lengthOf(0);
+    expect(context.log.error).to.not.have.been.called;
+  });
+
+  // Message-fallback branch: error carries the canonical "getaddrinfo ENOTFOUND" message but no
+  // .code (e.g. a fetch wrapper that strips the code). Still a definitive DNS failure → broken.
+  it('flags internal link as broken (status 0) on a DNS-resolution error message without a code', async () => {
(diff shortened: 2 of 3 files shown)


src/preflight/links-checks.js
```diff
@@ -117,6 +117,27 @@ function isBrokenStatus(status) {
...
+ * refused, request timeouts — means the host resolves and is reachable but did not return a
+ * usable HTTP response. That is indistinguishable from a valid page that blocks crawlers at the
+ * connection level (e.g. ups.com resets the HTTP/2 stream with NGHTTP2_INTERNAL_ERROR), so these
+ * must NOT be reported as broken (SITES-47125).
+ *
+ * @param {Error} err
+ * @returns {boolean}
+ */
+function isDnsResolutionFailure(err) {
+  return err.code === 'ENOTFOUND' || /ENOTFOUND/.test(err.message);
```
MysticatBot: issue (blocking): The regex fallback `/ENOTFOUND/.test(err.message)` matches any error whose message happens to contain the substring `ENOTFOUND`, even if the failure is not a DNS resolution error. For example, a retry-library wrapper producing `"Retry after ENOTFOUND was cached"` would match.

Node's DNS resolver always sets `err.code = 'ENOTFOUND'` on resolution failures. If the fallback exists because a specific fetch wrapper (e.g. `tracingFetch`) strips `.code`, tighten the regex to anchor against the canonical format and guard on code being absent:

```js
function isDnsResolutionFailure(err) {
  return err.code === 'ENOTFOUND'
    || (err.code == null && /getaddrinfo ENOTFOUND/.test(err.message));
}
```

If no wrapper strips `.code`, the regex is dead weight and can be removed entirely. Either way, the current form could re-introduce the same class of false positives this PR fixes, just triggered by a different error path.

MysticatBot (changes requested):
Hey @johobot,

**Verdict: Request changes** - one fixable concern with the regex fallback in `isDnsResolutionFailure`.
**Complexity: MEDIUM** - medium diff, single service.
_Changes: Discriminates DNS-resolution failures from other network errors in the preflight links audit catch block, fixing false positives on bot-protected sites (3 files)._
Note: Recommend a human read before merge - possible divergence from the documented design (heuristic, flagged by the conventions reviewer).

## Must fix before merge

1. [Important] Regex fallback in `isDnsResolutionFailure` is overly broad and could re-introduce false positives - `src/preflight/links-checks.js:138` (details inline)

<details>

- nit: The existing internal-link ENOTFOUND test (`flags internal link as broken (status 0) on a DNS-resolution error`) constructs its error without `.code = 'ENOTFOUND'`, relying solely on the message regex branch. Either add `.code` for consistency with the other DNS tests, or rename/comment it to explicitly document it as the message-fallback test case - `test/preflight/links-checks.test.js:326`
- suggestion: If the message-regex fallback is kept, add a dedicated test where `err.code` is absent but the message contains `getaddrinfo ENOTFOUND` to intentionally cover that branch - `test/preflight/links-checks.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 6m 0s | <b>Cost:</b> $5.66 | <b>Commit:</b> <code>e467ddb0536baf515aa96b039a39f919944a9e31</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/preflight/links-checks.js
johobot: Fixed in ba851986. Tightened to:

```js
function isDnsResolutionFailure(err) {
  return err.code === 'ENOTFOUND'
    || (err.code == null && /getaddrinfo ENOTFOUND/.test(err.message));
}
```

`err.code === 'ENOTFOUND'` stays the primary signal. The message fallback now only fires when `.code` is absent and matches the canonical `getaddrinfo ENOTFOUND` format, so a substring like "Retry after ENOTFOUND was cached" no longer matches. Added a regression test asserting that case is not flagged as broken, plus a test covering the code-absent message-fallback branch. 100% coverage on the file maintained; real-network e2e re-confirmed (ups.com skipped, nonexistent domain still status 0).

MysticatBot (approved):
Hey @johobot,

**Verdict: Approve** - prior blocking finding addressed, no new issues found.
**Complexity: MEDIUM** - medium diff, single service.
_Changes: Discriminates DNS-resolution failures from other network errors in the preflight links audit catch block, fixing false positives on bot-protected sites (3 files)._

<details>

- suggestion: Consider logging at DEBUG rather than INFO for the "probe inconclusive" path - in production with many bot-blocking external links, these INFO lines are high-volume and not actionable; the DNS-failure (broken) path already logs at INFO where action is needed - `src/preflight/links-checks.js:249`
- suggestion: The unit-level negative tests in `test/preflight/links-checks.test.js` assert `brokenExternalLinks` length 0 but do not assert the "probe inconclusive" log message, unlike the integration tests in `test/audits/preflight.test.js` - adding that assertion would confirm the correct branch executed, not just that nothing broke - `test/preflight/links-checks.test.js:326`
- suggestion: If `isDnsResolutionFailure` grows additional error codes (e.g. `EAI_AGAIN` for transient DNS failures), consider exporting it for targeted unit tests rather than exercising it only through the full `checkLinkStatus` call chain - `src/preflight/links-checks.js:136`
</details>

- Regex fallback tightened: now guarded by `err.code == null` and matches canonical `getaddrinfo ENOTFOUND` format only

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 32s | <b>Cost:</b> $3.91 | <b>Commit:</b> <code>ba85198609ac1802c68f8c2c53b34824bde9d038</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/preflight/links-checks.js
```diff
@@ -203,16 +230,24 @@ async function checkLinkStatus(href, pageUrl, context, options = {
...
+        status: 0,
+        ...toElementTargets(selectors),
+      };
+    }
+
+    // Any other network-level failure (connection reset, HTTP/2 stream error, TLS, timeout)
+    // means the host is reachable but did not return a usable response — indistinguishable
+    // from a valid page that blocks bots at the connection level. Do NOT report as broken
+    // (SITES-47125: ups.com and similar bot-protected sites were false-flagged as Status 0).
+    log.info(`[preflight-audit] ${linkType} link ${href} probe inconclusive (${finalErr.message}) — not reporting as broken`);
```
atopper: If there is a page with many bot-blocked links, this will output a lot of `info` statements.  Is that ok?  Could the caller collect them and output them or a summary as 1 log?

src/preflight/links-checks.js
atopper: Is this log still valid?  If we later detect a bot and skip it, will this warning cause some confusion?  (Probably no worse than it is presently...)

test/preflight/links-checks.test.js
```diff
@@ -317,7 +323,60 @@ describe('preflight/links-checks - runLinksChecks', () => {
...
+    const http2Error = Object.assign(
+      new Error('Stream closed with error code NGHTTP2_INTERNAL_ERROR'),
+      { code: 'ERR_HTTP2_STREAM_ERROR' },
+    );
+    fetchStub.onFirstCall().rejects(http2Error);
+    fetchStub.onSecondCall().rejects(http2Error);
+
+    const result = await runLinksChecks(
+      [pageUrl],
+      makeScrapedObjects('<a href="https://www.ups.com/ppwa/doWork?loc=en_US">link</a>'),
```
atopper, line 337: Straight from Claude (reject if you want to):
> The unit-level "not broken" tests (does NOT flag as broken when a connection-level error blocks the bot, does NOT flag as broken when both HEAD and GET fail with a generic error) assert brokenExternalLinks.lengthOf(0) and log.error not called, but don't assert log.info was called with "probe inconclusive". The integration tests in test/audits/preflight.test.js:144 and :392 do assert calledWithMatch(/probe inconclusive/). Adding the same assertion at the unit level would confirm the correct branch ran, not just that no false-positive was emitted.

_I didn't really understand the significance so asked for more info:_

>  The null return in src/preflight/links-checks.js:249–250 can be reached via two routes:
  1. isDnsResolutionFailure(finalErr) === false → logs "probe inconclusive" → return null (the intended path)
  2. Any earlier return null in checkLinkStatus (non-http protocol check at line 171, or a hypothetical future early exit)
If a refactor accidentally short-circuited out of the finalErr catch block before reaching the isDnsResolutionFailure call — say, the HEAD catch at line 201 silently swallowed the error for non-DNS cases and returned instead of falling through — brokenExternalLinks would still be length 0, log.error would still be clean, and all three unit tests would keep passing. The "probe inconclusive" log assertion is the signal that would distinguish "correct path executed" from "exited earlier."

>  The concrete fix
  Add one line to each of the three tests:
  // test/preflight/links-checks.test.js:341–342 (HTTP/2 reset)
  expect(result.auditResult.brokenExternalLinks).to.have.lengthOf(0);
  expect(context.log.error).to.not.have.been.called;
  expect(context.log.info).to.have.been.calledWith(sinon.match(/probe inconclusive/)); // ADD
  // test/preflight/links-checks.test.js:355–356 (generic network error)
  expect(result.auditResult.brokenExternalLinks).to.have.lengthOf(0);
  expect(context.log.error).to.not.have.been.called;
  expect(context.log.info).to.have.been.calledWith(sinon.match(/probe inconclusive/)); // ADD
  // test/preflight/links-checks.test.js:373–374 (substring "ENOTFOUND" regression)
  expect(result.auditResult.brokenExternalLinks).to.have.lengthOf(0);
  expect(context.log.error).to.not.have.been.called;
  expect(context.log.info).to.have.been.calledWith(sinon.match(/probe inconclusive/)); // ADD

> This mirrors what the integration tests already do (line 148 and 401) and makes the unit suite self-sufficient: a reviewer running only test/preflight/links-checks.test.js would have full confidence that the isDnsResolutionFailure branch was reached, not just that the result happened to be non-broken.

clotton (approved):
**Approve** — careful regression-focused review, including verification against the real `@adobe/fetch` client. No harmful regression found.

### Verified empirically (not just by reading)
The fix hinges on `ENOTFOUND` being detectable so genuinely-dead links are still flagged. I probed the real client:

| Probe | Result | PR behavior |
|---|---|---|
| Dead domain | `code=ENOTFOUND` **and** msg `getaddrinfo ENOTFOUND …` | → status 0 (broken) ✅ |
| Connection refused | `code=ECONNREFUSED` | → inconclusive (skip) ✅ |
| ups.com bot-block | `ERR_HTTP2_STREAM_ERROR` (GET intermittently 200) | → not broken ✅ |

So dead-domain detection is preserved on the real client — no false-negative regression. (An internal-vs-external approach I prototyped would have regressed exactly this by dropping *all* external failures; the error-type axis here is the better choice.)

### Regression vectors checked — all clear
- **SITES-40919** (internal DNS failure → status 0): preserved, with explicit test.
- **SITES-46696 throttle**: untouched code path; treating timeouts as inconclusive strengthens it.
- **Substring false-match**: already tightened in commit 2 to `getaddrinfo ENOTFOUND` (only when `.code` is absent) with a regression test 👍
- **Coverage/tests**: all four quadrants (internal/external × DNS-broken / connection-inconclusive) in unit + nock integration; CI green.

### One intentional tradeoff worth stating in the ticket
Non-DNS failures on a host that *resolves but is genuinely down* (reset/refused/TLS/timeout) are now never reported, for internal and external links alike. That's the deliberate price of removing the bot-block false positives — detection now means "DNS doesn't resolve, OR an HTTP error code" rather than "any failure." Right call; just make sure it's understood.

### Minor nits (non-blocking)
- `EAI_AGAIN` (transient DNS) falls into "inconclusive" — fine, just no flaky-DNS visibility.
- `isDnsResolutionFailure` assumes `err` is an Error with `.message` — safe for fetch, noting for completeness.

LGTM. 🚢

atopper (approved):
I added some comments to consider.

src/preflight/links-checks.js
johobot: Good call. I downgraded this to `log.debug`. On a page with lots of bot-blocked links this "probe inconclusive" line is the common, non-actionable path, so INFO was just flooding the logs. I left the DNS-failure branch above at INFO since that's the actionable "reporting as broken" case. I went with a level change instead of caller-side aggregation to keep the diff small. Happy to add a summary-count log as a follow-up if that'd be more useful.

src/preflight/links-checks.js
johobot: Agreed, it was misleading. This WARN fired for every bot-blocked host (which always throws on HEAD first) even though we then correctly skip it as not-broken. I downgraded it to `log.debug` for the same noise reason as the inconclusive line. It's accurate, since HEAD did fail and we did retry with GET, but it's not something worth warning on.

test/preflight/links-checks.test.js
johobot: Done. I added the branch assertion to all three unit tests (HTTP/2 reset, generic network error, and the "ENOTFOUND substring" regression). It asserts against `log.debug` now, since I also downgraded that line per your other comment. This confirms the branch where `isDnsResolutionFailure` returns false actually ran, rather than just checking that nothing got flagged. It matches what the integration tests already do.

clotton (approved):
Tested this end-to-end in the dev environment against the actual deployed code from this branch.

**Setup:** Deployed this PR branch to a personal Lambda alias (`clotton`) in the dev AWS account — a completely isolated execution environment that doesn't touch the shared `ci` or `latest` aliases. Injected controlled test scrapes directly into the preflight audit pipeline and observed the results in CloudWatch logs.

**Two scenarios tested:**

1. **Bot-block / timeout (the bug scenario):** Ran the links checker against a page with a link to `iana.org`, which timed out on both HEAD and GET requests — the same class of error as a site actively blocking our bot. The checker correctly logged `probe inconclusive — not reporting as broken` and produced **zero broken link findings**. Before this fix, this would have been a false positive.

2. **Genuinely dead domain (the regression check):** Ran the links checker against a page with a link to a domain that cannot possibly exist (`.invalid` TLD). The checker correctly got `getaddrinfo ENOTFOUND`, logged `DNS does not resolve — reporting as broken`, and produced a `status: 0` broken link finding. Truly dead links are still caught.

Both behaviours were observed in production Lambda code running the full preflight pipeline. The fix does exactly what it claims — bot-blocks are silenced, real broken links are not.

PR #2736, opened 2026-06-26
fix(cdn-analysis): warn when raw logs aggregate to zero rows (wrong format)

akshaymagapu:
## Problem

A wrong CDN log format (e.g. a non-JSON feed delivered to a JSON-SerDe table) is **silently dropped** today. The raw tables use `org.openx.data.jsonserde.JsonSerDe` with `ignore.malformed.json = true`, so any unparseable line becomes a row of all-NULL columns instead of erroring. The aggregation's LLM user-agent `WHERE` filter then matches nothing, the `INSERT` writes an empty partition, and the audit returns a **normal success** with no error anywhere in the pipeline.

The result: a customer who *is* delivering logs (in the wrong format) looks identical to a customer with genuinely no agentic traffic. The dashboard is empty and there is no signal to act on.

## Fix

After the aggregation inserts, if raw logs were present but the aggregated output partition is empty, emit one greppable warning:

```
[cdn-logs-analysis] empty aggregation despite raw logs - possible wrong log format siteId=... host=... serviceProvider=... cdnType=... rawDataPath=...
```

- **Naming-agnostic:** probes the S3 output partition via the existing `pathHasData` helper rather than per-CDN raw columns (`reqHost` / `"x-host-header"` / `properties.hostname` / …), so it works for every CDN without a column map.
- **No extra Athena query:** `pathHasData` is a single `ListObjectsV2` with `MaxKeys: 1`.
- **Minimal:** 4 functional lines in the per-provider loop.

This turns a silent failure into a dashboardable / alertable signal:

```
index=dx_aem_engineering sourcetype=dx_aem_sites_spacecat_backend_prod service=audit-worker
  "[cdn-logs-analysis] empty aggregation despite raw logs"
| stats count by siteId, host, cdnType, serviceProvider
```

## Testing

- Added two tests: warning fires on empty output; warning suppressed when the output partition has data (via `forceReprocess`).
- `src/cdn-analysis/handler.js` at 100% lines/branches/functions/statements; full cdn-analysis spec green (54 passing).

## Notes / limitations

- It is a `warn`, not a definitive wrong-format detector: an hour with genuinely zero agentic traffic also produces empty aggregation, so the message says *"possible"*. A site emitting it **every hour** = wrong format; sporadic = a quiet hour.

### DIFF
--- a/src/cdn-analysis/handler.js
+++ b/src/cdn-analysis/handler.js
@@ -532,6 +532,16 @@ export async function processCdnLogs(auditUrl, context, site, auditContext) {
       // eslint-disable-next-line no-await-in-loop
       await athenaClient.execute(sqlInsertReferral, database, `[Athena Query] Insert aggregated referral data for ${serviceProvider} into ${database}.${aggregatedReferralTable}`);
 
+      // Raw logs were present but the aggregation wrote no rows. The usual cause is a
+      // wrong log format: the JSON SerDe runs with ignore.malformed.json=true, so
+      // unparseable lines become all-NULL rows, the LLM user-agent filter matches
+      // nothing, and the audit "succeeds" empty with no error. Emit one greppable line
+      // so it can be alerted/dashboarded instead of failing silently.
+      // eslint-disable-next-line no-await-in-loop
+      if (!await pathHasData(s3Client, paths.aggregatedOutput)) {
+        log.warn(`[cdn-logs-analysis] empty aggregation despite raw logs - possible wrong log format siteId=${siteId} host=${host} serviceProvider=${serviceProvider} cdnType=${cdnType} rawDataPath=${rawDataPath}`);
+      }
+
       log.info(`${auditType} processed logs for siteId=${siteId} with:
         serviceProvider=${serviceProvider}
         cdnType=${cdnType}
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - well-scoped observability fix that surfaces a real silent failure with no new risk.
**Complexity: LOW** - small diff, single service, no risk signals.
_Changes: Adds a warning log when CDN raw logs are present but the aggregated output partition is empty after Athena INSERT, indicating a possible wrong log format (2 files)._

<details>

- suggestion: Only `paths.aggregatedOutput` is checked, not `paths.aggregatedReferralOutput`. Since the same SerDe format issue would cause both to be empty simultaneously, this is a reliable proxy today. Consider checking both if the team wants to detect asymmetric emptiness in the future. - `src/cdn-analysis/handler.js:537`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 24s | <b>Cost:</b> $4.52 | <b>Commit:</b> <code>533c0a0b18028ce8401c409e35e088d41fd9f028</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2737, opened 2026-06-26
fix(llmo-customer-analysis): create brand-presence schedule via shared drs-client helper | LLMO-5605

igor-grubic:
## 1. Abstract
Migrates the `llmo-customer-analysis` onboarding cascade's brand-presence schedule creation from an inline `fetch` payload to the shared `createBrandPresenceSchedule` drs-client helper. Behaviour-preserving.

## 2. Reasoning
Anti-drift. This makes the cascade use the **single definition** of the brand-presence schedule, shared with the self-serve activate-brand endpoint (spacecat-api-service, LLMO-5605). DRS dedups schedules on the provider-set `(site_id, brand_id, cadence, provider-set)`, so two hand-synced payloads in two repos would eventually drift and silently produce **duplicate** schedules. One helper both paths call removes that class of bug.

- `src/llmo-customer-analysis/handler.js`: brand-presence schedule creation now routes through `@adobe/spacecat-shared-drs-client`'s `createBrandPresenceSchedule` (POST + tolerate 409) instead of an inline `/schedules` + `/trigger` fetch.
- **Behaviour-preserving:** same weekly cadence, provider set (`brightdata` / `google_ai_overviews` / `openai_web_search`), `enable_brand_presence`, and immediate trigger.
- Legacy no-op fields dropped (a camelCase `siteId` and a `platforms` array in `provider_parameters`) — verified unread by DRS (the scheduler copies `provider_parameters` verbatim into the job, but brightdata consumes only `dataset_id` + `metadata.site`).
- Bumps `@adobe/spacecat-shared-drs-client` 1.12.1 → 1.13.0 (the release that adds the helper).

## 4. Required information
- Jira / issue: LLMO-5605 — https://jira.corp.adobe.com/browse/LLMO-5605
- Spec: Piece 2 design spec — https://github.com/adobe/project-elmo-ui/issues/2095
- Other: Piece 1 spec — https://github.com/adobe/project-elmo-ui/issues/2092

- **spacecat-shared** (drs-client) — *consumed*: depends on `createBrandPresenceSchedule` from 1.13.0 — https://github.com/adobe/spacecat-shared/pull/1721.
- **spacecat-api-service** — *contract / coordinated*: the activate endpoint (https://github.com/adobe/spacecat-api-service/pull/2706) and this cascade share the same helper definition; both must route through it so the DRS provider-set dedup keeps matching.

## 7. Test plan
(a) **Local** — tests reworked to assert the `createBrandPresenceSchedule` helper call (args + returned `scheduleId`) instead of the raw `/schedules` + `/trigger` fetches.

(b) **Per-env (dev):** run the `llmo-customer-analysis` onboarding for a site → confirm exactly one weekly brand-presence schedule is created in DRS (same cadence / provider-set as before) with the immediate trigger firing; re-run → confirm idempotency (DRS 409 tolerated, no duplicate schedule).

Part of the LLMO-5605 Piece 2 chain. Anti-drift migration, parallel with the api-service endpoint.
- drs-client 1.13.0 (spacecat-shared) — *released*: provides the helper — https://github.com/adobe/spacecat-shared/pull/1721.
- **This PR** (spacecat-audit-worker) — migrates the cascade to the helper; can merge independently now the dependency is released.
- spacecat-api-service activate endpoint — *related (parallel anti-drift)*: must use the same helper — https://github.com/adobe/spacecat-api-service/pull/2706.

Ordered sequence: drs-client 1.13.0 released (done) → this PR and the api-service endpoint each merge; both must stay on the shared helper to preserve dedup.


MysticatBot (approved):
Hey @igor-grubic,

**Verdict: Approve** - clean behaviour-preserving migration to the shared drs-client helper.
**Complexity: HIGH** - large patch (mostly deletions); dependency bump.
_Changes: Migrates brand-presence schedule creation from inline fetch logic to the shared `createBrandPresenceSchedule` drs-client helper, eliminating drift risk between the audit-worker and api-service paths (4 files)._

<details>

- suggestion: Tighten the error-message assertion in "should handle brand presence schedule creation failure" to include the DRS error text suffix (e.g. `sinon.match(/Failed to create\/trigger.*DRS API error/)`) so loss of error context propagation would be caught - `test/audits/llmo-customer-analysis.test.js`
- suggestion: Add `calledWithMatch({ siteId: 'site-123', triggerImmediately: true })` to the "already existed" test for parity with other brand-presence-schedule tests - `test/audits/llmo-customer-analysis.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 27s | <b>Cost:</b> $4.68 | <b>Commit:</b> <code>41c481f310102e7906ed5bee0fb3689637061be6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich (approved):
Hey @igor-grubic,

**Verdict: Approve** - clean migration to the shared `createBrandPresenceSchedule` helper; the dedup-key fields (`site_id`, `brand_id` + `spacecat_org_id` top-level, weekly cadence, canonical provider set) are byte-for-byte identical to the old inline payload and to the api-service activate path, so the DRS `(site_id, brand_id, cadence, provider-set)` dedup keeps matching across both repos. Enqueue gate untouched, so "no shebang" is preserved.

One thing to note, not blocking:

**nit: not strictly behaviour-preserving on the 409 path (it's actually a benign improvement).** The old inline code threw on any non-`ok` response including a 409, so a pre-existing brand-presence schedule surfaced as an error, was caught by the outer try/catch, and `brand-presence-schedule` was omitted from `triggeredSteps`. The 1.13.0 helper treats 409 as success (`alreadyExisted:true`), so the caller now marks the step triggered, returns `existing_schedule_id`, and still fires the immediate trigger on that branch (self-healing). This aligns with the spec's "POST + tolerate 409" dedup design and is the behaviour we want - but since the PR is framed as "behaviour-preserving," it'd be worth a one-line note in the PR body that the 409 handling changed from throw to tolerate-as-success. (Also optional: a one-line confirmation from the DRS provider owner that the dropped `provider_parameters.*.siteId` and `brightdata.platforms` fields are genuinely unread - low risk since `dataset_id` + `metadata.site` still carry the same information, but it's asserted only in the helper's own comment.)

PR #2752, opened 2026-07-08
fix(cwv): dispatch guidance for PENDING_VALIDATION suggestions (SITES-47558)

ramboz:
## What

On paid-tier ASO sites, CWV suggestions are created `PENDING_VALIDATION` and an SME reviews the generated guidance before approving them to `NEW`. `processAutoSuggest` gated the `guidance:cwv` dispatch on `status === 'NEW'`, so `PENDING_VALIDATION` suggestions never received guidance — the SME had nothing to review and the suggestion could never be approved (deadlock). Primary defect behind [SITES-47558](https://jira.corp.adobe.com/browse/SITES-47558).

Free/PLG sites (created `NEW`) are unaffected; this fixes paid-tier ASO sites.

## Changes

- `shouldSendAutoSuggestForSuggestion`: dispatch for `NEW` **and** `PENDING_VALIDATION`. For `PENDING_VALIDATION`, only (re)dispatch when guidance is missing or in the legacy aggregated format (issues lack `source_index`), so granular guidance is not regenerated on every weekly audit. `APPROVED`/other statuses stay ineligible; the `isCodeChangeAvailable` skip is unchanged.
- `processAutoSuggest`: stamp `data.suggestionStatus` on the outbound message so Mystique gates code-fix generation on SME approval.
- `docs/cwv-autofix-flow.md`: eligibility gate updated.

## Tests

Red→green regression coverage in `test/audits/cwv/auto-suggest.test.js`: PENDING_VALIDATION dispatch + `suggestionStatus` stamping (integration) and per-branch unit tests (no-guidance → dispatch, legacy-aggregated → dispatch, granular → skip, patch-present → skip). Full file green (31); CWV suite 48.

## Paired change

Requires the Mystique side, which gates code-fix (Task B) on `suggestionStatus === 'NEW'` so `PENDING_VALIDATION` receives guidance only — see SITES-47558. Deploy Mystique first, then this.

### DIFF
--- a/docs/cwv-autofix-flow.md
+++ b/docs/cwv-autofix-flow.md
@@ -51,10 +51,17 @@ The CWV audit is a 2-step `StepAudit` registered as audit type `cwv` ([`src/cwv/
 ## Eligibility gate
 
 A suggestion is sent to Mystique only when ALL of:
-- Status is `NEW`
-- No guidance has been written yet
+- Status is `NEW` **or** `PENDING_VALIDATION` (guidance must be generated for
+  `PENDING_VALIDATION` so an SME has something to review before approving to `NEW` —
+  gating on `NEW` deadlocked paid-tier sites, SITES-47558)
+- No code patch has run yet (`isCodeChangeAvailable !== true`)
+- For `PENDING_VALIDATION`: guidance is missing or in the legacy aggregated format
+  (issues lack `source_index`) — avoids regenerating granular guidance every weekly audit
 - `type: 'url'` (group-type suggestions are skipped)
 - The site has `cwv-auto-suggest` enabled in `Configuration`
+
+The outbound message carries `data.suggestionStatus` so Mystique gates code-fix
+generation on SME approval (`NEW` → code-fix; `PENDING_VALIDATION` → guidance only).
 
 ## Feature flags (per site, in `Configuration`)
 
--- a/src/cwv/auto-suggest.js
+++ b/src/cwv/auto-suggest.js
@@ -51,7 +51,8 @@ const CWV_AUTO_FIX_FEATURE_TOGGLE = 'cwv-auto-fix';
  *
  * CWV suggestion structure:
  * {
- *   status: 'NEW' | 'APPROVED' | 'SKIPPED' | 'FIXED' | 'ERROR' | 'REJECTED',
+ *   status: 'NEW' | 'PENDING_VALIDATION' | 'APPROVED' | 'SKIPPED' | 'FIXED' | 'ERROR'
+ *     | 'REJECTED',
  *   data: {
  *     type: 'url' | 'group',
  *     url?: string,                  // Present for type: 'url'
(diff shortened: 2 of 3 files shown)


tkotthakota-adobe:
Hi @ramboz review comments here:
Perspectives: Architecture, QA/Testing, Product/Consumer Impact, Engineering Practices. Skipped Security (no                   
  auth/secrets/user-input changes), SRE (no new I/O paths; existing SQS send is unchanged), LLM/AI Systems (no LLM calls in scope), AI-Native (focused fix, not new maintainability surface).                                                              
                                                                                                                                 
  Summary                                                                                                                        

  Fixes a deadlock where paid-tier ASO sites' CWV suggestions (born PENDING_VALIDATION) never received guidance because the dispatch gate required status === 'NEW'. The fix extends the gate to PENDING_VALIDATION, adds a no-re-dispatch guard based on source_index presence, and stamps suggestionStatus on the outbound message so Mystique can gate code-fix accordingly. Correct implementation, well-tested, well-documented. Nearly ready to merge; one Should Fix on test coverage.

  Strengths

  - Surgical change: 29 lines of logic in auto-suggest.js address the root cause with no side effects on other opportunity types 
  or statuses.
  - Re-dispatch guard is elegant: checking issues.some((i) => Number.isInteger(i.source_index)) rather than issues.length > 0  correctly distinguishes "has granular guidance" from "has legacy aggregated guidance" — avoids weekly re-spam while still upgrading legacy rows.
  - suggestionStatus stamped unconditionally for both NEW and PENDING_VALIDATION, so Mystique can rely on the field without case-splitting.                                                                                                                
  - JSDoc updated in the same PR alongside cwv-autofix-flow.md — documentation stays in sync with the fix.
  - Unit tests cover all four PENDING_VALIDATION branches of shouldSendAutoSuggestForSuggestion: no-guidance → dispatch; legacy-aggregated → dispatch; granular → skip; isCodeChangeAvailable → skip.                                                   
                                                                                                                                 
  Issues                                                                                                                         
                                                             
  Blockers

  None.

  Should Fix

  S1. No integration test for the PENDING_VALIDATION no-re-dispatch path — test/audits/cwv/auto-suggest.test.js                  
  
  The two integration tests for processAutoSuggest only cover the dispatch-yes case (PENDING_VALIDATION with empty issues).      
  There's no integration test that sends a PENDING_VALIDATION suggestion with granular guidance (source_index present) through processAutoSuggest and asserts that sqsStub was NOT called. The no-re-dispatch guard is unit-tested via shouldSendAutoSuggestForSuggestion, but processAutoSuggest is the SQS-send boundary — a future refactor that rewires the gate inside processAutoSuggest directly (instead of via the helper) would bypass the unit test and this integration path would be unguarded. Suggest adding:

  it('does NOT dispatch for PENDING_VALIDATION suggestions that already have granular guidance (no weekly re-spam)', async () =>
  {                                                                                                                              
    const opportunity = {
      ...                                                                                                                        
      getSuggestions: () => Promise.resolve([{               
        getId: () => 'sugg-001',                                                                                                 
        getStatus: () => 'PENDING_VALIDATION',
        getData: () => ({                                                                                                        
          type: 'url',                                       
          url: 'https://example.com/page1',                                                                                      
          metrics: [{ deviceType: 'mobile', lcp: 3500 }],
          issues: [{ type: 'lcp', value: 'some guidance', source_index: 0 }],                                                    
        }),                                                                                                                      
      }]),                                                                                                                       
    };                                                                                                                           
    await processAutoSuggest(context, opportunity, site);                                                                        
    expect(sqsStub.called).to.be.false;
  });                                                                                                                            
                                                             
  Perspective: QA/Testing                                                                                                        
                                                             
  Nice to Have

  N1. Fixture in test 6 has an irrelevant source_index — test/audits/cwv/auto-suggest.test.js:~847                               
  
  returns false for PENDING_VALIDATION when isCodeChangeAvailable is true includes source_index: 0 in the fixture, but the isCodeChangeAvailable check executes before the PENDING_VALIDATION branch — source_index never influences the result. Removing it makes the fixture's purpose clearer (the test is about isCodeChangeAvailable, not about granular guidance).                 
                                                             
  Perspective: QA/Testing                                                                                                        
  
  Recommendations                                                                                                                
                                                             
  The deployment ordering is documented in both PR bodies ("Deploy Mystique first"). This is the right approach — deploying audit-worker first would mean the original bug persists briefly (not a new regression), so the safe failure mode is acceptable. The prose documentation is sufficient; no code guard needed.                                                                  
                                                             
  Verdict

  Ready to merge: With fixes                                                                                                     
  Reasoning: S1 is a real coverage gap at the integration boundary — worth closing before merge given this is a reliability fix for a production deadlock.

tkotthakota-adobe (approved):

ramboz:
@tkotthakota-adobe thanks for the review! Addressed **S1** in 69beabbf: added an integration test on `processAutoSuggest` asserting a `PENDING_VALIDATION` suggestion that already has granular guidance (issues carry `source_index`) is **not** re-dispatched — guards the no-re-spam path at the SQS-send boundary, not just via the `shouldSendAutoSuggestForSuggestion` helper. File green at 32 passing.

PR #2759, opened 2026-07-09
fix: cancel preflight audit runs when disabled for site, surface errorCode

bhellema:

## Related Issues

SITES-47836

## Summary
- Reject/cancel preflight audit runs instead of throwing when the preflight handler is disabled for a site (or the site lacks the required product-code entitlement) — the `AsyncJob` is set to `CANCELLED` with metadata instead of erroring out on every retry.
- Register the preflight handler in the links-selector test configs.
- Add a small `PreflightError` catalog (`src/preflight/error-constants.js`) with `code` / `message` / `description` / `classification` per entry, modeled after Content Backflow's error enum pattern.
- Surface a stable `errorCode` (`PREFLIGHT-100`) plus a static `reason` message on the cancellation job's metadata payload, so downstream consumers (e.g. the preflight MFE) can map the code to a localized message instead of parsing a freeform string.

## Test plan
- [x] `npx mocha test/audits/preflight.test.js` — 143 passing
- [x] `npm run lint` on changed files

### DIFF
--- a/src/preflight/handler.js
+++ b/src/preflight/handler.js
@@ -130,11 +130,23 @@ export const preflightAudit = async (context) => {
   const { S3_SCRAPER_BUCKET_NAME } = context.env;
   const jobId = job.getId();
 
-  log.info(`Processing preflight audit for ${site}`);
+  log.info(`Processing preflight audit for ${site.getId()}`);
 
   if (!(await isAuditEnabledForSite('preflight', site, context))) {
-    log.error(`[preflight-audit] site: ${site.getId()}, job: ${jobId}. Preflight is disabled for this site.`);
-    throw new Error(`[preflight-audit] site: ${site.getId()}. Preflight handler is disabled for this site.`);
+    const reason = `preflight audits disabled for site ${site.getId()}`;
+    log.info(`[preflight-audit] site: ${site.getId()}, job: ${jobId}. ${reason}, skipping`);
+
+    const jobEntity = await AsyncJobEntity.findById(jobId);
+    jobEntity.setStatus(AsyncJob.Status.CANCELLED);
+    jobEntity.setMetadata({
+      payload: {
+        siteId: site.getId(),
+        reason,
+      },
+    });
+    jobEntity.setEndedAt(new Date().toISOString());
+    await jobEntity.save();
+    return;
   }
 
   const jobMetadata = job.getMetadata();
--- a/src/preflight/handler.js
+++ b/src/preflight/handler.js
@@ -22,6 +22,7 @@ import {
 } from './utils.js';
 import { getDomElementSelector, toElementTargets } from './utils/dom-selector.js';
 import { AUDIT_ALT_TEXT } from './audit-constants.js';
+import { PreflightError } from './error-constants.js';
 import canonical from './canonical.js';
 import metatags from './metatags.js';
 import links from './links.js';
@@ -142,6 +143,7 @@ export const preflightAudit = async (context) => {
       payload: {
         siteId: site.getId(),
         reason,
+        errorCode: PreflightError.PREFLIGHT_DISABLED.code,
       },
     });
     jobEntity.setEndedAt(new Date().toISOString());
--- a/src/preflight/handler.js
+++ b/src/preflight/handler.js
@@ -389,7 +389,7 @@ export const preflightAudit = async (context) => {
       },
     }));
 
-    log.debug(`[preflight-audit] site: ${site.getId()}, job: ${jobId}, step: ${step}. resultWithProfiling: ${JSON.stringify(resultWithProfiling)}`);
+    log.info(`[preflight-audit] site: ${site.getId()}, job: ${jobId}, step: ${step}. resultWithProfiling: ${JSON.stringify(resultWithProfiling)}`);
 
     const jobEntity = await AsyncJobEntity.findById(jobId);
     const anyProcessing = handlerResults.some((r) => r && r.processing === true);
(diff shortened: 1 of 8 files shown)


atopper:
Can we add a document giving guidance to developers/agents handling new response errors, to follow this type of pattern?  Any review hook that could be used to enforce this?

src/preflight/handler.js
```diff
@@ -130,6 +131,26 @@ export const preflightAudit = async (context) => {
   const { S3_SCRAPER_BUCKET_NAME } = context.env;
   const jobId = job.getId();
 
+  log.info(`Processing preflight audit for ${site.getId()}`);
+
+  if (!(await isAuditEnabledForSite('preflight', site, context))) {
```
atopper: This new block runs before the existing IN_PROGRESS check (line 177), so a redelivered/retried job that already reached a terminal state can have its real result silently overwritten with a CANCELLED payload.
I imagine someone would have to be playing with the 'enable' switch for this to happen for the same site to encounter this.

Suggestion:
Move the new block to run after the existing IN_PROGRESS guard, so both checks share the same precondition instead of the cancellation firing unconditionally on every invocation.

src/preflight/handler.js
```diff
@@ -130,6 +131,26 @@ export const preflightAudit = async (context) => {
...
+    jobEntity.setStatus(AsyncJob.Status.CANCELLED);
+    jobEntity.setMetadata({
+      payload: {
+        siteId: site.getId(),
+        reason: PreflightError.PREFLIGHT_DISABLED.message,
+        errorCode: PreflightError.PREFLIGHT_DISABLED.code,
+      },
+    });
+    jobEntity.setEndedAt(new Date().toISOString());
+    await jobEntity.save();
```
atopper: This needs a try/catch around it.  If `findById` (140) or `save()` (150) throws — bad connection, job not found, whatever — nothing catches it. The error just propagates out of preflightAudit unhandled, and the job might be left stuck IN_PROGRESS forever since none of the setStatus/setEndedAt/save() calls ran.

src/preflight/handler.js
```diff
@@ -130,6 +131,26 @@ export const preflightAudit = async (context) => {
   const { S3_SCRAPER_BUCKET_NAME } = context.env;
   const jobId = job.getId();
 
+  log.info(`Processing preflight audit for ${site.getId()}`);
```
atopper: Does this a consistent log suffix? `[preflight-audit]`

src/preflight/handler.js
```diff
@@ -384,7 +405,7 @@ export const preflightAudit = async (context) => {
       },
     }));
 
-    log.debug(`[preflight-audit] site: ${site.getId()}, job: ${jobId}, step: ${step}. resultWithProfiling: ${JSON.stringify(resultWithProfiling)}`);
+    log.info(`[preflight-audit] site: ${site.getId()}, job: ${jobId}, step: ${step}. resultWithProfiling: ${JSON.stringify(resultWithProfiling)}`);
```
atopper: Some upgraded to 'log.info' in here.  Will they fire a lot?  Do we need all of them?

src/preflight/handler.js
atopper: Apparently `isAuditEnabledForSite` can return 'false' for more reasons then the obvious one.

The entitlement check this PR relies on (checkProductCodeEntitlements in audit-utils.js) is written to "fail closed" — if it hits an error talking to the entitlement service (network blip, service hiccup, timeout, whatever), it doesn't crash or bubble the error up. It just quietly returns false, same as it would if the site genuinely isn't entitled.

Before this PR, that quiet false only affected which individual checks got skipped for that run — it didn't touch the job's overall status.

Now, this PR wires that same false result into the top-level check that decides whether to cancel the whole job. So if the entitlement service has a temporary outage, the code can't tell the difference between "this site really isn't allowed to run preflight" and "we couldn't reach the service to check." Either way, it now permanently cancels the job and tells the customer (via the MFE) "Preflight audit is not enabled for this site" — a hard, final-sounding message — when the real problem might just be a blip that would've resolved itself on retry.

test/audits/preflight.test.js
```diff
@@ -1452,6 +1455,28 @@ describe('Preflight Audit', () => {
...
+        payload: {
+          siteId: 'site-123',
+          reason: 'The Preflight audit is not enabled for this site.',
+          errorCode: 'PREFLIGHT-100',
+        },
+      });
+      expect(finalJobEntity.setEndedAt).to.have.been.called;
+      expect(finalJobEntity.save).to.have.been.called;
+    });
+
```
atopper, line 1479: Should we have more tests?  Ones that cover:
- A job that's already finished (COMPLETED, FAILED, or already CANCELLED) when this code runs again — there's no test proving the cancellation logic won't overwrite a job that's no longer in progress.
- AsyncJob.findById or .save() failing — there's no test exercising what happens if those calls reject; nothing confirms the job ends up in a sane state (e.g. FAILED) rather than hanging.
- The entitlement check itself failing (as opposed to the handler simply being toggled off) — the test never drives the path where checkProductCodeEntitlements errors out and fails safe to false, so there's no test confirming what happens to the job when that's the actual cause of "disabled."

bhellema:
> Can we add a document giving guidance to developers/agents handling new response errors, to follow this type of pattern? Any review hook that could be used to enforce this?

Yes! love this idea, I'm adding a specific preflight's version of the CLAUDE.md that will provide this instruction.

atopper (approved):

PR #2683, opened 2026-06-16
refactor(prerender): remove PageCitability dependency, read recently-processed pathnames from S3 scrape status

dhavkuma:
## Summary

- Removes `writeToCitabilityRecords` and the associated `PageCitability` DB writes from the prerender audit — the prerender audit no longer reads from or writes to the `PageCitability` table
- Refactors `getRecentlyProcessedPathnames` from an async DB query (`PageCitability.allByIndexKeys`) to a synchronous read from `siteStatus.pages` already fetched from S3 at the start of `submitForScraping`. A page is treated as recently processed if its `scrapedAt` falls within the rolling window and it either `needsPrerender === true` or was successfully scraped (`needsPrerender === false && scrapingStatus === 'success'`)
- Removes 820 lines of test code covering the deleted functionality and updates filtering tests to mock S3 status pages instead of `PageCitability` stubs

## Motivation

The `PageCitability` table was a cross-audit dependency: the prerender audit wrote to it, and the page-citability audit read from it to detect recently-processed URLs. Reading from the S3 scrape status (`siteStatus.pages`) is simpler, keeps the data in one place, and removes the DB round-trip and connection pool pressure from `submitForScraping`.

## Test plan

- [ ] All 239 prerender handler tests pass (`npm run test:spec -- test/audits/prerender/handler.test.js`)
- [ ] No remaining `PageCitability` references in `src/prerender/` or its tests
- [ ] Daily batching / recency filtering behaviour verified via updated unit tests (S3 status pages as source of truth)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2761, opened 2026-07-10
fix: restore writeToCitabilityRecords writes to PageCitability entity

dhavkuma:
## Summary

- Restores `writeToCitabilityRecords` and its call site in `processContentAndGenerateOpportunities`, which write citability metrics (score, content ratio, word counts, edge-deployed flag) to the `PageCitability` entity for successfully scraped URLs.
- This function was removed in #2683/#2696 as part of refactoring the prerender audit's recently-processed-URL detection to read from S3 scrape status instead of `PageCitability`. That read-side refactor is being kept as-is — only the write side is restored, since the page-citability audit still depends on these writes to detect recently-processed URLs.
- Restores the corresponding unit tests (`describe('writeToCitabilityRecords', ...)`, 16 cases covering create/update, query-param dedup keys, error handling, batching, and concurrent-write races) and the integration test verifying citability metrics flow from `analyzeHtmlForPrerender` through to the write.

## Test plan

- [x] `npm run test:spec -- test/audits/prerender/handler.test.js` — 290 passing, 0 failing
- [x] `npm run lint`
- [x] `handler.js` coverage: 99.44% lines / 99.21% branches / 100% functions (remaining gaps are pre-existing, unrelated to this change)

### DIFF
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -1070,6 +1070,7 @@ export async function processOpportunityAndSuggestions(
  * @param {Object} context - Audit context with dataAccess and log
  * @returns {Promise<void>}
  */
+// todo: remove after LLMO-5912
 export async function writeToCitabilityRecords(comparisonResults, siteId, context) {
   if (!comparisonResults?.length) {
     return;
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -1061,6 +1061,87 @@ export async function processOpportunityAndSuggestions(
 }
 
 /**
+ * Writes citability metrics to the PageCitability entity for all successfully scraped URLs.
+ * This enables the page-citability audit to detect recently-processed URLs via its 7-day
+ * staleness filter, avoiding duplicate scraping across both audits.
+ *
+ * @param {Array} comparisonResults - Results from compareHtmlContent (all scraped URLs)
+ * @param {string} siteId - Site ID
+ * @param {Object} context - Audit context with dataAccess and log
+ * @returns {Promise<void>}
+ */
+export async function writeToCitabilityRecords(comparisonResults, siteId, context) {
+  if (!comparisonResults?.length) {
+    return;
+  }
+
+  const { dataAccess, log } = context;
+  const { PageCitability } = dataAccess;
+
+  if (!PageCitability?.allBySiteId) {
+    log.debug(`${LOG_PREFIX} PageCitability not available, skipping citability record writes`);
+    return;
+  }
+
+  const existingRecords = await PageCitability.allBySiteId(siteId);
+  const existingRecordsMap = new Map(
+    existingRecords.map((r) => [normalizePathnameWithQuery(r.getUrl()), r]),
+  );
+
+  const successful = comparisonResults.filter((r) => !r.error);
+  const WRITE_BATCH_SIZE = 10;
+
+  const writeOne = async (result) => {
+    const {
+      url,
+      citabilityScore,
+      contentGainRatio,
+      wordDifference,
+      wordCountBefore,
+      wordCountAfter,
+      isDeployedAtEdge,
+    } = result;
+    try {
+      const existing = existingRecordsMap.get(normalizePathnameWithQuery(url));
+      if (existing) {
+        existing.setCitabilityScore(citabilityScore ?? null);
+        existing.setContentRatio(contentGainRatio ?? null);
+        existing.setWordDifference(wordDifference ?? null);
+        existing.setBotWords(wordCountBefore ?? null);
+        existing.setNormalWords(wordCountAfter ?? null);
+        existing.setIsDeployedAtEdge(isDeployedAtEdge ?? false);
+        await existing.save();
+      } else {
+        await PageCitability.create({
+          siteId,
+          url,
+          citabilityScore: citabilityScore ?? null,
+          contentRatio: contentGainRatio ?? null,
+          wordDifference: wordDifference ?? null,
+          botWords: wordCountBefore ?? null,
+          normalWords: wordCountAfter ?? null,
+          isDeployedAtEdge: isDeployedAtEdge ?? false,
+        });
+      }
+      return true;
+    } catch (e) {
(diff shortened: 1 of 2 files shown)


src/prerender/handler.js
```diff
@@ -1060,6 +1060,88 @@ export async function processOpportunityAndSuggestions(
...
+
+  const { dataAccess, log } = context;
+  const { PageCitability } = dataAccess;
+
+  if (!PageCitability?.allBySiteId) {
+    log.debug(`${LOG_PREFIX} PageCitability not available, skipping citability record writes`);
+    return;
+  }
+
+  const existingRecords = await PageCitability.allBySiteId(siteId);
```
MysticatBot: issue (blocking): `PageCitability.allBySiteId(siteId)` is not wrapped in a try-catch. If this call fails (DB pool exhaustion, timeout, network error), the exception propagates up through `writeToCitabilityRecords` and will abort the entire prerender audit for this site.

Since citability writes are a supplementary side-effect (the prerender audit does not depend on them), a transient failure here should not block the audit.

**Fix:** Wrap the body of `writeToCitabilityRecords` (after the early-returns) in a top-level try-catch:

```js
try {
  const existingRecords = await PageCitability.allBySiteId(siteId);
  // ... rest of function body ...
} catch (e) {
  log.warn(`${LOG_PREFIX} Failed to write citability records for siteId=${siteId}: ${e.message}`);
}
```

src/prerender/handler.js
```diff
@@ -1060,6 +1060,88 @@ export async function processOpportunityAndSuggestions(
...
+      log.warn(`${LOG_PREFIX} Failed to write PageCitability for ${url}: ${e.message}`);
+      return false;
+    }
+  };
+
+  let written = 0;
+  for (let i = 0; i < successful.length; i += WRITE_BATCH_SIZE) {
+    const batch = successful.slice(i, i + WRITE_BATCH_SIZE);
+    // eslint-disable-next-line no-await-in-loop
+    const results = await Promise.all(batch.map(writeOne));
```
MysticatBot: issue (blocking): This pattern (`Promise.all(batch.map(writeOne))` where each `writeOne` calls `existing.save()` or `PageCitability.create()`) is the exact N+1 anti-pattern documented in this repo's CLAUDE.md:

> "Never do `Promise.all(suggestions.map((s) => s.save()))` - use bulk methods instead. Use `Collection.saveMany(items)` instead of N concurrent saves."

With up to 320 URLs per audit run, this fires up to 320 individual HTTP requests to PostgREST (in 32 sequential batches of 10).

**Fix:** Either use `PageCitability.saveMany(updatedRecords)` for the update path (if the collection supports it), or add an explicit comment justifying the deviation given the temporary nature (marked for removal per LLMO-5912) and the batch-size cap. The latter is acceptable for restored temporary code, but the justification should be in the code so the next reviewer does not re-flag it.

MysticatBot (changes requested):
Hey @dhavkuma,

**Complexity: HIGH** - large diff (675 lines).
_Changes: Restores `writeToCitabilityRecords` and its call site to write citability metrics to the PageCitability entity, along with 16 unit tests and an integration test (2 files)._

## Must fix before merge
1. [Important] Missing top-level error handling around `allBySiteId` - transient DB failure will abort the entire prerender audit - `src/prerender/handler.js:1087` (details inline)
2. [Important] N+1 write pattern (`Promise.all(batch.map(writeOne))` with individual `.save()` calls) violates documented CLAUDE.md database conventions - `src/prerender/handler.js:1138` (details inline)

<details>

- nit: Unbounded `allBySiteId` fetch loads all PageCitability records for the site into memory - acceptable given temporary nature (LLMO-5912) but risks scaling issues if the TODO lingers - `src/prerender/handler.js:1087`
- suggestion: Deduplicate `successful` by normalized URL before the batch loop to avoid noisy constraint violations on duplicate URLs within a single invocation - `src/prerender/handler.js:1092`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 30s | <b>Cost:</b> $5.59 | <b>Commit:</b> <code>ea524bc3eb67aedb5035206bfc158c19c6b6e784</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/prerender/handler.js
dhavkuma: This restores the earlier working version of `writeToCitabilityRecords` (it wasn't a new design choice), and the function is temporary — it's tagged `// todo: remove after LLMO-5912` and will be deleted once that ticket lands. Given that, I've kept the batched-write pattern as-is (capped at `WRITE_BATCH_SIZE = 10` per chunk, with per-item try/catch) rather than rearchitecting it to `saveMany`, and added a comment above the loop calling out this justification so it doesn't get re-flagged.

MysticatBot (approved):
Hey @dhavkuma,

**Verdict: Approve** - prior blocking issues resolved, no new concerns.
**Complexity: HIGH** - large diff (706 lines).
_Changes: Restores `writeToCitabilityRecords` and its call site to write citability metrics to the PageCitability entity, along with 16 unit tests and an integration test (2 files)._

- Top-level try-catch now wraps the function body, preventing transient DB failures from aborting the prerender audit.
- N+1 batched-write pattern accepted as a justified temporary deviation (inline comment documents the rationale, code is tagged for removal via LLMO-5912).

<details>

- nit: `existingRecords.map(r => [normalizePathnameWithQuery(r.getUrl()), r])` will throw if any record has a null/undefined URL (data corruption scenario) - the outer try-catch prevents a crash but skips all remaining writes. A `.filter(r => r.getUrl())` before building the map would let valid records proceed. - `src/prerender/handler.js:1093`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 9m 3s | <b>Cost:</b> $3.14 | <b>Commit:</b> <code>a8c52a15ce9d968d7bfb1b286e17a625ceb78095</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2763, opened 2026-07-12
fix(backlinks): keep broken backlinks whose target uses overrideBaseURL

Kanishkavijay39:
## Summary
- The `broken-backlinks` audit's subdomain filter compared each SEMrush result's target host only against `site.getBaseURL()`, while the SEMrush query itself was run against `site.resolveFinalURL()` (which honors `config.fetchConfig.overrideBaseURL` when set).
- Whenever a site has `overrideBaseURL` configured and it differs from `base_url`, those two domains never match, so **every** real broken backlink was silently rejected as "different subdomain" — no error, no opportunity, no suggestion, ever.
- Found while investigating a real customer report (HDFC Bank, `applyonline.hdfcbank.com` / override `applyonline.hdfc.bank.in`): the audit ran clean twice, SEMrush returned ~19 real live broken backlinks both times (confirmed via Splunk logs), yet the stored result was empty both times. An existing opportunity had also been incorrectly auto-resolved as a side effect.
- Fix: a backlink is now kept if its target matches *either* the site's base URL *or* the configured override URL, so both legitimate query domains are recognized as "the site."

## Test plan
- [x] Reproduced the bug with an isolated unit test against the real (unmodified) `brokenBacklinksAuditRunner` before applying the fix — confirmed it returned `[]` for a real, valid backlink.
- [x] Added two regression tests: one confirming override-domain backlinks are now kept, one confirming genuinely unrelated subdomains are still excluded.
- [x] Full existing `backlinks.test.js` suite passes unchanged (81 tests).
- [x] Full repo test suite (`npm test`) passes with 100% coverage maintained.
- [x] `npm run lint` clean.

Jira: SITES-47904

### DIFF
--- a/src/backlinks/handler.js
+++ b/src/backlinks/handler.js
@@ -281,9 +281,19 @@ export async function brokenBacklinksAuditRunner(auditUrl, context, site) {
 
     // Drop backlinks whose target belongs to a different subdomain than the audited site.
     // www and the bare domain are treated as the same host and are never filtered.
+    // A backlink is kept if it matches either the site's base URL or, when configured, the
+    // override URL the SEMrush query itself was run against (see withUrlResolver /
+    // site.resolveFinalURL()). Without this, every backlink would be rejected for sites with
+    // config.fetchConfig.overrideBaseURL configured, since the query domain and the base URL
+    // domain never match in that case (SITES-47904).
     const siteBaseURL = site.getBaseURL();
+    const overrideBaseURL = site.getConfig()?.getFetchConfig?.()?.overrideBaseURL;
     const filteredBacklinks = excludedFilteredBacklinks?.filter((backlink) => {
-      if (isOnDifferentSubdomain(backlink.url_to, siteBaseURL)) {
+      const differsFromBase = isOnDifferentSubdomain(backlink.url_to, siteBaseURL);
+      const differsFromOverride = overrideBaseURL
+        ? isOnDifferentSubdomain(backlink.url_to, overrideBaseURL)
+        : true;
+      if (differsFromBase && differsFromOverride) {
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @Kanishkavijay39,

**Verdict: Approve** - Clean, well-scoped bug fix with good test coverage.
**Complexity: LOW** - small diff, single service.
_Changes: Fixes the subdomain filter in the broken-backlinks audit to also accept backlinks targeting the configured overrideBaseURL domain, which was previously silently rejecting all real backlinks for sites with that config set (2 files)._

<details>

- nit: Debug log at `src/backlinks/handler.js:293` still only mentions `siteBaseURL` when excluding a backlink, but the exclusion now depends on both `siteBaseURL` and `overrideBaseURL`. When debugging sites with overrideBaseURL configured, the operator would benefit from seeing both checked domains. Consider: `` `Excluding backlink ${backlink.url_to}: different subdomain from ${siteBaseURL}${overrideBaseURL ? ` and ${overrideBaseURL}` : ''}` ``
- suggestion: The optional chaining `site.getConfig()?.getFetchConfig?.()?.overrideBaseURL` treats `getFetchConfig` as potentially missing from the config object, but `site.getBaseURL()` is called without defensive chaining on the same site instance. If Config always exposes `getFetchConfig`, the extra `?.()` is noise that may confuse future readers into thinking the method is optional. Consider aligning with the existing access patterns in this file.

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 9s | <b>Cost:</b> $2.16 | <b>Commit:</b> <code>965dfd148e9cf279af7c0d23afb8eea38dd86a69</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jkf276 (approved):

PR #2765, opened 2026-07-13
feat(evergreen-offsite-opportunities): Make offsite opportunities a stable, persistent entity that is refreshed in place every week (LLMO-6122)

grigoreadobe:
## Scope

https://jira.corp.adobe.com/browse/LLMO-6122

Make each cited, Reddit, and YouTube offsite opportunity a stable evergreen entity that refreshes in place every week instead of receiving a new identifier on every run.

This preserves customer decisions, Opportunity Workspace links, and Brand Claims references while keeping dashboard data current.

## Implementation

**Stable evergreen opportunity**

`resolveEvergreenOffsiteOpportunity` selects the existing `NEW` opportunity by site and audit type. If historical duplicates exist, it keeps the most recently updated record and retires the others to `IGNORED` before refresh continues.

**Explicit offsite persistence**

`persistOffsiteOpportunity` creates the first opportunity or refreshes the resolved evergreen record. Refreshes preserve the opportunity ID while replacing audit identity and offsite analysis data, and the implementation is restricted to supported offsite audit types so unrelated opportunities are unaffected.

**Payload and suppression safety**

`isValidOffsiteAnalysis` rejects malformed or mismatched payloads before mutation. `isSuppressedRun` keeps an `IGNORED` incoming run separate, so it cannot overwrite the customer-visible evergreen opportunity or its suggestions.

**Supported handlers**

The cited-analysis, reddit-analysis, and youtube-analysis guidance handlers use the shared refresh workflow. Wikipedia analysis is unchanged.

## Testing

- Full unit and integration suite passes with required coverage.
- Lint and build pass.
- Dedicated tests cover validation, first creation, in-place refresh, duplicate retirement, suppressed runs, and all three supported handlers.
- Local Postgres/PostgREST scenarios verified surfaced, suppressed, duplicate, and repeated-refresh behavior.

## Out of Scope

- Snapshot creation, restore, and retention
- Suggestion retention or deterministic suggestion deduplication
- Migration of suggestions or links from retired duplicate opportunities
- Historical dashboard querying and rollback
- Evidence-binding stability across runs
- Comparative QA gates or phased rollout

### DIFF
--- a/src/cited-analysis/guidance-handler.js
+++ b/src/cited-analysis/guidance-handler.js
@@ -17,10 +17,15 @@ import { Audit } from '@adobe/spacecat-shared-data-access';
 
 import { syncSuggestions } from '../utils/data-access.js';
 import { createOpportunityData } from './opportunity-data-mapper.js';
-import { convertToOpportunity } from '../common/opportunity.js';
 import { postMessageOptional, buildAnalysisVisibilityMessage } from '../utils/slack-utils.js';
 import { resolveBrandResultForSite, applyScopeToOpportunity } from '../utils/brand-resolver.js';
 import { fetchAnalysisFromPresignedUrl } from '../utils/analysis-fetch.js';
+import {
+  isValidOffsiteAnalysis,
+  persistOffsiteOpportunity,
+  resolveEvergreenOffsiteOpportunity,
+  isSuppressedRun,
+} from '../common/offsite-refresh.js';
 
 const AUDIT_TYPE = Audit.AUDIT_TYPES.CITED_ANALYSIS;
 const LOG_PREFIX = '[Cited]';
@@ -95,9 +100,22 @@ export default async function handler(message, context) {
 
     log.info(`${LOG_PREFIX} Processing ${suggestions.length} suggestions for ${companyName}`);
 
-    const auditType = opportunityData.type || AUDIT_TYPE;
+    // Use the handler-owned type; the payload may only confirm it.
+    const auditType = AUDIT_TYPE;
+    const incomingStatus = opportunityData.status || 'NEW';
 
-    const opportunity = await convertToOpportunity(
+    // Validate before mutating the evergreen opportunity.
+    if (!isValidOffsiteAnalysis(analysisData, auditType)) {
+      log.error(`${LOG_PREFIX} Malformed analysis payload for siteId: ${siteId}; skipping update`);
+      return badRequest('Malformed analysis payload');
+    }
+
+    const evergreenOpportunity = await resolveEvergreenOffsiteOpportunity({
+      dataAccess, siteId, auditType, log,
+    });
+
+    // Suppressed runs create a hidden record; surfaced runs reuse the evergreen record.
+    const opportunity = await persistOffsiteOpportunity(
       baseUrl,
       {
         siteId,
@@ -107,18 +125,17 @@ export default async function handler(message, context) {
       context,
       createOpportunityData,
       auditType,
-      { opportunityData },
-      (oppty) => oppty.getAuditId() === auditId,
+      {
+        opportunityData,
+        existingOpportunity: isSuppressedRun(incomingStatus)
+          ? null
+          : evergreenOpportunity,
+      },
     );
 
-    // Persist the opportunity (with scope) BEFORE syncing suggestions. Reversed
-    // ordering avoids the partial-write window where syncSuggestions succeeds
-    // but opportunity.save() throws — that previously left suggestions orphaned
-    // against an unsaved opportunity, which re-runs would observe in an
-    // unrecoverable state.
+    // Save the scoped opportunity before syncing its suggestions.
     applyScopeToOpportunity(opportunity, brandResult, log, LOG_PREFIX);
-    const status = opportunityData.status || 'NEW';
-    opportunity.setStatus(status);
+    opportunity.setStatus(incomingStatus);
     opportunity.setData({
       ...opportunity.getData(),
       fullAnalysis: analysisData,
@@ -154,7 +171,7 @@ export default async function handler(message, context) {
           analysisName: 'cited-analysis',
           baseUrl,
           suggestionsCount: suggestions.length,
-          isVisible: status !== 'IGNORED',
+          isVisible: incomingStatus !== 'IGNORED',
           verdict: opportunityData.qaVerdict,
         });
 
--- a/src/cited-analysis/opportunity-data-mapper.js
+++ b/src/cited-analysis/opportunity-data-mapper.js
@@ -17,7 +17,7 @@ import { DATA_SOURCES } from '../common/constants.js';
  * Creates opportunity data for cited URL analysis.
  * When a BO JSON opportunity object is provided (from Mystique), uses its values.
  * Otherwise falls back to defaults.
- * @param {Object} props - The props object from convertToOpportunity
+ * @param {Object} props - The options passed by persistOffsiteOpportunity
  * @param {Object} [props.opportunityData] - The opportunity object from the BO JSON
  * @returns {Object} Opportunity data
  */
--- a/src/common/constants.js
+++ b/src/common/constants.js
@@ -10,6 +10,8 @@
  * governing permissions and limitations under the License.
  */
 
+import { Audit } from '@adobe/spacecat-shared-data-access';
+
 /**
  * Data sources for opportunities
  */
@@ -22,3 +24,12 @@ export const DATA_SOURCES = {
 };
 
 export const ELMO_LIVE_HOST = 'https://main--project-elmo-ui-data--adobe.aem.live';
+
+/**
+ * Offsite audit types: opportunities refreshed on a recurring schedule.
+ */
+export const OFFSITE_AUDIT_TYPES = new Set([
+  Audit.AUDIT_TYPES.CITED_ANALYSIS,
+  Audit.AUDIT_TYPES.REDDIT_ANALYSIS,
+  Audit.AUDIT_TYPES.YOUTUBE_ANALYSIS,
+]);
--- a/src/common/offsite-refresh.js
+++ b/src/common/offsite-refresh.js
@@ -1,0 +1,172 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { Opportunity as Oppty } from '@adobe/spacecat-shared-data-access';
+import { DATA_SOURCES, OFFSITE_AUDIT_TYPES } from './constants.js';
+import { checkGoogleConnection } from './opportunity-utils.js';
+
+/**
+ * Validates the payload shape and rejects a declared opportunity type that differs from
+ * the handler-owned audit type.
+ *
+ * @param {*} analysisData - The parsed analysis payload (inline or fetched from S3).
+ * @param {string} expectedType - The calling handler's own audit type (e.g. 'cited-analysis').
+ * @returns {boolean} true if the payload is well-formed enough to process.
+ */
+export function isValidOffsiteAnalysis(analysisData, expectedType) {
+  if (!analysisData || typeof analysisData !== 'object' || Array.isArray(analysisData)) {
+    return false;
+  }
+  const { opportunity } = analysisData;
+  if (opportunity !== undefined
+      && (typeof opportunity !== 'object' || opportunity === null || Array.isArray(opportunity))) {
+    return false;
+  }
+  if (opportunity?.type !== undefined && opportunity.type !== expectedType) {
+    return false;
+  }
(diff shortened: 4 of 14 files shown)


src/cited-analysis/guidance-handler.js
```diff
@@ -96,7 +101,29 @@ export default async function handler(message, context) {
...
+    if (!isValidOffsiteAnalysis(analysisData)) {
+      log.error(`${LOG_PREFIX} Malformed analysis payload for siteId: ${siteId}; skipping update`);
+      return badRequest('Malformed analysis payload');
+    }
 
+    // Lazily de-duplicates any pre-existing NEW-status duplicates of this type for this
+    // site (self-heals leftovers from before opportunities of this type became evergreen).
+    // Side-effect only — its return value no longer gates this run (see below).
+    await resolveVisibleOpportunity({
+      dataAccess, siteId, auditType, log,
```
rarescheseli, line 114: issue (blocking): `auditType` is still derived from external `opportunity.type`. Previously, the audit-ID comparator constrained which existing opportunity could be reused. This PR deduplicates by the supplied type and then defaults to matching any NEW opportunity of that type. A cited payload carrying `type: 'reddit-analysis'` can therefore retire Reddit duplicates and overwrite the canonical Reddit opportunity with cited data and suggestions. Use the handler-owned `AUDIT_TYPE`, or reject mismatched types before deduplication and matching. Apply the same validation to Reddit and YouTube.

src/cited-analysis/guidance-handler.js
```diff
@@ -108,7 +135,7 @@ export default async function handler(message, context) {
       createOpportunityData,
       auditType,
       { opportunityData },
-      (oppty) => oppty.getAuditId() === auditId,
+      requiresNewOpportunity(status) ? () => false : undefined,
```
rarescheseli: issue (blocking): An `IGNORED` run forces creation, but `convertToOpportunity` creates the row without its status and returns after that database write. The handler only sets `IGNORED` in a subsequent save. If that save fails, suppressed content remains at the default visible status. Replays also always create another row because this comparator never matches. Create the complete suppressed record as `IGNORED` atomically and make replay idempotent using the audit ID. Apply the same correction to Reddit and YouTube.

src/common/offsite-refresh-guard.js
```diff
@@ -0,0 +1,106 @@
...
+    return matching[0];
+  }
+
+  const [canonical, ...duplicates] = [...matching].sort(
+    (a, b) => new Date(b.getUpdatedAt()) - new Date(a.getUpdatedAt()),
+  );
+
+  log.info(`[OffsiteRefreshGuard] Found ${matching.length} NEW ${auditType} opportunities for siteId ${siteId}; retiring ${duplicates.length} duplicate(s), keeping ${canonical.getId()} as canonical`);
+
+  await Promise.all(duplicates.map(async (duplicate) => {
```
rarescheseli: issue (blocking): `Promise.all(...save())` directly violates the repository's no-N+1 rule. Individual save failures are also swallowed, after which conversion re-queries NEW opportunities and may refresh a duplicate that failed retirement. Update the models, persist them with `Opportunity.saveMany`, propagate failures, and use the returned canonical opportunity rather than discarding it.

src/common/opportunity.js
```diff
@@ -92,7 +92,13 @@ export async function convertToOpportunity(auditUrl, auditData, context, createO
...
-      } else if (auditType === Audit.AUDIT_TYPES.PRERENDER) {
+      } else if (auditType === Audit.AUDIT_TYPES.PRERENDER
+          || auditType === Audit.AUDIT_TYPES.CITED_ANALYSIS
+          || auditType === Audit.AUDIT_TYPES.REDDIT_ANALYSIS
+          || auditType === Audit.AUDIT_TYPES.YOUTUBE_ANALYSIS) {
+        // Offsite types (cited/reddit/youtube) are evergreen: each refresh must replace the
+        // dashboard/performance data (sentiment, SoV, citations, topics) wholesale, not just
+        // dataSources — otherwise the Performance view only ever reflects the first run.
         opportunity.setData({
           ...opportunity.getData(),
```
rarescheseli, line 97: issue (blocking): The PR promises wholesale replacement, but spreading existing data retains every top-level field omitted by the latest analysis. Removed metrics or topics therefore remain stale. Replace refresh-owned data with the current mapped data, preserving only explicitly identified persistent fields, and test that an old-only field disappears.

rarescheseli (changes requested):
Hey @grigoreadobe,

**Verdict: Request changes** - four state-integrity issues need correction before merge.
_Changes: makes cited, Reddit, and YouTube opportunities persistent across refreshes, adds suppressed-run isolation and lazy duplicate retirement, and expands tests (10 files)._

## Must fix before merge
1. [Critical] Type-only matching lets malformed payloads hijack and deduplicate another offsite opportunity - `src/cited-analysis/guidance-handler.js:117` (details inline)
2. [Critical] A suppressed run can remain customer-visible after a failed save - `src/cited-analysis/guidance-handler.js:138` (details inline)
3. [Important] Duplicate retirement fails open and performs prohibited per-entity writes - `src/common/offsite-refresh-guard.js:78` (details inline)
4. [Important] Refresh data is merged instead of replaced wholesale - `src/common/opportunity.js:103` (details inline)

src/cited-analysis/guidance-handler.js
grigoreadobe: Fixed in 77fbd337: `auditType` is now always the handler's own `AUDIT_TYPE` constant, never derived from the inbound payload. `isValidOffsiteAnalysis` now also takes the expected type and rejects the payload outright if `opportunity.type` is present and disagrees with it, so a cross-type payload is a no-op rather than a hijack. Applied identically to cited/reddit/youtube.

src/common/offsite-refresh-guard.js
grigoreadobe: Fixed in 77fbd337: duplicate retirement now goes through a single `Opportunity.saveMany(duplicates)` call (matches the repo's no-N+1 rule and the existing pattern in e.g. `paid-traffic-analysis/guidance-handler.js`), and the failure is no longer swallowed — it propagates out of `resolveVisibleOpportunity` and up through the handler's outer catch, so a failed retirement now aborts the run with badRequest instead of proceeding against a still-ambiguous state. Also stopped discarding the resolved canonical: it's now passed to `convertToOpportunity` as `props.existingOpportunity` and used verbatim (new optional param, backward compatible — only used when explicitly present in props), so there's no second, independent query that could disagree with the dedup step.

src/common/opportunity.js
grigoreadobe: Fixed in 77fbd337: split the offsite branch off from prerender's (unaffected, left as a merge) and made it a true wholesale replace — `opportunity.setData({ ...opportunityInstance.data })` with no spread of the prior `opportunity.getData()`. Added a test with an old-only field on the existing opportunity (e.g. `fullAnalysis`/`staleTopic`) asserting it's absent from the new `setData` call.

src/cited-analysis/guidance-handler.js
grigoreadobe: Split into two parts:

Atomicity — fixed in 77fbd337. `createOpportunityData` already computes `status` on the mapped instance (e.g. `opportunityData?.status || 'NEW'`); `convertToOpportunity`'s create branch just never copied it into the `Opportunity.create()` payload. It now does (conditionally, only when the mapper provides one, so every other audit type's create payload is unaffected). The row is created at the correct status in the single write; there's no longer a window (or permanent state, on a follow-up save failure) where a suppressed run is visible at the schema-default NEW.

Idempotency on replay — not fixed here, flagging as a known limitation for a follow-up rather than folding into this PR. The reason it's not a simple comparator swap: `convertToOpportunity` only ever queries `NEW`-status opportunities (`Opportunity.allBySiteIdAndStatus(..., Oppty.STATUSES.NEW)`) before matching — a previously-created suppressed record is `IGNORED`, so no comparator choice makes it visible to that query; the auditId comparator this PR removed only worked before because the pre-existing behavior always operated over that same NEW-only set. Making a replay converge on the same suppressed record instead of creating another one needs either extending that shared, ~15-caller function's status scope, or a caller-side auditId+type lookup against IGNORED opportunities before calling it — happy to pick this up as a fast follow if you'd rather not merge with it open.

src/common/offsite-refresh-guard.js
```diff
@@ -0,0 +1,125 @@
...
+export async function resolveVisibleOpportunity({
+  dataAccess, siteId, auditType, log,
+}) {
+  const { Opportunity } = dataAccess;
+  let opportunities;
+  try {
+    opportunities = await Opportunity.allBySiteIdAndStatus(siteId, Oppty.STATUSES.NEW);
+  } catch (e) {
+    log.error(`[OffsiteRefreshGuard] Failed to fetch opportunities for siteId ${siteId}: ${e.message}`);
+    return null;
```
rarescheseli: issue (blocking): This catch converts a lookup failure into `null`, the same value used for a successful query with no matching opportunity. The follow-up commit now passes that result as an explicit `existingOpportunity`, so `convertToOpportunity` skips its own lookup and creates a new row. A transient database read failure during a surfaced run can therefore create a duplicate `NEW` opportunity. Log and rethrow the lookup error, reserve `null` for a successful empty result, and update the guard and handler-level failure tests to verify conversion is not attempted.

rarescheseli (changes requested):
Hey @grigoreadobe,

**Verdict: Request changes** - one fail-open regression remains in the canonical-opportunity handoff.
_Changes: addresses prior review findings for type validation, atomic suppressed creation, bulk duplicate retirement, canonical reuse, and wholesale data replacement (10 files)._

## Must fix before merge
1. [Important] A failed canonical lookup is treated as no match and creates a duplicate visible opportunity - `src/common/offsite-refresh-guard.js:76` (details inline)

- Handler-owned audit types prevent cross-type matching and creation.
- Suppressed opportunities are created atomically with `IGNORED` status.
- Duplicate retirement uses `saveMany`, propagates write failures, and reuses the selected canonical opportunity.
- Offsite refresh data is replaced wholesale while prerender retains merge behavior.

src/common/offsite-refresh-guard.js
grigoreadobe: Fixed in 1ced88fc: the catch block now logs and rethrows instead of returning `null`, so `null` is reserved strictly for the successful, genuinely-empty query result. A lookup failure now propagates out of `resolveVisibleOpportunity`, through the handler's outer catch, and aborts the run with `badRequest` — matching the pattern already used everywhere else in this codebase for this same kind of failure.

This also surfaced (and I fixed) a pre-existing, unrelated gap: `MockContextBuilder.withOverrides()` shallow-replaces the whole `dataAccess` object, so most tests in the three guidance-handler suites had no `dataAccess.Opportunity` mock at all — previously invisible because the swallowed `TypeError` didn't affect their assertions, but no longer once the lookup error is rethrown. Added a default `Opportunity.allBySiteIdAndStatus` stub to each file's base context, plus a guard-level test asserting the rethrow and handler-level tests (cited/reddit/youtube) asserting no duplicate is created on a lookup failure.

MysticatBot (approved):
Hey @grigoreadobe,

**Verdict: Approve** - all prior findings are addressed and no new issues found.
**Complexity: HIGH** - large diff (1734 lines, 10 files).
_Changes: makes cited, Reddit, and YouTube offsite opportunities persistent per (site, type) with lazy deduplication, suppressed-run isolation, wholesale data replacement, and atomic status on create (10 files)._

<details>
<summary>Non-blocking (2): minor suggestions</summary>

- suggestion: Consider extracting the three-way offsite type check (`CITED_ANALYSIS || REDDIT_ANALYSIS || YOUTUBE_ANALYSIS`) in `convertToOpportunity` into a shared constant set so new offsite channels only need one update - `src/common/opportunity.js:115`
- suggestion: A one-line comment above the dedup sort in `resolveVisibleOpportunity` noting the invariant ("canonical = most recently updated, preserving freshest data") would help future readers - `src/common/offsite-refresh-guard.js:89`
</details>

- Handler-owned audit types prevent cross-type matching and creation.
- Suppressed opportunities are created atomically with IGNORED status.
- Duplicate retirement uses saveMany, propagates write failures, and reuses the selected canonical opportunity.
- Offsite refresh data is replaced wholesale while prerender retains merge behavior.
- Canonical-lookup failure now rethrows instead of returning null, preventing duplicate creation on transient DB errors.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 21s | <b>Cost:</b> $4.41 | <b>Commit:</b> <code>1ced88fc2cae3853f17165f59e75f9a6ad814730</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

grigoreadobe:
Non-functional cleanup: renamed `offsite-refresh-guard.js` to `offsite-refresh.js` (and its test file accordingly) — the module now does more than gatekeeping (validation, dedup, visible/suppressed resolution) for all three offsite handlers, so the `-guard` suffix no longer fit. Also standardized on `evergreen` instead of `canonical` throughout the code, comments, log messages, and test names, to match the `evergreenOpportunityId` terminology already agreed for the follow-up UoW5 work.

No behavior change — full test suite and lint pass.

grigoreadobe:
Addressed both non-blocking suggestions from the last automated review pass (commit b7df4fe3): the three-way cited/reddit/youtube type check in `convertToOpportunity` is now a single `OFFSITE_ANALYSIS_TYPES` set in `common/constants.js`, and the dedup sort in `offsite-refresh.js` has a one-line comment on the "most-recently-updated wins" invariant.

rarescheseli (approved):

PR #2766, opened 2026-07-14
[Draft] fix(prerender): strip FontFaceObserver/Next.js font-detection noise from scraped HTML (LLMO-6061)

sneharora1:
## Summary
- FontFaceObserver and Next.js `@next/font` inject transient test elements into the DOM that Puppeteer captures before the library cleans them up, inflating the scraped word count and causing false `needsPrerender: true` results
- Adds `removeFontDetectionNoise()` in `html-comparator.js` that strips these elements from `scrapedHtml` via cheerio before passing to `calculateStats`
- Strips two patterns: `mmMwWLliI0fiflO` prefix (font-metrics probe, covers `&N` suffix variants) and 10+ repeated `word` tokens (character-width probe)
- Adds `html-comparator.test.js` with 16 tests covering both `isFontDetectionLeaf` unit cases and `analyzeHtmlForPrerender` integration cases

## Test plan
- [x] `npx mocha --require test/setup-env.js test/audits/prerender/html-comparator.test.js` — 16 passing

Closes LLMO-6061

### DIFF
--- a/src/index-local.js
+++ b/src/index-local.js
@@ -10,16 +10,10 @@
  * governing permissions and limitations under the License.
  */
 /* eslint-disable no-console */
-process.env.SPACECAT_SKIP_VAULT_SECRETS = '1';
-
-const { main: universalMain } = await import('./index.js');
+import { main as universalMain } from './index.js';
 
 /**
  * Local testing entry point
- *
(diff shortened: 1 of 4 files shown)


sneharora1:
Closing this since we need to make this change at scraper level.

PR #2767, opened 2026-07-14
fix: repair TOC guidance:table-of-contents Mystique roundtrip (LLMO-6167)

anagarwa:
## Summary

Four independent bugs prevented the TOC `guidance:table-of-contents` pipeline from ever completing successfully end-to-end — each one masked the next until fixed, so no real production TOC audit has ever successfully round-tripped through Mystique and back.

1. **`src/toc/handler.js`** — the headings guard `Array.isArray(hastNodes)` rejected the real HAST root-node shape (`{type: 'root', children: [...]}`) that `transformRules.value` actually stores. Headings sent to Mystique were therefore always empty, so Mystique's crew correctly (per its own instructions) returned empty prompts for every suggestion in every real audit run.
2. **`src/toc/guidance-handler.js`** — destructured `opportunityId`/`hasPrompts` in camelCase, but Mystique's reply nests them in `data` as snake_case (`opportunity_id`/`has_prompts`). `Opportunity.findById(undefined)` threw a validation error on every real Mystique reply.
3. **`src/toc/guidance-handler.js`** — read `Suggestion.STATUSES` off the request-scoped `dataAccess.Suggestion` collection, which doesn't carry that static map (unlike the class imported directly from `@adobe/spacecat-shared-data-access`, which `handler.js` already does correctly for the outbound side). Crashed with `Cannot read properties of undefined (reading 'FIXED')` the first time a real message got past bug #2.
4. **`src/toc/guidance-handler.js`** — persisted Mystique's prompts as-is, including its newer `PromptItem` fields (`type`/`topic`/`category`/`reasoning`/`supporting_evidence`) and non-standard `origin`/`source` values (`"TOC-grounded"`/`"TABLE OF CONTENTS"`). Normalized to the existing suggestion data convention used elsewhere: `{id, origin: "ai", prompt, source: "config", regions}`.

## Test plan

- [x] Each bug reproduced locally first (confirmed failing on unfixed code with a realistic input matching the real message shape), then confirmed fixed.
- [x] Added regression tests for all 4 fixes in `test/audits/toc.test.js` and `test/audits/toc/guidance-handler.test.js`.
- [x] Full `toc.test.js` + `guidance-handler.test.js` suite: 214 passing, 0 failing.
- [x] **Full live end-to-end verification** against a real dev SpaceCat opportunity with real scraped headings and a real `Audit` record: triggered via `mode:ai-only generatePrompts:true` → `spacecat-audit-worker` sent real headings to Mystique → Mystique generated real CrewAI prompts → `guidance-handler.js` wrote them back onto the suggestions with the correct `{id, origin, prompt, source, regions}` shape, `hasPrompts: true`, no errors. Verified directly via the SpaceCat API on the persisted suggestion data.

Related: LLMO-6167 (this closes out the audit-worker side of the same TOC ai-only prompt-generation feature; the companion Mystique-side fix — `TocOpportunityData` missing from the `Opportunity.data` Pydantic union — is in `mystique` PR #3394, already merged).

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/src/toc/guidance-handler.js
+++ b/src/toc/guidance-handler.js
@@ -11,6 +11,7 @@
  */
 
 import { badRequest, notFound, ok } from '@adobe/spacecat-shared-http-utils';
+import { Suggestion } from '@adobe/spacecat-shared-data-access';
 import { isEligibleTocSuggestion } from './eligibility-utils.js';
 
 /**
@@ -21,13 +22,14 @@ import { isEligibleTocSuggestion } from './eligibility-utils.js';
  * prompts are grounded in the page's headings rather than the checkType, it's correct to apply
  * the same prompt set to every non-terminal suggestion for that URL.
  * @param {Array<Object>} suggestions - Existing Suggestion entities for the opportunity
- * @param {Object} Suggestion - The Suggestion data-access collection (for STATUSES)
+ * @param {Object} SuggestionClass - The Suggestion class (for STATUSES; not the request-scoped
+ * data-access collection, which doesn't carry it)
  * @returns {Map<string, Array<Object>>} URL -> matching, non-terminal suggestion entities
  */
-function groupEligibleSuggestionsByUrl(suggestions, Suggestion) {
+function groupEligibleSuggestionsByUrl(suggestions, SuggestionClass) {
   const byUrl = new Map();
   suggestions.forEach((suggestion) => {
-    if (!isEligibleTocSuggestion(suggestion, Suggestion)) {
+    if (!isEligibleTocSuggestion(suggestion, SuggestionClass)) {
       return;
     }
     const { url } = suggestion.getData() || {};
@@ -43,6 +45,25 @@ function groupEligibleSuggestionsByUrl(suggestions, Suggestion) {
 }
 
 /**
+ * Normalizes a single Mystique-generated prompt into the legacy, minimal shape used
+ * by existing TOC suggestion data (e.g. CrowdStrike): {id, origin, prompt, source,
+ * regions}. Mystique's current PromptItem schema adds type/topic/category/reasoning/
+ * supporting_evidence and sets its own origin/source values, but TOC intentionally
+ * keeps the older, narrower field set rather than adopting the richer shape.
+ * @param {Object} prompt - Raw prompt object from Mystique
+ * @returns {Object} Normalized prompt: {id, origin: "ai", prompt, source: "config", regions}
+ */
+function normalizeTocPrompt(prompt) {
+  return {
+    id: prompt?.id,
+    origin: 'ai',
+    prompt: prompt?.prompt,
+    source: 'config',
+    regions: Array.isArray(prompt?.regions) ? prompt.regions : [],
+  };
+}
+
+/**
  * Handles the Mystique guidance:table-of-contents callback and persists the generated
  * Impact Engine prompts back onto their matching TOC suggestions.
  * @param {Object} message - Message from Mystique with generated prompts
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @anagarwa,

**Verdict: Approve** - well-scoped bug fix with solid regression tests for all four failure modes.
**Complexity: MEDIUM** - medium diff, single-service bug fix.
_Changes: Repairs four independent bugs that prevented the TOC guidance:table-of-contents Mystique roundtrip from completing end-to-end (4 files)._

<details>

- nit: `normalizeTocPrompt` passes through `undefined` for `id` and `prompt` when the input object lacks those fields, while `regions` gets a `[]` default - consider filtering out entries where `prompt?.prompt` is falsy since a prompt without content is not useful to persist - `src/toc/guidance-handler.js:55`
- suggestion: the `opportunityId` extraction uses optional chaining (`data?.opportunityId ?? data?.opportunity_id`) while `suggestions` uses `const { suggestions } = data || {}` - unifying to a single destructure pattern would make the two casing-fallback workarounds read consistently - `src/toc/guidance-handler.js:80`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 11m 40s | <b>Cost:</b> $5.29 | <b>Commit:</b> <code>529e8e7014ad404727318f48b82ebca8f5a36dc6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @anagarwa,

**Verdict: Approve** - well-tested bug fix with sound self-healing logic; no blocking concerns.
**Complexity: MEDIUM** - medium diff, single-service bug fix.
_Changes: Repairs four independent bugs preventing the TOC guidance:table-of-contents Mystique roundtrip from completing, plus a fifth fix making hasPrompts self-healing by deriving it from the actual prompts array (4 files)._

<details>

- nit: `normalizeTocPrompt` passes through `undefined` for `id` and `prompt` when the input object lacks those fields - if Mystique ever sends a malformed entry like `{type: 'Non-Branded'}` without core fields, `promptsArray.length > 0` is truthy and the useless normalized entry replaces existing real prompts. Consider filtering out entries where `prompt?.prompt` is falsy before mapping, or using conditional property assignment so undefined keys are omitted. - `src/toc/guidance-handler.js:56`
- nit: existing prompts preserved via the fallback path (`existingData?.prompts ?? []`) bypass `normalizeTocPrompt`, so suggestions written before this PR retain the old un-normalized shape until Mystique next sends fresh prompts for them. Acceptable if no backfill is planned, but worth a code comment noting the intent. - `src/toc/guidance-handler.js:140`

</details>

- Suggestion naming/destructure consistency addressed via clearer `SuggestionCollection` vs imported `Suggestion` class distinction with inline comments explaining the split.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 52s | <b>Cost:</b> $7.58 | <b>Commit:</b> <code>adb0e7cbff9a4e69262f9fb9272843236f88dbf8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/toc/guidance-handler.js
```diff
@@ -97,29 +125,37 @@ export default async function handler(message, context) {
...
+    const { url, prompts } = promptResult;
     const matches = eligibleByUrl.get(url);
 
     if (!matches || matches.length === 0) {
       log.warn(`[TOC Guidance] No matching suggestion found for URL: ${url}`);
       return;
     }
 
-    const promptsArray = Array.isArray(prompts) ? prompts : [];
+    const promptsArray = (Array.isArray(prompts) ? prompts : []).map(normalizeTocPrompt);
```
ymadobe, line 136: `normalizeTocPrompt` persists content-free prompts and reopens the exact "permanently skipped" bug this PR targets - `src/toc/guidance-handler.js:136` (and `:56`)**
> MysticatBot flagged the `undefined` id/prompt passthrough as a cosmetic nit; the stronger take is that it undercuts the self-healing goal that is the whole point of this change. If Mystique replies with `prompts: [{}]` or an entry lacking prompt text, you persist `{id: undefined, origin: 'ai', prompt: undefined, source: 'config', regions: []}` - a structurally valid but semantically empty element. Because `hasPrompts` is derived from `finalPrompts.length`, that suggestion is now marked `hasPrompts: true` with a garbage payload, and the outbound self-heal (`data?.prompts?.length > 0`) also sees length 1, so it is never re-sent to Mystique. That is precisely the stuck state the array-derived flag was meant to prevent, entering through a different door. Given the whole PR exists because Mystique's serialization is unreliable, trusting every array element to be well-formed is inconsistent with the premise. Fix: filter to usable entries (and coerce to string to protect downstream render/inject sinks), e.g. `.map(normalizeTocPrompt).filter((p) => p.id && p.prompt)`, and add a test for `prompts: [{ topic: 'x' }]` with nothing existing to preserve, asserting persisted `prompts: []` / `hasPrompts: false`.

anagarwa: @ymadobe IMO thats a rare case and it would need to check too many edge case for something a rare case. I would skip this

ymadobe: okay, rest looks good .

ymadobe (approved):

PR #2769, opened 2026-07-15
feat(cdn-logs-report): generate referral URL categories for sites without CDN logs | LLMO-6257

omairtemurian:
## 1. Abstract
Extends the `cdn-logs-report` audit so sites with no CDN logs still get URL category classifications, sourcing the corpus from Postgres referral traffic instead of Athena and reusing the existing category-generation machinery.

## 2. Reasoning
Category classifications are only produced by this audit's CDN path, whose URL corpus comes from Athena (CDN access logs). A site whose traffic comes only from non-CDN sources (optel / GA4 / Adobe Analytics / CJA) never gets classifications, so the shipped Referral Traffic "Category" filter shows only "All". This is the producer side of LLMO-6257: generate those categories from referral traffic.

A new best-effort step in `runCdnLogsReport`, gated to sites without CDN data:

- `report-utils.js` gains two thin data-service clients: `fetchReferralTopUrls` (calls `rpc_referral_traffic_top_urls`) and `applyCategoryRulesToReferral` (calls `wrpc_apply_category_rules_to_referral`).
- `patterns-uploader.js` gains `generateReferralPatternsWorkbook`: fetch the site's top referral URLs, reuse `analyzeProducts` + the existing rule merge, persist rules via `replaceAgenticUrlClassificationRules`, then materialise them onto referral URLs via the apply RPC. Only category rules are (re)generated; existing page-type rules are preserved.
- `handler.js` gains the `generateReferralPatterns` wrapper. `hasCdnData` is now computed once and shared with the agentic step, so the two paths are mutually exclusive (no double LLM spend) and a CDN site never runs the referral path.

Behaviour delta: before, a referral-only site produced no category classifications; after, the daily audit generates and applies them. The classification sink is unchanged, so the read-path Category filter lights up with no read-side changes. The step is best-effort - a failure (LLM, RPC, PostgREST) is logged and never blocks the daily exports - and on re-runs it reuses existing rules (no LLM re-hit) and skips the whole-site rule replace so it does not churn customer-tuned categories.

## 4. Required information
- Jira / issue: [LLMO-6257](https://jira.corp.adobe.com/browse/LLMO-6257) (epic LLMO-5440)
- Spec: [docs/specs/2026-07-15-referral-category-generation.md](https://github.com/adobe/spacecat-audit-worker/blob/LLMO-6257-referral-category-generation/docs/specs/2026-07-15-referral-category-generation.md)

- `mysticat-data-service` - this change calls the two new RPCs (`rpc_referral_traffic_top_urls`, `wrpc_apply_category_rules_to_referral`) at runtime; contract dependency, added in companion PR adobe/mysticat-data-service pull 807.

## 7. Test plan
(a) Local: behavioural verification of the new step via the unit suite (data-service RPCs and the LLM mocked at the module boundary) - confirmed it is gated to non-CDN sites, is best-effort (a thrown RPC/LLM error is caught, logged, and does not block the exports), reuses existing product rules without re-hitting the LLM, and skips the whole-site rule replace on the reuse path. The full end-to-end (real Lambda invocation) has not been run locally.

(b) Per env (dev / stage / prod, after deploy, and only once the data-service companion is deployed): trigger a `cdn-logs-report` run for a referral-only site (no CDN aggregated data, but referral rows present) and confirm it generates `agentic_url_category_rules` and populates `agentic_url_classifications`, and that the site's Referral Traffic Category dropdown then fills. Confirm a CDN-covered site is unaffected (referral path skipped).

- Depends-on: adobe/mysticat-data-service pull 807 - this audit calls the two RPCs that PR adds, so they must exist in the deployed data-service first.
- Ordered sequence: merge/deploy the data-service PR (807) first, then this PR. Both are drafts under a prod-deploy hold - do not merge until the deploy can be babysat.

### DIFF
--- a/src/cdn-logs-report/utils/report-utils.js
+++ b/src/cdn-logs-report/utils/report-utils.js
@@ -152,3 +152,72 @@ export async function replaceAgenticUrlClassificationRules({
 
   return Array.isArray(data) ? data[0] : data;
 }
+
+/**
+ * Fetches a site's top distinct referral url_paths (ranked by summed pageviews
+ * across all referral sources) via the read RPC. This is the Postgres analogue of
+ * the CDN Athena top-URLs query — the URL corpus fed to category-rule generation
+ * for referral-only sites (LLMO-6257). Returns a flat array of path strings.
+ */
+export async function fetchReferralTopUrls({ site, context, limit = 200 }) {
+  const siteId = site.getId();
+  const postgrestClient = context?.dataAccess?.services?.postgrestClient;
+
+  if (!postgrestClient?.rpc) {
+    throw new Error('PostgREST client is required to fetch referral top URLs');
+  }
+
+  const { data, error } = await postgrestClient.rpc(
+    'rpc_referral_traffic_top_urls',
+    {
+      p_site_id: siteId,
+      p_limit: limit,
+    },
+  );
+
+  if (error) {
+    context?.log?.error?.(`Failed to fetch referral top URLs for site ${siteId}: ${error.message}`);
+    throw error;
+  }
+
+  return (Array.isArray(data) ? data : [])
+    .map((row) => row?.url_path)
+    .filter((path) => typeof path === 'string' && path.length > 0);
+}
+
+/**
+ * Materializes a site's active category rules onto its referral URLs via the writer
+ * RPC, upserting category_name into agentic_url_classifications (LLMO-6257). Returns
+ * the RPC payload ({ site_id, classified }).
+ */
+export async function applyCategoryRulesToReferral({
+  site,
+  context,
+  source = null,
+  since = null,
+  updatedBy = 'audit-worker:referral-patterns',
+}) {
+  const siteId = site.getId();
+  const postgrestClient = context?.dataAccess?.services?.postgrestClient;
+
+  if (!postgrestClient?.rpc) {
+    throw new Error('PostgREST client is required to apply category rules to referral');
+  }
+
+  const { data, error } = await postgrestClient.rpc(
+    'wrpc_apply_category_rules_to_referral',
+    {
+      p_site_id: siteId,
+      p_source: source,
+      p_since: since,
+      p_updated_by: updatedBy,
+    },
+  );
+
+  if (error) {
+    context?.log?.error?.(`Failed to apply category rules to referral for site ${siteId}: ${error.message}`);
+    throw error;
+  }
+
+  return Array.isArray(data) ? data[0] : data;
+}
--- a/docs/specs/2026-07-15-referral-category-generation.md
+++ b/docs/specs/2026-07-15-referral-category-generation.md
@@ -1,0 +1,130 @@
+# Referral Category Generation for Sites Without CDN Logs
+
+| Field | Value |
+|-------|-------|
+| **Status** | Proposed |
+| **Author** | Omair Temurian |
+| **Created** | 2026-07-15 |
+| **Jira** | [LLMO-6257](https://jira.corp.adobe.com/browse/LLMO-6257) (epic LLMO-5440) |
+
+---
+
+## Summary
+
+Extends the `cdn-logs-report` audit so sites whose referral data comes **only** from
+non-CDN sources (optel / GA4 / Adobe Analytics / CJA — stored in the
+`mysticat-data-service` Postgres, not Athena) still get URL **category** classifications.
+It reuses the existing CDN category-generation machinery (`analyzeProducts` → merge →
+`wrpc_replace_agentic_url_classification_rules`), but sources the URL corpus from Postgres
+and materializes the rules onto referral URLs via a new data-service RPC. The
+classification sink (`agentic_url_classifications`) is unchanged, so the shipped Referral
+Traffic category read-filter (LLMO-6025/6/7) works with zero read-side changes.
+
+---
+
+## Problem Statement
+
+The Referral Traffic category filter resolves category by joining referral rows to
+`agentic_url_classifications`. That table is populated **only** by the `cdn-logs-report`
+audit, whose entire URL corpus comes from Athena (CDN access logs). A site with referral
+data only from non-CDN sources therefore has no corpus → `analyzeProducts` never runs →
+no rules → no classifications → the category dropdown shows only "All".
+
+---
+
+## Goals
+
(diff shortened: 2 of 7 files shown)


src/cdn-logs-report/utils/report-utils.js
```diff
@@ -152,3 +152,72 @@ export async function replaceAgenticUrlClassificationRules({
...
+
+  if (!postgrestClient?.rpc) {
+    throw new Error('PostgREST client is required to fetch referral top URLs');
+  }
+
+  const { data, error } = await postgrestClient.rpc(
+    'rpc_referral_traffic_top_urls',
+    {
+      p_site_id: siteId,
+      p_limit: limit,
```
cwjwisse, line 174: issue (blocking): `fetchReferralTopUrls` calls `rpc_referral_traffic_top_urls` with only `{ p_site_id, p_limit }`, but the in-repo spec (`docs/specs/2026-07-15-referral-category-generation.md`) documents the signature as `rpc_referral_traffic_top_urls(p_site_id, p_since, p_limit)`, and the sibling `applyCategoryRulesToReferral` in this same file threads `p_since` (as `null`). The asymmetry leaves this call silently depending on the data-service RPC defaulting `p_since`.

I verified the companion PR (adobe/mysticat-data-service pull 807) currently declares `p_since DATE DEFAULT NULL`, so this is not a live outage today - but the code relies on that default without saying so. If a future migration makes `p_since` required, PostgREST rejects the arg-less call, `fetchReferralTopUrls` throws, and the best-effort catch swallows it: referral category generation goes off for every no-CDN site with only a log line.

Fix (in-repo either way): add a `since = null` param to `fetchReferralTopUrls` and pass `p_since: since` in the RPC args, matching the documented 3-arg signature and the sibling wrapper. Or, if the 2-arg form is intentional, correct the spec's documented signature to `(p_site_id, p_limit)` so spec and code agree.

cwjwisse (changes requested):
Hey @omairtemurian,

**Verdict: Request changes** - one should-fix before merge: the referral read RPC is called without the `p_since` arg its own spec documents, leaving a hidden dependency on a downstream SQL default.
**Complexity: HIGH** - large diff (~720 lines); single service, no high-risk-surface signals.
_Changes: adds a best-effort referral-URL category-generation step to the cdn-logs-report audit for sites without CDN logs, sourcing the corpus from two Postgres RPCs and reusing the existing category machinery (7 files)._
Note: Recommend a human read before merge - possible divergence from the documented design (heuristic: the referral-read RPC call shape does not match the signature in the in-repo spec). The bot review complements, not replaces, a human read here.

## Must fix before merge
1. [Important] `fetchReferralTopUrls` calls `rpc_referral_traffic_top_urls` with `{p_site_id, p_limit}`, dropping the `p_since` arg the in-repo spec documents and that the sibling apply-wrapper threads - `src/cdn-logs-report/utils/report-utils.js:174` (details inline)

<details>
<summary>Non-blocking (6): minor issues and suggestions</summary>

- suggestion: Gate fails open on S3 error. `pathHasData` swallows its errors and returns `false`, so a transient S3 blip makes a real CDN site read as no-CDN and run the referral path (bounded: it lands on the reuse/apply-only branch, no rule churn), while the "No agentic report data found" log becomes misleading. No test covers this S3-error branch. `src/cdn-logs-report/handler.js` (hasCdnData computation)
- suggestion: Referral corpus is not normalized like the Athena corpus. The CDN path strips query/fragment and caps path length before `analyzeProducts`; the referral `url_path` values reach the LLM unfiltered. These paths originate from client-emitted analytics (optel/GA4/AA/CJA), so they are lower-trust than server-observed CDN logs - the existing regex guardrails contain the blast radius to the tenant's own classification data, but adding the same normalization guard (or a comment noting `url_path` is pre-cleaned) would keep the two paths symmetric. `src/cdn-logs-report/patterns/patterns-uploader.js`
- nit: Duplicate error logging. The RPC wrappers log-and-throw, then the best-effort handler catch logs the same failure again (two lines per failure). `src/cdn-logs-report/utils/report-utils.js`
- nit: `site.getId?.()` vs `site.getId()` used inconsistently within the new referral function - pick one. `src/cdn-logs-report/patterns/patterns-uploader.js`
- suggestion: Move `new URL(site.getBaseURL()).hostname` into the fresh-generation branch - it is only consumed by `analyzeProducts`, so it is dead work on the reuse path and a malformed base URL would needlessly abort a reuse run that never needed it. `src/cdn-logs-report/patterns/patterns-uploader.js`
- suggestion: The spec mermaid diagram routes the reuse path through the whole-site replace, but the code correctly skips replace on reuse - branch the diagram on `reusedExistingRules` so it matches the shipped behavior. `docs/specs/2026-07-15-referral-category-generation.md`
</details>

Out of scope, worth tracking: the RPCs' actual per-tenant scoping and any ReDoS/statement-timeout bound live in the SQL bodies in companion PR adobe/mysticat-data-service#807, not this repo - the audit-worker passes `p_site_id` correctly, but the RPC internals are not verifiable from here. Track that verification on the companion PR.

akshaymagapu (changes requested):
`cdn-logs-report` audit might not be correct place, the patterns can be shared but the sites we enable here is purely for the ones that select CDN provider in UI. (and we should not enable for all sites as its waste of resources)

If the goal is to have categories create for referral traffic that dont have CDN data available, that should be part of other audit which runs for all sites of LLMO like referral traffic specific audits.

so idea is you run for all LLMO sites via that audit, if a site do not have any category patterns in DB, you use the pattern logic to compute via LLM and insert into URL category rules table.
(Only if not exist) and the sample urls of referral is good enough.

and on down stream DRS / Audit worker of Referral traffic related producers, you use these regexes to match against urls and store in the Referral traffic related tables (please dont use `agentic_url_classifications` as source of truth)

omairtemurian:
Closing — superseded by #2821.

Per @akshaymagapu's review here, embedding category generation in `cdn-logs-report` was the wrong place (that audit only runs for CDN-provider sites). The right approach is a standalone audit that runs for all referral/LLMO sites with create-if-missing LLM rule-gen — which is exactly what #2821 (`llmo-referral-category-rules` sweeper) does, now merged. So this cdn-embedded approach is replaced.

Thanks @cwjwisse and @akshaymagapu — your feedback drove the pivot to the sweeper.

PR #2771, opened 2026-07-15
feat(offsite-snapshot-opportunties): Make offsite opportunities snapshots on every refresh (LLMO-6152)

grigoreadobe:
## Scope

https://jira.corp.adobe.com/browse/LLMO-6152

Preserve offsite refresh history without changing the customer-visible evergreen opportunity introduced in #2765. A surfaced refresh snapshots the evergreen state it is about to replace; a suppressed refresh is stored directly as an inert snapshot.

This is a stacked PR on #2765 (`LLMO-6122`).

## Implementation

**Dedicated snapshot workflow**

Added `src/common/offsite-snapshot.js` with explicit `prepareSupersededRunSnapshot` and `prepareSuppressedRunSnapshot` workflows backed by shared metadata, tagging, and lookup helpers.

Managed snapshots are `IGNORED`, tagged `offsite-snapshot`, and contain `data.snapshot` metadata:

- `kind`: `superseded-refresh` or `suppressed-refresh`
- `evergreenOpportunityId`: included when an evergreen exists
- `triggerAuditId`: identifies the incoming refresh when available

**Superseded refresh snapshots**

Before a surfaced refresh overwrites an existing evergreen opportunity, its data, scope, audit identity, and suggestions—including review statuses and metadata—are copied to a snapshot. A first surfaced run has no previous state to preserve.

**Suppressed refresh snapshots**

A suppressed run is persisted as a tagged snapshot from its initial write. It links to the evergreen opportunity when one exists and never mutates that evergreen record.

**Redelivery handling**

When `triggerAuditId` is available, `findSnapshotByTriggerAuditId` reuses an existing snapshot for the same site, audit type, and triggering audit. An initially unlinked suppressed snapshot can receive its evergreen link on redelivery. Without an audit ID, the record is still managed snapshot history, but idempotency and traceability are unavailable.

## Testing

- Full unit and integration suite, lint, and build pass.
- Dedicated unit and real-composition integration tests cover both snapshot workflows and redelivery reuse.
- All three guidance handlers cover surfaced, suppressed, first-run, missing-audit-ID, and scope-preservation paths.
- Local Postgres/PostgREST scenarios verified real persistence, suggestion copying, and unlinked-to-linked redelivery.

## Out of Scope

- Guided snapshot restore (LLMO-6153)
- Automatic snapshot deletion (LLMO-6154)
- New Backoffice UI
- Retroactive tagging of existing ignored opportunities
- Wikipedia analysis
[...]

PR #2772, opened 2026-07-16
feat(offsite-snapshots-deletion): Outdated snapshots deletion after 30 days (LLMO-6154)

grigoreadobe:
## Scope

https://jira.corp.adobe.com/browse/LLMO-6154

Hard-delete managed offsite snapshots older than 30 days after a weekly refresh, keeping operator history bounded without adding a scheduler or changing the evergreen opportunity. See [ADR 004](docs/decisions/004-offsite-snapshot-retention-window.md) for the 30-day rationale and [the spec](docs/specs/2026-07-20-offsite-snapshot-retention.md) for the full design.

This is a stacked PR on #2771 (`LLMO-6152`), which creates the snapshots managed here.

## Implementation

**Shared retention module**

Added `src/common/offsite-retention.js`. `findExpiredSnapshots` loads `IGNORED` opportunities for the site and selects only records of the current audit type carrying the reserved `offsite-snapshot` tag with `createdAt` strictly before the 30-day cutoff. Exactly 30 days old is retained.

The reserved tag is the ownership marker. Optional snapshot provenance metadata is not required for deletion, so a malformed managed snapshot does not become permanent.

**Bulk deletion (not per-record)**

`deleteExpiredSnapshots` deletes at most `MAX_DELETIONS_PER_RUN` (50) of the oldest eligible snapshots per invocation, via two bulk `removeByIds` calls: suggestions first (read sequentially per snapshot, then bulk-removed), then the snapshot opportunities themselves. This mirrors the existing `removeOpportunityWithSuggestions` pattern elsewhere in the repo and avoids the N+1 concurrent-write fan-out that a `Promise.all` over individual `.remove()` calls would produce against the shared connection pool — worst exactly when a dormant site resumes refreshing after an accumulated backlog. The trade-off: a single `removeByIds` failure now fails the whole batch rather than isolating per-record failures; this is documented as a deliberate deviation in the module and the spec.

**Operational logging**

Deletion and lookup failures are logged via the shared `createOffsiteLogger` (structured `key=value` fields), consistent with the rest of the offsite modules. A success summary (`eligible`/`deleted`) is emitted only when at least one snapshot was actually deleted, to avoid a log line on every refresh across three audit types × N sites when there's nothing to report.

**Refresh integration**

Snapshot deletion runs after persistence and suggestion synchronization in cited-analysis, reddit-analysis, and youtube-analysis. Handler-level guards ensure deletion failures cannot turn a successful refresh into a failed response.

`isOffsiteSnapshot` is shared with snapshot redelivery lookup so ownership rules remain consistent.

## Testing

- Full unit and integration suite, lint, and build pass with required coverage.
- Tests cover ownership filtering, strict cutoff boundaries, ordering, lookup failures, bulk deletion (with and without suggestions), the per-run deletion cap, propagated bulk-delete failures, structured logs, and empty passes.
- All three handlers verify deletion after surfaced and suppressed refreshes, non-blocking failure behavior, and that retention is skipped when the refresh itself fails before reaching the success path.
- Local Postgres/PostgREST scenarios verified deletion of 31-day snapshots and dependent suggestions while retaining recent, boundary, untagged, and wrong-type records.

## Out of Scope

- Count-based or same-day snapshot consolidation
- A scheduled or infrastructure-level sweep for sites whose audit is **disabled or offboarded** (as opposed to merely **dormant**): a dormant site self-heals on its next refresh, since retention runs as a side effect of that refresh; a disabled/offboarded site never refreshes again, so its snapshots are never revisited. Bounding that case needs a separate scheduled sweep, tracked as a follow-up.
- Soft deletion or tombstones
- Per-site or per-customer retention configuration
- Snapshot restore and picker behavior (LLMO-6153)
- Wikipedia analysis
- One-time migration or backfill sweep

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Rares Cheseli", "Dominique Jäggi"]
```

### DIFF
--- a/src/cited-analysis/guidance-handler.js
+++ b/src/cited-analysis/guidance-handler.js
@@ -34,6 +34,7 @@ import {
 import {
   createOffsiteLogger, errorField, AUDIT, PEER,
 } from '../utils/offsite-logging.js';
+import { deleteExpiredSnapshots } from '../common/offsite-retention.js';
 
 const AUDIT_TYPE = Audit.AUDIT_TYPES.CITED_ANALYSIS;
 // Human prefix for the two shared, untouched utils that still log via a passed-in prefix
@@ -222,6 +223,15 @@ export default async function handler(message, context) {
       count: suggestions.length,
     });
     logOffsiteLlmUsage(log, HUMAN_PREFIX, siteId, opportunityData.llmUsage);
+
+    // Retention must not fail an otherwise successful refresh.
+    try {
+      await deleteExpiredSnapshots({
+        dataAccess, siteId, auditType, log,
+      });
+    } catch (error) {
+      log.error(`[Offsite][Retention] Unexpected failure siteId=${siteId} auditType=${auditType} error=${error.message}`);
+    }
 
     if (auditId) {
       const auditRecord = await AuditModel.findById(auditId);
--- a/src/cited-analysis/guidance-handler.js
+++ b/src/cited-analysis/guidance-handler.js
@@ -230,7 +230,9 @@ export default async function handler(message, context) {
         dataAccess, siteId, auditType, log,
       });
     } catch (error) {
-      log.error(`[Offsite][Retention] Unexpected failure siteId=${siteId} auditType=${auditType} error=${error.message}`);
+      ologOpp.failure('retention_delete', `Unexpected retention failure for auditType ${auditType}`, {
+        peer: PEER.POSTGRES, direction: 'outbound', ...errorField(error),
+      }, error);
     }
 
     if (auditId) {
--- a/src/common/offsite-retention.js
+++ b/src/common/offsite-retention.js
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { subDays, differenceInDays } from 'date-fns';
+import { Opportunity as Oppty } from '@adobe/spacecat-shared-data-access';
+import { isOffsiteSnapshot } from './offsite-snapshot.js';
+
+export const SNAPSHOT_RETENTION_DAYS = 30;
+
+/**
+ * Finds managed offsite snapshots older than the retention cutoff, oldest first.
+ */
+export async function findExpiredSnapshots({
+  dataAccess, siteId, auditType, log,
+}) {
+  const { Opportunity } = dataAccess;
+
+  let ignoredOpportunities;
+
+  try {
+    ignoredOpportunities = await Opportunity
+      .allBySiteIdAndStatus(siteId, Oppty.STATUSES.IGNORED);
+  } catch (error) {
+    log.error(`[Offsite][Retention] Failed to find snapshots siteId=${siteId} auditType=${auditType} error=${error.message}`);
+    return [];
+  }
+
+  const retentionCutoff = subDays(new Date(), SNAPSHOT_RETENTION_DAYS);
+
+  return (ignoredOpportunities || [])
+    .filter((opportunity) => isOffsiteSnapshot(opportunity, auditType)
+      && new Date(opportunity.getCreatedAt()) < retentionCutoff)
+    .sort((firstOpportunity, secondOpportunity) => (
+      new Date(firstOpportunity.getCreatedAt()) - new Date(secondOpportunity.getCreatedAt())
+    ));
+}
+
+/**
+ * Deletes expired snapshots without interrupting the refresh that invoked retention.
+ */
+export async function deleteExpiredSnapshots({
+  dataAccess, siteId, auditType, log,
+}) {
+  const expiredSnapshots = await findExpiredSnapshots({
+    dataAccess, siteId, auditType, log,
+  });
+
+  const deletionResults = await Promise.all(expiredSnapshots.map(async (snapshot) => {
+    const snapshotAgeDays = differenceInDays(new Date(), new Date(snapshot.getCreatedAt()));
+    const triggerAuditId = snapshot.getData()?.snapshot?.triggerAuditId || 'unknown';
+
+    try {
+      // Dependent suggestions cascade-delete with the snapshot opportunity.
+      await snapshot.remove();
+
+      log.info(`[Offsite][Retention] Deleted snapshot opportunityId=${snapshot.getId()} `
+        + `siteId=${siteId} auditType=${auditType} triggerAuditId=${triggerAuditId} `
+        + `snapshotAgeDays=${snapshotAgeDays}`);
+
+      return true;
+    } catch (error) {
+      log.error(`[Offsite][Retention] Failed to delete snapshot opportunityId=${snapshot.getId()} `
+        + `siteId=${siteId} auditType=${auditType} triggerAuditId=${triggerAuditId} `
+        + `snapshotAgeDays=${snapshotAgeDays} error=${error.message}`);
+
+      return false;
+    }
+  }));
+
+  const deletedSnapshotCount = deletionResults.filter(Boolean).length;
+  const failedSnapshotCount = expiredSnapshots.length - deletedSnapshotCount;
+
+  log.info(`[Offsite][Retention] Snapshot deletion summary siteId=${siteId} `
+    + `auditType=${auditType} eligible=${expiredSnapshots.length} `
+    + `deleted=${deletedSnapshotCount} failed=${failedSnapshotCount}`);
+
+  return deletedSnapshotCount;
+}
--- a/src/common/offsite-retention.js
+++ b/src/common/offsite-retention.js
@@ -10,11 +10,27 @@
  * governing permissions and limitations under the License.
  */
 
-import { subDays, differenceInDays } from 'date-fns';
+import { subDays } from 'date-fns';
 import { Opportunity as Oppty } from '@adobe/spacecat-shared-data-access';
 import { isOffsiteSnapshot } from './offsite-snapshot.js';
+import {
+  createOffsiteLogger, errorField, AUDIT, PEER,
+} from '../utils/offsite-logging.js';
 
+// See docs/decisions/004-offsite-snapshot-retention-window.md for the rationale.
 export const SNAPSHOT_RETENTION_DAYS = 30;
+
+// Bounds the blast radius of a single invocation — e.g. a dormant site's first refresh after
+// months, when many snapshots become eligible at once. findExpiredSnapshots sorts oldest-first,
+// so any remainder beyond this cap drains across subsequent refreshes rather than all at once.
+export const MAX_DELETIONS_PER_RUN = 50;
+
+// Mirrors the mapping in offsite-snapshot.js — kept local to avoid a cross-module circular import.
+const AUDIT_SLUG_BY_TYPE = {
+  'cited-analysis': AUDIT.CITED,
+  'reddit-analysis': AUDIT.REDDIT,
+  'youtube-analysis': AUDIT.YOUTUBE,
+};
 
 /**
  * Finds managed offsite snapshots older than the retention cutoff, oldest first.
@@ -23,6 +39,7 @@ export async function findExpiredSnapshots({
   dataAccess, siteId, auditType, log,
 }) {
   const { Opportunity } = dataAccess;
+  const olog = createOffsiteLogger(log, { audit: AUDIT_SLUG_BY_TYPE[auditType] ?? 'unknown', siteId });
 
   let ignoredOpportunities;
 
@@ -30,7 +47,9 @@ export async function findExpiredSnapshots({
     ignoredOpportunities = await Opportunity
       .allBySiteIdAndStatus(siteId, Oppty.STATUSES.IGNORED);
   } catch (error) {
-    log.error(`[Offsite][Retention] Failed to find snapshots siteId=${siteId} auditType=${auditType} error=${error.message}`);
+    olog.failure('retention_lookup', `Failed to find snapshots for auditType ${auditType}`, {
+      peer: PEER.POSTGRES, direction: 'inbound', ...errorField(error),
+    });
     return [];
   }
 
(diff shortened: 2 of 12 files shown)


src/common/offsite-retention.js
```diff
@@ -0,0 +1,87 @@
...
+ * Deletes expired snapshots without interrupting the refresh that invoked retention.
+ */
+export async function deleteExpiredSnapshots({
+  dataAccess, siteId, auditType, log,
+}) {
+  const expiredSnapshots = await findExpiredSnapshots({
+    dataAccess, siteId, auditType, log,
+  });
+
+  const deletionResults = await Promise.all(expiredSnapshots.map(async (snapshot) => {
```
rarescheseli, line 57: issue (blocking): This fans out every expired snapshot's deletion concurrently: `await Promise.all(expiredSnapshots.map(async (snapshot) => { ... await snapshot.remove(); ...}))`. CLAUDE.md's "Database Query Patterns - Avoiding N+1" section documents the data-access layer as PostgreSQL-via-PostgREST with a 200-connection pool (10 ECS tasks x 20), lists this exact shape verbatim as BAD ("N concurrent removes - await Promise.all(suggestions.map((s) => s.remove()))"), and prescribes Collection.removeByIds(ids) (a single chunked DELETE .in(id, ids)) as the replacement.

It's worse than the surface count here: each snapshot Opportunity has suggestions copied onto it (via addSuggestions in the companion snapshot PR), and remove() cascades by removing every dependent individually, so the real concurrent operation count is roughly (expired snapshots) x (suggestions per snapshot + 1) - all against the shared pool, on every cited/reddit/youtube refresh. Steady-state batches are small, but the scenario this feature exists for - a site whose audits were disabled for months then resumed, or any site catching up on an accumulated backlog - is exactly when this fires at scale, and the PR's own "dormant sites are cleaned on their next refresh" design assumes that catch-up case will happen.

Fixing this by simply swapping to Opportunity.removeByIds(ids) would introduce a new bug: removeByIds is a raw bulk DELETE that does not honor cascade/removeDependents, so it would orphan the copied suggestions rather than cleaning them up. This repo already has the correct pattern for this exact situation - removeOpportunityWithSuggestions in src/experimentation-opportunities/guidance-high-organic-low-ctr-handler.js bulk-deletes the dependent suggestions first via Suggestion.removeByIds(...), then removes the opportunity. Mirror that pattern here, batched across the expired set. If per-record failure isolation is a hard requirement a batched approach can't preserve, bound the concurrency instead (chunked batches or a small worker pool) rather than firing everything at once, and document the deliberate deviation from the CLAUDE.md pattern in a code comment so it isn't re-flagged as an oversight later.

src/common/offsite-retention.js
```diff
@@ -0,0 +1,87 @@
+/*
```
rarescheseli, line 1: issue (blocking): This new module introduces an automatic, irreversible deletion policy - permanently removing Opportunity records (and, via cascade, their dependent Suggestions) on a schedule tied to refresh activity. CLAUDE.md states: "Significant new functionality must be developed spec-first. Add a spec to /docs/specs/ following the template...". No spec exists under /docs/specs/ describing this policy's scope, cutoff semantics, or cascade behavior; permanent data deletion is exactly the class of change this rule exists for. Add a spec under /docs/specs/, or state explicitly in the PR why one isn't warranted for a change of this kind.

src/common/offsite-retention.js
```diff
@@ -0,0 +1,87 @@
...
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+import { subDays, differenceInDays } from 'date-fns';
+import { Opportunity as Oppty } from '@adobe/spacecat-shared-data-access';
+import { isOffsiteSnapshot } from './offsite-snapshot.js';
+
+export const SNAPSHOT_RETENTION_DAYS = 30;
```
rarescheseli, line 17: issue (blocking): SNAPSHOT_RETENTION_DAYS = 30 establishes a permanent-deletion policy with no explanation of why 30 days, and no consideration on record of what happens to an in-flight restore request against a snapshot this window has already deleted. CLAUDE.md states: "Technical or architectural decisions with long-term impact (non-obvious behavior choices, trade-offs) must be recorded as ADRs in /docs/decisions/... Use status values: Proposed / Accepted / Deprecated / Superseded." A hardcoded retention window that permanently destroys data a related, not-yet-built feature will need to read is exactly this kind of non-obvious, long-term trade-off. Add an ADR under /docs/decisions/ (numbered after the existing 001) with status Accepted.

src/common/offsite-retention.js
```diff
@@ -0,0 +1,87 @@
...
+  return (ignoredOpportunities || [])
+    .filter((opportunity) => isOffsiteSnapshot(opportunity, auditType)
+      && new Date(opportunity.getCreatedAt()) < retentionCutoff)
+    .sort((firstOpportunity, secondOpportunity) => (
+      new Date(firstOpportunity.getCreatedAt()) - new Date(secondOpportunity.getCreatedAt())
+    ));
+}
+
+/**
+ * Deletes expired snapshots without interrupting the refresh that invoked retention.
```
rarescheseli, line 48: issue (blocking): Retention only runs as a side effect of a successful guidance-handler refresh, so it never fires for a site whose audit has been disabled or that has been offboarded entirely - as opposed to a merely dormant site, which self-heals on its next refresh. The PR's own description says "dormant sites are cleaned on their next refresh," which reads as if this covers the inactive case in general, but a permanently inactive site's snapshots (and their cascade-linked suggestions) accumulate forever with nothing to catch them, since scheduled or infrastructure-level sweepers are explicitly out of scope for this PR. Since bounding storage growth is this PR's whole purpose, this leaves a real class of sites unbounded. Correct the disclosure to distinguish dormant from disabled/offboarded, and track a follow-up for the backstop (a scheduled sweep) the same way the deletion policy itself was tracked as LLMO-6154 rather than left implicit.

rarescheseli (changes requested):
Hey @grigoreadobe,

**Verdict: Request changes** - the deletion mechanism itself needs to change before merge (a documented anti-pattern that compounds badly in exactly the backlog scenario this feature targets), plus process and disclosure-accuracy gaps.
_Changes: Adds a retention module (`src/common/offsite-retention.js`) that permanently deletes offsite-snapshot Opportunity records older than 30 days after every successful cited/reddit/youtube refresh, closing the LLMO-6154 follow-up from the companion snapshot-history PR (10 files)._

## Must fix before merge
1. [Critical] Concurrent per-record snapshot deletion violates the repo's documented N+1 pattern and multiplies via cascade fan-out - `src/common/offsite-retention.js:57` (details inline)
2. [Important] New automatic, irreversible deletion policy ships without a required spec - `src/common/offsite-retention.js:1` (details inline)
3. [Important] Hardcoded 30-day retention window is an undocumented architectural trade-off with no ADR - `src/common/offsite-retention.js:17` (details inline)
4. [Important] The "cleaned on next refresh" disclosure conflates dormant sites with permanently disabled ones, leaving an unbounded growth path - `src/common/offsite-retention.js:48` (details inline)

<details>
<summary>Non-blocking (9): minor issues and suggestions</summary>

- nit: the oldest-first sort in `findExpiredSnapshots` has no consumer under the current unbounded `Promise.all` - either use it to cap deletions per run or drop it - `src/common/offsite-retention.js:42`
- nit: `snapshotAgeDays` and `triggerAuditId` are computed outside the per-item try block, so a malformed `getCreatedAt()`/`getData()` on one snapshot would reject the whole `Promise.all` rather than being isolated like other per-item failures - `src/common/offsite-retention.js:58`
- nit: a snapshot whose `getCreatedAt()` returns null/undefined is silently retained forever (`new Date(undefined) < cutoff` is false) - unlikely given the datastore always populates createdAt, but untested - `src/common/offsite-retention.js:41`
- nit: this PR escalates the consequence of the pre-existing, unauthenticated `siteId` trust boundary (unchanged here) from create/read to permanent cascade delete - not a fix for this PR, but worth a line in the risk notes - `src/cited-analysis/guidance-handler.js:40`
- nit: at-least-once SQS delivery plus concurrent invocations could have two handler runs select the same expired snapshot; the second `remove()` fails and is caught/logged, and a mid-cascade failure could leave a parent deleted with some dependents already gone - not fatal, worth an idempotency note - `src/common/offsite-retention.js:63`
- nit: no test asserts that retention is skipped when the refresh itself fails before reaching the success path - the call-site placement makes this look intentional, just unasserted - `test/audits/cited-analysis/guidance-handler.test.js`
- suggestion: add one line reserving how retention and the not-yet-built snapshot restore feature will coordinate, since a snapshot can be hard-deleted out from under a user about to restore it - `src/common/offsite-retention.js:48`
- suggestion: the cascade-delete assumption ("dependent suggestions cascade-delete through the existing data model") is entirely unverified at this layer since `remove()` is mocked as an always-resolving stub - confirm it's covered by an integration-level test, or link to where it's verified - `src/common/offsite-retention.js:63`
- suggestion: confirm log-based alerting watches for repeated "[Offsite][Retention] Failed to find snapshots" errors - on lookup failure the deletion summary logs `eligible=0 deleted=0 failed=0`, indistinguishable from "nothing was actually expired," so a site whose lookups fail on every refresh would silently never clean up - `src/common/offsite-retention.js:33`
</details>

- `findExpiredSnapshots` repeats the same full-site IGNORED-scan-then-filter-in-memory pattern already flagged on the companion snapshot-history PR - fixing the lookup strategy touches that PR's functions and is tracked there, not here.

src/cited-analysis/guidance-handler.js
```diff
@@ -223,6 +224,15 @@ export default async function handler(message, context) {
     });
     logOffsiteLlmUsage(log, HUMAN_PREFIX, siteId, opportunityData.llmUsage);
 
+    // Retention must not fail an otherwise successful refresh.
+    try {
+      await deleteExpiredSnapshots({
+        dataAccess, siteId, auditType, log,
+      });
+    } catch (error) {
+      log.error(`[Offsite][Retention] Unexpected failure siteId=${siteId} auditType=${auditType} error=${error.message}`);
```
rarescheseli: let's use the new logger intead.

src/common/offsite-retention.js
```diff
@@ -0,0 +1,87 @@
...
+}) {
+  const { Opportunity } = dataAccess;
+
+  let ignoredOpportunities;
+
+  try {
+    ignoredOpportunities = await Opportunity
+      .allBySiteIdAndStatus(siteId, Oppty.STATUSES.IGNORED);
+  } catch (error) {
+    log.error(`[Offsite][Retention] Failed to find snapshots siteId=${siteId} auditType=${auditType} error=${error.message}`);
```
rarescheseli: we could use the new offsite logger in this file.

src/common/offsite-retention.js
MysticatBot: issue (blocking): `await Promise.all(expiredSnapshots.map(async (snapshot) => { await snapshot.remove(); }))` is the exact pattern the repo CLAUDE.md lists under "Never do this":

> BAD: N concurrent removes - await Promise.all(suggestions.map((s) => s.remove()));

The prescribed fix is `Collection.removeByIds(ids)` (a single chunked DELETE). Here the impact is compounded: each `remove()` cascades by individually deleting dependent Suggestions, so the real concurrent-request count is roughly `N * (1 + suggestions_per_snapshot)` against a shared 200-connection pool.

The catch-up scenario this feature exists to address (a dormant site resuming refreshes with months of accumulated snapshots) is exactly when N is large - potentially tens of snapshots per audit type, times three handlers running concurrently.

**Fix:** Collect all expired snapshot IDs, bulk-remove their Suggestions first via `Suggestion.removeByIds(ids)`, then bulk-remove the Opportunities via `Opportunity.removeByIds(snapshotIds)`. This mirrors the existing `removeOpportunityWithSuggestions` pattern in `src/experimentation-opportunities/guidance-high-organic-low-ctr-handler.js`. If per-record failure isolation is a hard requirement, serialize with a for-of loop or chunk into small batches rather than firing all at once.

src/common/offsite-retention.js
MysticatBot, line 1: issue (blocking): This module introduces an automatic, irreversible deletion policy - permanently removing Opportunity records (and, via cascade, their dependent Suggestions) triggered on every successful refresh. CLAUDE.md states:

> **Significant new functionality** must be developed spec-first. Add a spec to `/docs/specs/` following the template...

No spec exists under /docs/specs/ for this retention policy. Permanent data deletion triggered automatically on a recurring basis qualifies as significant new functionality.

**Fix:** Add a spec at `docs/specs/` covering: retention window rationale, cutoff semantics (strict < vs <=), cascade scope, which record types are affected, interaction with the planned restore feature (LLMO-6153), and the disabled-site gap.

src/common/offsite-retention.js
MysticatBot, line 21: issue (blocking): `SNAPSHOT_RETENTION_DAYS = 30` establishes a permanent-deletion cliff with no recorded justification. CLAUDE.md states:

> **Technical or architectural decisions with long-term impact** (non-obvious behavior choices, trade-offs) must be recorded as ADRs in `/docs/decisions/` numbered sequentially (001, 002, ...).

Why 30 and not 14 or 90? What constraints drove this number? How does it interact with the not-yet-built restore feature (LLMO-6153) that may need to read a snapshot this window has already deleted?

**Fix:** Add an ADR to `/docs/decisions/` (next sequential number) with status Accepted. Record the rationale for 30 days, the relationship to restore, and whether per-customer overrides are anticipated.

MysticatBot (changes requested):
Hey @grigoreadobe,

> ⚠ Degraded review - no spec document was found for this change (searched the PR links, the touched repos docs, the architecture/guidelines docs, and linked Jira). This review covers code-level quality but could not validate the change against an agreed design, so confidence is reduced. Add a spec link (PR template section 4) and re-request review for a full-confidence pass.

**Verdict: Request changes** - the N+1 deletion pattern remains unresolved from the prior review and violates an explicit repo rule, and the required spec/ADR documentation is still absent.
**Complexity: HIGH** - large diff (700+ lines, 10 files); introduces an automatic irreversible deletion policy with database concurrency implications.
_Changes: Adds a retention module that permanently hard-deletes managed offsite-snapshot Opportunity records older than 30 days (and their cascade-deleted Suggestions) after each successful cited/reddit/youtube refresh (10 files)._

Note: CI checks are currently passing but branch was force-pushed since the prior review - performing full re-review.

## Must fix before merge
1. [Critical] Concurrent per-record deletion via Promise.all violates the explicit CLAUDE.md N+1 ban and risks connection pool exhaustion under backlog catch-up - `src/common/offsite-retention.js:57` (details inline)
2. [Important] New automatic irreversible deletion policy ships without the required spec document - `src/common/offsite-retention.js:1` (details inline)
3. [Important] Hardcoded 30-day retention window has no ADR recording the architectural rationale - `src/common/offsite-retention.js:17` (details inline)

<details>
<summary>Non-blocking (6): minor issues and suggestions</summary>

- nit: time-dependent tests in offsite-retention.test.js use real clock via daysAgo() helper but assert on exact snapshotAgeDays values in log output - will flake if test straddles midnight on CI; use sandbox.useFakeTimers in the deleteExpiredSnapshots describe block - `test/common/offsite-retention.test.js:580`
- nit: summary log fires at INFO for every handler invocation even when zero snapshots are eligible, adding noise across 3 audit types x N sites x every weekly run - gate on expiredSnapshots.length > 0 or reduce to debug - `src/common/offsite-retention.js:82`
- nit: file uses raw log.error/log.info string formatting while the rest of the offsite module has migrated to createOffsiteLogger with structured fields - `src/common/offsite-retention.js:33`
- suggestion: cap deletions per invocation (e.g. 50) to bound blast radius on first deployment when months of accumulated snapshots become eligible at once - the sorted-oldest-first ordering already supports draining across multiple refreshes - `src/common/offsite-retention.js:57`
- suggestion: clarify in the PR description that "dormant sites are cleaned on their next refresh" applies only to dormant sites that will eventually refresh, not to permanently disabled or offboarded sites whose snapshots accumulate indefinitely - the sweep is out of scope but the phrasing could be read as covering all inactive sites
- nit: the SNAPSHOT_RETENTION_DAYS constant test (`expect(SNAPSHOT_RETENTION_DAYS).to.equal(30)`) is tautological - if the constant changes the test is updated in lockstep; the boundary behavior tests already cover the meaningful semantics - `test/common/offsite-retention.test.js:599`
</details>


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 15s | <b>Cost:</b> $6.21 | <b>Commit:</b> <code>79f62171733c575cb39739259beec0ce01ec1bc1</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @grigoreadobe,

**Verdict: Approve** - all prior blocking findings are resolved; the bulk-deletion pattern, spec, and ADR are in place and correct.
**Complexity: HIGH** - large diff (1248 lines, 12 files); automatic irreversible deletion policy with bulk DB operations.
_Changes: Adds a retention module that hard-deletes managed offsite-snapshot Opportunity records (and their cascade-linked Suggestions) older than 30 days after each successful refresh, using bulk removeByIds to avoid N+1, capped at 50 per run (12 files)._

Note: Recommend a human read before merge - this change adds a new ADR and spec document under docs/decisions/ and docs/specs/ (architectural significance). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `AUDIT_SLUG_BY_TYPE` is duplicated between `offsite-snapshot.js` and `offsite-retention.js` (comment explains why: circular import avoidance) - consider extracting to a shared constants file if a third consumer appears - `src/common/offsite-retention.js:27`
- nit: if `getSuggestions()` throws for one snapshot in the sequential loop, the entire batch is abandoned and retried next refresh with the same failing record at the front (oldest-first); a per-record try/catch around the read would let the batch skip a problematic record and make progress on the rest - `src/common/offsite-retention.js:75`
</details>

- N+1 concurrent deletion replaced with bulk `Suggestion.removeByIds` + `Opportunity.removeByIds`
- Spec added at `docs/specs/2026-07-20-offsite-snapshot-retention.md`
- ADR added at `docs/decisions/004-offsite-snapshot-retention-window.md`
- Structured logger (`createOffsiteLogger`) adopted throughout the retention module and handler catch blocks
- Deletion cap (`MAX_DELETIONS_PER_RUN = 50`) added to bound blast radius on backlog catch-up
- Dormant vs disabled/offboarded distinction clarified in spec's "Known limitation" section

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 6m 24s | <b>Cost:</b> $1.69 | <b>Commit:</b> <code>287c8626204af056452f05f9bf13e3bfacad60fb</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rarescheseli (commented):
Hey @grigoreadobe,

**Verdict: Comment** - no blocking findings; a few minor test/observability gaps worth a look.
**Complexity: HIGH** - large diff (1042 additions, 12 files); adds an irreversible hard-delete policy with DB bulk-delete concurrency implications.
_Changes: Adds a retention module that permanently hard-deletes managed offsite-snapshot Opportunity records (and their cascade-linked Suggestions) older than 30 days via bulk `removeByIds`, run as a side effect after each successful cited/reddit/youtube refresh and capped at 50 deletions per run (12 files)._

Note: Recommend a human read before merge - this change adds a new ADR and spec document under docs/decisions/ and docs/specs/ (architectural significance). The bot review is a complement to, not a replacement for, a human read here.

<details>
<summary>Non-blocking (11): minor issues and suggestions</summary>

- nit: the per-snapshot `getSuggestions()` read in the deletion loop has no per-record failure isolation - a persistently failing record would block deletion for that `(siteId, auditType)` on every refresh, and the rejection path is untested; low likelihood and refresh-safe, but worth a follow-up test plus a per-record guard - `src/common/offsite-retention.js:99`
- nit: `retention_lookup` failure log omits the 4th `error` arg other call sites in this module pass, dropping stack capture in CloudWatch for this one path - `src/common/offsite-retention.js:50`
- nit: `expect(SNAPSHOT_RETENTION_DAYS).to.equal(30)` is a tautological test; the boundary tests already cover the real behavior - `test/common/offsite-retention.test.js:67`
- suggestion: add a comment noting the partial bulk-delete failure window is self-healing (an orphaned snapshot is re-found and retried next refresh) - `src/common/offsite-retention.js:103-108`
- suggestion: consider a one-cycle log-only/dry-run gate for this worker's first irreversible hard-delete, given a regression in `isOffsiteSnapshot` would have no recovery path - `src/common/offsite-retention.js:80`
- nit: the spec's dormant-vs-disabled/offboarded framing is stated at whole-site granularity, but retention is scoped per `(siteId, auditType)` - an active site that disables just one audit type would accumulate that type's snapshots forever - `docs/specs/2026-07-20-offsite-snapshot-retention.md:96-104`
- nit: `AUDIT_SLUG_BY_TYPE` is duplicated between `offsite-retention.js` and `offsite-snapshot.js` (comment explains circular-import avoidance; there's no bulk-read SDK primitive to consolidate around either) - `src/common/offsite-retention.js:29`
- nit: the retention try/catch block is copy-pasted verbatim across all three guidance handlers - mentioning only, this is three-similar-lines territory, not a real duplication problem - `src/cited-analysis/guidance-handler.js:231`
- suggestion: the boundary tests build `createdAt` via raw millisecond arithmetic while production computes the cutoff with `subDays` (date-fns); safe under the current fixture's date/timezone but could silently drift if the fixture date changes - `test/common/offsite-retention.test.js:178`
- suggestion: the bulk-delete tests only ever use one suggestion per snapshot; a multi-suggestion case would confirm the ID-flattening and ordering into the single bulk call - `test/common/offsite-retention.test.js:219`
- suggestion: the `Opportunity.removeByIds`-failure test doesn't assert `Suggestion.removeByIds` ran first, which would document the intentional non-atomic delete ordering - `test/common/offsite-retention.test.js:352`

</details>

- Pre-existing unauthenticated `siteId` trust boundary now carries cascade-delete blast radius (unchanged by this PR; bounded by a hardcoded `auditType`, the `offsite-snapshot` tag, and the handler's existing `Site.findById` check).
- Pre-existing ADR numbering collision (two existing docs both numbered "002") predates this PR.
- `findExpiredSnapshots`'s full-site-scan-then-filter-in-memory lookup repeats a pattern already flagged on the companion snapshot-history PR - tracked there, not here.

- N+1 concurrent deletion replaced with bulk `Suggestion.removeByIds` + `Opportunity.removeByIds`
- Spec added at `docs/specs/2026-07-20-offsite-snapshot-retention.md`
- ADR added at `docs/decisions/004-offsite-snapshot-retention-window.md`
- Dormant vs disabled/offboarded distinction clarified in the spec's "Known limitation" section

rarescheseli (approved):

PR #2782, opened 2026-07-20
fix(offsite): reuse active brand opportunity instead of duplicating it (23505)

AndreiAlexandruParaschiv:
Fixes the duplicate-key failure that left youtube/cited/reddit analysis opportunities with **zero suggestions**.

## Analysis — 23505 on the unique active-scope constraint

Splunk showed:

```
code: 23505
duplicate key value violates unique constraint "idx_opportunities_unique_active_scope"
Key (scope_type, scope_id, type)=(brand, 019d6b16-42da-7999-8161-0492a07dca15, cited-analysis) already exists.
```

`idx_opportunities_unique_active_scope` is a partial unique index on `(scope_type, scope_id, type)` for `status IN ('NEW','IN_PROGRESS')` — i.e. a brand may have only **one** active opportunity of a given analysis type.

The guidance handlers asked `convertToOpportunity` to find the existing opportunity with a comparison fn that matched by **`auditId`**. `auditId` is regenerated on every analysis run, so a brand-scoped re-run never matched the row created by the previous run:

1. `convertToOpportunity` finds nothing → creates a **second** opportunity row.
2. On save, the handler stamps the resolved **brand scope** onto it.
3. That collides with the still-active row from the previous run → **23505**.
4. The handler throws **before** `syncSuggestions`, so no suggestions persist.

`convertToOpportunity`'s own fetch also only looked at `NEW`, so an opportunity a user had moved to `IN_PROGRESS` was invisible and re-collided on the next run.

## Fix

When a brand is resolved:

- Resolve the existing **active** brand opportunity by scope via `Opportunity.allByScope('brand', brandId)`, matching `NEW` **and** `IN_PROGRESS`, and pass it into `convertToOpportunity` so it is updated **in place**.
- Match by **brand scope** `(scopeType, scopeId)` instead of `auditId`.
- Fall back to the previous `auditId` match for the site-level (no-brand) path.

`convertToOpportunity` gains an optional `existingOpportunity` param; when provided it skips the `NEW`-only fetch.

## Changes

- `src/{youtube,reddit,cited}-analysis/guidance-handler.js`: scope-aware lookup + reuse of the active brand opportunity.
- `src/common/opportunity.js`: accept a pre-resolved `existingOpportunity`.
- Test suites: brand-scope comparison-fn cases + reuse of an `IN_PROGRESS` brand opportunity.

## Test plan

- `mocha` youtube/cited/reddit guidance-handler suites — 114 passing
- lint clean
- Verify in env: a localized brand-scoped RACQ cited/reddit/youtube re-run updates the existing opportunity and persists suggestions (no 23505).

## Follow-ups (separate PRs)

- **#2783** (stacked cleanup): drop the redundant `fullAnalysis` blob + refresh the dashboard when an opportunity is reused.
- Mystique's silent S3 delivery failure (`save_result_to_s3` swallows exceptions → null `presignedUrl`) can independently cause "no opportunity" for reddit/cited; tracked separately in the `mystique` repo.

### DIFF
--- a/src/cited-analysis/guidance-handler.js
+++ b/src/cited-analysis/guidance-handler.js
@@ -33,7 +33,7 @@ const LOG_PREFIX = '[Cited]';
  */
 export default async function handler(message, context) {
   const { log, dataAccess } = context;
-  const { Site, Audit: AuditModel } = dataAccess;
+  const { Site, Audit: AuditModel, Opportunity } = dataAccess;
   // Note: any inbound `brandId` from Mystique is informational only. Scope is
   // re-resolved server-side via resolveBrandResultForSite; trusting the inbound
   // value would let a tampered message re-attribute the opportunity.
@@ -97,6 +97,20 @@ export default async function handler(message, context) {
 
     const auditType = opportunityData.type || AUDIT_TYPE;
 
+    // Brand-scoped analyses are unique per (scopeType, scopeId, type) in Postgres
+    // (idx_opportunities_unique_active_scope, covering NEW + IN_PROGRESS). auditId
+    // changes on every run, so match/reuse the existing active brand opportunity by
+    // scope and update it in place; creating a second row and stamping the same
+    // brand scope collides (23505 duplicate key) and aborts before syncSuggestions.
+    const resolvedBrandId = brandResult?.brand?.brandId;
+    let existingOpportunity;
+    if (resolvedBrandId && typeof Opportunity?.allByScope === 'function') {
+      const scoped = await Opportunity.allByScope('brand', resolvedBrandId);
+      existingOpportunity = scoped.find(
+        (oppty) => oppty.getType() === auditType
+          && ['NEW', 'IN_PROGRESS'].includes(oppty.getStatus()),
+      );
+    }
     const opportunity = await convertToOpportunity(
       baseUrl,
       {
@@ -108,7 +122,10 @@ export default async function handler(message, context) {
       createOpportunityData,
       auditType,
       { opportunityData },
-      (oppty) => oppty.getAuditId() === auditId,
+      (oppty) => (resolvedBrandId
+        ? oppty.getScopeType?.() === 'brand' && oppty.getScopeId?.() === resolvedBrandId
+        : oppty.getAuditId() === auditId),
+      existingOpportunity,
     );
 
     // Persist the opportunity (with scope) BEFORE syncing suggestions. Reversed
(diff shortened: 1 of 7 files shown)


AndreiAlexandruParaschiv:
closing in favor of #2765

MysticatBot (approved):

**Verdict: Approve** - correct fix for the 23505 duplicate-key bug with solid test coverage.
**Complexity: MEDIUM** - multi-handler bug fix touching a shared utility.
_Changes: Pre-resolves the existing active brand-scoped opportunity by scope and passes it into `convertToOpportunity` to avoid duplicate-key violations on re-runs (7 files)._

<details>

- suggestion: Extract the 14-line brand-opportunity resolution block into a shared helper (e.g. `resolveActiveBrandOpportunity` in `src/common/`) - the same logic is copy-pasted across cited, reddit, and youtube handlers. Triplication was the root cause of the original bug (matching logic drifted); consolidating reduces future drift risk. PR description notes follow-up in stacked cleanup PR. - `src/cited-analysis/guidance-handler.js:103`
- nit: The defensive guard `typeof Opportunity?.allByScope === 'function'` silently falls through to the old code path if `allByScope` is missing (e.g. data-access version mismatch). Consider logging a warning when the guard evaluates to false so operators know the fix is inoperative. - `src/cited-analysis/guidance-handler.js:108`
- suggestion: `convertToOpportunity` now has 8 positional parameters. The trailing optional params (`props`, `comparisonFn`, `existingOpportunity`) would be clearer as a single options object. Worth considering when the follow-up PR touches this function. - `src/common/opportunity.js:33`
- nit: The brand-matching `comparisonFn` is unreachable when `existingOpportunity` is provided (the fetch is skipped entirely). It only fires in the non-brand fallback. Not a bug, but worth a one-line comment clarifying the two paths are independent, not belt-and-suspenders. - `src/cited-analysis/guidance-handler.js:122`
- suggestion: Add an explicit test for the "brand resolved but `allByScope` returns empty array" scenario to confirm a new opportunity is created without error. Codecov shows line coverage, but this critical first-run path deserves a named test case. - `test/audits/cited-analysis/guidance-handler.test.js:198`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 34s | <b>Cost:</b> $3.95 | <b>Commit:</b> <code>7e139a71fbebb6ae24c795ba2067eb4c7f8935c3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2783, opened 2026-07-20
chore(offsite): drop redundant fullAnalysis blob from analysis opportunities

AndreiAlexandruParaschiv:
Rebased onto `main` now that the evergreen offsite work (#2765, LLMO-6122) has merged. Reduced to the one change that is still needed.

## What changed vs. the original #2783

- **Kept:** dropping the redundant `fullAnalysis` blob.
- **Dropped:** the "refresh dashboard on reuse" change — it is now redundant. #2765's `persistOffsiteOpportunity` already replaces the opportunity data in place on reuse (`evergreenOpportunity.setData({ ...mappedOpportunity.data })`), so the dashboard stays fresh without any extra code.

## Drop the redundant `fullAnalysis` blob

After `persistOffsiteOpportunity` (LLMO-6122) already stores the dashboard data the UI renders, the youtube/reddit/cited guidance handlers still wrote the entire raw Mystique payload back onto the opportunity as `fullAnalysis` via a second `setData()` + `save()`. That blob duplicates:

- `data.dashboard` — what the UI actually renders, and
- the suggestions — which are already stored as their own records.

`fullAnalysis` is not read by the UI for these analysis types (only the separate Wikipedia handler still uses it, left untouched). After localization enlarged the payloads, this duplicate also risked bloating the stored item. The write is removed from all three handlers.

## Changes

- `src/{youtube,reddit,cited}-analysis/guidance-handler.js`: remove the `fullAnalysis` write (the surrounding `setStatus`/scope/`save` flow is unchanged).
- Test suites: "stores fullAnalysis" assertions replaced with "item stays lean (no fullAnalysis)".

## Test plan

- `mocha` youtube/cited/reddit guidance-handler suites — 134 passing
- lint clean

## Notes

- `wikipedia-analysis` still writes `fullAnalysis` (it is consumed there) — intentionally out of scope.
- Supersedes the earlier duplicate-key fix (#2782, closed) which was replaced by #2765.


MysticatBot (approved):

**Verdict: Approve** - clean removal of redundant data persistence, well-scoped and well-explained.
**Complexity: LOW** - small cleanup, single service.
_Changes: Removes redundant `fullAnalysis` blob writes from the youtube/reddit/cited guidance handlers, since `persistOffsiteOpportunity` already persists the dashboard data the UI renders (6 files)._

<details>

- nit: The negative assertion `expect(mockOpportunity.setData).to.not.have.been.calledWith(sinon.match.has('fullAnalysis'))` passes vacuously when `setData` is never called at all - it documents intent but does not assert the stronger contract that `setData` is not called in this path. Acceptable as-is, just weaker than the old positive assertion. - `test/audits/cited-analysis/guidance-handler.test.js:900`
- suggestion: The 4-line explanatory comment replacing the removed code is longer than the code it replaced. A single line like `// fullAnalysis intentionally omitted; dashboard + suggestions already persisted upstream.` would convey the same intent with less visual weight. - `src/cited-analysis/guidance-handler.js:139`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 9m 14s | <b>Cost:</b> $3.54 | <b>Commit:</b> <code>b655cc08ab0b8e4335ada92f14bda24f2f4274c1</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2784, opened 2026-07-20
feat(evergreen-offsite-opportunities): reuse active IN_PROGRESS opportunity on refresh (LLMO-6122)

grigoreadobe:
## Scope

https://jira.corp.adobe.com/browse/LLMO-6122

Follow-up to the merged evergreen offsite work: reuse a customer-activated (`IN_PROGRESS`) offsite opportunity on refresh instead of only reusing `NEW`.

This folds in the fix from the closed `#2782` so cited, Reddit, and YouTube refreshes converge on the single active opportunity regardless of its status.

## Problem

An offsite opportunity is unique per `(scope_type, scope_id, type)` while active, where active means `NEW` **or** `IN_PROGRESS` (`idx_opportunities_unique_active_scope`). The resolver only queried `NEW`, so once a customer opened an opportunity in the Workspace (moving it to `IN_PROGRESS`), the next weekly refresh:

1. did not see the active row and created a second opportunity;
2. stamped the same brand scope on it and saved;
3. collided on the unique index (`23505 duplicate key`);
4. threw before `syncSuggestions`, leaving the opportunity with zero suggestions.

## Implementation

- `resolveEvergreenOffsiteOpportunity` now resolves the evergreen opportunity across both active statuses (`NEW` and `IN_PROGRESS`) via one query per status, then applies the existing keep-most-recent / retire-duplicates convergence.
- Brand-scoped opportunities carry their `siteId`, so the existing site-based lookup resolves the same rows the unique index protects; no new query surface is needed.
- Production change is confined to `src/common/offsite-refresh.js`.

## Testing

- Focused, lint, and full unit/integration suite pass with 100% coverage.
- Resolver and all three guidance-handler suites cover reusing an `IN_PROGRESS` opportunity in place (not retired) and convergence when duplicates span both active statuses.
- Local Postgres/PostgREST E2E: a brand-scoped opportunity activated to `IN_PROGRESS` is reused in place on a surfaced refresh (same opportunity ID, suggestions persisted, no `23505`, exactly one active row); prior evergreen scenarios (first run, repeat refresh, suppressed run isolation, duplicate retirement) still pass.

## Known behavior / follow-up

On a surfaced refresh the reused opportunity's status is set from the incoming run, so an `IN_PROGRESS` opportunity returns to `NEW`. This matches the prior handler behavior and the closed `#2782`; preserving customer activation across refreshes would be a separate change.

## Out of Scope

- Preserving `IN_PROGRESS` across refreshes (status-downgrade avoidance)
- Snapshot creation, restore, and retention
- Suggestion retention or deterministic suggestion deduplication
- Wikipedia analysis

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/common/offsite-refresh.js
+++ b/src/common/offsite-refresh.js
@@ -14,6 +14,12 @@ import { Opportunity as Oppty } from '@adobe/spacecat-shared-data-access';
 import { DATA_SOURCES, OFFSITE_AUDIT_TYPES } from './constants.js';
 import { checkGoogleConnection } from './opportunity-utils.js';
 
+// An offsite opportunity is unique per (scopeType, scopeId, type) while active, where active
+// means NEW or IN_PROGRESS (idx_opportunities_unique_active_scope). Both statuses must be
+// treated as the reusable evergreen; a customer-activated (IN_PROGRESS) opportunity that is
+// not reused would be recreated and collide with the existing row on save.
+const ACTIVE_OFFSITE_STATUSES = [Oppty.STATUSES.NEW, Oppty.STATUSES.IN_PROGRESS];
+
 /**
  * Validates the payload shape and rejects a declared opportunity type that differs from
  * the handler-owned audit type.
@@ -110,9 +116,12 @@ export async function persistOffsiteOpportunity(
 }
 
 /**
- * Resolves the evergreen offsite opportunity (status NEW).
+ * Resolves the evergreen offsite opportunity (active status: NEW or IN_PROGRESS).
  *
- * If duplicates exist, keeps the most recently updated one and retires the rest to IGNORED.
+ * A customer-activated (IN_PROGRESS) opportunity is reused in place, not treated as absent —
+ * otherwise the refresh would create a second row and collide with it on the active-scope
+ * unique index. If duplicates exist, keeps the most recently updated one and retires the
+ * rest to IGNORED.
  *
  * @param {Object} params
  * @param {Object} params.dataAccess - The data access object (from audit-worker context).
@@ -126,15 +135,18 @@ export async function resolveEvergreenOffsiteOpportunity({
   dataAccess, siteId, auditType, log,
 }) {
   const { Opportunity } = dataAccess;
-  let opportunities;
+  let activeOpportunities;
   try {
-    opportunities = await Opportunity.allBySiteIdAndStatus(siteId, Oppty.STATUSES.NEW);
+    const opportunitiesByStatus = await Promise.all(
+      ACTIVE_OFFSITE_STATUSES.map((status) => Opportunity.allBySiteIdAndStatus(siteId, status)),
+    );
+    activeOpportunities = opportunitiesByStatus.flatMap((list) => list || []);
   } catch (e) {
     log.error(`[OffsiteRefresh] Failed to fetch opportunities for siteId ${siteId}: ${e.message}`);
     throw e;
   }
 
-  const matchingOpportunities = (opportunities || [])
+  const matchingOpportunities = activeOpportunities
     .filter((opportunity) => opportunity.getType() === auditType);
   if (matchingOpportunities.length === 0) {
     return null;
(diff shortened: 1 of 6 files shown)


src/common/offsite-refresh.js
```diff
@@ -147,7 +159,7 @@ export async function resolveEvergreenOffsiteOpportunity({
     (a, b) => new Date(b.getUpdatedAt()) - new Date(a.getUpdatedAt()),
```
rarescheseli, line 159: **issue (blocking):** The survivor is chosen by `updatedAt` recency alone, with no status preference:

```js
const [evergreenOpportunity, ...duplicates] = [...matchingOpportunities].sort(
  (a, b) => new Date(b.getUpdatedAt()) - new Date(a.getUpdatedAt()),
);
```

This PR is what first puts a NEW and an IN_PROGRESS row into the same pool, so this sort now has to choose across statuses - and it can choose wrong. If a NEW row's `updatedAt` is more recent than a customer-activated IN_PROGRESS row (a stray/legacy NEW, or an anomalous multi-row state like the one PR 2782 describes), the IN_PROGRESS opportunity is retired to IGNORED. That silently discards the customer's activated opportunity and breaks any ID-based reference to it - the exact harm this PR exists to prevent.

On an exact `updatedAt` tie the outcome is decided by list order (`ACTIVE_OFFSITE_STATUSES` lists NEW first, and `Array.prototype.sort` is stable), so NEW wins deterministically - the "keep the customer's row" behavior is emergent, not deliberate.

This is distinct from the acknowledged out-of-scope status-downgrade: that item is about the survivor's status field returning to NEW; this is about which row survives.

Fix - make the tiebreak status-aware, then fall back to recency:

```js
.sort((a, b) => {
  const aActive = a.getStatus() === Oppty.STATUSES.IN_PROGRESS;
  const bActive = b.getStatus() === Oppty.STATUSES.IN_PROGRESS;
  if (aActive !== bActive) return aActive ? -1 : 1;
  return new Date(b.getUpdatedAt()) - new Date(a.getUpdatedAt());
});
```

Please also add the missing inverse-recency test (stale IN_PROGRESS + more-recent NEW, asserting IN_PROGRESS is kept and NEW retired) and an equal-`updatedAt` tie test - the suite currently only covers "IN_PROGRESS is more recent," which is the safe direction.

Honest caveat on frequency: `idx_opportunities_unique_active_scope` prevents a committed same-scope NEW+IN_PROGRESS pair, and offsite is one-brand-per-site, so in steady state this branch is rare - it fires mainly on legacy/pre-index or anomalous multi-row data. But resolving those anomalous states is exactly what this convergence branch is for, so it should keep the customer's row deliberately.

rarescheseli (changes requested):
Hey @grigoreadobe,

**Verdict: Request changes** - the reuse-on-refresh fix is correct and well-tested, but the cross-status convergence can retire the very IN_PROGRESS opportunity this PR sets out to preserve.
_Changes: broadens the evergreen offsite resolver to reuse an active IN_PROGRESS opportunity (not just NEW) on refresh, avoiding a unique-index collision that dropped suggestions; adds coverage across the resolver and the cited/reddit/youtube guidance handlers (6 files)._

## Must fix before merge
1. [Important] Convergence picks the survivor by `updatedAt` recency alone, so a more-recently-updated NEW row can retire a customer-activated IN_PROGRESS opportunity to IGNORED - `src/common/offsite-refresh.js:159` (details inline)

<details>

- nit: the `allBySiteIdAndStatus ... calledTwice` assertion is a count-based guard coupled to `ACTIVE_OFFSITE_STATUSES.length`; prefer `calledWith(siteId, 'IN_PROGRESS')` (and NEW) so it states intent and survives adding a third active status - `test/audits/cited-analysis/guidance-handler.test.js` (and the reddit/youtube equivalents)
- suggestion: consider a short ADR in `docs/decisions/` (or a follow-up ticket) capturing the documented IN_PROGRESS->NEW status-downgrade trade-off, so the "customer activated this" semantics stay discoverable for later work
</details>

- Convergence retires "duplicates" by `(siteId, type)` without a scope filter; safe under today's one-brand-per-site assumption, but would retire legitimate distinct-scope opportunities if offsite ever hosts multiple brand scopes per site. A scope-aware resolve lives partly in `@adobe/spacecat-shared-data-access`.
- Residual resolve-then-persist race: two concurrent refreshes that both resolve `null` both call `Opportunity.create`; one wins, the other hits 23505 and fails before syncSuggestions. Pre-existing and self-heals on the next run; a true fix needs upsert / ON CONFLICT in data-access.
- IN_PROGRESS->NEW status downgrade of the reused row (already flagged as a follow-up in the PR description).

PR #2758, opened 2026-07-09
fix(sync): freeze SKIPPED/REJECTED suggestions in syncSuggestions (SITES-44646)

absarasw:
## Summary

- **Root cause**: `syncSuggestions` had no guard for terminal-status suggestions in its matched-suggestions loop. When a merge function produced any difference in data (updated backlink counts, new metrics, etc.), SKIPPED and REJECTED suggestions were still added to `toUpdate` and saved via `saveMany` — bumping `updatedAt` on every scan.
- **PR #2593 was a partial fix**: its `deepEqual` guard prevents spurious bumps only when data is *identical*. Any data delta still triggered a save.
- **This fix**: return early for `SKIPPED` and `REJECTED` suggestions before merge functions run. Their `updatedAt` must reflect the moment the operator acted, not subsequent audit runs.
- **`FIXED` is intentionally excluded**: some audits (e.g. `permissions/suggestion-data-mapper.js`) implement regression detection by transitioning `FIXED → NEW` via a custom `mergeStatusFunction`. That path must remain reachable.

## Changes

- [`src/utils/data-access.js`](src/utils/data-access.js): added `FROZEN_STATUSES` early-return guard in the `matchedSuggestions.forEach` loop of `syncSuggestions`
- [`test/utils/data-access.test.js`](test/utils/data-access.test.js): corrected 3 existing REJECTED tests that were asserting the buggy behavior (saveMany called on data change), added tests for SKIPPED freeze and FIXED regression path

## Test plan

- [ ] `npx mocha test/utils/data-access.test.js` — 138 passing, 0 failing
- [ ] SKIPPED suggestion with changed data → not saved, `updatedAt` unchanged
- [ ] REJECTED suggestion with changed data → not saved, `updatedAt` unchanged
- [ ] FIXED suggestion with custom mergeStatusFunction → still transitions to NEW (regression detection intact)
- [ ] NEW/PENDING_VALIDATION/OUTDATED suggestions still update normally

Fixes: [SITES-44646](https://jira.corp.adobe.com/browse/SITES-44646)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2787, opened 2026-07-21
fix(internal-links): freeze SKIPPED/REJECTED suggestions in custom sync loop (SITES-44646)

absarasw:
## Summary

Follow-up to [PR #2758](https://github.com/adobe/spacecat-audit-worker/pull/2758).

The internal-links audit has its own suggestion-sync loop in [suggestions-generator.js](src/internal-links/suggestions-generator.js) — it does **NOT** go through the shared `syncSuggestions`, so the `FROZEN_STATUSES` guard added there does not apply here. Every matched suggestion (including SKIPPED and REJECTED) was still going through `setData` + `setUpdatedBy('system')` + `saveMany` on every audit run, bumping `updatedAt` even though the operator's decision should be permanent.

This is the residual leak driving the ~30% REJECTED / ~39% SKIPPED weekly refresh rates the ESE team is still seeing on the ASO dashboard after #2758 landed.

## Fix

Mirror the `FROZEN_STATUSES` early-filter from the shared `syncSuggestions` in the internal-links sync loop. Terminal-status suggestions are dropped before `setData` / `setUpdatedBy` / `saveMany` run.

## Test plan

- [x] `npx mocha test/audits/internal-links/suggestions-generator.test.js` — 43 passing
- [x] Renamed the misleading `keeps rejected suggestions unchanged` test (it was asserting `saveMany.calledOnce` — the *buggy* behavior) to reflect the correct freeze contract: no `setData`, no `setUpdatedBy`, no `saveMany`.
- [x] Added a matching SKIPPED test.
- [x] Lint clean.

Fixes: [SITES-44646](https://jira.corp.adobe.com/browse/SITES-44646)

### DIFF
--- a/src/internal-links/opportunity-suggestions.js
+++ b/src/internal-links/opportunity-suggestions.js
@@ -100,18 +100,12 @@ export function createOpportunityAndSuggestionsStep({
         log.info('no broken internal links found, updating opportunity to RESOLVED');
         await opportunity.setStatus(opptyStatuses.RESOLVED);
         const suggestions = await opportunity.getSuggestions();
-        // Preserve operator decisions: SKIPPED / REJECTED (and other terminal
-        // statuses) must not be flipped to OUTDATED — that would both destroy
-        // the decision and bump updatedAt, corrupting the ASO "Moved to
-        // Rejected / Skipped" metrics (SITES-44646).
+        // Preserve operator decisions on SKIPPED / REJECTED — flipping those to
+        // OUTDATED would destroy the decision and bump updatedAt, corrupting
+        // the ASO "Moved to Rejected / Skipped" metrics (SITES-44646).
         const suggestionsToOutdate = (suggestions || []).filter((s) => ![
-          suggestionStatuses.OUTDATED,
-          suggestionStatuses.FIXED,
-          suggestionStatuses.ERROR,
           suggestionStatuses.SKIPPED,
           suggestionStatuses.REJECTED,
-          suggestionStatuses.APPROVED,
-          suggestionStatuses.IN_PROGRESS,
         ].includes(s.getStatus()));
         if (isNonEmptyArray(suggestionsToOutdate)) {
           await Suggestion.bulkUpdateStatus(suggestionsToOutdate, suggestionStatuses.OUTDATED);
--- a/src/internal-links/opportunity-suggestions.js
+++ b/src/internal-links/opportunity-suggestions.js
@@ -100,16 +100,17 @@ export function createOpportunityAndSuggestionsStep({
         log.info('no broken internal links found, updating opportunity to RESOLVED');
         await opportunity.setStatus(opptyStatuses.RESOLVED);
         const suggestions = await opportunity.getSuggestions();
-        // Preserve operator decisions: SKIPPED / REJECTED (and other terminal
-        // statuses) must not be flipped to OUTDATED — that would both destroy
-        // the decision and bump updatedAt, corrupting the ASO "Moved to
-        // Rejected / Skipped" metrics (SITES-44646).
+        // Rule: don't bump updatedAt on suggestions whose state isn't changing.
+        // - NEW / PENDING_VALIDATION / SKIPPED / REJECTED: the underlying issue is
+        //   now gone → transition to OUTDATED (legitimate state change, updatedAt
+        //   bumps accordingly). This is the "Moved to OUTDATED" signal ASO needs.
+        // - FIXED / APPROVED / IN_PROGRESS: represent completed or in-flight work
+        //   that must not be overwritten by an audit sweep.
+        // - OUTDATED / ERROR: unchanged.
         const suggestionsToOutdate = (suggestions || []).filter((s) => ![
           suggestionStatuses.OUTDATED,
           suggestionStatuses.FIXED,
           suggestionStatuses.ERROR,
-          suggestionStatuses.SKIPPED,
-          suggestionStatuses.REJECTED,
           suggestionStatuses.APPROVED,
           suggestionStatuses.IN_PROGRESS,
         ].includes(s.getStatus()));
--- a/src/internal-links/opportunity-suggestions.js
+++ b/src/internal-links/opportunity-suggestions.js
@@ -100,17 +100,16 @@ export function createOpportunityAndSuggestionsStep({
         log.info('no broken internal links found, updating opportunity to RESOLVED');
         await opportunity.setStatus(opptyStatuses.RESOLVED);
         const suggestions = await opportunity.getSuggestions();
-        // Rule: don't bump updatedAt on suggestions whose state isn't changing.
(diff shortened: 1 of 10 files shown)


MysticatBot (approved):
Hey @absarasw,

**Verdict: Approve** - clean, well-scoped bug fix that correctly mirrors the shared FROZEN_STATUSES guard in the internal-links custom sync loop.
**Complexity: MEDIUM** - medium diff, tests-heavy.
_Changes: Freezes SKIPPED/REJECTED (and other terminal) suggestions from being overwritten during internal-links audit sync and resolve flows, fixing the spurious updatedAt bumps driving incorrect ASO dashboard refresh rates (5 files)._

<details>

- suggestion: Consider extracting a shared constant for the frozen/terminal status lists. The FROZEN_STATUSES array (SKIPPED, REJECTED) now lives in both `src/utils/data-access.js` and `src/internal-links/suggestions-generator.js`, and the broader "do not outdate" exclusion list in `opportunity-suggestions.js` is a superset. A single exported constant would prevent silent drift if a third status is ever frozen. - `src/internal-links/suggestions-generator.js:405`
- nit: `FROZEN_STATUSES` is defined inside the function body of `syncBrokenInternalLinksSuggestions`, re-created on every invocation. Moving it to module scope (as the shared `syncSuggestions` does in `data-access.js`) would improve discoverability and consistency. - `src/internal-links/suggestions-generator.js:405`

</details>

Note: CI checks are passing.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 17s | <b>Cost:</b> $4.91 | <b>Commit:</b> <code>889dffa951179527635f5c1f203d02a10e833d59</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/internal-links/opportunity-suggestions.js
```diff
@@ -100,8 +100,21 @@ export function createOpportunityAndSuggestionsStep({
...
-          await Suggestion.bulkUpdateStatus(suggestions, suggestionStatuses.OUTDATED);
+        // Preserve operator decisions: SKIPPED / REJECTED (and other terminal
+        // statuses) must not be flipped to OUTDATED — that would both destroy
+        // the decision and bump updatedAt, corrupting the ASO "Moved to
+        // Rejected / Skipped" metrics (SITES-44646).
+        const suggestionsToOutdate = (suggestions || []).filter((s) => ![
+          suggestionStatuses.OUTDATED,
+          suggestionStatuses.FIXED,
+          suggestionStatuses.ERROR,
+          suggestionStatuses.SKIPPED,
```
Kanishkavijay39, line 107: Skipped is visible to customer. If it's no more broken we shoild move to outdated.

src/internal-links/opportunity-suggestions.js
```diff
@@ -100,8 +100,21 @@ export function createOpportunityAndSuggestionsStep({
...
+          suggestionStatuses.REJECTED,
+          suggestionStatuses.APPROVED,
+          suggestionStatuses.IN_PROGRESS,
```
Kanishkavijay39: Not a terminal state, An opportunity in_progress should go to fix or error?
Not sure should be there in the list or not.

src/internal-links/suggestions-generator.js
```diff
@@ -404,8 +402,24 @@ export async function syncBrokenInternalLinksSuggestions({
     statusToSetForOutdated: SuggestionDataAccess.STATUSES.OUTDATED,
   });
 
+  // SKIPPED and REJECTED suggestions are operator decisions that must not be
+  // overwritten by subsequent scans — their updatedAt must reflect the moment
+  // the operator acted. Mirrors the FROZEN_STATUSES guard in the shared
+  // syncSuggestions (SITES-44646).
+  const FROZEN_STATUSES = [
+    SuggestionDataAccess.STATUSES.SKIPPED,
```
Kanishkavijay39, line 410: Skipped is frozen status?

It can be move back to new by the customer.

Kanishkavijay39 (changes requested):

src/internal-links/opportunity-suggestions.js
absarasw: Good catch — applied in commit 1b3606ce. Removed `SKIPPED` and `REJECTED` from the exclusion list so both transition to OUTDATED when the underlying issue is gone. The `updatedAt` bump here is now a legitimate state change (ticket rule is *don't bump when state isn't changing*; here state IS changing).

src/internal-links/opportunity-suggestions.js
absarasw: Kept `IN_PROGRESS` (and `APPROVED`) in the exclusion because they represent in-flight or already-approved customer work — flipping mid-flight to OUTDATED would be a surprising state overwrite. Those transitions (IN_PROGRESS → FIXED / ERROR) happen when the fix work actually completes, not from an audit sweep. This matches the pattern in the shared `resolveOpportunityIfNoIssues` in `src/utils/data-access.js`.

src/internal-links/suggestions-generator.js
absarasw: The freeze here only prevents the **audit code** from silently touching SKIPPED on every scan (issue still live → no state change → no `updatedAt` bump — that's the SITES-44646 rule). Customer-initiated SKIPPED → NEW via the ASO UI is a completely separate code path (direct `suggestion.setStatus(NEW)` API call) that this filter doesn't restrict.

src/internal-links/opportunity-suggestions.js
absarasw: Correction / update: commit `1b3606ce` was reverted (see commit `945149f6`). SKIPPED and REJECTED are again preserved in the exclusion list — the audit no longer flips them to OUTDATED on the RESOLVED path, and their `updatedAt` is unaffected.

Why: the SITES-44646 bug is that when an operator SKIPs / REJECTs a suggestion but the underlying issue keeps being detected on every audit, `updatedAt` was being bumped weekly on those terminal-status suggestions — corrupting the ASO "Moved to Rejected / Skipped" metrics. That's fixed by the `FROZEN_STATUSES` guard in `suggestions-generator.js` (the sync loop that runs when broken links are still detected). Transitioning SKIPPED/REJECTED to OUTDATED when the whole opportunity resolves is a broader semantic change with fleet-wide impact (the shared `handleOutdatedSuggestions` and `resolveOpportunityIfNoIssues` follow the same "preserve terminal statuses" pattern today), so it's kept out of this PR to keep scope tight.

absarasw: https://jira.corp.adobe.com/browse/SITES-48663 Opened the ticket to handle the issue of IGNORED statuses properly.

Kanishkavijay39 (approved):

PR #2788, opened 2026-07-21
fix: classify .md/.markdown URLs as md content type (LLMO-6399)

akshaymagapu:
## Problem

In the Agentic Traffic dashboard, filtering by the **MD** content type returns no results, while filtering the same view by URL containing `.md` returns hits. The two should be equivalent.

Root cause: `inferContentType(urlPath)` in `src/cdn-logs-report/utils/agentic-traffic-mapper.js` classifies URLs by file extension but has **no branch for `.md`**. Markdown URLs fall through to `return 'other'`, so they are stored as content type `other` and never match the `md` filter that the dashboard always offers. Markdown is already a first-class known content type upstream in `cdn-logs-analysis` (its SQL filters `response_content_type` to `text/html|application/pdf|text/markdown`), so it deserves its own bucket.

## Change

Add a `.md`/`.markdown` -> `md` branch to `inferContentType`, placed after the `.txt` check and before the no-extension catch-all:

```js
if (/\.(md|markdown)$/.test(path)) {
  return 'md';
}
```

## Tests

Extended the existing extension-classification case with `/foo.md`, `/foo.markdown`, and `/FOO.MD` (case-insensitivity). Full suite passes (9114 passing), 100% coverage gate met; changed file at 100/100/100/100.

## Follow-ups (not in this PR)

- Existing rollup rows classified `other` need a data backfill to reclassify to `md`.
- project-elmo-ui hardcodes the content-type list in two places (`src/constants/filters.ts` and `src/utils/filterHandlers.ts`); keep them in sync with the mapper output.

Ref: LLMO-6399

### DIFF
--- a/src/cdn-logs-report/utils/agentic-traffic-mapper.js
+++ b/src/cdn-logs-report/utils/agentic-traffic-mapper.js
@@ -73,6 +73,9 @@ function inferContentType(urlPath = '') {
   if (/\.txt$/.test(path)) {
     return 'txt';
   }
+  if (/\.(md|markdown)$/.test(path)) {
+    return 'md';
+  }
   if (!path || path.endsWith('/') || !path.split('/').pop().includes('.')) {
     return 'html';
   }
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - clean, well-scoped bug fix with proper test coverage.
**Complexity: TRIVIAL** - 9 lines, 2 files, single domain.
_Changes: Adds `.md`/`.markdown` extension classification to `inferContentType` so markdown URLs map to the `md` content type instead of falling through to `other` (2 files)._

<details>

- nit: The new `.md`/`.markdown` branch sits after `.txt` but the function groups markup/document types (`.xml`, `.pdf`, `.html`/`.htm`) together above `.txt`. Placing `.md` next to `.html`/`.htm` would be slightly more intuitive for future readers - cosmetic only. - `src/cdn-logs-report/utils/agentic-traffic-mapper.js:76`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 22s | <b>Cost:</b> $2.28 | <b>Commit:</b> <code>6231b35e10ca29254c879fc890d217bfe2eb55d8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2789, opened 2026-07-21
fix(broken-backlinks): normalise camelCase/snake_case data + harden reconcile (SITES-48410)

absarasw:
## Summary

Three related fixes for `reconcileDisappearedSuggestions` on the broken-backlinks path, motivated by [SITES-48410](https://jira.corp.adobe.com/browse/SITES-48410).

### 1. Schema-drift normalisation (root cause)

`suggestion.data` has two shapes in the wild:
- **Audit worker** (this repo) writes **snake_case**: `url_to`, `url_from`, `urls_suggested`, `traffic_domain`
- **ASO UI edit mutation** rewrites the record in **camelCase**: `urlTo`, `urlFrom`, `urlsSuggested`, `urlEdited`, `isEdited`

Every audit-side consumer that reads snake_case silently short-circuits once a customer edits the suggestion. `checkIfBacklinkFixedWithSuggestion` returns `false` at the `if (!urlTo)` guard without ever calling `fetch()`, so **customer redirects matching `urlEdited` exactly go uncredited**.

Added `normaliseBacklinkFields(data)` and use it in `checkIfBacklinkFixedWithSuggestion`, `buildBacklinkFixEntityPayload`, and `checkIfBacklinkResolvedOnProduction`.

### 2. Widen reconcile candidacy for edited/OUTDATED suggestions

If the customer deploys their redirect **after** a prior audit swept the suggestion to OUTDATED, it's stuck — reconcile only accepts `NEW`, and `handleOutdatedSuggestions` excludes OUTDATED. Accept `status=OUTDATED` when `data.isEdited === true` (customer explicitly picked a target = high-confidence attribution). Non-edited OUTDATED records stay excluded to avoid attribution drift.

### 3. Reorder FixEntity write before status flip

Current path flips `status=FIXED` first via `Suggestion.saveMany`, then best-effort `addFixEntities` with swallow-and-log. Any transient failure leaves `status=FIXED` with **no FixEntity** — invisible to the `/fixes` API. Reordered: build+persist FixEntities first; only on success do we flip status. On failure, suggestions stay in prior status for the next audit to retry.

Companion to [SITES-48404](https://jira.corp.adobe.com/browse/SITES-48404) — Anmol's write-up covers three mystique apply-side paths with the same anti-pattern; this is Path 4 (audit-worker reconcile).

## Evidence

Live-network probe (2026-07-21) against 6 lilly.com broken-backlink URLs the customer trial-summary PDF reported as resolved:

| # | url_to | Live resolution | Matches | ASO status |
|---|---|---|---|---|
| 1 | `/our-medicines/humalog-and-lispro` | 301 → `supply.lilly.com/` | `urlEdited` exact | OUTDATED |
| 2 | `/news/stories/authors/daniel-skovronsky` | 301 → `/about/leadership/daniel-skovronsky` | `urlEdited` exact | OUTDATED |
| 3 | `/responsibility` | 301 → `/about/impact/improving-lives-and-communities` | `urlEdited` exact | OUTDATED |
| 4 | `/news/stories/ukraine-support` | 200 as-is | — | OUTDATED |
| 5 | `/news/stories/inflation-reduction-act-...` | 200 as-is | — | OUTDATED |
| 6 | `/index` (★ ASO autofix) | 301 → `/` | `urlsSuggested[0]` exact | FIXED ✅ |

3 of 6 are real customer-deployed redirects to the exact `urlEdited` target picked in the ASO UI, all currently OUTDATED with no FixEntity. Row 6 (unedited, matched `urlsSuggested[0]`) is the only one credited today — confirming the code works when suggestion data stays snake_case.

## Not in scope

The UI-side edit mutation that first introduces the camelCase form remains as-is — this PR makes the backend robust to either casing so the fix doesn't couple to a UI change.

A separate one-shot backfill utility (not in this PR) will replay reconcile against existing OUTDATED-with-isEdited suggestions to credit historical customer fixes retroactively.

## Test plan

- [x] `npx mocha test/utils/data-access.test.js --grep 'reconcileDisappearedSuggestions'` — 11 passing (3 new)
- [x] `npx mocha test/audits/backlinks.test.js --grep 'checkIfBacklinkFixedWithSuggestion|buildBacklinkFixEntityPayload|checkIfBacklinkResolvedOnProduction'` — 16 passing (2 new)
- [x] `npx eslint src/backlinks/handler.js src/utils/data-access.js` — clean
- [ ] Post-merge on next audit run: verify at least one of the 3 Lilly Class A URLs (`26a6e578`, `f9296d8c`, `019e8321-d52f-71c5`) flips OUTDATED → FIXED with FixEntity

### DIFF
--- a/src/backlinks/handler.js
+++ b/src/backlinks/handler.js
@@ -171,23 +171,22 @@ async function filterOutValidBacklinks(backlinks, log) {
  * fields, causing every audit-side consumer that reads snake_case to silently
  * short-circuit on `undefined`. This helper accepts either casing.
  *
+ * Precedence is uniform: snake_case (the form the audit-worker writes) is
+ * preferred; camelCase is the UI-mutation fallback. `urlEdited` and `isEdited`
+ * are intentionally asymmetric — they are UI-only fields that the audit-worker
+ * never writes, so there is no snake_case equivalent to fall back to.
+ *
  * @param {Object} data - The suggestion's `.getData()` payload.
  * @returns {{urlTo: string|undefined, urlFrom: string|undefined,
  *            urlsSuggested: string[], urlEdited: string|undefined}}
  */
 function normaliseBacklinkFields(data) {
-  let urlsSuggested = [];
-  if (Array.isArray(data?.urlsSuggested)) {
-    urlsSuggested = data.urlsSuggested;
-  /* c8 ignore next 2 */
-  } else if (Array.isArray(data?.urls_suggested)) {
-    urlsSuggested = data.urls_suggested;
-  }
+  const urlsSuggestedRaw = data?.urls_suggested ?? data?.urlsSuggested;
   return {
     urlTo: data?.url_to ?? data?.urlTo,
     urlFrom: data?.url_from ?? data?.urlFrom,
-    urlsSuggested,
-    urlEdited: data?.urlEdited,
+    urlsSuggested: Array.isArray(urlsSuggestedRaw) ? urlsSuggestedRaw : [],
+    urlEdited: data?.urlEdited, // UI-only field; no snake_case equivalent
   };
 }
 
--- a/src/backlinks/handler.js
+++ b/src/backlinks/handler.js
@@ -164,6 +164,34 @@ async function filterOutValidBacklinks(backlinks, log) {
 }
 
 /**
+ * Normalises a suggestion's data payload against the historical schema drift
+ * between the audit-worker (snake_case: `url_to`, `url_from`, `urls_suggested`,
+ * `traffic_domain`) and the ASO UI edit mutation (camelCase: `urlTo`, `urlFrom`,
+ * `urlsSuggested`, `trafficDomain`). Edited-in-UI records lose their snake_case
+ * fields, causing every audit-side consumer that reads snake_case to silently
+ * short-circuit on `undefined`. This helper accepts either casing.
+ *
+ * @param {Object} data - The suggestion's `.getData()` payload.
+ * @returns {{urlTo: string|undefined, urlFrom: string|undefined,
+ *            urlsSuggested: string[], urlEdited: string|undefined}}
+ */
+function normaliseBacklinkFields(data) {
+  let urlsSuggested = [];
+  if (Array.isArray(data?.urlsSuggested)) {
+    urlsSuggested = data.urlsSuggested;
+  /* c8 ignore next 2 */
+  } else if (Array.isArray(data?.urls_suggested)) {
+    urlsSuggested = data.urls_suggested;
+  }
+  return {
+    urlTo: data?.url_to ?? data?.urlTo,
+    urlFrom: data?.url_from ?? data?.urlFrom,
+    urlsSuggested,
+    urlEdited: data?.urlEdited,
+  };
+}
+
+/**
  * Checks if a broken backlink issue was fixed using our AI suggestion.
  * Verifies if the broken URL now redirects to one of the suggested targets.
  *
@@ -172,12 +200,8 @@ async function filterOutValidBacklinks(backlinks, log) {
  * @returns {Promise<boolean>} - True if the issue was fixed with our suggestion.
  */
 export async function checkIfBacklinkFixedWithSuggestion(suggestion, log) {
-  const data = suggestion?.getData?.();
-  const urlTo = data?.url_to;
-  const suggestedTargets = Array.isArray(data?.urlsSuggested) ? data.urlsSuggested : [];
-  const targets = data?.urlEdited
-    ? [data.urlEdited, ...suggestedTargets]
-    : suggestedTargets;
+  const { urlTo, urlsSuggested, urlEdited } = normaliseBacklinkFields(suggestion?.getData?.());
+  const targets = urlEdited ? [urlEdited, ...urlsSuggested] : urlsSuggested;
   if (!urlTo || targets.length === 0) {
     return false;
   }
@@ -207,7 +231,9 @@ export async function checkIfBacklinkFixedWithSuggestion(suggestion, log) {
  * @returns {Object} - The fix entity payload.
  */
 export function buildBacklinkFixEntityPayload(suggestion, opportunity, isAuthorOnly, site) {
-  const data = suggestion?.getData?.();
+  const {
+    urlTo, urlFrom, urlsSuggested, urlEdited,
+  } = normaliseBacklinkFields(suggestion?.getData?.());
   return {
     opportunityId: opportunity.getId(),
     status: isAuthorOnly
@@ -217,9 +243,9 @@ export function buildBacklinkFixEntityPayload(suggestion, opportunity, isAuthorO
     executedAt: new Date().toISOString(),
     changeDetails: {
       system: site.getDeliveryType(),
-      pagePath: data?.url_from,
-      oldValue: data?.url_to || '',
-      updatedValue: data?.urlEdited || data?.urlsSuggested?.[0] || '',
+      pagePath: urlFrom,
+      oldValue: urlTo || '',
+      updatedValue: urlEdited || urlsSuggested[0] || '',
     },
     suggestions: [suggestion?.getId?.()],
   };
@@ -234,11 +260,11 @@ export function buildBacklinkFixEntityPayload(suggestion, opportunity, isAuthorO
  * @returns {Promise<boolean>} - True if the issue is resolved on production.
  */
 export async function checkIfBacklinkResolvedOnProduction(suggestion, log) {
-  const url = suggestion?.getData?.()?.url_to;
-  if (!url) {
+  const { urlTo } = normaliseBacklinkFields(suggestion?.getData?.());
+  if (!urlTo) {
     return false;
   }
-  const stillBrokenItems = await filterOutValidBacklinks([{ url_to: url }], log);
+  const stillBrokenItems = await filterOutValidBacklinks([{ url_to: urlTo }], log);
   return stillBrokenItems.length === 0; // resolved if NO items are still broken
 }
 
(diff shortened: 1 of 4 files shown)


src/backlinks/handler.js
```diff
@@ -163,6 +163,34 @@ async function filterOutValidBacklinks(backlinks, log) {
...
+ * @param {Object} data - The suggestion's `.getData()` payload.
+ * @returns {{urlTo: string|undefined, urlFrom: string|undefined,
+ *            urlsSuggested: string[], urlEdited: string|undefined}}
+ */
+function normaliseBacklinkFields(data) {
+  let urlsSuggested = [];
+  if (Array.isArray(data?.urlsSuggested)) {
+    urlsSuggested = data.urlsSuggested;
+  /* c8 ignore next 2 */
+  } else if (Array.isArray(data?.urls_suggested)) {
```
MysticatBot: issue (blocking): The `/* c8 ignore next 2 */` pragma opts out of coverage for the `urls_suggested` (snake_case) branch. The repo CLAUDE.md requires:

> Coverage enforcement is strict: **100% lines, branches, and statements** across all `src/**/*.js`. Every code path must be tested.

This branch is reachable (data with `urls_suggested` but no `urlsSuggested`) and straightforward to test:

```js
it('should normalise snake_case urls_suggested', async () => {
  nock('https://example.com').get('/broken').reply(301, '', { Location: 'https://example.com/fix' });
  nock('https://example.com').get('/fix').reply(200);
  const suggestion = {
    getData: () => ({ url_to: 'https://example.com/broken', urls_suggested: ['https://example.com/fix'] }),
  };
  const result = await checkIfBacklinkFixedWithSuggestion(suggestion, mockLog);
  expect(result).to.be.true;
});
```

Remove the pragma and add a test case exercising the snake_case path.

MysticatBot (changes requested):
Hey @absarasw,

**Verdict: Request changes** - one convention violation to resolve.
**Complexity: MEDIUM** - medium diff, single domain (backlinks reconcile path).
_Changes: normalises camelCase/snake_case schema drift in broken-backlinks suggestion data, widens reconcile candidacy for edited OUTDATED suggestions, and reorders FixEntity writes before status flips to prevent attribution drift (4 files)._

## Must fix before merge
1. [Important] Coverage pragma violates repo convention - `src/backlinks/handler.js:~183` (details inline)

<details>

- nit: Precedence order in `normaliseBacklinkFields` is inconsistent - `urlTo`/`urlFrom` prefer snake_case (`data?.url_to ?? data?.urlTo`) while `urlsSuggested` checks camelCase first. In practice records have one casing, not both, so this is a readability issue. Consider making all fields uniform. - `src/backlinks/handler.js:187`
- suggestion: `urlEdited` has no snake_case fallback (`url_edited`) unlike the other fields. If this field is always camelCase (UI-originated only), a brief comment explaining the asymmetry would help the next reader. - `src/backlinks/handler.js:191`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 9m 40s | <b>Cost:</b> $6.18 | <b>Commit:</b> <code>e6f37fc8f224461051061824c4724f9bbd920d5d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/backlinks/handler.js
absarasw: Addressed in 922561ad — removed the pragma and added a dedicated test (`should accept snake_case urls_suggested`) that exercises the branch. Local `c8 --check-coverage --lines 100 --statements 100 --branches 100 --functions 100` now passes on both touched files.

absarasw:
@MysticatBot — thanks for the review. Addressed in 922561ad:

- **[Blocking]** Removed `/* c8 ignore */` pragma and added a test that exercises the `urls_suggested` (snake_case) branch. Local coverage on both touched files is now 100% lines / statements / branches / functions.
- **[nit] Precedence uniformity** — flipped `urlsSuggested` to snake_case-first so every field in `normaliseBacklinkFields` follows the same order (`data?.foo ?? data?.fooCamel`). Rationale in the updated JSDoc: snake_case is what the audit-worker natively writes; camelCase is the UI-mutation fallback.
- **[suggestion] `urlEdited` asymmetry** — added an inline comment explaining that `urlEdited` (and `isEdited`) are UI-only fields the audit-worker never writes, so there's deliberately no snake_case fallback to look for.

MysticatBot (approved):
Hey @absarasw,

**Verdict: Approve** - well-scoped bug fix with solid test coverage; all prior findings addressed.
**Complexity: MEDIUM** - medium diff, single domain (backlinks reconcile path).
_Changes: normalises camelCase/snake_case schema drift in broken-backlinks suggestion data, widens reconcile candidacy for edited OUTDATED suggestions, and reorders FixEntity writes before status flips to prevent attribution drift (4 files)._

- Coverage pragma removed and dedicated test added for the snake_case `urls_suggested` branch
- Precedence order made uniform (snake_case-first for all fields in `normaliseBacklinkFields`)
- `urlEdited` asymmetry explained with inline comment (UI-only field, no snake_case equivalent)

<details>

- nit: If `buildFixEntityPayload` throws for every suggestion in the batch, `fixEntityObjects` stays empty, the `addFixEntities` guard is skipped, and suggestions get marked FIXED without a FixEntity - the same drift the reorder intends to prevent. Unlikely in practice (payload construction throws are code bugs, not infra transients), but `if (fixEntityObjects.length === 0 && typeof buildFixEntityPayload === 'function') return;` after the loop would close the window. - `src/utils/data-access.js:731`
- nit: If `addFixEntities` succeeds but `Suggestion.saveMany` throws, suggestions stay in prior status for next-audit retry which may produce duplicate fix entities. Whether this matters depends on downstream dedup in `addFixEntities`. The trade-off is documented and the window is narrow. - `src/utils/data-access.js:745`
- suggestion: `checkIfBacklinkResolvedOnProduction` now routes through `normaliseBacklinkFields` but existing tests only exercise it with `url_to` (snake_case). A single test passing `{ urlTo: '...' }` would confirm the camelCase fallback works for this function too. - `test/audits/backlinks.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 42s | <b>Cost:</b> $6.92 | <b>Commit:</b> <code>922561ad6461924f85ead166d7c824e8bac6b142</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/utils/data-access.js
```diff
@@ -667,13 +678,20 @@ export async function reconcileDisappearedSuggestions({
...
+    // high-confidence attribution — see function docstring).
     const candidates = disappearedSuggestions.filter((s) => {
-      if (!newStatus || s?.getStatus?.() !== newStatus) {
-        return false;
+      const status = s?.getStatus?.();
+      if (newStatus && status === newStatus) {
+        return true;
       }
-      return true;
+      if (outdatedStatus && status === outdatedStatus && s?.getData?.()?.isEdited === true) {
```
Kanishkavijay39, line 691: In every audit run the same suggestion will get checked which is edited outdated.
Is there any way we only check this in next audit run once if not match we don't check it again and again in all the next audit runs.

As outdated will be picked everytime.

Kanishkavijay39 (approved):
LGTM, minor suggestion for everytime outdated suggestions get picked.

src/utils/data-access.js
Kanishkavijay39, line 691: Ig, we can remove check on this outdatedStatus making if condition more simple

absarasw:
@MysticatBot — thanks for the approve. Follow-ups:

- **Nit #1 (all-throws drift)** — fixed in 80935b06. Now tracks a `payloadThrows` counter and returns early only when it equals the batch size, so the case you described (every payload build throws → `fixEntityObjects` empty → status still flipped) is guarded. New test `should NOT flip suggestion status when ALL buildFixEntityPayload calls throw (no drift)` locks it in.
- **Nit #2 (saveMany after addFixEntities succeeds)** — leaving as-is. The trade-off is documented and the window is narrow; downstream dedup in `addFixEntities` is the natural next hop and belongs in a follow-up if we ever observe the duplicate.
- **Suggestion #3 (camelCase test for `checkIfBacklinkResolvedOnProduction`)** — added in 80935b06 (`should accept camelCase urlTo when data is UI-edited`).

Local: 100/100/100/100 on both touched files, 30 targeted tests green.

src/utils/data-access.js
absarasw: Fair concern. Two things soften the cost in practice:

- **Bounded population** — the OUTDATED branch is gated on `data.isEdited === true` (customer explicitly picked a target in the UI). That's a small set per site (a handful, not hundreds).
- **Bounded concurrency** — `MAX_CONCURRENT_CHECKS = 5` caps parallel HTTP checks per audit run.

Skipping the OUTDATED case entirely (your simpler condition) is what the code did before this PR — and it's exactly what caused the 3 Lilly customer fixes to be uncredited. The customer deployed the redirect AFTER the initial NEW→OUTDATED sweep, and there was no path back. So the retries are load-bearing for late-deployed customer fixes.

That said, retrying on every audit forever is inelegant. I'll file a follow-up to add a `data.lastReconcileAttemptAt` timestamp so we only re-check every N days (e.g. weekly). That way we still catch late fixes without hitting the customer's origin every audit run. Not blocking for this PR.

absarasw: Same underlying concern as your thread above — replied there in detail. TL;DR: removing the OUTDATED branch is what the code did before this PR, and it's the exact reason the 3 Lilly customer fixes went uncredited (customers deployed the redirect AFTER the initial NEW→OUTDATED sweep, with no code path back to FIXED). The OUTDATED+isEdited branch is load-bearing.

The retry-throttle idea (a `lastReconcileAttemptAt` timestamp so we only re-check every N days) is the cleaner shape and I'll file it as a follow-up.

PR #2798, opened 2026-07-22
feat(referral-traffic): generate + classify referral URL categories write-time-in-service (optel + cdn) | LLMO-6257

omairtemurian:
## 1. Abstract
The audit-worker slice of LLMO-6257 Phase 2: generates a site's shared URL category rules (create-if-missing) and classifies referral URLs **write-time-in-service** for the optel and cdn sources, emitting a category-only classification dataset for the projector to import into `referral_url_classifications`.

## 2. Reasoning
P1 decoupled the Referral Traffic category filter onto the referral-owned `referral_url_classifications` table. P2 populates it for the sources audit-worker produces. Because audit-worker runs where the category rules are readable, its sources classify in-service (compute the category as the daily export runs) and ship the result through the projector — the model @cwjwisse's P1 review landed on. This covers both optel (`llmo-referral-traffic-daily`) and cdn (`cdn-logs-report`).

- **Rule generation (create-if-missing).** `generateReferralCategoryRules` builds the site's shared `agentic_url_category_rules` from the referral corpus if none exist (LLM `analyzeProducts` → `mergePatternRules` → replace), skipping when rules already exist. Wired best-effort into the `llmo-referral-traffic-daily` runner (a generation hiccup never blocks the daily traffic export). (CDN sites already get rules from the existing `generateAgenticPatterns` Athena path.)
- **Classification (`classify.js`).** A small pure matcher — `classifyUrlPath` (first rule by `sort_order`, uncompilable regex = no-match, JS analogue of the SQL `_safe_regex_match`) + `buildClassificationRows` (distinct `(host, url_path)`, matched-only).
- **Emit, per source.** After the traffic export, each source classifies its run's URLs and emits a **second** dataset — a `host,url_path,category_name,updated_by` CSV + a projector message on `pipeline_id=referral_url_classifications` (FIFO group `referral_url_classifications:<siteId>`), `updated_by=spacecat:optel` / `spacecat:cdn`. Best-effort (the traffic export has already succeeded), cleanup-on-dispatch-failure. The cdn emit uses the importer S3 client and a source-namespaced dedup id so it cannot collide with optel's message on the shared FIFO group.

Behaviour delta: before, optel/cdn referral URLs got no ongoing category. After, each daily export also ships a classification dataset that (via the projector + the data-service import RPC) populates `referral_url_classifications`, so those sites resolve categories in the Referral Traffic filter.

## 4. Required information
- Jira / issue: [LLMO-6257](https://jira.corp.adobe.com/browse/LLMO-6257)
- Spec: `docs/specs/2026-07-21-referral-category-generation-p2.md` (in this PR; synced to as-built) + the merged design-of-record `mysticat-data-service docs/plans/2026-07-17-llmo-6257-referral-category-generation.md`
- Other: data-service P2 PR https://github.com/adobe/mysticat-data-service/pull/836 (the import RPC) · projector P2 PR (routes the pipeline)

- `mysticat-data-service` — **contract.** The emitted `host,url_path,category_name,updated_by` CSV is imported by `wrpc_import_referral_url_classifications` (data-service P2 PR #836).
- `mysticat-projector-service` — **contract.** Routes `pipeline_id=referral_url_classifications` to that RPC (projector P2 PR).

- Exercised the classify + emit for both sources at the handler level: confirmed the classification CSV columns/body, the `pipeline_id=referral_url_classifications` message shape + FIFO group + source-namespaced dedup id, org_id present/absent, the no-rules / no-match skips, and the best-effort behaviour (a classification failure logs a warning and still returns a successful audit) with cleanup of the uploaded CSV on dispatch failure. The full writer→reader path is cross-service, so it is verified per-env, not locally.

## 7. Test plan
**Local (done):** the `llmo-referral-traffic-daily` and `cdn-logs-report` handler suites exercise rule-gen (create-if-missing skip), classify, and emit for optel + cdn, including the best-effort and cleanup paths.

**Per-env (dev → stage → prod):** once this, the projector config, and the data-service import RPC are all deployed, run a daily referral export for an optel and a cdn site with active category rules and confirm a `referral_url_classifications` message is emitted and the table gains rows for that site; confirm a site without rules simply skips (no message). Watch prod most closely (lower envs lack representative data).

- **related** data-service P2 https://github.com/adobe/mysticat-data-service/pull/836 (import RPC that consumes the emitted CSV) and the projector P2 PR (routes the pipeline). Each merges independently.
- Ordered sequence: any merge order; the emitted classification datasets are only imported once data-service #836 + the projector config are live on the same environment (until then the messages queue / DLQ harmlessly).


src/llmo-referral-traffic-daily/handler.js
```diff
@@ -177,6 +198,74 @@ export async function triggerTrafficAnalysisDailyImport(context) {
...
+  const classificationUri = `s3://${bucket}/${classificationKey}`;
+
+  await s3Client.send(new PutObjectCommand({
+    Bucket: bucket,
+    Key: classificationKey,
+    Body: serializeClassificationCsv(classificationRows),
+    ContentType: 'text/csv',
+  }));
+
+  const dedupId = createHash('sha256')
```
MysticatBot, line 234: issue (blocking): The dedup ID hash input is `${siteId}:${date}:referral_url_classifications` with no source suffix. The CDN version correctly uses `...referral_url_classifications:cdn`. Both sources serialize on the same FIFO group (`referral_url_classifications:${siteId}`), so their dedup IDs must be source-namespaced to prevent silent SQS deduplication collisions within a 5-minute window.

Fix: add `:optel` suffix:
```js
.update(`${siteId}:${date}:referral_url_classifications:optel`)
```

src/llmo-referral-traffic-daily/handler.js
```diff
@@ -72,6 +75,18 @@ export function serializeCsv(rows) {
   return [header, ...body].join('\r\n');
 }
 
+// Category-only classification CSV (LLMO-6257 P2) — imported into
+// referral_url_classifications by the projector via wrpc_import_referral_url_classifications.
+const CLASSIFICATION_CSV_COLUMNS = ['host', 'url_path', 'category_name', 'updated_by'];
```
MysticatBot: issue (blocking): `CLASSIFICATION_CSV_COLUMNS` and `serializeClassificationCsv` are duplicated identically in `referral-daily-export.js`. Both feed the same downstream projector import RPC (`wrpc_import_referral_url_classifications`). If a column is added to one and not the other, the projector will ingest malformed data.

Fix: export both from `classify.js` (which already owns the row shape via `buildClassificationRows`) and import in both handler files.

src/llmo-referral-traffic-daily/classify.js
```diff
@@ -0,0 +1,81 @@
...
+// _safe_regex_match): rules are pre-sorted (sort_order ASC, name ASC) by
+// fetchAgenticUrlClassificationRules, the first compilable rule whose regex matches
+// wins, an uncompilable regex is treated as no-match, and an unmatched URL gets no
+// classification (null). The result is imported into referral_url_classifications
+// through the projector single-writer FIFO.
+
+// (?i) is an inline modifier Postgres ~* honors; JS RegExp needs it stripped + /i.
+// Mirrors compileAthenaRegex so JS classification matches the SQL/read-path behaviour.
+function compileRuleRegex(regex) {
+  return new RegExp(String(regex).replace(/^\(\?i\)/, ''), 'i');
```
MysticatBot: issue (blocking): `new RegExp(regex, "i")` compiles and executes an arbitrary regex from the database without validating backtracking complexity. A rule regex with catastrophic backtracking (e.g., `(a+)+$`) will hang the Node.js event loop indefinitely.

The existing `patterns-uploader.js` guards (alternation branch count, coverage %) are heuristic content-quality checks, not security controls against exponential-time patterns. The rules are LLM-generated today, but the classifier has no independent validation.

Fix (pick one):
- Use `safe-regex2` to reject super-linear patterns before compilation
- Switch to the `re2` npm package for linear-time execution guarantees
- Add a maximum regex length and ban nested quantifiers (`(a+)+`, `(a*)*`) before compilation

If the team accepts the residual risk (bounded to one site per daily run, controlled LLM source), this can be tracked as a follow-up ticket rather than blocking merge.

MysticatBot (changes requested):
Hey @omairtemurian,

**Verdict: Request changes** - three fixable issues below, two of them one-liners.
**Complexity: HIGH** - large diff (1314 lines across 11 files).
_Changes: Adds write-time-in-service referral URL category generation and classification for both optel and cdn sources, emitting a second classification dataset through the projector FIFO for import into referral_url_classifications (11 files)._
Note: Recommend a human read before merge - this change amends an architectural spec (docs/specs/**). The bot review is a complement to, not a replacement for, a human read here.
Note: CI build is pending on the latest push.

## Must fix before merge
1. [Important] Dedup ID missing source suffix - optel emit uses unsuffixed ID while cdn uses `:cdn`, creating asymmetry that risks silent SQS dedup collision when a third source is added - `src/llmo-referral-traffic-daily/handler.js:237` (details inline)
2. [Important] Duplicated `serializeClassificationCsv` and `CLASSIFICATION_CSV_COLUMNS` across two files that feed the same downstream import RPC - drift risk when schema changes - `src/llmo-referral-traffic-daily/handler.js:80` (details inline)
3. [Important] ReDoS risk from unvalidated database-stored regex patterns compiled and executed without time bounds - `src/llmo-referral-traffic-daily/classify.js:24` (details inline)

<details>

- nit: CDN `emitReferralClassifications` wraps both S3 upload and SQS dispatch in the same try/catch, so an S3 failure triggers `cleanupCsvFromS3` on a key never written (harmless no-op but semantically misleading vs the optel version) - `src/cdn-logs-report/referral-daily-export.js`
- nit: `classifyUrlPath` recompiles every rule's regex on each call; the established pattern in this codebase (`COMPILED_COUNTRY_PATTERNS`) is compile-once-reuse - `src/llmo-referral-traffic-daily/classify.js`
- suggestion: Add a defensive guard in `buildClassificationRows` to skip rows where `host` or `url_path` is falsy, preventing corrupt dedup keys if upstream contracts weaken - `src/llmo-referral-traffic-daily/classify.js:64`
- suggestion: Optel handler test should assert classification CSV header matches expected column order (`host,url_path,category_name,updated_by`) to pin the contract - `test/audits/llmo-referral-traffic-daily/handler.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 7m 49s | <b>Cost:</b> $5.91 | <b>Commit:</b> <code>a8c793b55666519558bd0d9d5398ba28fa492acf</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @omairtemurian,

**Verdict: Approve** - all three prior blocking findings are addressed; no new blocking issues.
**Complexity: HIGH** - large diff (1314 lines across 11 files).
_Changes: Adds write-time-in-service referral URL category generation and classification for both optel and cdn sources, emitting a second classification dataset through the projector FIFO for import into referral_url_classifications (11 files)._
Note: Recommend a human read before merge - this change amends an architectural spec (docs/specs/**). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: CATASTROPHIC_QUANTIFIER heuristic has known bypass patterns (lazy quantifiers like `(a+?)+`, interstitial chars like `(\d+\.)+\d+`, and overlapping alternation like `(a|a)+` all evade the guard). Lowest-effort fix: add `?` to the character class (`/[*+}?]\)[*+{]/`). Risk is low given the LLM-generated rule source, data-service statement_timeout backstop, and best-effort execution context. - `src/llmo-referral-traffic-daily/classify.js:28`
- nit: `escapeCsvValue` does not sanitize formula-triggering prefixes (`=`, `+`, `-`, `@`). Not exploitable today (CSV is consumed exclusively by the projector import RPC, never human-opened), but worth noting if the CSV ever gains a user-facing download path. - `src/llmo-referral-traffic-daily/classify.js:96`
</details>

- Dedup ID source suffix: optel emit now includes `:optel` in the hash input, symmetric with CDN's `:cdn`.
- Duplicated serializer: `serializeClassificationCsv` and `CLASSIFICATION_CSV_COLUMNS` consolidated into shared `classify.js`, imported by both consumers.
- ReDoS risk: `CATASTROPHIC_QUANTIFIER` heuristic guard rejects nested/adjacent unbounded quantifiers before compilation, with dedicated tests covering skip-and-continue and null-return behaviors.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 3s | <b>Cost:</b> $6.12 | <b>Commit:</b> <code>337ba3a5ef1c323adce2539f9fa656954711e337</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @omairtemurian,

**Verdict: Approve** - no new blocking issues; the canonicalization commit is clean.
**Complexity: HIGH** - large diff (1467 lines across 11 files).
_Changes: Adds shared `canonicalizeUrlPath` to normalize url_path across optel and cdn producers, eliminating cross-source fragmentation in the referral_url_classifications join (11 files)._
Note: Recommend a human read before merge - this change amends an architectural spec (docs/specs/**). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `extractCountryCode(urlPath)` now receives canonicalized input (query stripped), making the two query-param country patterns (`query_country`, `query_locale` in `DEFAULT_COUNTRY_PATTERNS`) permanently unreachable for the optel source. Not a regression (RUM row.path never had query params in practice), but a one-line comment near the call would prevent future confusion. - `src/llmo-referral-traffic-daily/handler.js:87`
- suggestion: Add a test case for percent-encoded paths (e.g., `/shoes%2Fsneakers`) to document the "no decode" contract. The URL constructor may decode percent-encoded characters, and two paths differing only by `%2F` vs `/` will canonicalize differently. Pinning this in a test makes the decision explicit. - `test/audits/llmo-referral-traffic-daily/classify.test.js`
- suggestion: Add a test case for query-only or fragment-only input (e.g., `?utm=abc` or `#top`) to document that these return `/`. Messy log data could surface these forms. - `test/audits/llmo-referral-traffic-daily/classify.test.js`
</details>

- Dedup ID source suffix: optel emit now includes `:optel` in the hash input, symmetric with CDN's `:cdn`.
- Duplicated serializer: `serializeClassificationCsv` and `CLASSIFICATION_CSV_COLUMNS` consolidated into shared `classify.js`, imported by both consumers.
- ReDoS risk: `CATASTROPHIC_QUANTIFIER` heuristic guard rejects nested/adjacent unbounded quantifiers before compilation, with dedicated tests covering skip-and-continue and null-return behaviors.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 40s | <b>Cost:</b> $5.84 | <b>Commit:</b> <code>70d7d1267150467c8bacd3a4e81bdd881bdbfa61</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cwjwisse (changes requested):
## Review — LLMO-6257 P2 (audit-worker): generate + classify referral URL categories (optel + cdn)

Reviewed against the merged design-of-record (`mysticat-data-service docs/plans/2026-07-17-llmo-6257-referral-category-generation.md`) and the sibling PRs (data-service #836, projector #237). The classify/emit design is solid: `canonicalizeUrlPath` is applied consistently at each producer's single `url_path` derivation point so a source's traffic and classification CSVs agree; best-effort wrapping means a classify/emit failure never fails the already-succeeded daily traffic export; the CSV columns (`host,url_path,category_name,updated_by`) match the data-service import RPC byte-for-byte; and the shared FIFO group with source-namespaced dedup ids is intentional and correct given the idempotent, convergent upsert on the data-service side.

**Requesting changes** — one blocker, two majors, plus a couple of doc/robustness minors.

---

### Blocker

**B2 — create-if-missing rule generation calls an RPC that does not exist, so it is dead code in production.**
`generateReferralCategoryRules` → `fetchReferralTopUrls` → `POST /rpc/rpc_referral_traffic_top_urls` (`src/cdn-logs-report/utils/report-utils.js:170`, wired at `patterns-uploader.js:254`). That function is absent from data-service `db/schema.sql`, every migration, and the #836 branch. Because the call is wrapped best-effort in the runner, it 404s and is swallowed with a `warn` on every run — the flagship "create-if-missing" chunk never actually seeds rules for any site. The unit tests mock the RPC boundary, so this is invisible to CI.
This is the load-bearing gap for the whole feature: a never-served non-CDN site (ga4/aa/cja only, no rules yet) gets in-DB classification at import in #836, but only if rules exist — and rules are seeded only by this path. Please add `rpc_referral_traffic_top_urls` in data-service (see #836) or repoint to an existing read RPC, and add a test that does not mock the RPC away.

---

### Major

**M3 — the JS ReDoS guard has no SQL-side counterpart, producing silent cross-source category disagreement.**
`classify.js:28-40` treats any rule whose regex matches `CATASTROPHIC_QUANTIFIER` (`/[*+}]\)[*+{]/`, i.e. `(a+)+`, `(.*)+`, `(\w+){2,}`) as permanently unmatchable before compiling it. The data-service side (`_safe_regex_match`, `mysticat-data-service db/schema.sql:1628-1636`, used by both the P1 CDN backfill and the new DRS in-DB path) has no equivalent pre-check — it evaluates the pattern and relies on `statement_timeout`. So for any active rule of that shape, the write-time-in-service producers (optel + cdn) always skip it while the SQL-side sources can classify it — the same shared URL resolves to different categories depending on which source wrote it. Likelihood is low (LLM-generated rules are mostly simple, and `isValidGeneratedRegex` already filters some shapes) but it is real and unacknowledged.
Fix: add the identical shape-guard on the SQL side (a `_regex_is_safe()` companion used by both `wrpc_backfill_referral_categories` and `wrpc_apply_referral_categories`) so all sources agree, or explicitly document this as an accepted known divergence in `docs/specs/2026-07-21-referral-category-generation-p2.md`.

**M4 — the canonicalization change is a silent one-time discontinuity for already-persisted rows.**
The spec notes that before this PR optel wrote raw `row.path` (query kept) and cdn stripped only the query string; this PR moves both to the new canonical form (`handler.js:87`, `referral-daily-export.js:88`). Rows already in `referral_traffic_optel`/`referral_traffic_cdn` under the old form will now permanently mismatch newly-written canonical rows for the same logical page — fragmenting per-page aggregation across the cutover date and, worse, never joining `referral_url_classifications` (which only ever holds canonical paths), so pre-cutover traffic reads as permanently uncategorized. This is not called out anywhere as an accepted cost (contrast with how the P1 CDN-cutover cost is explicitly named).
Fix: name the pre/post-cutover `url_path` discontinuity as an accepted one-time cost in the spec, or scope a backfill.

---

### Should-verify
- Rule generation is wired only into the optel runner (`llmo-referral-traffic-daily`); cdn already has rules via the agentic Athena path. Please confirm the optel daily audit is actually scheduled for sites whose only traffic is ga4/adobe_analytics/cja — otherwise those sites never get rules generated at all, which compounds B2 and leaves the exact LLMO-6257 gap open. Also confirm the corpus fetch (once B2 is fixed) unions all of a site's sources, not just optel, per spec line 53.

### Minor
- Optel's classification cleanup delete is unguarded: `handler.js:252`'s `DeleteObjectCommand` runs in the `catch` before `throw err`; if the delete itself throws it masks the original SQS-dispatch error and leaks the CSV. cdn's `cleanupCsvFromS3` (`referral-daily-export.js:132-140`) already guards this. Bring optel up to the same standard, ideally by sharing one helper.
- `emitReferralClassifications` logs `fetchAgenticUrlClassificationRules` returning `{ error: true }` (DB fetch failure) identically to "genuinely zero rules" (`handler.js:205-210`, `referral-daily-export.js:162-167`). A real DB outage is indistinguishable from a legitimately rule-less site in the logs. Recommend a distinct warning (and test) when `rulesResult?.error` is true.
- The shared FIFO group deliberately diverges from the per-source group convention used for the traffic messages in the same files, and from the design-of-record's generic "(pipeline_id, site_id) serializes within a source but not across sources" claim (which silently doesn't hold when `pipeline_id` is shared). Worth a one-line note in the spec so a future reader doesn't assume the general rule applies.

---

_Reviewed via a fan-out of spec-compliance + audit-worker-specialist agents against the merged design-of-record; the missing-RPC finding verified directly against data-service schema.sql and both PR branches._

MysticatBot (approved):
Hey @omairtemurian,

**Verdict: Approve** - all three cwjwisse findings addressed by this commit are cleanly resolved; no new blocking issues.
**Complexity: HIGH** - large diff (1467 lines across 11 files).
_Changes: Guards rule-fetch errors with distinct warn logs in both handlers, wraps optel cleanup in try/catch, and documents the canonicalization cutover cost in the spec (5 files)._
Note: Recommend a human read before merge - this change amends an architectural spec (docs/specs/**). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: The two `rulesResult?.error` guards have slightly different comment styles (CDN one-liner vs optel two-liner with rationale). Cosmetic only, but the optel variant's extra sentence ("so a real outage isn't silently indistinguishable from 'no rules'") is useful context worth mirroring in the CDN guard for parity. - `src/cdn-logs-report/referral-daily-export.js:163`
</details>

- Canonicalization cutover cost (M4): Documented as an accepted, self-healing one-time cost in spec section 3, with backfill explicitly rejected.
- Rule-fetch error logging: Both handlers now return early on `rulesResult?.error` with a distinct warn-level log, differentiating a DB outage from a legitimately rule-less site.
- Optel cleanup unguarded: Cleanup DeleteObjectCommand now wrapped in try/catch so a delete failure never masks the original dispatch error (parity with CDN's `cleanupCsvFromS3`).

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 26s | <b>Cost:</b> $4.30 | <b>Commit:</b> <code>90370abb928075d90a262ecd772266cf06d68bf0</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cwjwisse (changes requested):
## Re-review (round 2) — audit-worker #2798 @ 90370abb9

Most of round 1 is resolved. The remaining items are the stale in-repo spec doc and a decision on the regex-parity edge case; the code itself is in good shape.

### Round-1 findings — resolved
- **B2 (blocker: rule-gen called a nonexistent RPC) — RESOLVED (contract-level).** data-service #836 now ships `rpc_referral_traffic_top_urls(p_site_id, p_limit) RETURNS (url_path, total_pageviews)`, and `fetchReferralTopUrls` (`src/cdn-logs-report/utils/report-utils.js:161-170`) calls it with exactly `{p_site_id, p_limit}` and reads `row.url_path` — exact match, pinned by `test/audits/cdn-logs-report/report-utils.test.js`. Operational note (not a code defect): #836 is still open — if this ships first, the RPC 404s but the best-effort wrapper in `referralTrafficDailyRunner` (`handler.js:298-302`) makes rule-gen a graceful no-op, no audit failure. Call out the merge-order dependency in the deploy plan.
- **M4 (canonicalization cutover orphans pre-cutover rows) — RESOLVED.** `90370abb9` adds the "Accepted one-time cutover cost" paragraph to the P2 spec (~line 57): pre-cutover rows read uncategorized until they age out of the query window, named as a self-healing one-time cost mirroring the P1 CDN cutover, with the backfill option explicitly considered and rejected. Accepted and scoped — good.
- **Minors — all RESOLVED.** Optel cleanup delete is now in its own try/catch so it can't mask the original dispatch error (`handler.js:258-264`, with a test for "rethrows the original dispatch error even when cleanup also fails"); rule-fetch `{error:true}` now logs a distinct warn separate from "no rules" in both producers (`handler.js:206-211`, `referral-daily-export.js:163-167`, tests added); shared-FIFO-group dedup-id namespacing confirmed intact.

### Remaining — should fix before merge
**Stale in-repo spec doc (`docs/specs/2026-07-21-referral-category-generation-p2.md`).** Lines ~15, 21, 31, 55, 71, 78 still describe GA4/AA/CJA classifying **in-DB at referral-traffic import time via `wrpc_apply_referral_categories`**, with chunk 5 marked "✅ done." That path was reverted in data-service #836 (`51087359a`) before this PR's latest commit — the RPC no longer exists, and DRS-source classification is not implemented anywhere. This doc is the authoritative cross-team reference (a future DRS engineer will read it), so shipping it as-is would tell them P2 is fully done for all five sources when three have no implementation. Please update the "as-built" section to the final all-sources write-time-in-service model and mark DRS classification as a separate/pending DRS-side deliverable, not "done."

**M3 (regex-guard parity) — PARTIAL; needs a decision.** The chunk-5 unwind shrank the blast radius: there is no longer a live in-DB DRS classifier, so `classify.js:28`'s `CATASTROPHIC_QUANTIFIER` guard now only disagrees with the SQL side for the **one-time, CDN-fenced P1 backfill** (`wrpc_backfill_referral_categories`, `_safe_regex_match`). `_safe_regex_match` catches only *uncompilable* regexes, not catastrophic-backtracking ones; Postgres's engine is more resistant than JS but not immune. So for a catastrophic-shaped active rule, a JS write-time classify (skips the rule) and a SQL backfill re-run (matches, or times out aborting the whole run) can still disagree. Narrow and low-likelihood, but unaddressed and undocumented (unlike M4). Pick one: (a) document it as an accepted narrow-scope (P1-backfill-only) divergence the way M4 was documented, or (b) close it at the source — reject catastrophic-quantifier regexes at rule-write time (`mergePatternRules` / rule validation) so no engine ever stores one, which eliminates the cross-engine parity question entirely. (b) is the cleaner fix.

### Follow-up (not blocking this PR)
- **C2 — cross-language canonicalization conformance fixtures (design-of-record hard gate, §65/105).** The JS side is well covered by `classify.js` unit tests, but the versioned cross-language contract + Python-side fixtures are still owed and become *more* necessary post-unwind, since DRS will emit into the same shared `wrpc_import_referral_url_classifications` CSV and its Python `url_path` derivation must match `canonicalizeUrlPath`'s output exactly. This lands with the DRS-side work; track it as the pre-go-live gate it is.
- **Chunk 6 (ongoing CDN write-time classification) is beyond the spec's literal Phase-2 scope** (spec lists CDN as P1-only one-time backfill; P2 non-CDN = optel + ga4/aa/cja). It plausibly closes a real gap (post-backfill CDN pages would otherwise never re-classify), and the in-repo spec explains the rationale — but the design-of-record wasn't amended to bless it. Worth an explicit confirmation and a one-line addition to the merged plan.

**Verdict: changes requested** — sync the stale spec doc and resolve M3 (document or fix-at-source) before merge; C2 + chunk-6 sign-off as tracked follow-ups.

_Re-reviewed via spec-compliance + audit-worker-specialist agents at HEAD 90370abb9, cross-checked against data-service #836 @ 0b71d130e._

MysticatBot (approved):
Hey @omairtemurian,

**Verdict: Approve** - no new blocking issues; cwjwisse's round-2 findings are cleanly resolved.
**Complexity: HIGH** - large diff (1467 lines across 11 files).
_Changes: Rejects catastrophic-backtracking regex shapes at rule-write time (closing M3 at the source) and syncs the in-repo spec to reflect the DRS unwind (5 files)._
Note: Recommend a human read before merge - this change amends an architectural spec (docs/specs/**). The bot review is a complement to, not a replacement for, a human read here.
Note: CI build is pending on the latest push.

<details>

- nit: `compileRuleRegex` now calls `String(regex).replace(/^\(\?i\)/, '')` twice - once inside `isCatastrophicQuantifier` and once on the return line. A local `const source = ...` before both would make the single-source-of-truth coupling explicit and prevent future divergence. Functionally harmless today. - `src/llmo-referral-traffic-daily/classify.js:38`
</details>

- Stale in-repo spec doc: DRS sections updated from "done (in-DB classify)" to "deferred (write-time-in-service on DRS side, not in these PRs)".
- M3 regex-guard parity: closed at the source - `isCatastrophicQuantifier` exported from `classify.js` and called at rule-write time in `isValidGeneratedRegex`, so catastrophic shapes are never stored. Option (b) per cwjwisse's round-2 guidance.
- Canonicalization cutover cost (M4): documented as accepted, self-healing one-time cost.
- Rule-fetch error logging: both handlers distinguish DB outage from legitimately rule-less site.
- Optel cleanup unguarded: now wrapped in try/catch matching CDN's pattern.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 5s | <b>Cost:</b> $3.96 | <b>Commit:</b> <code>3a59cc1c982e5e2a68f189fde83f7b713edcac3f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

cwjwisse (approved):
## Re-review (round 3) — audit-worker #2798 @ 3a59cc1c9

Both remaining items resolved, and via the cleaner fix. Approving.

- **M3 (JS/SQL regex-guard parity) — RESOLVED at the source.** `isCatastrophicQuantifier` is now exported from `classify.js` and wired into `isValidGeneratedRegex` (`patterns-uploader.js`), so a catastrophic-backtracking (ReDoS) rule shape is **rejected at rule-write time** and never stored. Since no engine ever sees such a rule, the JS-skip vs SQL-evaluate divergence can't arise — this closes the parity gap at the source rather than papering over it, which is the better of the two options. `compileRuleRegex` still calls the shared helper defensively at match time. Tests added on both sides (`patterns-uploader.test.js`, `classify.test.js`).
- **Stale spec doc — RESOLVED.** `docs/specs/2026-07-21-referral-category-generation-p2.md` is fully synced to the as-built state: all sources classify write-time-in-service; the DRS in-DB `wrpc_apply_referral_categories` attempt is marked rejected/unwound; chunk 5 is ⏳ deferred to the DRS side (Python, reading rules via a spacecat-api proxy); the topology table, decisions, and chunk list all match. The canonicalization §3 now correctly frames JS↔Python parity as a **real pre-go-live gate** (no longer "moot") owned by the DRS-side work, and the open-items section explicitly names the create-if-missing gap for ga4/aa/cja-only sites as part of that deliverable.

Round-1/2 items stay resolved: B2 (corpus RPC now exists and the contract matches), M4 (accepted cutover-cost doc), and the three minors (optel cleanup guard, rule-fetch-error log, FIFO dedup). CI is green (build, branch-deploy, codecov, pr-review all SUCCESS).

Outstanding LLMO-6257 work is all DRS-side and correctly out of scope for this PR: the Python write-time classifier + spacecat-api rules proxy, the Python side of the canonicalization conformance fixtures, and rule-gen coverage for ga4/aa/cja-only sites. All are now documented as deferred rather than silently missing.

**Verdict: approve.** Merge-order reminder still stands — deploy data-service #836 before the audit-worker producer so there's no PGRST202 retry/DLQ churn during the window.

_Re-reviewed at HEAD 3a59cc1c9; the fix commit diffed line-by-line and CI confirmed green._

PR #2800, opened 2026-07-22
fix(cwv): only outdate suggestions whose page was observed this run (SITES-48436)

tkotthakota-adobe:
## Problem
[SITES-48436](https://jira.corp.adobe.com/browse/SITES-48436) — on bmw.fr, CWV suggestions with clearly-failing cached metrics (CLS up to 1.26, INP up to 900ms) were marked `OUTDATED`, hiding real, actionable CWV issues from the ASO UI.

Root cause: the CWV audit marks a suggestion `OUTDATED` purely because its URL is **absent** from the current run's failing set (`handleOutdatedSuggestions` keys on URL-absence). But a URL can drop out of a run for reasons that have nothing to do with being fixed:
- the HEAD probe was blocked/timed out (bmw.fr is behind a WAF),
- the page fell out of the top-15 / ≥7000-pv/week selection, or
- RUM returned sparse/no data for it.

The audit treated all of these as "resolved" and aged the suggestion out, even though the issue is still live. This is a latent defect for any WAF/CDN-protected or long-tail site, not just bmw.fr.

## Fix
`syncOpportunitiesAndSuggestions` now builds a `scrapedUrlsSet` of the URLs actually observed in this run's RUM data (`auditResult.cwv`, before the failing-metrics filter) and passes it to `syncSuggestions`. This activates the existing "only outdate URLs actually observed this run" guard in `handleOutdatedSuggestions` — the same guard prerender/llm-error-pages already use.

Result:
- page **observed this run and now passing** → `OUTDATED` ✓ (genuinely resolved)
- page **not observed this run** (WAF/HEAD-fail, dropped from selection, sparse RUM) → **preserved** ✓ (absence is not evidence it's fixed)

No shared-code change — `data-access.js` is untouched; only the CWV sync opts into the guard. Group/pattern suggestions have no scraped-URL identity and are unaffected.

## Trade-off
"Only outdate when proven fixed" also means a genuinely-removed page is no longer auto-outdated — it's preserved until a run observes it passing. This is the conservative direction the ticket wants (never hide a live issue). Age-out of truly-removed pages, if needed, is a separate last-seen/retention follow-up.

## Tests
- Regression (red→green): still-failing page absent from the run is **not** OUTDATED.
- Positive: page observed this run and now passing **is** OUTDATED.
- Updated the existing `NON_EXISTING_URL` test, which had asserted the buggy behavior.
- 47 passing, 100% coverage on `opportunity-sync.js`, lint clean.

## Recovery (not in this PR)
bmw.fr's already-stuck rows are being recovered separately (moved to `pending_validation`; the validator re-promotes still-failing ones to `NEW`). This PR prevents recurrence when the CWV audit is re-run.

### DIFF
--- a/src/cwv/opportunity-sync.js
+++ b/src/cwv/opportunity-sync.js
@@ -169,6 +169,20 @@ export async function syncOpportunitiesAndSuggestions(context) {
     .map(filterToFailingDeviceMetrics);
   log.info(`[syncOpportunitiesAndSuggestions] site ${site.getId()} - ${cwvData.length} of ${auditResult.cwv.length} CWV entries have failing metrics`);
 
+  // Set of page URLs actually observed in THIS run's RUM data (the full reported
+  // set, before the failing-metrics filter). Passed to syncSuggestions so a
+  // suggestion is only aged out to OUTDATED when its page was measured this run
+  // and dropped from the failing set because it now passes. A URL absent from
+  // this set — bot/WAF-blocked HEAD, dropped from the top-N/threshold selection,
+  // or sparse/empty RUM — is NOT evidence the issue is resolved, so it must not
+  // be marked OUTDATED (SITES-48436). Group/pattern rows carry no scraped-URL
+  // identity and are unaffected by this guard.
+  const scrapedUrlsSet = new Set(
+    auditResult.cwv
+      .filter((entry) => entry.type === 'url')
+      .map((entry) => entry.url),
+  );
+
   // Build minimal audit data object for opportunity creation
   const auditData = {
     siteId: site.getId(),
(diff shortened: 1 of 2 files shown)


ramboz (approved):
Looks good to me, but I only see us writing the `scrapedUrlsSet`, not reading from it anywhere. I imagine this is done later in mystique based on the `auditData`?

tkotthakota-adobe:
@ramboz good catch on it not being visible — it *is* read, just in existing (unchanged) code that doesn't show in the diff, so you have to follow the call chain. It's consumed entirely within this worker invocation, not in mystique.

`opportunity-sync.js` passes `scrapedUrlsSet` into `syncSuggestions` (`src/utils/data-access.js:435`), which forwards it to `handleOutdatedSuggestions` (`:472`), where the filter at `src/utils/data-access.js:294-301` only marks a suggestion OUTDATED if its URL is in the set:

```js
.filter((existing) => {
  if (scrapedUrlsSet) {
    const suggestionUrl = existing.getData()?.url;
    return suggestionUrl && scrapedUrlsSet.has(suggestionUrl);
  }
  return true;
});
```

It's a runtime param, not persisted to `auditData` and not read by mystique. Before this change CWV called `syncSuggestions` with no `scrapedUrlsSet`, so that filter defaulted to `return true` — outdating every absent URL, which is the bug. This is the same guard prerender/llm-error-pages already use; CWV just wasn't opting in.

PR #2801, opened 2026-07-22
feat(offsite): self-heal empty URL store on individual analysis runs

AndreiAlexandruParaschiv:
## Problem

Running an offsite analysis individually — `@spacecat run audit <site> reddit-analysis` (or youtube/cited) — **fails silently** when the URL store has no URLs for that bucket. The store is normally populated by `offsite-brand-presence`, so a standalone run hits `StoreEmptyError` and returns `success:false` with **no Slack signal** (observed on `lot.com`: the analysis card posts, then nothing).

There are two distinct empty states, handled very differently before this PR:

| State | Meaning | Before |
|---|---|---|
| `DrsNoContentAvailableError` | URLs in store, not scraped yet | **self-heals** via `requestOffsiteScrape` |
| `StoreEmptyError` | **zero URLs** in store | **fails silently** |

## Fix

`StoreEmptyError` now mirrors the existing self-heal: it triggers a **domain-scoped** `offsite-brand-presence` run (`requestOffsiteScrape` with `messageData.domainScope`) that collects + scrapes **only that bucket** — `reddit.com` / `youtube.com` / `top-cited` — and re-triggers the analysis when DRS completes. **No need to run `offsite-brand-presence` for all three buckets manually.**

- Posts a Slack note: `:mag: … no stored URLs yet; collecting & scraping Reddit URLs first, will retry automatically.`
- Loop guard: if the run is already a re-triggered one (`drsScrapeRequested`) and the store is *still* empty → posts a terminal `:warning: … no Reddit URLs found to analyze.` and stops (no infinite scrape→analyze loop).

This reuses the exact mechanism the `DrsNoContentAvailableError` path already uses, so the two empty states are now consistent.

## Tests

- Unit tests only. Per handler: replaced the old "returns error on empty store" test with a first-run self-heal test (asserts scoped `offsite-brand-presence` dispatch + Slack note) and added a terminal drsScrapeRequested case. All three `handler.js` at **100%** coverage; 124 passing; lint clean.
- (Full local `npm test` still aborts on the unrelated pre-existing `filterBySiteScope` import error in `src/prerender/handler.js` — not touched here.)

### DIFF
--- a/src/cited-analysis/handler.js
+++ b/src/cited-analysis/handler.js
@@ -306,13 +306,37 @@ async function runCitedAnalysisAudit(url, context, site, auditContext = {}) {
     };
   } catch (error) {
     if (error instanceof StoreEmptyError) {
-      log.error(`${LOG_PREFIX} Store data missing: ${error.message}`);
+      const { slackContext } = auditContext;
+      const { channelId, threadTs } = slackContext || {};
+      // A scoped scrape already ran and the store is STILL empty → the brand has no
+      // cited URLs to analyze. Report a terminal message instead of looping.
+      if (auditContext.drsScrapeRequested) {
+        log.error(`${LOG_PREFIX} URL store still empty after scrape: ${error.message}`);
+        await postMessageOptional(
+          context,
+          channelId,
+          `:warning: *cited-analysis* for *${site.getBaseURL()}* — no cited URLs found to analyze.`,
+          { threadTs },
+        );
+        return {
+          auditResult: { success: false, error: error.message, storeName: error.storeName },
+          fullAuditRef: url,
+        };
+      }
+      // First individual run with an empty store: collect + scrape just this bucket via a
+      // domain-scoped offsite-brand-presence run, which re-triggers this analysis when DRS
+      // completes — no need to run offsite-brand-presence for all buckets manually.
+      log.info(`${LOG_PREFIX} URL store empty, requesting a scoped scrape for top-cited`);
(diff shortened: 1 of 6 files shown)


MysticatBot (approved):

**Verdict: Approve** - clean pattern extension with solid loop guard; one minor test gap.
**Complexity: MEDIUM** - medium diff; tests-heavy signal.
_Changes: Adds self-heal logic to the StoreEmptyError catch in all three offsite analysis handlers (cited, reddit, youtube), mirroring the existing DrsNoContentAvailableError pattern with Slack notifications and a loop guard (6 files)._

<details>

- nit: First-run tests do not assert that `slackContext` is forwarded in the SQS message to `requestOffsiteScrape` - if thread-threading ever regresses, these tests will not catch it - `test/audits/cited-analysis/handler.test.js:357` (same gap in reddit and youtube test files)
- suggestion: Extract the duplicated StoreEmptyError handling (~25 lines, identical logic across three handlers) into a shared utility alongside the existing DrsNoContentAvailableError pattern - reduces drift surface for future changes
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 4s | <b>Cost:</b> $9.56 | <b>Commit:</b> <code>79e6597ade52328dfd7669aa3d895d39cd47a752</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2803, opened 2026-07-23
fix(sync): LLMO-6483 preserve manually-edited suggestions during re-audit

baranovskyalexandr:
## Summary

When a customer edits a suggestion via the PATCH API, `updatedBy` is stamped with their email. `syncSuggestions` now treats such suggestions as frozen — skipping data merge, OUTDATED transitions, and reconcile-disappeared processing — preventing subsequent audit runs from overwriting customer edits.

### What changed

- **`src/utils/data-access.js`**
  - New shared helper `isManuallyEditedSuggestion` — true when `updatedBy` is set and ≠ `'system'`
  - **syncSuggestions matched-key path**: skips data merge for manually-edited suggestions
  - **handleOutdatedSuggestions**: excludes manually-edited suggestions from OUTDATED transition
  - **reconcileDisappearedSuggestions**: excludes manually-edited suggestions from FIXED transition (prevents `updatedBy='system'` from erasing the manual-edit marker)
  - JSDoc disambiguates entity-level `updatedBy` from the alt-text domain-specific `data.recommendations[].isManuallyEdited` flag

- **`src/experimentation-opportunities/guidance-high-organic-low-ctr-handler.js`**
  - `hasManuallyModifiedSuggestions` now delegates to the shared `isManuallyEditedSuggestion` helper (removes duplicate predicate)

- **`test/utils/data-access.test.js`** — 5 new tests:
  - manually-edited NEW suggestion → data merge skipped
  - manually-edited APPROVED suggestion → data merge skipped
  - manually-edited IN\_PROGRESS suggestion → data merge skipped
  - `updatedBy: 'system'` → updated as usual
  - manually-edited suggestion whose key disappears → NOT marked OUTDATED
  - PII regression: customer email must not appear in any log call

### How it works

| `updatedBy` value | Behavior |
|---|---|
| `null` / `undefined` | Normal merge (backward-compatible) |
| `'system'` | Normal merge (set by audit runs) |
| `'user@company.com'` | **Frozen** — data, status, and rank preserved across all three mutation paths |

## Related Issues

- [LLMO-6483](https://jira.corp.adobe.com/browse/LLMO-6483) — Manually edited suggestions should not be overwritten by subsequent audit runs

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Duy Nguyen", "Valerii Naida", "Dominique Jäggi"]
```

### DIFF
--- a/src/experimentation-opportunities/guidance-high-organic-low-ctr-handler.js
+++ b/src/experimentation-opportunities/guidance-high-organic-low-ctr-handler.js
@@ -14,20 +14,12 @@ import { notFound, ok } from '@adobe/spacecat-shared-http-utils';
 import { Suggestion as SuggestionModel } from '@adobe/spacecat-shared-data-access';
 import { convertToOpportunityEntity } from './opportunity-data-mapper.js';
 import { HIGH_ORGANIC_LOW_CTR_OPPTY_TYPE } from './handler.js';
-import { warnOnInvalidSuggestionData } from '../utils/data-access.js';
+import { isManuallyEditedSuggestion, warnOnInvalidSuggestionData } from '../utils/data-access.js';
 
 const MAX_HIGH_ORGANIC_LOW_CTR_OPPORTUNITIES = 3;
 
-/**
- * Checks if any suggestions in the array were manually modified (updatedBy !== 'system')
- * @param {Array} suggestions - Array of suggestion objects
- * @returns {boolean} - True if any suggestion was manually modified
- */
 function hasManuallyModifiedSuggestions(suggestions) {
-  return suggestions.some((suggestion) => {
-    const suggestionUpdatedBy = suggestion.getUpdatedBy();
-    return suggestionUpdatedBy && suggestionUpdatedBy !== 'system';
-  });
+  return suggestions.some(isManuallyEditedSuggestion);
 }
 
 /**
--- a/src/utils/data-access.js
+++ b/src/utils/data-access.js
@@ -36,9 +36,12 @@ export const AUTHOR_ONLY_OPPORTUNITY_TYPES = [
 
 /**
  * True when a suggestion was last touched by a non-system actor (typically a
- * customer email stamped by the API). Used to skip regenerating/overwriting
+ * customer email stamped by the PATCH API). Used to skip regenerating/overwriting
  * that suggestion's data on subsequent audits (LLMO-6483).
- * Applies regardless of suggestion status (NEW, APPROVED/deployed, etc.).
+ *
+ * This checks the entity-level `updatedBy` field — distinct from the alt-text
+ * domain-specific `data.recommendations[].isManuallyEdited` flag used in
+ * image-alt-text handlers.
  *
  * @param {Object} suggestion - Suggestion entity (or mock).
  * @returns {boolean}
@@ -428,9 +431,11 @@ export const isTBYBSite = checkIsTBYBSite;
  * Handles outdated suggestions by updating their status, either to OUTDATED or the provided one.
  * Updates existing suggestions with new data if they match based on the provided key.
  * For REJECTED suggestions that appear again, preserves REJECTED status.
- * Suggestions with updatedBy set to a non-system actor (customer edits) are not
- * regenerated on re-audit — data/status/rank are left unchanged. Applies to both
- * NEW and deployed (e.g. APPROVED / IN_PROGRESS) suggestions (LLMO-6483).
+ * Suggestions with updatedBy set to a non-system actor (customer edits) are
+ * protected from data overwrites when they are still present in the current
+ * audit data (matched-key path). They are also excluded from the OUTDATED and
+ * reconcile-disappeared paths. Note: this is distinct from the alt-text
+ * domain-specific `isManuallyEdited` data flag (LLMO-6483).
  *
  * Prepares new suggestions from the new data and adds them to the opportunity.
  * Maps new data to suggestion objects using the provided mapping function.
@@ -710,6 +715,9 @@ export async function reconcileDisappearedSuggestions({
     // (customer explicitly picked a redirect target, so a live match is
     // high-confidence attribution — see function docstring).
     const candidates = disappearedSuggestions.filter((s) => {
+      if (isManuallyEditedSuggestion(s)) {
+        return false;
+      }
       const status = s?.getStatus?.();
       if (newStatus && status === newStatus) {
         return true;
--- a/src/utils/data-access.js
+++ b/src/utils/data-access.js
@@ -312,7 +312,8 @@ export const handleOutdatedSuggestions = async ({
         return suggestionUrl && scrapedUrlsSet.has(suggestionUrl);
       }
       return true;
-    });
+    })
+    .filter((existing) => !isManuallyEditedSuggestion(existing));
 
   // prevents JSON.stringify overflow
   log.info(`[SuggestionSync] Final count of suggestions to mark as ${statusToSetForOutdated}: ${existingOutdatedSuggestions.length}`);
--- a/src/utils/data-access.js
+++ b/src/utils/data-access.js
@@ -35,6 +35,20 @@ export const AUTHOR_ONLY_OPPORTUNITY_TYPES = [
 ];
 
 /**
+ * True when a suggestion was last touched by a non-system actor (typically a
+ * customer email stamped by the API). Used to skip regenerating/overwriting
+ * that suggestion's data on subsequent audits (LLMO-6483).
+ * Applies regardless of suggestion status (NEW, APPROVED/deployed, etc.).
+ *
+ * @param {Object} suggestion - Suggestion entity (or mock).
+ * @returns {boolean}
+ */
+export function isManuallyEditedSuggestion(suggestion) {
+  const updatedBy = suggestion?.getUpdatedBy?.();
+  return Boolean(updatedBy && updatedBy !== 'system');
+}
+
+/**
  * Validates suggestion data against the Joi schema for the given opportunity type.
  * Logs a warning if validation fails but does not block the operation.
  *
@@ -412,7 +426,10 @@ export const isTBYBSite = checkIsTBYBSite;
  * Synchronizes existing suggestions with new data.
  * Handles outdated suggestions by updating their status, either to OUTDATED or the provided one.
  * Updates existing suggestions with new data if they match based on the provided key.
- * For REJECTED suggestions that appear again, preserves REJECTED status
+ * For REJECTED suggestions that appear again, preserves REJECTED status.
+ * Suggestions with updatedBy set to a non-system actor (customer edits) are not
+ * regenerated on re-audit — data/status/rank are left unchanged. Applies to both
+ * NEW and deployed (e.g. APPROVED / IN_PROGRESS) suggestions (LLMO-6483).
  *
  * Prepares new suggestions from the new data and adds them to the opportunity.
  * Maps new data to suggestion objects using the provided mapping function.
@@ -526,11 +543,11 @@ export async function syncSuggestions({
       return;
     }
 
-    // Suggestions manually edited by a customer (updatedBy is set to their
-    // email, not 'system') must not have their data overwritten by a re-audit.
-    const updatedBy = existing.getUpdatedBy?.();
-    if (updatedBy && updatedBy !== 'system') {
-      log.debug(`Skipping manually-edited suggestion ${existingKey} (updatedBy: ${updatedBy})`);
+    // Do not regenerate a customer-edited suggestion on re-audit (LLMO-6483).
+    // Applies to NEW and deployed statuses alike. updatedBy may be a user email
+    // — never log it (PII).
+    if (isManuallyEditedSuggestion(existing)) {
+      log.debug(`Skipping manually-edited suggestion ${existingKey}`);
       return;
     }
 
--- a/src/utils/data-access.js
+++ b/src/utils/data-access.js
@@ -526,6 +526,14 @@ export async function syncSuggestions({
       return;
     }
 
+    // Suggestions manually edited by a customer (updatedBy is set to their
+    // email, not 'system') must not have their data overwritten by a re-audit.
+    const updatedBy = existing.getUpdatedBy?.();
+    if (updatedBy && updatedBy !== 'system') {
+      log.debug(`Skipping manually-edited suggestion ${existingKey} (updatedBy: ${updatedBy})`);
+      return;
+    }
+
     const newDataItem = newDataByKey.get(existingKey);
     // mergeDataFunction must return a new object (not mutate existingData)
     // for deepEqual comparison to work correctly
--- a/test/utils/data-access.test.js
+++ b/test/utils/data-access.test.js
@@ -1093,8 +1093,14 @@ describe('data-access', () => {
         (msg) => /Skipping manually-edited suggestion/.test(msg),
       );
       expect(skipMsg).to.be.a('string');
-      // Must not log the customer email (PII)
-      expect(skipMsg).to.not.include('customer@example.com');
+      // Must not log the customer email (PII) in ANY log call
+      const allLogMessages = [
+        ...mockLogger.debug.args,
+        ...mockLogger.info.args,
+        ...mockLogger.warn.args,
+      ].flat().filter((a) => typeof a === 'string');
+      const piiLeak = allLogMessages.find((msg) => msg.includes('customer@example.com'));
+      expect(piiLeak, 'customer email must not appear in any log message').to.be.undefined;
     });
 
     it('should not overwrite a manually-edited deployed (APPROVED) suggestion', async () => {
--- a/test/utils/data-access.test.js
+++ b/test/utils/data-access.test.js
@@ -1053,6 +1053,82 @@ describe('data-access', () => {
       );
     });
 
+    it('should not overwrite a manually-edited suggestion (updatedBy is a user email)', async () => {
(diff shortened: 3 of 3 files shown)


naydav:
# Review: preserve manually-edited suggestions during re-audit (LLMO-6483)

## Summary
Small, well-scoped, well-tested change. The centralized guard in `syncSuggestions` correctly fixes the primary reported problem — customer edits being overwritten on re-audit — for every opportunity type that flows through `syncSuggestions`. Core logic is sound and CI is green. No blockers; the notable items are a coverage gap on the *outdated* path and a ticket-number typo in the PR title.

## Strengths
- **Right layer for the fix.** Both pre-existing mechanisms (alt-text's `isManuallyEdited` data flag, experimentation's page-level `updatedBy !== 'system'` check) genuinely bypass `syncSuggestions`, so summarization and others were unprotected. Centralizing here is the correct call.
- **The core invariant checks out.** On **Suggestion entities**, the only `updatedBy` value the audit worker ever writes is `'system'`. The other non-`'system'` values in the repo (`'drs'` in `drs-config-writer.js`, `'audit-worker:...'` in `patterns-uploader.js`) are on config JSON / DB pattern rows, not Suggestion entities, so they don't reach this guard.
- **PII-awareness is excellent.** The new log line logs only `existingKey`, never `updatedBy`, with an explicit comment — and a test asserts the email never appears in the skip message.
- **Backward compatible.** `null`/`undefined` and `'system'` both merge normally; legacy and system-created suggestions are unaffected.
- **Test quality is high.** Covers NEW / APPROVED / IN_PROGRESS plus the two negative cases, following the file's existing sandbox conventions. `codecov/patch` passes.

## Should Fix

**1. The `OUTDATED` path bypasses the manual-edit guard** — `src/utils/data-access.js:489` / `:275-283`
`handleOutdatedSuggestions` runs *before* the new guard and ages out any existing suggestion whose key isn't in the new audit data. Its status exclusion list (`[OUTDATED, FIXED, ERROR, SKIPPED, REJECTED, APPROVED, IN_PROGRESS]`) does **not** exclude `NEW` or `PENDING_VALIDATION`. So a customer edits a `NEW` suggestion (`updatedBy = email`, still `NEW`), the page later drops out of the audit set (common for top-N-pages audits where the page set churns), and the edited suggestion flips to `OUTDATED` — a status/`updatedAt` change on a customer edit. This partially contradicts the stated goal. Notably, the alt-text handler already guards against exactly this (`src/image-alt-text/handler.js:436` — "Never mark manually edited suggestions as outdated"), so the centralized fix is currently *less* protective than the per-handler code it supersedes.

Please confirm whether this is intended. If `OUTDATED` is considered correct when a page truly leaves scope, a one-line comment noting the intentional gap is enough. Otherwise, add the same predicate to the outdated filter:
```js
.filter((existing) => !isManuallyEditedSuggestion(existing))
```

**2. Ticket-number typo in the PR title** — traceability
The title reads `LLMO-5483`, but the branch, body, commits, and code comments all say `LLMO-6483`. Since the repo squash-merges (subject defaults to the PR title), the wrong ticket would land in the release notes — which the PR checklist explicitly calls out. Please fix the title before merge.

## Nice to Have

**3. Deduplicate the manual-edit predicate** — `src/experimentation-opportunities/guidance-high-organic-low-ctr-handler.js:26-31`
`hasManuallyModifiedSuggestions` reimplements the identical predicate. It operates at a different granularity (skip the whole opportunity if *any* suggestion is edited), so it can't be replaced outright, but its inner test could call the new exported `isManuallyEditedSuggestion` so the definition lives in one place. Follow-up, not blocking.

**4. Strengthen the PII regression guard** — `test/utils/data-access.test.js:1088`
The test asserts the *skip message* doesn't contain the email — but that message only ever contained `existingKey`, so it's close to tautological. Asserting the email appears in *no* log call is stronger and would also surface #5:
```js
const allLogArgs = ['debug', 'info', 'warn', 'error']
  .flatMap((l) => mockLogger[l].args.flat());
expect(allLogArgs.join(' ')).to.not.include('customer@example.com');
```

**5. Verify the adjacent debug log doesn't already serialize the email** — `src/utils/data-access.js:498`
`log.debug(\`Existing suggestions = ...: ${safeStringify(existingSuggestions)}\`)` serializes full suggestion entities at debug level. If the entity serialization includes `updatedBy`, that line already leaks the customer email — independent of this PR. Pre-existing and out of scope, but worth a quick check given the PR's care with PII (the test in #4 would catch it).

## Recommendation
The design rests on one cross-repo invariant: **audit-created suggestions have `updatedBy ∈ {null, 'system'}`, and only the API sets an email.** I confirmed the audit-worker side; the Suggestion model's creation *default* lives in `@adobe/spacecat-shared-data-access`. If that model ever defaults `updatedBy` to a creating user/service account, this guard would silently freeze *every* suggestion after first creation. The existing experimentation-handler convention strongly suggests the invariant holds, but a one-line confirmation against the model schema (and ideally an integration test at the data-access boundary) would de-risk it.

## Verdict
**Ready to merge:** With fixes — fix the PR-title ticket number and make an explicit decision on the `OUTDATED`-path gap (#1); the rest are follow-ups.

naydav:
# Review: PR #2803 — preserve manually-edited suggestions during re-audit (LLMO-6483)

## Architecture Review

### Summary of the change

The PR adds an exported predicate `isManuallyEditedSuggestion(suggestion)` and a guard inside the matched-suggestion loop of `syncSuggestions` (`src/utils/data-access.js`). Any suggestion whose `updatedBy` is truthy and not `'system'` is skipped during the merge. Placing this at the shared `syncSuggestions` layer is the correct instinct — it is the common synchronization choke point most opportunity handlers route through, so a cross-cutting "don't clobber customer edits" invariant belongs there rather than being copied into each handler.

### Strengths (genuine)

- **Right choke point for the matched-update case.** Centralizing in `syncSuggestions` rather than per-handler is the sound architectural choice, and the helper is pure, exported, and independently testable.
- **Backward-compatible and low-coupling.** `null`/`undefined`/`'system'` all fall through to normal merge, so existing suggestions and audit-written ones are unaffected. The guard reads through the entity's public accessor (`getUpdatedBy?.()`) rather than reaching into internals.
- **PII-aware boundary.** The skip log deliberately omits the `updatedBy` value (which may be a customer email), and a test asserts this — good discipline for a log emitted from a shared utility.

### 🟠 Major

**1. Protection is asymmetric across the suggestion-mutation paths in this module** — `src/utils/data-access.js:549` (guard) vs. `:275-283` (`handleOutdatedSuggestions` status filter) and `:776-780` (`reconcileDisappearedSuggestions`)

The guard lives only in the *matched* path (suggestion key present in the new audit data). Two other mutation paths in the same module bypass it:

- **`handleOutdatedSuggestions`** runs first (`:489`) and ages out any suggestion whose key is *absent* from the new data. Its exclusion list (`:275-283`) covers APPROVED/IN_PROGRESS/SKIPPED/REJECTED/etc. but not `updatedBy`. So a manually-edited **NEW/PENDING_VALIDATION** suggestion whose key drops out of an audit run (page falls off top-pages, key normalization shifts, issue transiently not re-detected) is still flipped to OUTDATED — an overwrite of the customer's state, which is exactly what the PR's stated invariant ("should not be overwritten by subsequent audit runs") forbids.
- **`reconcileDisappearedSuggestions`** (`:776-780`, reached via `syncSuggestionsWithPublishDetection`) flips disappeared NEW suggestions to FIXED and stamps `updatedBy = 'system'`, erasing the manual-edit marker.

This is precisely the gap the alt-text handler already patched in *its own* outdated path (`src/image-alt-text/handler.js:437`, `guidance-missing-alt-text-handler.js:96`), which is direct evidence that the disappeared/outdated path is a real case the team has needed to cover. As written, "protects all opportunity types" holds only for suggestions the audit re-detects each run.

*Suggested fix:* either extend the same predicate to `handleOutdatedSuggestions` (add an `isManuallyEditedSuggestion` clause to the filter chain at `:284-307`) and to the reconcile candidate filter, or explicitly document that the guard's scope is intentionally limited to matched re-generation and that disappearance-driven status changes are out of scope. Deployed statuses (APPROVED/IN_PROGRESS) are already safe via the outdated exclusion list, so the concrete exposure is NEW/PENDING edited suggestions — decide consciously rather than leaving it implicit.

### 🟡 Minor

**1. Open-world predicate couples correctness to an unenforced convention** — `src/utils/data-access.js:46-49`

`updatedBy && updatedBy !== 'system'` classifies *every* non-`'system'` value as a customer edit. The real contract — "audit-owned suggestions carry `updatedBy === 'system'`" — is an implicit convention shared with the API service (which stamps emails) and the data-access layer, and it is not owned or expressed anywhere as a shared constant. The repo already stamps non-`'system'` actor strings on *other* entities (`'drs'`, `'audit-worker:agentic-patterns'` in `drs-config-writer.js` / `patterns-uploader.js`); the day any automated producer writes a non-`'system'` actor onto a Suggestion that flows through `syncSuggestions`, those suggestions freeze permanently with no un-freeze path. An allow-list of known system actors (or at minimum a named constant with a comment pinning the contract) is more robust than a deny-everything-but-`'system'` check.

*Suggested fix:* invert to `isSystemActor(updatedBy)` backed by a small known-actor set, or introduce a shared `SYSTEM_ACTOR = 'system'` constant and document, next to the helper, that the API/data-access layer owns the "email == human edit" convention this depends on.

**2. Two divergent "manually edited" signals now coexist unreconciled** — `src/utils/data-access.js:46` (`isManuallyEditedSuggestion`, `updatedBy`-based) vs. `src/image-alt-text/handler.js:437` (`data.recommendations[0].isManuallyEdited`, domain-flag-based)

After this PR, "manually edited" means two structurally different things depending on the module: an entity-level `updatedBy` field (new, alt-text-agnostic) and an alt-text domain flag buried in suggestion data. They live in different mutation paths and aren't linked. This is a cohesion/naming hazard — a future maintainer reading `isManuallyEditedSuggestion` will reasonably assume it is the single source of truth for "manually edited," which it is not. Not a defect in this PR's behavior, but worth an explicit note that the two mechanisms are intentionally distinct (or a plan to converge them).

*Suggested fix:* add a comment on the helper distinguishing it from the alt-text `isManuallyEdited` flag, or track convergence so the two signals don't drift further.

**3. Missed consolidation of the pre-existing duplicate predicate** — `src/experimentation-opportunities/guidance-high-organic-low-ctr-handler.js:26-31`

`hasManuallyModifiedSuggestions` is now a line-for-line duplicate of the logic in the new shared helper (`updatedBy && updatedBy !== 'system'`). The PR introduces the helper precisely to centralize this predicate but leaves the one existing duplicate untouched. The experimentation flow is page-level delete/recreate so it can't use the `syncSuggestions` guard, but it can reuse the predicate: `suggestions.some(isManuallyEditedSuggestion)`.

*Suggested fix:* import and reuse `isManuallyEditedSuggestion` in the experimentation handler to collapse the duplication.

### 🟢 Nice-to-have

**1. Non-overridable policy sits inside an otherwise heavily-parameterized function** — `src/utils/data-access.js:549` (guard placed before `mergeStatusFunction` at `:577`)

`syncSuggestions` is deliberately extensible via injected strategies (`mergeDataFunction`, `mergeStatusFunction`, `statusToSetForOutdated`, `bypassValidationForPlg`, `newSuggestionStatus`). The new guard is a hardcoded, unconditional policy that short-circuits *before* `mergeStatusFunction`, so it also silently overrides any custom regression detection (e.g. FIXED→NEW in permissions) for edited suggestions, and no caller can opt out. It also introduces a second, actor-based "frozen" dimension layered on the existing status-based `FROZEN_STATUSES` concept. This is defensible for a safety invariant, but it departs from the module's strategy-injection design.

*Suggested fix:* either accept an opt-out flag consistent with the existing parameter style, or add a one-line note at the guard stating the freeze is intentionally unconditional and intentionally precedes `mergeStatusFunction` (so the interaction with custom status strategies is a documented decision, not an accident).

No blockers. The core placement is architecturally sound; the material item to resolve or consciously scope out before merge is finding (1) (coverage asymmetry with the outdated/reconcile paths), since it determines whether the PR's headline invariant actually holds end-to-end.

---

## Documentation Review

All code locations are in `src/utils/data-access.js` unless stated otherwise.

### 🟡 Minor

**1. `syncSuggestions` docblock overstates the protection (accuracy/completeness)** — `data-access.js:430-432`

The added sentence says customer-edited suggestions "are not regenerated on re-audit — data/status/rank are left unchanged." Read in context (it sits immediately after the outdated-handling sentences), this reads as a blanket guarantee that a manually-edited suggestion's status is never changed. It isn't: the new guard lives only in the matched-key update loop, which runs *after* `handleOutdatedSuggestions`. That earlier sweep is not gated by the `updatedBy` check and does not exclude `NEW`/`PENDING_VALIDATION`, so a manually-edited `NEW` suggestion whose key disappears from the current audit data can still be flipped to `OUTDATED`. The guarantee only holds while the suggestion still appears in the audit data.

*Suggested fix:* scope the sentence, e.g. "Suggestions still present in the current audit data whose `updatedBy` is a non-system actor (customer edits) are not regenerated — data/status/rank are left unchanged."

**2. "deployed" terminology is inconsistent and overloaded (consistency)** — `data-access.js:41`, `432`, `547`

The new docs label `APPROVED`/`IN_PROGRESS` as "deployed" suggestion statuses, but line 1003 in the same file already calls those exact statuses "terminal states," and "deployed"/`DEPLOYED` is separately the name of a `FixEntity` status (line 822). Same statuses, two different informal names, plus collision with an actual status enum.

*Suggested fix:* pick one term for the `APPROVED`/`IN_PROGRESS` grouping and use it consistently (e.g. reuse "terminal states" from line 1003, or drop "deployed" and just say "e.g. APPROVED / IN_PROGRESS").

**3. PR title cites the wrong ticket (traceability)** — PR #2803 title

The title says `LLMO-5483`; the commit messages, the PR "Related Issues" section, and all three in-code references (JSDoc line 40, docblock line 432, inline comment line 546) say `LLMO-6483`. The title is the sole outlier and will mislink the automated Jira traceability.

*Suggested fix:* correct the PR title to `LLMO-6483`.

**4. PR description undercounts the tests (accuracy)** — PR #2803 description, "What changed"

The description says "two new tests," but the diff adds five `it(...)` blocks (`data-access.test.js:61, 104, 140, 176, 211`): manually-edited NEW, APPROVED, IN_PROGRESS, `updatedBy: null`, and `updatedBy: 'system'`.

*Suggested fix:* update the description to list all five (or say "five new tests").

### 🟢 Nice-to-have

**1. PR description doesn't mention the newly exported helper** — PR #2803 description

"What changed" describes an inline guard (`if existing.getUpdatedBy() is truthy and ≠ 'system'`) but the final code extracts and *exports* a new public function `isManuallyEditedSuggestion`. A reader diffing description against code will find the description stale.

*Suggested fix:* note that the check is factored into an exported `isManuallyEditedSuggestion(suggestion)` helper.

**2. JSDoc could disambiguate from the pre-existing `isManuallyEdited` flag** — `data-access.js:37-45`

The new function `isManuallyEditedSuggestion` (checks `updatedBy`) is one keyword away from the existing `isManuallyEdited` *data flag* used in the alt-text handlers (`src/image-alt-text/handler.js:437`, `src/image-alt-text/guidance-missing-alt-text-handler.js:96`), which is an entirely different mechanism the PR description itself calls out. A one-line note in the JSDoc ("distinct from the alt-text `data.isManuallyEdited` flag; this reads `updatedBy`") would prevent conflation.

**3. Opportunities wiki (referenced in the PR checklist) — could not verify**

If that wiki documents `syncSuggestions` preservation semantics the way it presumably covers the existing `SKIPPED`/`REJECTED` freeze, it should gain the manually-edited freeze. I can't access it to confirm whether such a section exists.

### Notes on documentation that is good

- The inline comment at `data-access.js:546-548` is exemplary: it states the *why* (customer-edit preservation) and flags the non-obvious PII constraint ("`updatedBy` may be a user email — never log it"), and that constraint is actually enforced (the log at line 550 uses `existingKey`, and `data-access.test.js:100-101` asserts the email is not in the message).
- The `isManuallyEditedSuggestion` JSDoc is concise and accurate; the bare `@returns {boolean}` is fine for a clearly named predicate.
- All five new test titles truthfully describe their assertions (the "should update…" titles assert `setData`/`saveMany` called once; the "should not overwrite…" titles assert they are not called).
- The docblock edit also fixed a missing period on the pre-existing REJECTED line (`data-access.js:429`).
- No type-declaration gap: `data-access.js` has no companion `.d.ts`, and `src/common/index.d.ts` does not cover this module, so the new export needs no declaration update.
- README update is not required here: the README (`README.md:443`) documents `syncSuggestions` at a usage level only and does not enumerate the existing `SKIPPED`/`REJECTED` freeze semantics either, so omitting the manually-edited freeze keeps it consistent with current doc scope.

---

## Final Verdict

**Ready to merge:** ⚠️ With fixes

Driven by the Architecture review's one 🟠 Major finding (1) — the manual-edit guard doesn't cover the `handleOutdatedSuggestions` / `reconcileDisappearedSuggestions` paths, so the "never overwritten" invariant doesn't hold end-to-end for NEW/PENDING edits that drop out of an audit; the Documentation review raised only Minor and Nice-to-have items.

duynguyen (approved):
Approved. The manual-edit guards correctly preserve matched, outdated, and reconcile-disappeared suggestions.

Note: a re-audit item with a new key can still be created alongside a manually edited suggestion with the old key; treating that as non-blocking for this change.

---
*Reviewed with ❤️ by [Codex](https://openai.com/codex) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

PR #2805, opened 2026-07-23
fix(experimentation-ess): return created audit from persistOnlyMetadata (SITES-47215)

Dereje24:
## Related Issues

[SITES-47215](https://jira.corp.adobe.com/browse/SITES-47215)

## Problem

`experimentation-ess-all` fails on every run (reproduced in prod: bamboohr, ran 344.6s then errored with `TypeError: Cannot read properties of undefined (reading 'getId')` at `RunnerAudit.processAuditResult`).

Root cause: the framework persists the audit and then dereferences the persister's return value — `const audit = await this.persister(...)` then `audit.getId()` (`src/common/base-audit.js:143,149,165`). The `experimentation-ess-all` persister `persistOnlyMetadata` called `Audit.create(...)` but did **not return it**, so `audit` was `undefined`. `experimentation-ess-daily` uses the default persister (which returns a `getId` shim), which is why only `-all` failed. Regression dates to the 2026-06-10 persist-result refactor of `base-audit.js` (`4db760efe`).

## Fix

`persistOnlyMetadata` now returns the created audit (matching `defaultPersister`).

## Tests

Adds `test/audits/experimentation-ess-all.test.js` — the first test for this module (the dir is `/* c8 ignore */`). Guards that the persister returns the created audit and preserves the `auditResult: []` override.

## Verification

- New test passes; `eslint` clean on the source change.
- Not runtime-verified end-to-end here (needs redeploy + re-run of `experimentation-ess-all`).

## Not in scope (tracked in SITES-47215)

This fixes only the fatal crash. Still open: the cheerio↔DOM mismatch in `common.js getMetadata`, the cross-account IAM denial invoking `statistics-service` (p-values), and independent verification of the V2→V3 (Aurora) data migration.

### DIFF
--- a/src/experimentation-ess/all.js
+++ b/src/experimentation-ess/all.js
@@ -19,12 +19,15 @@ const DAYS = 180;
 
 let log = console;
 
-async function persistOnlyMetadata(auditData, context) {
+export async function persistOnlyMetadata(auditData, context) {
   // persists only the audit metadata, as the
   // whole audit result will be bigger than the allowed size in dynamo
   const { dataAccess } = context;
   const { Audit } = dataAccess;
-  await Audit.create({
+  // must return the created Audit: the framework (base-audit.js processAuditResult)
+  // dereferences the persister's return value via audit.getId(). Returning undefined
+  // here throws "Cannot read properties of undefined (reading 'getId')" (SITES-47215).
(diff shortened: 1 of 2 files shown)


Dereje24:

PR #2806, opened 2026-07-23
fix(experimentation-ess): query metadata via cheerio instead of doc.head.querySelectorAll (SITES-47215)

Dereje24:
## Related Issues

[SITES-47215](https://jira.corp.adobe.com/browse/SITES-47215) (fix 2 of 3; see also #2805 for the persister/`getId` fix)

## Problem

During `experimentation-ess-all`, every experiment-page metadata fetch logs `TypeError: Cannot read properties of undefined (reading 'querySelectorAll')` (reproduced on prod and dev). `getExperimentMetaDataFromExperimentPage` parses the page with `cheerioLoad(html)` and passes the cheerio instance to `getMetadata`, but `getMetadata` used DOM APIs — `doc.head.querySelectorAll(...)` and `m.content`. A cheerio instance has no `.head`, so the lookup throws. The error is swallowed per-URL, so experiments are persisted with **missing metadata** (status, dates, variants, conversion event) rather than failing loudly.

## Fix

Rewrite `getMetadata` to use the cheerio API (`$('head meta[...]')` / `.attr('content')`), preserving the existing head-scoped lookup, name-vs-property attribute selection, and comma-join of multiple matches.

## Tests

Adds `test/audits/experimentation-ess-common.test.js` (module was previously untested — it's `/* c8 ignore */`): reads name- and property-based meta tags from a cheerio instance, returns `''` for missing/empty tags, and joins multiple matches.

## Verification

- Unit test passes; `eslint` clean.
- Dev re-run of `experimentation-ess-all` to confirm the `querySelectorAll` errors are gone and metadata is populated (pending deploy).

## Not in scope (SITES-47215)

Independent of this: #2805 (persister `getId`, merged/queued separately) and the cross-account IAM / resource-based policy on `statistics-service` for p-values.

### DIFF
--- a/src/experimentation-ess/common.js
+++ b/src/experimentation-ess/common.js
@@ -36,13 +36,16 @@ let log = console;
 /**
  * Retrieves the content of metadata tags.
  * @param {string} name The metadata name (or property)
- * @param {Document} doc Document object to query for metadata. Defaults to the window's document
+ * @param {import('cheerio').CheerioAPI} $ Cheerio instance for the loaded page (see cheerioLoad)
  * @returns {string} The metadata value(s)
  */
-function getMetadata(name, doc) {
+export function getMetadata(name, $) {
   const attr = name && name.includes(':') ? 'property' : 'name';
-  const meta = [...doc.head.querySelectorAll(`meta[${attr}="${name}"]`)]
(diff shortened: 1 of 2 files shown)


Dereje24:

PR #2807, opened 2026-07-23
perf(referral-traffic): skip corpus fetch when rules exist and compile classifier regex once | LLMO-6257

cwjwisse:
## 1. Abstract
Performance follow-ups to the merged LLMO-6257 P2 audit-worker work: skip the referral corpus fetch when a site already has category rules, and compile the classifier's rule regexes once instead of per URL. Behavior-preserving.

## 2. Reasoning
`generateReferralCategoryRules` fetched the corpus (`fetchReferralTopUrls`) before the create-if-missing "rules already exist" check, so every daily `llmo-referral-traffic-daily` run paid a five-source partitioned scan even for sites that already have rules (the steady state). Separately, `classifyUrlPath` recompiled each rule's regex inside the per-URL loop, making `buildClassificationRows` O(URLs × rules) `new RegExp()` compilations per run.

- **`patterns-uploader.js`** — reorder `generateReferralCategoryRules` so the existing-rules fetch and the `existingTopicPatterns.length > 0` skip run **before** `fetchReferralTopUrls`. Same skip conditions, same return values, same persisted state — the read-only corpus RPC is simply not called on the skip path.
- **`classify.js`** — compile each rule's regex once (`compileRules` → `{name, re|null}`), then match with `matchCompiledRules`, making `buildClassificationRows` O(URLs + rules). `classifyUrlPath` keeps its signature and exact semantics (first-match-by-order, unsafe/uncompilable regex → no-match parity with SQL `_safe_regex_match`, unmatched → no row). The compiled `RegExp` uses the `i` flag only, so `.test()` is stateless and safe to reuse across URLs.
- **Tests** — assert `fetchReferralTopUrls` is not called on the rules-exist and rule-fetch-error short-circuit paths.

## 4. Required information
- Jira / issue: [LLMO-6257](https://jira.corp.adobe.com/browse/LLMO-6257)
- Spec: `docs/specs/2026-07-21-referral-category-generation-p2.md`
- Pairs with the data-service sibling PR (adds the `traffic_date <= current_date` upper bound to `rpc_referral_traffic_top_urls`).

- `mysticat-data-service` — sibling PR bounds the corpus RPC this caller invokes.

- The corpus RPC's cost was measured live in prod: execution 27ms / planning 64ms. Skipping it once rules exist removes that per-run cost for the steady-state majority of sites.

## 7. Test plan
- **Local (done):** ESLint clean on both files; `llmo-referral-traffic-daily` + `cdn-logs-report` handler/classify/patterns-uploader suites green, including the new corpus-skip assertions.
- **Per-env:** confirm a rules-less site still generates rules on its daily run, and a site with rules skips generation without calling the corpus RPC (log: "already exist - skipping generation").

Independent; each promotes on its own. Pair with the data-service sibling PR for the full effect.

### DIFF
--- a/src/cdn-logs-report/patterns/patterns-uploader.js
+++ b/src/cdn-logs-report/patterns/patterns-uploader.js
@@ -261,13 +261,11 @@ export async function generatePatternsWorkbook(options) {
 export async function generateReferralCategoryRules({ site, context }) {
   const { log } = context;
 
-  const paths = await fetchReferralTopUrls({ site, context });
-  if (paths.length === 0) {
-    log.info('No referral URLs found in DB - skipping referral pattern generation');
-    return false;
-  }
-  log.info(`Fetched ${paths.length} referral URLs for pattern generation`);
-
+  // Create-if-missing gate FIRST: a site that already has category rules skips
+  // before the corpus fetch. fetchReferralTopUrls runs a 5-source partitioned
+  // scan, so fetching it up front would pay that cost on every daily run for
+  // every site even once rules exist. Check existing rules, then fetch the corpus
+  // only when we actually need to generate.
   const existingPatterns = await fetchAgenticUrlClassificationRules(site, context);
   if (existingPatterns && existingPatterns.error) {
     log.info(`Skipping referral patterns for ${site.getId()}; DB rule fetch failed`);
@@ -285,6 +283,13 @@ export async function generateReferralCategoryRules({ site, context }) {
     log.info(`Referral category rules already exist for site ${site.getId()} - skipping generation`);
     return false;
   }
+
+  const paths = await fetchReferralTopUrls({ site, context });
+  if (paths.length === 0) {
+    log.info('No referral URLs found in DB - skipping referral pattern generation');
+    return false;
+  }
+  log.info(`Fetched ${paths.length} referral URLs for pattern generation`);
 
   const domain = new URL(site.getBaseURL()).hostname;
   const productRegexes = await analyzeProducts(domain, paths, context);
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @cwjwisse,

**Verdict: Approve** - clean, behavior-preserving performance optimization with solid test coverage for the new short-circuit paths.
**Complexity: LOW** - small diff, single-service.
_Changes: Reorders the existing-rules check before the expensive corpus fetch in `generateReferralCategoryRules` and compiles classifier regex patterns once per run instead of per URL (3 files)._

<details>

- nit: compiled rules array retains `{name, re: null}` entries that can never match; filtering them out during `compileRules` would save per-URL iterations for large rule sets - `src/llmo-referral-traffic-daily/classify.js:108`
- suggestion: add a one-line comment on `compileRuleRegex` noting it MUST NOT use the `g` or `y` flag, since compiled regexes are now reused across URLs and global `.test()` advances `lastIndex` - `src/llmo-referral-traffic-daily/classify.js`
- suggestion: grep callers of `classifyUrlPath` to confirm none invoke it in a loop (the wrapper now recompiles per call; any loop caller silently regresses to O(URLs * rules)) - `src/llmo-referral-traffic-daily/classify.js:130`

</details>

Note: QA review did not complete and is not included.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 18s | <b>Cost:</b> $3.88 | <b>Commit:</b> <code>f1561f811b29cda4e26d1b4ef8512f75c1bb4500</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2808, opened 2026-07-23
fix(experimentation-ess): update existing experiment instead of duplicate insert (SITES-47215)

Dereje24:
## Related Issues

[SITES-47215](https://jira.corp.adobe.com/browse/SITES-47215) (fix 3 of 3; see also #2805 persister/`getId` and #2806 cheerio metadata)

## Problem

`experimentation-ess-all` fails on every **re-run** for a site with:
`[experiment] Failed to create — [23505] duplicate key value violates unique constraint "unique_site_exp" (site_id, exp_id)` → `DataAccessError: Failed to create at ExperimentCollection.create` (reproduced on dev).

The `postProcessor` fetches the existing experiment (`allBySiteIdAndExpId`), merges fields, then **always** calls `Experiment.create()` — even when the experiment already exists. It never updates. The `"…has been upserted"` log is misleading; there is no update path.

This is a v2→v3 data-access regression: under v2 (DynamoDB/ElectroDB, retired) `create` was a `put`/upsert and re-runs silently overwrote; under v3 (PostgREST/Aurora, current) `create` is a strict `INSERT`, so the second run per site violates the unique constraint. Net effect: the audit works once per site, then fails on every subsequent run.

## Fix

In `postProcessor`, when an existing experiment is found, update it in place via the entity setters + `save()`; only call `Experiment.create()` for genuinely new experiments. Merge logic is unchanged.

## Tests

Adds `test/audits/experimentation-ess-postprocessor.test.js`:
- new experiment → `Experiment.create` is called
- existing experiment → `Experiment.create` is **not** called and `save()` is called with the merged values (regression guard for the duplicate-key failure)

## Verification

- Unit tests pass; `eslint` clean.
- End-to-end: dev re-run of `experimentation-ess-all` on a site that already has experiments should now complete via the update path instead of the duplicate-key error (pending deploy).

## Note

Pre-existing (out of scope): the merge reads `existingExperiment.startDate/.url/.variants` as direct properties; if the v3 entity only exposes getters these fall back to the new values. Preserved as-is here; flagged for a separate follow-up.

### DIFF
--- a/src/experimentation-ess/common.js
+++ b/src/experimentation-ess/common.js
@@ -723,9 +723,26 @@ export async function postProcessor(auditUrl, auditData, context) {
           }
         }
       }
+      // update the existing record in place instead of inserting a duplicate: under v3
+      // (Aurora/PostgREST) Experiment.create is a strict INSERT, so re-running the audit
+      // would violate the unique_site_exp (site_id, exp_id) constraint. v2 (DynamoDB)
+      // create was an upsert, which masked this (SITES-47215).
+      existingExperiment
+        .setName(experimentData.name)
+        .setUrl(experimentData.url)
+        .setStartDate(experimentData.startDate)
+        .setEndDate(experimentData.endDate)
+        .setStatus(experimentData.status)
+        .setType(experimentData.type)
+        .setVariants(experimentData.variants)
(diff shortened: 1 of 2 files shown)


Dereje24:

PR #2813, opened 2026-07-24
fix(offsite): describe each flow step accurately in Slack/logs

AndreiAlexandruParaschiv:
## Problem

The offsite-brand-presence flow (**URL Store → DRS scrape → Mystique**) posted Slack messages that contradicted each other within the same thread. Concretely, a single run looked like this:

- `3:29` ✅ *DRS jobs … job_id …* (a scrape **was** triggered)
- `3:40` 🔍 *DRS content already available … **no scrape job needed***
- `4:31` 🏁 *DRS jobs complete … DRS scraping elapsed: ~62m*
- `4:45` ✅ *audit finished … **DRS scrape: 62m 38s***

The `reddit`/`youtube`/`cited` analysis audits **always** printed *"DRS content already available … no scrape job needed"* whenever the URL store had content — even when a DRS scrape had just run this cycle and its duration was reported moments later. A scrape was very much needed and did run; the analysis audit just didn't *request* one itself. The blanket wording hid that distinction and read as self-contradictory.

## Scope

**Logging & messaging only — no change to what gets scraped or when.**

## Changes

- **Analysis audits** now key off the DRS phase-timing anchors (`auditContext.timings`) to describe exactly what happened:
  - Fresh scrape this cycle (poll-dispatched) → *"DRS scrape finished; N scraped URL(s) ready. Sending to Mystique."*
  - Reusing a prior scrape (direct/scheduled run) → *"reusing previously scraped DRS content (N URL(s) available); no new scrape needed. Sending to Mystique."*
- **`filterUrlsByDrsStatus`** now returns `{ urls, counts }` with an available / scraping / not-found breakdown (surfaced in logs and Slack) instead of collapsing `scraping` and `not_found` into one. The breakdown is attached to `DrsNoContentAvailableError` so the "requesting a scrape" path can explain *why* nothing was available.
- **`offsite-brand-presence`**: added a URL-Store step message (*"collected & stored N URL(s)…"*) and reworded the jobs message to *"DRS scraping started … running in the background"* so it no longer reads as complete.
- **DRS status poll**: header reworded to *"DRS scraping complete"*; added an *"Analysis dispatched to Mystique for: …"* footer.
- Added the previously-missing Slack messages for the *"stored but not scraped yet → starting scrape"* and *"no content after scraping"* analysis paths.
- New shared helpers (`buildAnalysisScrapeStatusMessage`, `formatDrsExtras`, `scrapedThisCycle`) keep the three analysis handlers consistent.

## Testing

- `npm test` — full suite green, **100% coverage** maintained on every changed source file.
- Added cases for the fresh-scrape branch, the status-count breakdown, and the new helpers across all six affected suites.

### DIFF
--- a/src/cited-analysis/handler.js
+++ b/src/cited-analysis/handler.js
@@ -26,6 +26,9 @@ import {
   computeBrandTokens,
   isExcludedCitedHost,
   toApexHost,
+  buildAnalysisScrapeStatusMessage,
+  formatDrsExtras,
+  scrapedThisCycle,
 } from '../utils/offsite-audit-utils.js';
 import { CITED_ANALYSIS_DRS_CONFIG } from '../offsite-brand-presence/constants.js';
 import { computeTopicsFromBrandPresence } from '../utils/offsite-brand-presence-enrichment.js';
@@ -196,7 +199,7 @@ async function fetchStoreData(siteId, context, site) {
 
   const drsClient = DrsClient.createFrom(context);
   const { datasetIds } = CITED_ANALYSIS_DRS_CONFIG;
-  const urls = await filterUrlsByDrsStatus(
+  const { urls, counts } = await filterUrlsByDrsStatus(
     curatedUrls,
     datasetIds,
     siteId,
@@ -204,7 +207,7 @@ async function fetchStoreData(siteId, context, site) {
     log,
     LOG_PREFIX,
   );
-  log.info(`${LOG_PREFIX} ${urls.length} cited URLs available in DRS`);
+  log.info(`${LOG_PREFIX} ${urls.length} cited URLs available in DRS${formatDrsExtras(counts)}`);
 
   const topics = await computeTopicsFromBrandPresence(siteId, context, site);
   log.info(`${LOG_PREFIX} Computed ${topics.length} topics from brand presence data`);
@@ -227,6 +230,7 @@ async function fetchStoreData(siteId, context, site) {
   return {
     urls,
     sentimentConfig: { topics, guidelines },
+    drsCounts: counts,
   };
 }
 
@@ -276,23 +280,33 @@ async function runCitedAnalysisAudit(url, context, site, auditContext = {}) {
 
     const storeData = await fetchStoreData(siteId, context, site);
     log.info(`${LOG_PREFIX} Successfully fetched all store data for ${citedConfig.companyName}`);
+    // Whether this run's DRS scrape produced the content (poll-dispatched) or we are reusing
+    // a prior scrape (direct/scheduled run) changes the log and Slack wording so the thread
+    // reads as a coherent sequence rather than a contradictory "no scrape needed".
+    const scrapedNow = scrapedThisCycle(auditContext);
     log.info(
-      `${LOG_PREFIX} DRS content available for ${storeData.urls.length} URL(s) `
-        + '— no scrape job needed, proceeding to Mystique',
+      scrapedNow
+        ? `${LOG_PREFIX} DRS scrape finished this cycle; ${storeData.urls.length} URL(s) ready, proceeding to Mystique`
+        : `${LOG_PREFIX} Reusing previously scraped DRS content for ${storeData.urls.length} URL(s); no new scrape needed, proceeding to Mystique`,
     );
 
     const urlLimit = resolveMystiqueUrlLimit(auditContext, log, LOG_PREFIX);
 
     const { slackContext } = auditContext;
 
-    // Manual Slack-triggered runs get a notification explaining that no DRS scrape
-    // job was needed (content already available). No-ops on scheduled runs where
+    // Manual Slack-triggered runs get a notification describing the exact DRS state (fresh
+    // scrape this cycle vs. reused prior content). No-ops on scheduled runs where
     // slackContext is absent.
     await postMessageOptional(
       context,
       slackContext?.channelId,
-      `:mag: *cited-analysis* for *${site.getBaseURL()}* — DRS content already available `
-        + `(${storeData.urls.length} URL(s)); no scrape job needed, sending to Mystique.`,
+      buildAnalysisScrapeStatusMessage({
+        analysisName: 'cited-analysis',
+        baseUrl: site.getBaseURL(),
+        urlCount: storeData.urls.length,
+        counts: storeData.drsCounts,
+        scrapedNow,
+      }),
       { threadTs: slackContext?.threadTs },
     );
 
@@ -349,15 +363,31 @@ async function runCitedAnalysisAudit(url, context, site, auditContext = {}) {
     }
 
     if (error instanceof DrsNoContentAvailableError) {
+      const { slackContext } = auditContext;
+      const { channelId, threadTs } = slackContext || {};
       if (auditContext.drsScrapeRequested) {
+        // A scrape already ran this cycle and DRS still reports no scraped content → terminal.
         log.error(`${LOG_PREFIX} No DRS content available after scraping: ${error.message}`);
+        await postMessageOptional(
+          context,
+          channelId,
+          `:warning: *cited-analysis* for *${site.getBaseURL()}* — DRS reported no scraped content after scraping; nothing to analyze.`,
+          { threadTs },
+        );
         return {
           auditResult: { success: false, error: error.message },
           fullAuditRef: url,
         };
       }
-      log.info(`${LOG_PREFIX} No DRS content yet, requesting a scrape for top-cited`);
-      await requestOffsiteScrape(context, siteId, 'top-cited', auditContext.slackContext, enableBrandProfile);
+      log.info(`${LOG_PREFIX} URLs stored but not scraped in DRS yet, requesting a scrape for top-cited`);
+      await postMessageOptional(
+        context,
+        channelId,
+        `:mag: *cited-analysis* for *${site.getBaseURL()}* — cited URLs are stored but not scraped in DRS yet${formatDrsExtras(error.counts)}; `
+          + 'starting a DRS scrape for top-cited, will analyze automatically when it finishes.',
+        { threadTs },
+      );
+      await requestOffsiteScrape(context, siteId, 'top-cited', slackContext, enableBrandProfile);
       return {
         auditResult: { success: false, status: 'pending_scrape', error: error.message },
         fullAuditRef: url,
--- a/src/offsite-brand-presence/drs-status-handler.js
+++ b/src/offsite-brand-presence/drs-status-handler.js
@@ -207,11 +207,11 @@ async function triggerAnalysisAudits(
  *   error: string|undefined}>} statuses - Resolved per-job statuses
  * @returns {string} Slack message text
  */
-function buildSummary(baseURL, statuses, drsStartedAt) {
+function buildSummary(baseURL, statuses, drsStartedAt, triggeredAuditTypes = []) {
   const allTerminal = statuses.every((s) => DRS_TERMINAL_STATUSES.has(s.status));
   const header = allTerminal
-    ? `:checkered_flag: *offsite-brand-presence* DRS jobs *complete* for *${baseURL}*:`
-    : `:hourglass_flowing_sand: *offsite-brand-presence* DRS jobs *status update* for *${baseURL}* (timed out waiting):`;
+    ? `:checkered_flag: *offsite-brand-presence* — *DRS scraping complete* for *${baseURL}*:`
+    : `:hourglass_flowing_sand: *offsite-brand-presence* — *DRS scraping status update* for *${baseURL}* (timed out waiting):`;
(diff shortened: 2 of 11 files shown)


MysticatBot (changes requested):

**Verdict: Request changes** - one blocking issue with the ad-hoc error property assignment pattern.
**Complexity: HIGH** - large diff (1213 lines, 11 files).
_Changes: Makes DRS/Slack messaging context-aware so analysis audits report whether a scrape ran this cycle or prior content is being reused, fixing contradictory thread messages (11 files)._

## Must fix before merge

1. [Important] Ad-hoc `.counts` mutation on Error instance is fragile - `src/utils/offsite-audit-utils.js:~395` (details inline)

<details>

- nit: Inconsistent slackContext access pattern within the same function (optional chaining in happy path, destructure-with-fallback in error path) - `src/cited-analysis/handler.js:356`
- suggestion: `notFound` is derived via `total - available - scraping`, which silently buckets any future DRS status (e.g. `error`, `queued`) under that name; consider `pending` or track an explicit `other` bucket - `src/utils/offsite-audit-utils.js:~399`
- suggestion: Add a `@typedef DrsStatusCounts` JSDoc for the `{total, available, scraping, notFound, determined}` shape now that it flows through multiple modules - `src/utils/offsite-audit-utils.js`
- nit: The three error-path Slack message blocks in cited/reddit/youtube handlers are near-identical; the happy path was deduplicated into `buildAnalysisScrapeStatusMessage` but the error path was not - `src/cited-analysis/handler.js:356`

</details>

src/utils/offsite-audit-utils.js
```diff
@@ -289,21 +388,33 @@ export async function filterUrlsByDrsStatus(urls, datasetIds, siteId, drsClient,
...
   if (!atLeastOneLookupSucceeded) {
     log?.warn(`${prefix} All DRS lookups failed or returned null for datasets [${datasetIds.join(', ')}], skipping availability filter`);
-    return urls;
+    return { urls, counts: undeterminedDrsCounts(urls.length) };
   }
 
-  if (availableUrls.size === 0) {
-    throw new DrsNoContentAvailableError(
+  const total = urls.length;
+  const available = availableUrls.size;
```
MysticatBot, line 420: issue (blocking): `error.counts = counts` mutates the Error instance by attaching an untyped ad-hoc property after construction. Three separate catch sites then read `error.counts`. If any intermediary re-throws the error, wraps it in a generic Error, or passes it through a serialization boundary (structured clone, error-capture middleware), `.counts` vanishes silently. The `formatDrsExtras(undefined)` fallback returns `""` so nothing crashes, but the richer status info intended by this PR is lost.

**Fix:** Accept `counts` as a constructor parameter in `DrsNoContentAvailableError` and expose it as an instance property:

```js
export class DrsNoContentAvailableError extends Error {
  constructor(message, counts) {
    super(message);
    this.name = 'DrsNoContentAvailableError';
    this.counts = counts;
  }
}
```

Then throw with `new DrsNoContentAvailableError(msg, counts)` instead of post-hoc mutation. This makes the contract explicit, visible in the constructor signature, and survives serialization.

MysticatBot (approved):

**Verdict: Approve** - prior blocking finding resolved; remaining items are non-blocking nits.
**Complexity: HIGH** - large diff (1250 lines, 11 files).
_Changes: Makes DRS/Slack messaging context-aware so analysis audits report whether a scrape ran this cycle or prior content is being reused, fixing contradictory thread messages (11 files)._

<details>

- nit: Inconsistent `slackContext` access pattern within cited/reddit/youtube handlers - optional chaining in happy path, destructure-with-fallback in error path - `src/cited-analysis/handler.js:356`
- suggestion: `notFound` is derived via `total - available - scraping`, which silently buckets any future DRS status under that name; a one-line comment noting this assumption would prevent confusion - `src/utils/offsite-audit-utils.js:409`
- suggestion: The three analysis handler error blocks (DrsNoContentAvailableError catch) are near-identical; the happy path was deduplicated into `buildAnalysisScrapeStatusMessage` but the error path was not - consider a shared helper in a follow-up - `src/cited-analysis/handler.js:356`
- nit: `buildSummary` gained a `triggeredAuditTypes` parameter but no direct unit test for the new "Analysis dispatched" line - currently covered by integration tests - `src/offsite-brand-presence/drs-status-handler.js:124`

</details>

- DRS counts passed via `DrsNoContentAvailableError` constructor parameter instead of ad-hoc property mutation.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 32s | <b>Cost:</b> $3.96 | <b>Commit:</b> <code>debc5d7e1648db3f058115c3e96b77961fc022d1</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

grigoreadobe (approved):

src/offsite-brand-presence/drs-status-handler.js
```diff
@@ -227,6 +227,9 @@ function buildSummary(baseURL, statuses, drsStartedAt) {
   if (elapsed) {
     lines.push(`• DRS scraping elapsed: ~${elapsed}`);
   }
+  if (triggeredAuditTypes.length > 0) {
+    lines.push(`• Analysis dispatched to Mystique for: ${triggeredAuditTypes.join(', ')}`);
```
tb-adbe, line 231: issue (important): `triggeredAuditTypes`/`nextTriggered` includes both dispatched types (line 192) and skipped types (the `hasRecentAudit` branch, line 171-173) - `triggerAnalysisAudits` explicitly returns "dispatched or intentionally skipped" per its own docstring. This line renders every entry as dispatched, so a skipped analysis can be reported here as if Mystique received it. Split into dispatched/skipped, or reword to be outcome-neutral.

src/cited-analysis/handler.js
```diff
@@ -349,15 +363,31 @@ async function runCitedAnalysisAudit(url, context, site, auditContext = {}) {
...
           fullAuditRef: url,
         };
       }
-      log.info(`${LOG_PREFIX} No DRS content yet, requesting a scrape for top-cited`);
-      await requestOffsiteScrape(context, siteId, 'top-cited', auditContext.slackContext, enableBrandProfile);
+      log.info(`${LOG_PREFIX} URLs stored but not scraped in DRS yet, requesting a scrape for top-cited`);
+      await postMessageOptional(
+        context,
+        channelId,
+        `:mag: *cited-analysis* for *${site.getBaseURL()}* — cited URLs are stored but not scraped in DRS yet${formatDrsExtras(error.counts)}; `
```
tb-adbe, line 386: issue (important): This new message (and the sibling `:warning:` message at line 374, plus the identical pair in reddit-analysis/youtube-analysis handlers) has no test asserting its content, including the `formatDrsExtras(error.counts)` interpolation. The existing tests that hit this catch branch only assert on `sqs.sendMessage`/`log.error`. None of them supply `slackContext` either, so the `slackContext || {}` truthy branch is unexercised here and in youtube-analysis. A wiring mistake (e.g. dropping `error.counts`) would pass CI as-is.

tb-adbe (changes requested):

## Strengths

- The core rework is well-designed: `buildAnalysisScrapeStatusMessage`, `formatDrsExtras`, and `scrapedThisCycle` in src/utils/offsite-audit-utils.js cleanly centralize wording logic that was previously triplicated across the cited/reddit/youtube handlers, and they build on the pre-existing `timings` phase-boundary contract rather than inventing a new signal.
- The previously-flagged blocking issue (ad-hoc `error.counts = counts` mutation on `DrsNoContentAvailableError`) is genuinely fixed - `counts` is now a constructor parameter, and this is verified end-to-end by tests covering both direct construction and the throw/catch boundary it was meant to protect.
- `filterUrlsByDrsStatus`'s new `{ urls, counts }` return shape is applied consistently across all call sites and all three handler test suites, with no orphaned callers of the old signature.
- Good edge-case discipline in the utils tests: the "scraping and not_found statuses in the same response" case and the false/undefined/partial-timings cases for `scrapedThisCycle` are genuinely new scenarios, not just happy-path re-runs.
- This is a real fix, not cosmetic: it removes a genuine contradiction (claiming "no scrape job needed" when a scrape had, in fact, just run) that would mislead anyone reading the Slack thread as an operational signal.

## Issues

### Important (Should Fix)

1. **The new "Analysis dispatched to Mystique for: ..." footer conflates dispatched and skipped audit types** - src/offsite-brand-presence/drs-status-handler.js:171-197 (`triggerAnalysisAudits`), rendered at drs-status-handler.js:230-231 (`buildSummary`). `triggerAnalysisAudits`'s own docstring says it returns "audit types that were dispatched or intentionally skipped." An audit type lands in `handled` (and therefore in `nextTriggered`, which flows into the new footer line) both when an SQS message is actually sent (line 192) and when it's skipped because a recent audit already exists for that site (the `hasRecentAudit` branch at line 171-173). The new summary line unconditionally renders every entry as "dispatched," so a Slack thread can read "Analysis dispatched to Mystique for: reddit-analysis" when reddit-analysis was in fact skipped this cycle due to cooldown dedup. This is exactly the class of misleading-Slack-text bug this PR exists to fix elsewhere. Fix: split `handled` into `dispatched`/`skipped` arrays and only list `dispatched` in the footer, or reword the line to be outcome-neutral (e.g. "Analysis handled for: X") if the distinction isn't worth surfacing.

2. **The two new Slack notifications this PR adds have no test assertions on their content, and the `slackContext`-present branch is untested in two of three handlers.** In src/cited-analysis/handler.js:365-390 (and the equivalent blocks in reddit-analysis/handler.js and youtube-analysis/handler.js), the PR adds the ":warning: ... DRS reported no scraped content after scraping" message (line 374) and the ":mag: ... stored but not scraped in DRS yet ... starting a DRS scrape" message (lines 386-387), including the `formatDrsExtras(error.counts)` interpolation. These are this PR's headline "previously-missing Slack messages." The existing tests that exercise this catch branch (test/audits/cited-analysis/handler.test.js:299-342 and the reddit/youtube equivalents) only assert on `sqs.sendMessage` / `log.error`, never on `postMessageOptional`'s arguments - so the exact wording, emoji, and the `error.counts` to `formatDrsExtras` wiring are unverified where they're actually used. On top of that, none of these tests supply a `slackContext`, so the `const { channelId, threadTs } = slackContext || {}` truthy branch is never exercised for cited-analysis or youtube-analysis (reddit-analysis has one pre-existing test that supplies `slackContext` here, but it doesn't assert on `postMessageOptional` output either). A wiring mistake that dropped `error.counts` or a typo in either message would pass CI as-is. Fix: add a `calledWithMatch` assertion on the message text for both branches in all three handler test files, with at least one case per handler that supplies `slackContext` and has `error.counts` populated to prove the breakdown renders in situ.

### Minor (Nice to Have)

- src/utils/offsite-audit-utils.js - `buildOffsiteTimingLines` and `scrapedThisCycle` each independently recompute `Number.isFinite(drsStartedAt) && Number.isFinite(drsCompletedAt)` against the same `timings` shape. Not a problem with two call sites, but if a third consumer of this pattern shows up, extracting a single `hasDrsTimings(timings)` helper would prevent drift.
- test/audits/offsite-brand-presence.test.js:1466-1473 asserts the store message contains "selected", "to scrape this run", and the base URL, but never checks the per-bucket breakdown format built at handler.js:724-730 (`perBucket`). A regression that mangled the join separator or dropped a domain would go unnoticed - consider asserting on one bucket entry (e.g. `youtube.com: 1`).

## Recommendations

- cited-analysis/handler.js, reddit-analysis/handler.js, and youtube-analysis/handler.js remain three parallel copies of the fetch/DRS-branching/Slack-messaging logic. This PR deduplicated the happy path into `buildAnalysisScrapeStatusMessage` but added the two new error-path messages identically to all three files again. Given `OFFSITE_DOMAINS` already models domains as data, a shared orchestrator that the three thin handlers call into would remove the recurring 3x diff/review tax and the risk of silent drift between copies (which is close to what caused the bug this PR fixes). Not worth doing as part of this PR, but worth tracking as a follow-up.
- For the dispatched-vs-skipped ambiguity above, a `Map<type, 'dispatched'|'skipped'|'failed'>` return value from `triggerAnalysisAudits` instead of a flat `string[]` would make this class of mismatch structurally harder to reintroduce.

## Assessment
Ready to merge? With fixes
Reasoning: The previously-blocking issue is genuinely resolved and the core messaging redesign is solid, but the PR introduces one new inaccurate-message risk that falls squarely within its own stated goal, plus ships its two headline new Slack messages without any assertion on their content or the slackContext-present path. Both are small, targeted fixes rather than a redesign.

## Next Steps
1. Fix the dispatched-vs-skipped conflation in the "Analysis dispatched to Mystique for" footer.
2. Add content assertions (including a slackContext-present case) for the two new Slack messages across all three handlers.
3. Minor items are optional.

PR #2814, opened 2026-07-24
fix(form-a11y): compute a real per-form selector instead of a generic 'form'

clotton:
## Summary
- `detectFormAccessibility` hardcoded `formSource: 'form'` for every page sent to Mystique. Playwright's `page.query_selector('form')` — like the browser's native `querySelector` — returns the *first* `<form>` in DOM order, so any page with more than one form (e.g. a header search widget ahead of the actual content form) had the wrong form silently analyzed. Confirmed live on an internal test site: the real broken form was never reached at all, only the header search widget was.
- Now fetches each page's scrape (same S3 pattern already used in `handler.js`/`headings.js`), finds every `<form>` via `cheerio`, and computes a real selector per form using the existing `getDomElementSelector` utility — sending zero-to-N entries per URL instead of always exactly one generic entry.
- Falls back to today's exact single-generic-selector behavior whenever the scrape can't be fetched/parsed or yields no forms, so this degrades safely rather than dropping detection (this is also why the ~65 pre-existing tests needed no changes — they all exercise the fallback path already).
- Adds `generateFormAccessibilityFilename` (distinct from the shared `generateAccessibilityFilename`, which stays untouched since `accessibility.js` depends on its one-arg signature) so multiple forms on the same page get distinct S3 result-file keys instead of colliding — preserves the exact existing filename for the generic `'form'` fallback case.
- Companion fix on the Mystique side (same PR review cycle) resolves the matching S3-key-accumulation bug and uses the identical MD5-of-selector filename scheme (verified byte-for-byte identical between Node's `crypto` and Python's `hashlib`).

## Test plan
- [x] `test/audits/preflight-form-accessibility.test.js` — 75/75 passing, 100% line/branch/function coverage on `form-accessibility.js`
- [x] All ~65 pre-existing tests pass unchanged (they exercise the fallback path)
- [x] New tests cover: multi-form selector computation, single-form (non-fallback) selector, zero-form skip, scrape-unavailable fallback, unparseable-URL fallback, HTML-parse-failure fallback, no-selector-generated fallback, and cross-form opportunity merging
- [x] Verified end-to-end locally against a real multi-form production test page with the matching Mystique-side fix: both forms independently detected and correctly written to separate S3 keys (6 real accessibility issues surfaced on the previously-unreachable broken form, 1 on the search widget)

### DIFF
--- a/src/preflight/form-accessibility.js
+++ b/src/preflight/form-accessibility.js
@@ -10,22 +10,112 @@
  * governing permissions and limitations under the License.
  */
 
+import { createHash } from 'crypto';
 import { isNonEmptyArray } from '@adobe/spacecat-shared-utils';
 import { DeleteObjectsCommand } from '@aws-sdk/client-s3';
+import { load as cheerioLoad } from 'cheerio';
 
 import { saveIntermediateResults } from './utils.js';
 import { sleep } from '../support/utils.js';
 import { getObjectFromKey, getObjectKeysUsingPrefix } from '../utils/s3-utils.js';
 import { generateAccessibilityFilename } from './accessibility.js';
+import { getDomElementSelector } from './utils/dom-selector.js';
 
 export const PREFLIGHT_FORM_ACCESSIBILITY = 'form-accessibility';
+
+/**
+ * Generates a stable, unique filename for a single form's result file.
+ *
+ * Distinct from `generateAccessibilityFilename` (shared with the plain
+ * accessibility.js audit, one file per URL) — a page can have multiple forms,
+ * so the filename must also depend on the form's selector, not just the URL.
+ *
+ * Preserves the exact filename `generateAccessibilityFilename` would produce
+ * when formSource is the generic 'form' fallback (no selector was computed) —
+ * only diverges (with a disambiguating hash suffix) once a page genuinely has
+ * more than one form and each gets its own real selector.
+ *
+ * @param {string} url - The page URL.
+ * @param {string} formSource - The CSS selector identifying the form.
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @clotton,

**Verdict: Approve** - well-scoped bug fix with thorough fallback design and strong test coverage.
**Complexity: MEDIUM** - medium diff, multi-path fallback logic across 2 files.
_Changes: Replaces the hardcoded generic 'form' selector with per-form CSS selectors computed from scraped HTML, fixing multi-form page detection (2 files)._

<details>

- nit: Guard the empty-entries path in the orchestrator - when all pages have zero forms, `formEntries` is `[]`, polling exits immediately (fine), but `processFormAccessibilityOpportunities` proceeds to cleanup and sends an empty `DeleteObjectsCommand` which produces a caught S3 `MalformedXML` error. An `if (formEntries.length === 0) return;` after Step 1 would skip the wasteful path. - `src/preflight/form-accessibility.js:379`
- suggestion: Add an integration test for the full orchestrator (`formAccessibility` default export) with zero-form pages to verify the empty-entries edge case end-to-end. - `test/audits/preflight-form-accessibility.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 1s | <b>Cost:</b> $4.85 | <b>Commit:</b> <code>3fda5382f2029bcc4a701be0b521f30431cc0542</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2818, opened 2026-07-27
fix(llmo): populate previousConfigVersion in drs-prompt-generation handler

akshaymagapu:
## Summary
- The SQS message that re-triggers `llmo-customer-analysis` after a DRS prompt-generation job completes never set `previousConfigVersion`, so `isFirstTimeOnboarding` evaluated `true` on every completion regardless of actual onboarding history. This fired the "Onboarding Detected" Slack notification repeatedly for already-onboarded sites (observed dozens of times for the same sites in #llmo-onboarding-automation).
- Fix reads the existing LLMO config version before the legacy config write happens, same approach `updateLlmoConfig` already uses in spacecat-api-service, gated to `source=onboarding` since that's the only path that uses it. Non-fatal on read failure (logs a warning, falls back to omitting the field).

## Test plan
- [x] Added tests covering: prior version present, no prior version (genuine first-time), read failure (non-fatal), and non-onboarding sources skip the read entirely
- [x] `npm test` passes with 100% coverage maintained
- [x] `npm run lint` clean

### DIFF
--- a/src/drs-prompt-generation/handler.js
+++ b/src/drs-prompt-generation/handler.js
@@ -11,6 +11,7 @@
  */
 
 import { ok } from '@adobe/spacecat-shared-http-utils';
+import { llmoConfig } from '@adobe/spacecat-shared-utils';
 import writeDrsPromptsToLlmoConfig from './drs-config-writer.js';
 import { postMessageSafe } from '../utils/slack-utils.js';
 
@@ -155,6 +156,23 @@ export default async function drsPromptGenerationHandler(message, context) {
 
   log.info(`DRS prompt generation completed for site ${siteId}, job ${drsJobId}, result: ${resultLocation}`);
 
+  // Capture the config version that existed *before* this job's write (below), so
+  // llmo-customer-analysis can tell a genuine first-time onboarding from a repeat one.
+  // Must be read before processDrsResult writes a new version, and is only needed
+  // for source=onboarding since that's the only case that reaches the SQS send below.
+  let previousConfigVersion;
+  if (source === 'onboarding') {
+    try {
+      const { s3Client, env } = context;
+      const prevConfig = await llmoConfig.readConfig(siteId, s3Client, {
+        s3Bucket: env.S3_IMPORTER_BUCKET_NAME,
+      });
+      previousConfigVersion = prevConfig.exists ? prevConfig.version : undefined;
+    } catch (error) {
+      log.warn(`Failed to read existing LLMO config version for site ${siteId}: ${error.message}`);
+    }
+  }
+
   let configVersion;
   const shouldWriteLegacyConfig = source !== 'onboarding' || onboardingMode !== 'v2';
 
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - clean, well-scoped bug fix with solid test coverage.
**Complexity: MEDIUM** - medium line count, single-service.
_Changes: Populates `previousConfigVersion` in the SQS message sent to `llmo-customer-analysis` after DRS prompt generation completes, so downstream can distinguish genuine first-time onboarding from repeat completions (2 files)._

<details>

- nit: `log.warn` at line 170 only interpolates `error.message`, dropping the stack trace. If this fires in production, operators will want the stack to tell "S3 timeout" from "bug in readConfig serialization." Consider passing the error object as structured metadata if the logger supports it (e.g. `log.warn('...message...', { error })`) - `src/drs-prompt-generation/handler.js:170`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 49s | <b>Cost:</b> $2.57 | <b>Commit:</b> <code>1cc12bc934de54c18c114d2750aacb49280d5978</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

dzehnder (approved):
Hey @akshaymagapu,

**Verdict: Approve** - correct, tightly-scoped bug fix; the producer to consumer contract and both v1/v2 onboarding paths were verified against source.
**Complexity: MEDIUM** - medium line count (~125 added), single-service, no risk signals.
_Changes: Reads the pre-existing LLMO config version before the DRS config write and passes `previousConfigVersion` into the `llmo-customer-analysis` SQS message, so downstream can distinguish genuine first-time onboarding from repeat completions (2 files)._

<details>

- suggestion: transient `readConfig` failure leaves `previousConfigVersion` undefined, so the consumer computes `isFirstTimeOnboarding = true` and re-fires "Onboarding Detected" for an already-onboarded site - strictly better than today's always-fire behavior and correctly non-fatal, but a one-line comment (or a metric) noting the fallback biases toward the first-time path would save a future debugger - `src/drs-prompt-generation/handler.js:170`
- nit: `exists: true` with a falsy `version` collapses to the same "no prior version" result as `exists: false` (`readConfig` returns `version: res.VersionId || undefined`), conflating an existing-but-unversioned config with a genuine first-time onboarding. Real-world risk is low since the writer guarantees a `VersionId` on a versioned bucket, but the two states are indistinguishable and untested - `src/drs-prompt-generation/handler.js:169`
- nit: the ternary `prevConfig.exists ? prevConfig.version : undefined` is redundant - `readConfig` already returns `version: undefined` when `exists` is false, so `previousConfigVersion = prevConfig.version` suffices (the explicit form is defensible as intent-signalling) - `src/drs-prompt-generation/handler.js:169`
- suggestion: the read-failure test asserts only `previousConfigVersion` absent plus status 200; also assert `configVersion` is still present and the downstream trigger still fires, to confirm a read failure does not collateral-damage the normal path - `test/drs-prompt-generation/handler.test.js`
- nit: the added "includes imsOrgId" test covers pre-existing passthrough behavior unrelated to this fix - harmless and within budget, just slightly out of scope for a fix PR - `test/drs-prompt-generation/handler.test.js`
</details>

<details>
<summary>Test quality snapshot</summary>

- New code: 19 LOC across 1 file; new tests: 106 LOC across 1 file (ratio 5.58)
- Test cases: 5 new it() blocks; max 5 in `test/drs-prompt-generation/handler.test.js`
- Assertion density: 1.80 per test; mock density: 0.00 per test
- Fired signals: none
</details>

PR #2819, opened 2026-07-27
fix(prerender): strip tracking query params from RCV dedup keys

anuj-adobe:
## Summary
- Fixes duplicate entries in the RCV (Recover Content Visibility) opportunity caused by tracking/marketing query parameters (`utm_*`, `gclid`, `gbraid`, `wbraid`, `fbclid`, `msclkid`, `_gl`, `srsltid`, Mailchimp/Klaviyo/Bronto params, cache-busters, etc.) being treated as part of a page's identity when building suggestion/opportunity dedup keys.
- Adds `TRACKING_PARAM_PATTERNS` + a `stripTrackingParams()` helper to `src/prerender/utils/utils.js`, mirroring [`TRACKING_PARAM_PATTERNS` in tokowaka-worker](https://github.com/adobe-rnd/tokowaka-worker/blob/main/src/utils/request-utils.js#L220) so the audit worker and the edge worker agree on which query params are tracking-only noise.
- Wires the stripping into `normalizePathnameWithQuery` (used by `buildSuggestionKey`, `syncSuggestions` matching, and scraped-URL-set lookups) and into `mergeAndGetUniqueHtmlUrls`'s `includeQueryParams` path (CSV/included-URL and cross-source dedup), so two URLs differing only by tracking params now resolve to the same suggestion/opportunity, while non-tracking query-param variants (e.g. `?filter=iphone` vs `?filter=mac`) are still kept distinct as before.

## Root cause
- Organic/agentic top-page URL lists already dedup by pathname only (all query params dropped), but "included"/CSV URL lists and the cross-source merge intentionally preserve query params so legitimate variants aren't collapsed. Tracking params riding along on included/CSV URLs (or URLs reaching the audit from real traffic with ad/campaign params attached) fell into that "preserved" bucket and produced separate suggestions/opportunity rows for what was actually the same page.
- `normalizePathnameWithQuery`, the canonical identity key used across suggestion sync and scrape lookups, also included the full raw query string, so the same issue surfaced there independent of which URL source was used.

## Test plan
- [x] Added unit tests in `test/audits/prerender/utils.test.js` covering: single/multiple tracking params stripped, tracking params stripped while non-tracking params preserved, URLs differing only by tracking params normalize to the same key, and non-tracking query strings are left untouched (order preserved).
- [x] `npx mocha --recursive test/audits/prerender/` — 643 passing
- [x] `npm run lint` — clean
- [x] `npx c8 --include='src/prerender/utils/utils.js' mocha --recursive test/audits/prerender/` — 100% lines/branches/functions

Jira: LLMO-5897

### DIFF
--- a/src/prerender/utils/utils.js
+++ b/src/prerender/utils/utils.js
@@ -67,8 +67,84 @@ export function toPathname(url) {
 }
 
 /**
+ * Tracking/marketing query parameter patterns that should not cause two otherwise-identical
+ * URLs to be treated as distinct pages (e.g. in RCV suggestion/opportunity dedup keys).
+ * Mirrors TRACKING_PARAM_PATTERNS in tokowaka-worker's src/utils/request-utils.js so the audit
+ * and the edge worker agree on which query params are tracking-only noise.
+ */
+const TRACKING_PARAM_PATTERNS = [
+  // Google Ads
+  /^gad_source$/, // Google Ads click source identifier
+  /^gad_campaignid$/, // Google Ads campaign ID (auto-tagging)
+  /^gbraid$/, // Google Ads app-to-web attribution (iOS)
+  /^wbraid$/, // Google Ads web-to-app attribution (iOS)
+  /^gclid$/, // Google Ads click identifier
+  /^gclsrc$/, // Google Ads click source type (e.g. aw.ds for Search Ads 360)
+  /^dclid$/, // Google Display & Video 360 (DoubleClick) click identifier
+  /^srsltid$/, // Google Shopping / Merchant Center result identifier
+
+  // Google Analytics
+  /^_gl$/, // Google Analytics cross-domain linker parameter
+
+  // Microsoft Advertising (Bing Ads)
+  /^msclkid$/, // Microsoft Advertising click identifier
+
+  // Meta (Facebook / Instagram)
+  /^fbclid$/, // Meta (Facebook) click identifier
+
+  // Zanox / Awin (affiliate network)
+  /^zanpid$/, // Zanox/Awin affiliate partner click identifier
+
+  // Klaviyo (email marketing)
+  /^_kx$/, // Klaviyo email tracking identifier
+
+  // Mailchimp (email marketing)
+  /^mc_[a-z]+$/, // Mailchimp campaign tracking params (mc_cid, mc_eid, etc.)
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @anuj-adobe,

**Verdict: Approve** - well-scoped bug fix with solid test coverage, no blocking issues.
**Complexity: MEDIUM** - medium diff, utility + tests.
_Changes: Strips tracking/marketing query parameters from RCV dedup keys so URLs differing only by ad-tracking params no longer produce duplicate suggestions/opportunities (2 files)._

Note: CI checks are currently failing (deploy/stage) - confirm these are unrelated before merge.

<details>

- nit: `stripTrackingParams` returns the raw original `search` when no tracking params are found but returns a `URLSearchParams.toString()` reconstruction when some are stripped - encoding may differ between the two paths (e.g. `%20` vs `+`), which could produce different dedup keys for the same logical page depending on whether a tracking param was present. Low probability in practice. - `src/prerender/utils/utils.js:132`
- nit: Regex patterns are case-sensitive (`/^utm_[a-z]+$/` won't match `UTM_SOURCE`). This mirrors tokowaka-worker's behavior (stated design goal), but if the upstream ever adds case-insensitive matching, this copy would need updating too. - `src/prerender/utils/utils.js:116`

</details>

<details>
<summary>Recommendations</summary>

- suggestion: Consider extracting `TRACKING_PARAM_PATTERNS` to a shared package (e.g. `spacecat-shared`) so both this repo and tokowaka-worker import from a single source of truth. If that's out of scope, a `// Last synced: <date/commit>` comment would make drift visible during review.
- suggestion: A `.toLowerCase()` on the key before regex testing would handle the rare case of uppercase tracking params from user-supplied CSVs or redirects - but only adopt this if tokowaka-worker does the same, to preserve parity.

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 40s | <b>Cost:</b> $5.44 | <b>Commit:</b> <code>b54ef08f4207326ec356720d0d673275b2f4f2ab</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jindaliiita (approved):

PR #2817, opened 2026-07-26
docs(referral-traffic): spec for referral category generation from DRS sources (GA4/AA/CJA) | LLMO-6257

omairtemurian:
## Abstract
Adds a design spec for generating referral-traffic categories for the DRS analytics sources — GA4, Adobe Analytics, and CJA. This is the Phase-2 follow-up to LLMO-6257 Phase 1 (optel + cdn), which shipped to prod.

## Reasoning
Referral traffic from the DRS sources currently shows no category on the dashboard. A prod scan on 2026-07-26 confirmed the shape: categories appear only on sites that have optel data — every source combination without optel, including 142 cdn-only sites, is at 0%. This spec proposes how to close that gap so DRS-only (and cdn-only) sites get categories.

## Overview
Documentation only — adds `docs/specs/2026-07-26-referral-category-drs-generation.md`; no code or behavior changes. The spec identifies two gaps and proposes a fix for each:
- **Rule generation** is coupled to the optel audit, so sites without optel never get a rulebook. Fix A decouples it so rule-gen runs for any site with referral traffic (this also fixes the cdn-only sites broken in prod today). It only broadens the trigger — rule-gen stays in the audit-worker, nothing moves into the database. The per-site sweeper audit already merged in PR #2821; weekly fan-out + enablement config are still pending.
- **Classification** does not happen for DRS sources. Fix B has DRS classify its own URLs write-time in Python (mirroring the existing JS classifier), reading rules via the **existing** api-service endpoint `GET /sites/:siteId/agentic-categories` (no new endpoint — only DRS auth to provision) and emitting the same classification CSV into the existing pipeline.

The downstream path (projector → data-service → api → UI) is reused unchanged.

## Links
- Jira: LLMO-6257 (parent LLMO-5440)
- Follows the Phase-1 spec: `docs/specs/2026-07-21-referral-category-generation-p2.md`

## Affected projects
Documentation lands in spacecat-audit-worker. The proposed work would span: **spacecat-audit-worker** (decouple the rule-gen trigger — sweeper merged in PR #2821), **llmo-data-retrieval-service** (DRS Python classify), and **spacecat-api-service** (reuse the existing `GET /sites/:siteId/agentic-categories` — only provision DRS `site:read` auth, no new endpoint). Reused unchanged downstream: mysticat-projector-service, mysticat-data-service, project-elmo-ui.

## Test plan
[...]

PR #2821, opened 2026-07-27
feat(referral-traffic): rule-gen sweeper audit for all referral sites (llmo-referral-category-rules) | LLMO-6257

omairtemurian:
## Abstract
Adds a new per-site audit, `llmo-referral-category-rules`, that (re)generates a site's shared referral category rules when it has none yet. This is Fix A of the LLMO-6257 Phase-2 DRS work — decoupling rule generation from the optel audit so every site with referral traffic can get a rulebook.

## Reasoning
Referral category rules are only generated inside the optel daily audit today, so sites without optel — cdn-only sites, and (once built) DRS-only sites — never get a rulebook, and therefore never get categories. A prod scan on 2026-07-26 confirmed this: categories appear only on sites that have optel; every non-optel source (including 142 cdn-only sites) is at 0%. This audit gives the system a source-agnostic way to generate rules for any referral site, which fixes the cdn-only gap in prod today and unblocks the DRS work.

## Overview
A small `RunnerAudit` whose runner calls the existing `generateReferralCategoryRules({ site, context })` for one site and records the outcome. That function is already source-agnostic (its corpus unions all referral sources) and create-if-missing (idempotent), so this change only gives it a new trigger — it does not change how rules are generated, and it moves nothing into the database. On a per-site failure the runner records the error and returns rather than throwing, so a single site's failure can't trigger retries that would re-invoke the LLM. Registered as audit type `llmo-referral-category-rules` in the handler map; no existing audit is modified.

This handler is ready but does not run on its own yet — turning it on is a follow-up configuration step (register it as a scheduled job at the chosen cadence and enable it for referral sites in the Configuration), which lands after this deploys.

## Links
- Jira: LLMO-6257 (parent LLMO-5440)
- Design spec: adobe/spacecat-audit-worker#2817

## Affected projects
Change is confined to spacecat-audit-worker. It relies at runtime on the existing data-service corpus/rules RPCs (unchanged) and, once scheduled, on the jobs-dispatcher Configuration (a follow-up). The DRS Python classifier that will consume these rules is separate (DRS team).

## Test plan
Unit tests cover the runner's three outcomes: fresh rules generated, no-op when the site already has rules, and a generation failure being recorded without throwing. After deploy, the sweep is switched on via the scheduling/enablement configuration and verified by confirming categories begin appearing for cdn-only (and later DRS-only) sites through the referral `filter-dimensions` API — the same check used to confirm the optel/cdn path in prod.

## Deployment order
Independent and safe to ship on its own — it registers a new audit but nothing dispatches it yet, so there is no behavior change until the follow-up scheduling config is applied. Ship this first; the scheduling/enablement config follows once it is live.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -79,6 +79,7 @@ import readabilityOpportunities from './readability/opportunities/handler.js';
 import unifiedReadabilityGuidance from './readability/shared/unified-guidance-handler.js';
 import llmoReferralTraffic from './llmo-referral-traffic/handler.js';
 import llmoReferralTrafficDaily from './llmo-referral-traffic-daily/handler.js';
+import llmoReferralCategoryRules from './llmo-referral-category-rules/handler.js';
 import llmErrorPages from './llm-error-pages/handler.js';
 import llmErrorPagesGuidance from './llm-error-pages/guidance-handler.js';
 import paidTrafficAnalysis from './paid-traffic-analysis/handler.js';
@@ -190,6 +191,7 @@ const HANDLERS = {
   'page-intent': pageIntent,
   'llmo-referral-traffic': llmoReferralTraffic,
   'llmo-referral-traffic-daily': llmoReferralTrafficDaily,
+  'llmo-referral-category-rules': llmoReferralCategoryRules,
   'llm-error-pages': llmErrorPages,
   'guidance:llm-error-pages': llmErrorPagesGuidance,
   'optimization-report-callback': optimizationReportCallback,
--- a/src/llmo-referral-category-rules/handler.js
+++ b/src/llmo-referral-category-rules/handler.js
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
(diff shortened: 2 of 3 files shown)


MysticatBot (approved):
Hey @omairtemurian,

**Verdict: Approve** - clean implementation following established audit patterns.
**Complexity: MEDIUM** - 147 added lines across 3 files, single-service.
_Changes: Adds a standalone `llmo-referral-category-rules` RunnerAudit that wraps the existing `generateReferralCategoryRules` function with error isolation to prevent SQS retry storms on per-site LLM failures (3 files)._

<details>

- nit: Runner function signature declares 3 parameters but `RunnerAudit` passes 4 (including `auditContext`) - `src/llmo-referral-category-rules/handler.js:30`. Harmless in JS (extra args are silently ignored) but deviates from the documented convention; consider `(auditUrl, context, site, _auditContext)` for consistency with the invocation contract.
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 9m 20s | <b>Cost:</b> $4.83 | <b>Commit:</b> <code>b8b7adc059dc5c249602f38970da7363c839a05f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2823, opened 2026-07-28
fix(cwv): suppress redundant guidance re-dispatch for unchanged failing metrics

tkotthakota-adobe:
https://jira.corp.adobe.com/browse/SITES-48889

## Problem

Every weekly CWV audit re-sends a `guidance:cwv` message to Mystique for each `NEW` suggestion without a code patch — even when the identical metric (e.g. bad LCP) was already flagged and already has guidance. A page failing LCP in week *n* is re-guided in weeks *n+1*, *n+2*, … paying for fresh guidance/code-fix generation with **no new signal**. Bad LCP this week almost certainly means bad LCP next week; there is no reason to re-run the expensive downstream work unless the flagged metric set actually changes.

The suggestion DB sync already deep-equal-skips unchanged rows, so the waste is purely in the Mystique dispatch.

## Change

`shouldSendAutoSuggestForSuggestion` now stamps a **dispatch fingerprint** on `NEW` dispatches:

```js
data.autoSuggestDispatch = { failingMetrics: [...sorted], hadCodeInfo }
```

and suppresses the next re-dispatch when the failing-metric set is unchanged **AND** every failing metric is already covered by existing guidance (`data.issues[*].type`).

**It still re-dispatches (no functionality lost) when:**
- a new bad metric appears (e.g. CLS newly goes bad) — the fingerprint set changes;
- site code becomes available (`hadCodeInfo` flips `false -> true`) so a code patch that could not be generated before might now succeed;
- guidance never actually arrived for a still-failing metric (slow/failed Mystique run) — preserving the deliberate code-fix retry loop.

The stamp is persisted via a single batched `Suggestion.saveMany` (not N+1), stamped **NEW-only** (so a leftover `PENDING_VALIDATION`-era fingerprint cannot suppress the first post-approval dispatch that authorizes code-fix generation), and preserved across re-audits by `mergeCwvData`'s shallow spread (so it never trips the sync deep-equal skip).

## Goal

Cost reduction — stop paying for redundant weekly guidance/code-fix generation on pages whose flagged CWV metrics have not changed — **with essentially no loss of functionality**.

One bounded case is deliberately narrowed: if code was already available at the first dispatch (so `hadCodeInfo` is already `true` and never flips), guidance succeeded for every failing metric, but Mystique's *code-fix* generation transiently failed and no patch landed (`isCodeChangeAvailable` stays `false`), the fingerprint is unchanged and re-dispatch is suppressed until the failing-metric set changes. This is an accepted trade-off — restoring a blanket weekly retry for it is exactly the cost this guard removes — and it assumes code-fix generation is reliable whenever guidance succeeds and code is present. Guidance failures (the common failure mode) are still retried via the `allFailingGuided` gate.

## Testing

- `auto-suggest.js` at **100%** line/branch/statement/function coverage.
- 12 new unit + e2e tests (suppression branches, first-run/legacy dispatch, new-metric dispatch, code-available re-dispatch, PENDING not stamped, batched save, best-effort stamp-persist failure).
- Full suite: **9300 passing**, lines/branches/statements 100%.

## Note

A follow-up (tracked separately) will address the equivalent gap on the V2/blackboard path, where the 168h TTL on `a_cwv_guidance`/`a_cwv_autofix`/`a_cwv_autofix_verified` sits exactly on the weekly cadence and re-runs the full Opus chain for unchanged metrics; `idea_fingerprint` there only dedupes stored results, it never gates the agents.

### DIFF
--- a/src/cwv/auto-suggest.js
+++ b/src/cwv/auto-suggest.js
@@ -10,6 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
+import { deepEqual, isObject } from '@adobe/spacecat-shared-utils';
 import { getCodeInfo } from '../accessibility/utils/data-processing.js';
 import { METRICS, THRESHOLDS } from './kpi-metrics.js';
 
@@ -41,6 +42,24 @@ function getFailingMetricInfo(allMetrics) {
   };
 }
 
+/**
+ * Deterministic "dispatch fingerprint" describing the inputs that justify (re)sending a
+ * CWV suggestion to Mystique. Stamped on the suggestion after a NEW dispatch and compared
+ * on the next weekly audit so we only re-dispatch when a relevant input changed:
+ * - the set of failing metrics grew/changed (e.g. CLS newly goes bad), or
+ * - site code became available (`hadCodeInfo` false -> true) so a code patch that couldn't
+ *   be generated before might now succeed.
+ * Both the guard (read) and `processAutoSuggest` (stamp) build it identically so a stamped
+ * marker re-compares exactly next run. Mirrors the V2/blackboard `idea_fingerprint` idea.
+ *
+ * @param {string[]} failingMetrics - metrics failing this run (lcp/cls/inp)
+ * @param {boolean} hasCodeInfo - whether a code repo/path is available for this site
+ * @returns {{ failingMetrics: string[], hadCodeInfo: boolean }}
+ */
+function buildDispatchFingerprint(failingMetrics, hasCodeInfo) {
+  return { failingMetrics: [...failingMetrics].sort(), hadCodeInfo: !!hasCodeInfo };
+}
+
 const CWV_AUTO_SUGGEST_MESSAGE_TYPE = 'guidance:cwv';
 
 /**
@@ -69,10 +88,17 @@ const CWV_AUTO_SUGGEST_MESSAGE_TYPE = 'guidance:cwv';
  * is nothing for the SME to review and the suggestion can never be approved.
  *
  * For NEW, we dispatch whenever no code patch has been produced yet
- * (`data.isCodeChangeAvailable !== true`). We deliberately ignore
- * `issues[*].value` here — text guidance being present is not proof that a code
- * patch ran successfully, and `mergeCwvData` preserves that field across every
- * re-audit, which used to silently block retries forever.
+ * (`data.isCodeChangeAvailable !== true`) AND a relevant input has changed since the
+ * last dispatch. Cost control (same flagged metric week after week shouldn't pay for
+ * fresh guidance every audit): we SUPPRESS the re-dispatch when the currently-failing
+ * metric set is already fully covered by existing guidance (`data.issues[*].type`) AND
+ * the dispatch fingerprint is unchanged from the last stamped one
+ * (`data.autoSuggestDispatch`). We still re-dispatch when a new bad metric appears or
+ * when site code becomes available (both flip the fingerprint), and — crucially — when
+ * guidance never actually arrived for a still-failing metric (`allFailingGuided` false),
+ * so a slow/failed Mystique run is still retried. We deliberately ignore
+ * `issues[*].value` (its mere presence is not proof a code patch ran) — the retry is
+ * bounded by the fingerprint + code-patch availability, not by guidance text existing.
  *
  * For PENDING_VALIDATION, re-dispatch is gated on the existing guidance being
  * missing or in the legacy aggregated format (no per-issue `source_index`), so a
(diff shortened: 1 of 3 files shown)


src/cwv/auto-suggest.js
```diff
@@ -99,6 +127,23 @@ export function shouldSendAutoSuggestForSuggestion(suggestion) {
...
+    return true;
+  }
+  const guidedMetrics = new Set(
+    (Array.isArray(data.issues) ? data.issues : []).map((i) => i.type),
+  );
+  const allFailingGuided = failingMetrics.every((m) => guidedMetrics.has(m));
+  const stored = data.autoSuggestDispatch;
+  const fingerprintUnchanged = isObject(stored)
+    && deepEqual(stored, buildDispatchFingerprint(failingMetrics, hasCodeInfo));
+  if (allFailingGuided && fingerprintUnchanged) {
```
ramboz, line 154: The old guard dispatched every week while `isCodeChangeAvailable !== true`, deliberately ignoring guidance presence so the code-fix retry loop stayed alive. This suppression re-gates on guidance presence (`allFailingGuided` from `issues[*].type`), which is the intended cost fix — but it does narrow one case the PR body claims has "zero loss of functionality":

> code was available from the start (`hadCodeInfo` already `true`), guidance succeeded for every failing metric, but Mystique's code-fix generation transiently failed / never landed a patch (`isCodeChangeAvailable` stays `false`).

Here the metric set is unchanged and `hadCodeInfo` doesn't flip, so the fingerprint matches and re-dispatch is suppressed forever (until metrics change). The `hadCodeInfo false→true` escape only covers *newly*-available code, not a failed generation on already-available code.

Is that acceptable? It likely is if code-fix generation is reliable whenever guidance succeeds + code is present, so the guidance-succeeded-but-patch-missing state doesn't persist in practice. Worth confirming that assumption, or softening the "zero loss of functionality" wording to acknowledge this bounded case. Not a blocker.

src/cwv/auto-suggest.js
```diff
@@ -205,6 +256,22 @@ export async function processAutoSuggest(context, opportunity, site) {
...
+        suggestion.setData({
+          ...suggestionData,
+          autoSuggestDispatch: buildDispatchFingerprint(failingMetrics, hasCodeInfo),
+        });
+        dispatchedNew.push(suggestion);
+      }
+    }
+
+    if (dispatchedNew.length > 0) {
+      await Suggestion.saveMany(dispatchedNew);
```
ramboz: Nit: a throw from this stamp `saveMany` propagates to the outer `catch` and fails the whole audit step, even though every SQS message already went out successfully. The fail direction is safe (next week just re-dispatches, since the fingerprint wasn't persisted), so functionally fine — but consider a local try/catch that logs a warning instead of failing the audit on a best-effort persistence step. Optional.

ramboz (approved):
## Review — approve ✅

Clean, well-scoped cost fix on the CWV legacy V1 dispatch path. The core idea — stamp a deterministic dispatch fingerprint on `NEW` dispatches and suppress the weekly re-dispatch only when the failing-metric set is unchanged **and** every failing metric is already guided — is correct, and the guard/stamp symmetry (both call `buildDispatchFingerprint`, both sort) makes the deep-equal comparison stable.

I verified the two load-bearing claims outside the diff:

- **Persistence across re-audits:** `mergeCwvData` is `{ ...existingData, ...newDataItem }` and the raw CWV entry carries no `autoSuggestDispatch`, so last week's fingerprint survives the merge and doesn't perturb the sync deep-equal skip. ✔
- **Save ordering / `calledTwice`:** `syncOpportunityAndSuggestionsStep` runs `syncOpportunitiesAndSuggestions` (its own `saveMany`) *then* `processAutoSuggest` (the stamp `saveMany`). ✔

Nice touches worth calling out: `NEW`-only stamping correctly avoids a `PENDING_VALIDATION`-era fingerprint suppressing the first post-approval dispatch; `dispatchedNew` gates the `saveMany` so an absent `dataAccess.Suggestion` can't NPE; and because the stamp save happens *after* the SQS sends, a stamp-persist failure fails safe toward re-dispatch (no lost functionality). Coverage and the 11 new tests are thorough.

Two non-blocking points inline — one on the boundary of the "zero loss of functionality" claim, one nit. Neither blocks merge.

src/cwv/auto-suggest.js
tkotthakota-adobe: Good catch on the boundary. Confirmed the analysis: the `hadCodeInfo false→true` escape only covers newly-available code, so guidance-succeeded-but-code-fix-transiently-failed on already-available code stays suppressed until the metric set changes. A code fix that re-dispatched here would reintroduce exactly the weekly cost this guard removes, so I've kept the behavior and instead documented it as a bounded, accepted trade-off — added a paragraph to the `shouldSendAutoSuggestForSuggestion` docstring (72c... latest push) and softened the PR body from "zero loss of functionality" to "essentially no loss," spelling out this case. Guidance failures (the common mode) are still retried via `allFailingGuided`.

src/cwv/auto-suggest.js
tkotthakota-adobe: Done. Wrapped the stamp `saveMany` in a local try/catch that logs a warning instead of propagating to the outer catch, so a best-effort persistence failure no longer fails the audit step after every SQS message already went out. Added a test asserting a `saveMany` throw doesn't propagate (`does not fail the audit when the stamp saveMany throws`).

PR #2825, opened 2026-07-29
feat(cwv): bow out of legacy opportunity/suggestion sync when deliveryConfig.cwvEngine=blackboard

Dereje24:
## What

Per-site legacy→blackboard CWV cutover — the **suspenders** half of Mystique's
Spec 009-04 / ADR-0022. When a site's CWV opportunity is owned by Mystique's
blackboard producer cascade (`deliveryConfig.cwvEngine === "blackboard"`), the legacy
CWV audit's `syncOpportunityAndSuggestionsStep` now **bows out**: it creates no
`type:"cwv"` opportunity / `CODE_CHANGE` suggestion rows and sends no `guidance:cwv`
message, returning `{ status: 'complete' }` early. Absent / `"legacy"` / null
`deliveryConfig` is unchanged (default legacy path).

## Why

Both flows write the **same** SpaceCat `type:"cwv"` opportunity + suggestion rows. If
both run for one site they collide: Mystique's projector keys its parent on
`(type, scope_type='site', scope_id)`, so against this audit's null-scoped opportunity
it creates a **second** active `cwv` opportunity rather than reusing it; and the two
writers key suggestions in disjoint spaces (projector `site:{id}:cwv:{hash}:{metric}`
vs. this repo's `data.url`/`pattern`), so they never dedup. A collision-free cutover
therefore needs the source to stop writing — this change.

`cwvEngine` mirrors the existing `altTextEngine` / `formsA11yEngine` per-site switches;
flipping it is the migration action, flipping it back restores this flow on the next audit.

## Testing

`test/audits/cwv.test.js` — two new cases: blackboard bow-out (no opportunity/suggestion
row created, no SQS message, PLG-alert query not reached) and legacy passthrough (null
`deliveryConfig` still creates the opportunity). Full `cwv.test.js` suite green (49
passing); `eslint src/cwv/handler.js` clean.

## Coordination

Mystique owns the belt (its `_resolve_cwv_engine` gate, Spec 009-01) and the full
cross-repo contract, projector-ownership evidence, and a duplicate-row detection recipe:
`docs/opportunities/cwv/design-cwv-blackboard-migration.md` §9.4 (Spec 009-04 / ADR-0022).
Safe cutover ordering flips `cwvEngine` and confirms this bow-out **before** enrolling the
site in Mystique's scheduler.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Julien Ramboz", "Dominique Jäggi"]
```

### DIFF
--- a/src/cwv/handler.js
+++ b/src/cwv/handler.js
@@ -59,6 +59,22 @@ export async function syncOpportunityAndSuggestionsStep(context) {
   const { Suggestion } = dataAccess;
   const siteId = site.getId();
 
+  // Legacy-source bow-out (Spec 009-04 / ADR-0022). When a site's CWV opportunity is owned
+  // by the Mystique blackboard engine (deliveryConfig.cwvEngine === 'blackboard'), this
+  // legacy audit must NOT create the shared type:"cwv" opportunity / CODE_CHANGE suggestion
+  // rows or send the guidance:cwv message — otherwise the two flows write the same SpaceCat
+  // rows and collide (the blackboard projector keys its parent on
+  // (type, scope_type='site', scope_id), so it creates a *second* active cwv opportunity
+  // rather than reusing this null-scoped one). The blackboard producer cascade owns
+  // detection→guidance→autofix and projects the customer-facing suggestions itself.
+  const cwvEngine = site.getDeliveryConfig?.()?.cwvEngine;
+  if (cwvEngine === 'blackboard') {
+    log.info(`[audit-worker-cwv] siteId: ${siteId} | Step 2: bowing out — deliveryConfig.cwvEngine=blackboard, CWV opportunity is Mystique-owned; skipping opportunity/suggestion sync and auto-suggest`);
+    return {
+      status: 'complete',
+    };
+  }
+
   log.info(`[audit-worker-cwv] siteId: ${siteId} | Step 2: Syncing opportunities and suggestions`);
 
   const opportunity = await syncOpportunitiesAndSuggestions(context);
--- a/src/cwv/handler.js
+++ b/src/cwv/handler.js
@@ -10,7 +10,11 @@
  * governing permissions and limitations under the License.
  */
 
-import { Audit, Suggestion as SuggestionModel } from '@adobe/spacecat-shared-data-access';
+import {
+  Audit,
+  Suggestion as SuggestionModel,
+  Opportunity as OpportunityModel,
+} from '@adobe/spacecat-shared-data-access';
 import { AuditBuilder } from '../common/audit-builder.js';
 import { wwwUrlResolver } from '../common/index.js';
 import { buildCWVAuditResult } from './cwv-audit-result.js';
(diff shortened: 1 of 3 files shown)


src/cwv/handler.js
```diff
@@ -59,6 +59,22 @@ export async function syncOpportunityAndSuggestionsStep(context) {
...
+  // Legacy-source bow-out (Spec 009-04 / ADR-0022). When a site's CWV opportunity is owned
+  // by the Mystique blackboard engine (deliveryConfig.cwvEngine === 'blackboard'), this
+  // legacy audit must NOT create the shared type:"cwv" opportunity / CODE_CHANGE suggestion
+  // rows or send the guidance:cwv message — otherwise the two flows write the same SpaceCat
+  // rows and collide (the blackboard projector keys its parent on
+  // (type, scope_type='site', scope_id), so it creates a *second* active cwv opportunity
+  // rather than reusing this null-scoped one). The blackboard producer cascade owns
+  // detection→guidance→autofix and projects the customer-facing suggestions itself.
+  const cwvEngine = site.getDeliveryConfig?.()?.cwvEngine;
+  if (cwvEngine === 'blackboard') {
```
ramboz: Non-blocking, cutover-lifecycle question: this bows out cleanly, but for a site that already ran on `legacy` and *then* flips to `blackboard`, this step no longer touches the pre-existing `type:"cwv"` opportunity + suggestions — so they linger in whatever status they were last left in (never marked `OUTDATED`/resolved), and Mystique's projector creates a *second*, separately-keyed opportunity. Brand-new blackboard sites are unaffected. Is the stale legacy opportunity meant to be cleaned up as an explicit migration step on the Mystique side, or should the bow-out resolve/outdate an existing one on the way out? Worth a line in the cutover runbook either way.

Also minor: Step 1 (`collectCWVDataAndImportCode`) still runs RUM collection + triggers the `code` import for blackboard sites — harmless, but slightly wasteful per audit run. Intended?

ramboz (approved):
## Review: LGTM ✅

Clean, well-scoped, well-tested. Bows the legacy CWV flow out *before* any DB read/write when `deliveryConfig.cwvEngine === "blackboard"`, which is exactly the right cut point for a collision-free per-site cutover.

**Strengths**
- Early return at the top of Step 2 — minimal blast radius, no partial writes.
- Uses the established `getDeliveryConfig?.()?.<field>` idiom (matches `metatags`/`product-metatags` `useHostnameOnly` reads) and mirrors the existing `altTextEngine`/`formsA11yEngine` per-site switches.
- Strict `=== 'blackboard'` gate: absent / `null` / `"legacy"` all fall through unchanged (verified by the null-`deliveryConfig` passthrough test).
- Tests assert the full bow-out contract: no `Opportunity` fetch/create, no `sqs.sendMessage`, PLG-alert query not reached — plus the legacy passthrough. `cwv.test.js` green, lint clean, CI green (8 success / 10 skipped conditionals), mergeable.
- Doc + inline comment clearly capture the cross-repo contract and the projector-ownership rationale.

No blockers — approving. One non-blocking cutover question inline.

src/cwv/handler.js
Dereje24: bow-out should resolve and collectCWVDataAndImportCode shouldn't need to re-run twice as blackboard already will run need. Need to double check.

Dereje24:
Addressed the review in `1050cb5e` (thanks @ramboz + the cutover-lifecycle question):

**1. Bow-out now *resolves*, not just skips.** For a site flipped legacy→blackboard, Step 1 (`collectCWVDataAndImportCode`) resolves any pre-existing legacy `type:"cwv"` opportunity (`setStatus(RESOLVED)`) and outdates its still-live `NEW`/`IN_PROGRESS` suggestions, while **preserving customer/system-touched states** (`FIXED`/`SKIPPED`/`ERROR`). The resolve lives in Step 1 (not Step 2) because Step 1 always runs on the initial trigger — Step 2 is gated on the import-worker round-trip, so a resolve there would be skipped if the import lagged. Net: no lingering active legacy row, and Mystique's projector's scoped row becomes the sole active `cwv` opportunity.

**2. Step 1 skips the RUM/PSI collection** for blackboard sites (returns an empty audit result — verified nothing consumes the persisted `cwv` audit result; the trend audits read RUM directly).

**On the import-worker hop:** I kept it. The `IMPORT_WORKER` destination's `formatPayload` requires a valid `type`, and the StepAudit framework unconditionally chains non-last steps, so omitting the import cleanly isn't possible in-repo without an import-worker-side change (and I can't verify import-worker's reaction to a no-`type` message). The clean way to skip the audit **entirely** (RUM + import + sync) for a migrated site is the `Configuration` `cwv` handler-disable — documented in the bow-out section. Happy to open an import-worker follow-up if you'd prefer the hop skipped too.

**Also:** corrected the stale "both flags are checked in this repo" line in `docs/cwv-autofix-flow.md` (`processAutoSuggest` no longer re-checks `cwv-auto-suggest`/`cwv-auto-fix`; enablement is `isHandlerEnabledForSite('cwv', site)` upstream).

**Design note:** resolve-on-bow-out supersedes Mystique Spec 013-02's transactional in-place adoption for the CWV path (they're mutually exclusive) — flagged in the Mystique design §9.4.3 for 013-02 to absorb.

Tests: `cwv.test.js` now 53 passing (+4 Step-1 cases). Lint clean.

PR #2826, opened 2026-07-29
fix: restrict CDN log filter keys to a column allowlist

akshaymagapu:
## Problem
`buildSiteFilters()` builds Athena filter clauses from the customer-configurable `cdnlogsFilter` entries. The `key` of each entry was used directly when composing the SQL, with no constraint on its contents, and filter `value` strings were not escaped. An unexpected or malformed `key` could therefore produce a query that does not match the intended, well-formed shape.

## Change
- Validate each filter `key` against an allowlist of the known CDN log columns (`url`, `host`, `x_forwarded_host`, `user_agent`, `referer`, `cdn_provider`). Anything else throws early.
- Match case-insensitively, since Athena folds column identifiers to lowercase (e.g. a stored `X_forwarded_host` resolves to the same column).
- Escape single quotes in filter `value` strings.

## Tests
- Added coverage for allowlist rejection, all allowed columns, case normalization, and value escaping.
- Full suite passes; `cdn-utils.js` at 100% coverage; lint clean.

Ref: LLMO-6621

### DIFF
--- a/src/utils/cdn-utils.js
+++ b/src/utils/cdn-utils.js
@@ -499,6 +499,23 @@ export async function shouldRecreateTable(
   }
 }
 
+// Columns present on the aggregated/raw CDN log tables that filters are allowed to
+// target. `key` comes straight from customer-controlled site config (cdnlogsFilter),
+// so it must never be interpolated into the query without being checked against this
+// allowlist (see VULN-37491 - SQL injection via cdnlogsFilter[].key).
+const ALLOWED_FILTER_KEYS = new Set([
+  'url',
+  'host',
+  'x_forwarded_host',
+  'user_agent',
+  'referer',
+  'cdn_provider',
+]);
+
+function escapeSqlLiteral(value) {
+  return String(value).replace(/'/g, "''");
+}
+
 export function buildSiteFilters(filters, site) {
   if (!filters || filters.length === 0) {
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - clean, well-scoped security fix that correctly closes the SQL injection vector.
**Complexity: LOW** - small diff, single-service.
_Changes: Restricts CDN log filter keys to a column allowlist and escapes single quotes in filter values to close SQL injection via `cdnlogsFilter[].key` (2 files)._

<details>

- nit: The thrown error interpolates the raw unsanitized `key` value (`throw new Error(\`Invalid CDN log filter key: ${key}\`)`) - if this message ever reaches structured logs or an HTTP response, the injected payload appears verbatim. Consider using `normalizedKey` or omitting the value entirely. - `src/utils/cdn-utils.js:537`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 39s | <b>Cost:</b> $3.08 | <b>Commit:</b> <code>2dc2a3986fa3a26834aae242fa714d5322c11652</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2828, opened 2026-07-29
fix: escape base URL host and path in CDN log site filters

akshaymagapu:
## Problem
`buildSiteFilters()`'s no-custom-filter branch interpolates the site's base URL **host** and **path** directly into an Athena regex string literal (`REGEXP_LIKE(host, '(?i)^(www.)?<host>$')`). Both derive from the customer-controlled base URL via `new URL()`, which preserves characters like `'`, so a crafted base URL could break out of the string literal — the same class of issue already addressed for explicit filter keys/values.

## Change
- Escape single quotes in the base URL `host` (`rootHost`) before interpolation.
- Escape single quotes in the base URL `path` (in addition to the existing regex-metacharacter escaping).

Behavior is unchanged for normal hostnames/paths (no quotes to escape); only quote characters are neutralized.

## Tests
- Added coverage for quote-bearing host and path.
- Full suite passes; `cdn-utils.js` at 100% coverage; lint clean.

Ref: LLMO-6621

### DIFF
--- a/src/cdn-analysis/handler.js
+++ b/src/cdn-analysis/handler.js
@@ -26,6 +26,7 @@ import {
   getCdnAwsRuntime,
   pathHasData,
   shouldRecreateTable,
+  escapeForAthenaRegexLiteral,
 } from '../utils/cdn-utils.js';
 import { getImsOrgId } from '../utils/data-access.js';
 import { weekClosingSundayKey } from '../utils/date-utils.js';
@@ -510,7 +511,10 @@ export async function processCdnLogs(auditUrl, context, site, auditContext) {
           hour,
           hourFilter,
           bucket: bucketName,
-          host,
+          // host is the site's base-URL host, interpolated into a REGEXP_LIKE
+          // pattern in insert-aggregated.sql; escape it so a crafted base URL
+          // cannot break out of the string literal or broaden the match (VULN-37491).
+          host: escapeForAthenaRegexLiteral(host),
           serviceProvider,
         }),
         loadSql(cdnType, 'insert-aggregated-referral', {
--- a/src/utils/cdn-utils.js
+++ b/src/utils/cdn-utils.js
@@ -520,7 +520,10 @@ export function buildSiteFilters(filters, site) {
   if (!filters || filters.length === 0) {
     const baseURL = site.getBaseURL();
     const { host, pathname } = new URL(baseURL);
-    const rootHost = host.replace(/^www\./, '');
+    // host/pathname come from the customer-controlled base URL and are interpolated
+    // into the Athena regex string literal below. Escape single quotes so they cannot
+    // break out of the literal (VULN-37491, base_url vector).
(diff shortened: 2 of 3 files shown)


MysticatBot (changes requested):
Hey @akshaymagapu,

**Verdict: Request changes** - the SQL literal escape is correct but `rootHost` still allows regex metacharacter injection.
**Complexity: LOW** - small targeted security fix, single service.
_Changes: Escapes single quotes in base URL host and path to prevent Athena SQL literal breakout in CDN log filter queries (2 files)._

## Must fix before merge
1. [Important] Missing regex-metacharacter escaping on `rootHost` allows cross-tenant filter broadening via regex alternation - `src/utils/cdn-utils.js:523` (details inline)

<details>

- suggestion: Extract a shared helper like `escapeForAthenaRegex(value)` that chains regex-metacharacter escaping then SQL-literal escaping in the correct order. Both `rootHost` and `escapedPath` need the same two-layer defense, and a helper makes it impossible to forget one layer. - `src/utils/cdn-utils.js:515`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 38s | <b>Cost:</b> $3.74 | <b>Commit:</b> <code>e3cdb43d8092cc626a080afa8d1bf34c038ebe4a</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - prior findings addressed, escaping logic is correct and complete.
**Complexity: LOW** - small targeted security fix, single service.
_Changes: Escapes single quotes and regex metacharacters in base URL host and path to prevent Athena SQL literal breakout and regex match broadening in CDN log site filters (3 files)._

<details>

- suggestion: Add a standalone unit test for `escapeForAthenaRegexLiteral` covering edge cases (empty string, only quotes, only backslashes, mixed metacharacters + quotes). The function is exported and security-critical but currently only exercised indirectly through `buildSiteFilters`. A dedicated test would catch composition regressions (e.g., swapped call order) independently of the higher-level caller. - `test/utils/cdn-utils.test.js`
</details>

- Regex-metacharacter escaping on `rootHost` now handled by `escapeForAthenaRegexLiteral()` which chains `escapeRegexMetacharacters()` then `escapeSqlLiteral()`.
- Shared helper extracted as requested - `escapeForAthenaRegexLiteral` is reused across both `buildSiteFilters` and the handler's SQL template interpolation path.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 19s | <b>Cost:</b> $6.53 | <b>Commit:</b> <code>b4f845d339196a821e787d8ab48bb2778a25c40c</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2833, opened 2026-07-30
feat(autofix): honor isEdited for summarization, readability & FAQs (LLMO-6537)

baranovskyalexandr:
## Summary

- Preserve customer-edited suggestions during re-audit across summarization, readability, and FAQs handlers using a durable `data.isEdited` flag set only by the UI edit-save action (never inferred from `updatedBy`).
- Each handler's `mergeDataFunction` preserves domain-specific edited fields (`summarizationText`, `improvedText`, `item`) and the write-once original snapshot while allowing non-edited metadata to refresh.
- Readability's `mergeStatusFunction` short-circuits on `isEdited` to prevent flipping edited suggestions to SKIPPED.
- The OUTDATED sweep in `syncSuggestions` and the reconcile path in `reconcileDisappearedSuggestions` both exempt `isEdited` suggestions from being aged out or auto-reconciled to FIXED.
- Removes the `updatedBy`-based protection from PR #2803 (LLMO-6483) — `updatedBy` gets mutated by non-edit actions (rollback, move Ignored→New), causing false freezing.
- Adds missing `edgeOptimizeStatus` guard to FAQs handler (parity with summarization/readability).

## Test plan

- Unit tests for all three guidance handlers (`mergeDataFunction` with `isEdited`, `edgeDeployed`, `edgeOptimizeStatus`)
- Unit tests for readability guidance handler's `mergeStatusFunction` with `isEdited` short-circuit
- Unit tests for readability identify handler's `mergeDataFunction` edge + isEdited guards
- Unit tests for `isEdited` OUTDATED sweep exemption in `syncSuggestions`
- Unit tests for `reconcileDisappearedSuggestions`: NEW+isEdited skipped, OUTDATED+isEdited still reconciles
- All existing edge-guard and sync tests continue to pass (296 passing)

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Duy Nguyen", "@sneharora1", "Dominique Jäggi"]
```


src/experimentation-opportunities/guidance-high-organic-low-ctr-handler.js
```diff
@@ -14,12 +14,20 @@ import { notFound, ok } from '@adobe/spacecat-shared-http-utils';
...
 const MAX_HIGH_ORGANIC_LOW_CTR_OPPORTUNITIES = 3;
 
+/**
+ * Checks if any suggestions in the array were manually modified (updatedBy !== 'system')
+ * @param {Array} suggestions - Array of suggestion objects
+ * @returns {boolean} - True if any suggestion was manually modified
+ */
 function hasManuallyModifiedSuggestions(suggestions) {
-  return suggestions.some(isManuallyEditedSuggestion);
+  return suggestions.some((suggestion) => {
```
sneharora1, line 27: Should we handle this as well?

src/summarization/guidance-handler.js
```diff
@@ -64,6 +64,19 @@ async function addSuggestions(
...
+      // Do not overwrite a customer-edited summary. isEdited is set only by the UI
+      // edit-save action (never inferred from updatedBy); preserve the edited text and
+      // the write-once original snapshot while letting the rest of the data refresh
+      // (LLMO-6537).
+      if (existingData.isEdited) {
+        return {
+          ...existingData,
+          ...newData,
+          summarizationText: existingData.summarizationText,
+          originalSummarizationText: existingData.originalSummarizationText,
```
sneharora1: **Bootstrap gap for `original*` fields**

This preserves `existingData.originalSummarizationText` — but when is that field first written?

For suggestions created **before this PR is deployed**, a user editing for the first time will have `existingData.originalSummarizationText === undefined`. The merge returns `originalSummarizationText: undefined`, losing the system-generated original permanently. That's the one opportunity to capture it for pipeline-feedback use.

Two options:
1. **UI sets it on first edit** — when `handleSaveEdit` fires and `data.originalSummarizationText` is not yet present, include `originalSummarizationText: existingData.data.summarizationText` in the PATCH payload alongside `isEdited: true`.
2. **Audit worker bootstraps it** — in this `mergeDataFunction`, fall back to `existingData.summarizationText` when `existingData.originalSummarizationText` is missing:
```js
originalSummarizationText: existingData.originalSummarizationText ?? existingData.summarizationText,
```

Same gap applies to `originalItem` (FAQs) and `originalImprovedText` (readability). Option 2 is the simpler fix since it doesn't require a coordinated UI change — but worth an explicit decision either way.

baranovskyalexandr: Good catch — this is now covered from both sides:

1. **UI (option 1):** `handleSaveEdit` already writes a write-once snapshot on first edit: `...(data.originalSummarizationText == null && { originalSummarizationText: data.summarizationText })`. Same for readability (`originalImprovedText`) and FAQs (`originalItem`).

2. **Worker (option 2, just added in 910eefa):** As a belt-and-suspenders fallback for pre-existing suggestions where the UI hasn't set the field yet, `mergeDataFunction` now falls back: `originalSummarizationText: existingData.originalSummarizationText ?? existingData.summarizationText`. Applied to all three handlers.

src/experimentation-opportunities/guidance-high-organic-low-ctr-handler.js
baranovskyalexandr: No, this handler has a fundamentally different update model — it deletes and recreates all suggestions on every re-audit (line 157–159), so there's nothing to merge. The local `hasManuallyModifiedSuggestions` (line 26) is its own separate guard that skips the entire update when any suggestion has been manually touched. It's scoped to experimentation only and predates `isEdited`.

The `isEdited` pattern (soft-merge via `mergeDataFunction`) doesn't apply here — there's no inline content editing in the experimentation UI, so nothing to mark as edited. The current `updatedBy` guard is overcautious but safe: it may block re-audit updates after a rollback/move, but that's a minor inconvenience vs. losing data. Rearchitecting this handler from delete-recreate to merge is out of scope for LLMO-6537.

duynguyen (approved):
Approved.

Nit: please add a regression test documenting the intentional FAQ key-drift behavior: when a customer edits the question/answer and data.isEdited is set, a re-audit preserves that customer row while the prompt-bank question is created or refreshed as a separate system suggestion.

PR #2837, opened 2026-07-31
feat(offsite): structured key=value logging for offsite-brand-presence

rarescheseli:

## Related Issues

Thanks for contributing!

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Alexandru-Florin Tudoran", "Catalin Grigore", "Dominique Jäggi"]
```


src/utils/offsite-logging.js
```diff
@@ -0,0 +1,163 @@
...
+  SPACECAT: 'spacecat',
+};
+
+// Canonical order so a human scanning logs always sees the dimensions in the same place.
+const FIELD_ORDER = [
+  'domain', 'audit', 'event', 'outcome', 'peer', 'direction',
+  'siteId', 'auditId', 'opportunityId', 'jobId',
+];
+
+function renderField(key, value) {
```
grigoreadobe, line 96: issue (blocking): `appendFields` only escapes values passed through the `fields` parameter; the `message` string is concatenated verbatim with no escaping (`return parts.length > 0 ? \`${message} ${parts.join(' ')}\` : message;`). Multiple call sites interpolate attacker-influenceable data straight into `message` - e.g. `offsite-brand-presence/handler.js` (`\`Excluding ${rawUrl}\`` where `rawUrl` comes from DRS-scraped third-party page content), `cited-analysis/handler.js` (`\`Excluding ${entry.url}\``), `cited-analysis/guidance-handler.js` (`data.errorMessage` from the Mystique response), and `drs-status-handler.js` (`baseURL`/`err.message` repeatedly). A crafted URL such as `http://evil.example/x outcome=success auditId=<id>` reaching one of these call sites produces a log line with two `outcome=` tokens - one forged, one real - directly undermining the `stats ... by outcome` dashboards this PR is built to feed.

Separately, `renderField`'s quoting (`/[\s"=]/.test(str) ? \`${key}="${str.replace(/"/g, "'")}"\` : ...`) wraps but never strips an embedded newline, so a value containing `\n outcome=success` becomes a literal line break inside the log call - most shippers (CloudWatch, Fluent Bit without multiline stitching) split on `\n`, turning one log call into two lines, the second an unquoted, forged, standalone record.

Fix: (1) strip/escape control characters (\r, \n) in `renderField` in addition to the existing "->' substitution; (2) stop interpolating externally-sourced strings (rawUrl, entry.url, baseURL, data.errorMessage, error.message) directly into `message` templates - route them through `fields` instead so they get quoted, or add a small `sanitizeForLog(str)` helper applied consistently; (3) add adversarial tests in test/utils/offsite-logging.test.js for a value containing `" outcome=success`, an embedded newline, and multiple `=` characters, asserting no forged token can escape the quoted value.

src/cited-analysis/guidance-handler.js
```diff
@@ -193,7 +238,7 @@ export default async function handler(message, context) {
 
     return ok();
   } catch (error) {
-    log.error(`[Cited] Error processing cited analysis: ${error.message}`, error);
+    olog.failure('guidance_complete', `Error processing cited analysis: ${error.message}`, { errorName: error.name });
```
grigoreadobe: issue (blocking): The old code (`log.error(msg, error)`) passed the raw Error object as a second arg, so util.inspect still rendered the full stack trace for human/CloudWatch debugging even though it wasn't Splunk-extractable. The new `olog.failure('guidance_complete', ..., { errorName: error.name })` drops the stack entirely - only the exception class name survives, in exactly the catch-all path for unexpected/unclassified failures where the stack matters most. The same pattern repeats in reddit-analysis/guidance-handler.js and youtube-analysis/guidance-handler.js.

Fix: append a bounded error.stack (e.g. first few lines) as an additional field or into the free-text portion, or keep passing the raw error as a genuine second arg alongside the taxonomy string purely for stack capture.

src/offsite-brand-presence/handler.js
```diff
@@ -575,10 +594,13 @@ async function triggerDrsScraping(
   redditCommentsParams = {},
 ) {
   const { log } = context;
+  const olog = createOffsiteLogger(log, { audit: AUDIT.BRAND_PRESENCE, siteId });
   const drsClient = DrsClient.createFrom(context);
 
   if (!drsClient.isConfigured()) {
-    log.error(`${LOG_PREFIX} DRS_API_URL or DRS_API_KEY not configured, skipping DRS scraping`);
+    olog.warn('drs_submit', 'DRS_API_URL or DRS_API_KEY not configured, skipping DRS scraping', {
```
grigoreadobe: issue (blocking): This line was previously `log.error(...)` ("DRS_API_URL or DRS_API_KEY not configured, skipping DRS scraping") and is now `olog.warn(...)` - a real severity downgrade for a genuine operational misconfiguration, not just a naming/taxonomy change. Any existing alert keyed on level=error for this line would stop firing. The ADR frames this PR as "silent failures made loud," but this is the opposite direction for an existing loud line.

If the downgrade to warn is intentional (aligning with the sibling "no imsOrgId" skip case a few lines below, which already uses warn), please call that out explicitly in the PR description so anyone owning the corresponding alert knows to update it; otherwise restore error/failure level.

src/offsite-brand-presence/drs-status-handler.js
```diff
@@ -278,7 +339,9 @@ export default async function offsiteBrandPresenceDrsStatusHandler(message, cont
       const result = await drsClient.getJob(job.jobId);
       return { ...job, status: result?.status, error: result?.error_message };
     } catch (err) {
-      log.warn(`${LOG_PREFIX} getJob failed for ${job.jobId}: ${err.message}`);
+      olog.warn('drs_poll', `getJob failed for ${job.jobId}: ${err.message}`, {
+        peer: PEER.DRS, jobId: job.jobId, errorName: err.name,
```
grigoreadobe: issue (blocking): The new `jobId: job.jobId` field here is a per-DRS-job identifier, but the platform-wide logWrapper (per this repo's own @adobe/spacecat-shared-utils precedent, cited in ADR 002) already injects `[jobId=...]` from a different concept entirely - the async-queue/message job id. Two structurally unrelated identifiers now render as `jobId=` in the same log stream; a Splunk `stats ... by jobId` or an alert keyed on jobId will silently mix them.

Rename this one (e.g. drsJobId) before merge - cheap now, expensive once dashboards/alerts exist against the current name.

test/utils/offsite-audit-utils.test.js
```diff
@@ -238,37 +265,37 @@ describe('offsite-audit-utils', () => {
...
       };
 
-      const result = await filterUrlsByDrsStatus(urls, ['ds1'], siteId, drsClient, log, '[T]');
+      const result = await filterUrlsByDrsStatus(urls, ['ds1'], siteId, drsClient, olog);
 
       expect(result.urls).to.deep.equal(urls);
       expect(result.counts.determined).to.equal(false);
-      expect(log.warn).to.have.been.calledWithMatch(/DRS lookup returned null/);
-      expect(log.warn).to.have.been.calledWithMatch(/All DRS lookups failed or returned null/);
+      expect(olog.warn).to.have.been.calledWith('drs_availability', sinon.match(/DRS lookup returned null/), sinon.match.object);
```
grigoreadobe: issue (blocking): This assertion and its siblings a few lines below ('DRS lookup failed for datasetId=ds1', 'All DRS lookups failed', and the url_limit_resolve debug assertion) use `sinon.match.object` for the extra-fields argument, verifying only the event name and message text, not that `peer: PEER.DRS`, `datasetId`, `errorName`, `reason`, etc. actually thread through correctly from offsite-audit-utils.js. A dropped or mistyped field at these call sites would not be caught, even though the same test file elsewhere (`sinon.match({ reason: 'not_configured' })`) demonstrates the tighter pattern.

Replace sinon.match.object with content-specific matchers, e.g. `sinon.match({ peer: PEER.DRS, datasetId: 'ds1', errorName: 'Error' })`.

grigoreadobe (changes requested):
Hey @rarescheseli,

**Verdict: Request changes** - a genuine log-injection vulnerability in the new shared logging helper needs a scoped fix before this taxonomy is relied on for Splunk dashboards/alerts, plus a few real observability regressions worth closing first.
**Complexity: HIGH** - large diff (2030+ lines across 27 files), single service, no elevated-risk domains touched but crosses the size threshold on its own.
_Changes: Adds a shared `key=value` structured logging taxonomy across the offsite audit pipeline (cited/reddit/youtube analysis, brand-presence) and fills several previously-silent failure-logging gaps, backed by its own spec and ADR (27 files)._
Note: Recommend a human read before merge - this change modifies `docs/decisions/002-offsite-structured-logging-taxonomy.md` and `docs/specs/2026-07-31-offsite-structured-logging.md`, both architecturally-significant paths. The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Critical] `renderField`/`appendFields` only escape values passed through `fields`, never the free-text `message` string, and multiple call sites interpolate externally-sourced content (scraped URLs, Mystique error text) directly into the message - allowing a crafted value to forge fake `key=value` tokens (including `outcome=success`) or, via an embedded newline, split a single log call into two forged log lines - `src/utils/offsite-logging.js:57` (details inline)
2. [Important] The three guidance-handler top-level catch-alls dropped the error stack trace entirely (`log.error(msg, error)` -> `olog.failure(..., { errorName: error.name })`), a real debuggability regression in exactly the unclassified-failure path this taxonomy is meant to make more visible, not less - `src/cited-analysis/guidance-handler.js:241` (and the reddit/youtube equivalents) (details inline)
3. [Important] An existing `error`-level log ("DRS_API_URL or DRS_API_KEY not configured") was silently downgraded to `warn`, a real severity/alerting change that contradicts this PR's own "behavior unchanged, logging only" scope - `src/offsite-brand-presence/handler.js:601` (details inline)
4. [Important] The new DRS-poll `jobId` field collides with the platform-wide async-queue `jobId` that `logWrapper` already injects into the same log stream, so a Splunk query/alert keyed on `jobId` will silently mix two unrelated identifier namespaces - `src/offsite-brand-presence/drs-status-handler.js:343` (details inline)
5. [Important] Several `offsite-audit-utils.test.js` assertions use `sinon.match.object` for the extra-fields argument instead of matching specific field content, so a dropped/mistyped field (e.g. `errorName`, `datasetId`) at these call sites would not be caught - `test/utils/offsite-audit-utils.test.js:278` (details inline)

<details>
<summary>Non-blocking (9): minor issues and suggestions</summary>

- nit: `.warn()` vs `.skip()` are used inconsistently for equivalent "benign skip" conditions (e.g. drs-status-handler's "No jobs to poll" vs cited-analysis's "URL store empty" scoped-scrape wait), so a Splunk `outcome=skip` query now spans both `info` and `warn` levels for conceptually the same case - worth a one-line guideline in the ADR - `src/offsite-brand-presence/drs-status-handler.js`, `src/cited-analysis/handler.js`
- nit: `url_limit_resolve`'s success log omits the resolved `urlLimit` value as a field despite it being the whole point of the event and already computed - `src/cited-analysis/guidance-handler.js`, `src/reddit-analysis/guidance-handler.js`
- suggestion: `classifyFetchFailure`'s validation-vs-fetch classification relies on regex-matching `error.message` substrings owned by a different module (`analysis-fetch.js`), repeated across all three guidance handlers - a typed/tagged error would remove this fragile coupling
- nit: a single suggestion-sync failure now logs two `outcome=failure` lines (inner `suggestion_sync` catch, then the outer `guidance_complete` catch); likely intentional drill-down, but worth a one-line comment saying so - cited/reddit/youtube `guidance-handler.js`
- suggestion: the three near-identical analysis-handler implementations mechanically replicate the new logging boilerplate 3x rather than sharing a wrapper; correctly out of scope for this logging-only PR, but worth a follow-up ticket now that the duplication has grown
- nit: `renderField`'s quote substitution (`"` -> `'`) is lossy, not general-purpose escaping; a code comment warning against reuse for arbitrary user text would help (ties into the Critical fix above)
- nit: `String(value)` coercion in `renderField` silently turns an accidentally-passed object into `[object Object]` rather than surfacing a bug
- suggestion: centralize the repeated `error.message`/`error.name` interpolation pattern across all five handler files into a small `errorField(err)` helper - reduces duplicated logic and closes most of the remaining injection surface in one move
- nit: ADR-002's metadata table omits `Author`/`Deciders`/`Supersedes`/`Superseded by` fields the repo's own ADR template specifies (ADR-001 includes them) - `docs/decisions/002-offsite-structured-logging-taxonomy.md`

</details>

src/cited-analysis/guidance-handler.js
```diff
@@ -27,9 +27,27 @@ import {
...
+ * failures are `fetch`. The messages come from analysis-fetch.js / assertPresignedUrl.
+ *
+ * @param {Error} error
+ * @returns {'validation'|'fetch'}
+ */
+function classifyFetchFailure(error) {
+  return /presignedUrl|not JSON|too large|content-type/i.test(error.message)
+    ? 'validation'
+    : 'fetch';
+}
```
rarescheseli, line 52: this should be moved in an util since it's present in all offsite `guidance-handler.js` files.

src/cited-analysis/guidance-handler.js
```diff
@@ -27,9 +27,27 @@ import {
   resolveEvergreenOffsiteOpportunity,
   isSuppressedRun,
 } from '../common/offsite-refresh.js';
+import { createOffsiteLogger, AUDIT, PEER } from '../utils/offsite-logging.js';
 
 const AUDIT_TYPE = Audit.AUDIT_TYPES.CITED_ANALYSIS;
-const LOG_PREFIX = '[Cited]';
+// Human prefix for the two shared, untouched utils that still log via a passed-in prefix
+// string (logOffsiteLlmUsage + applyScopeToOpportunity). All other logging in this file goes
+// through the bound offsite logger (createOffsiteLogger), which emits the same prefix.
```
rarescheseli, line 37: a bit too verbose, could be rephrased.

src/cited-analysis/handler.js
```diff
@@ -35,8 +35,15 @@ import { computeTopicsFromBrandPresence } from '../utils/offsite-brand-presence-
...
 import { postMessageOptional } from '../utils/slack-utils.js';
+import {
+  createOffsiteLogger, withAuditPersistLog, AUDIT, OUTCOME, PEER,
+} from '../utils/offsite-logging.js';
 
-const LOG_PREFIX = '[Cited]';
+// Human prefix for the one offsite-audit-utils helper that still logs via a passed-in prefix
+// string (resolveEnableBrandProfile, shared with the offsite-brand-presence handler). All other
+// logging in this file goes through the bound offsite logger (createOffsiteLogger), which emits
+// the same `[offsite:<audit>]` prefix.
```
rarescheseli, line 47: rephrase

src/cited-analysis/guidance-handler.js
rarescheseli: maybe could be moved to `src/utils/offsite-audit-utils.js`

src/utils/offsite-logging.js
rarescheseli: Fixed. `sanitizeForLog` strips C0 control chars (incl. \r\n) from both field values and the free-text message before anything is emitted (src/utils/offsite-logging.js:74). `renderField` (src/utils/offsite-logging.js:91) applies it and quotes any value containing whitespace/`=`/`"`, so a crafted value can no longer forge a second `key=value` token or split the line. Beyond the helper itself, every call site that previously interpolated external text into the message (`rawUrl`, `entry.url`, Mystique's `data.errorMessage`, and every `${error.message}`/`${err.message}` tail across the handlers, `offsite-refresh.js`, and `offsite-audit-utils.js`) now routes that text through quoted fields instead — the new `errorField(err)` helper (src/utils/offsite-logging.js:138) centralizes the error case. Added adversarial tests for a value containing `\" outcome=success`, an embedded newline, and multiple `=` characters (test/utils/offsite-logging.test.js).

src/cited-analysis/guidance-handler.js
rarescheseli: Fixed. The three guidance-handler catch-alls now pass the raw `Error` as a genuine 4th argument to `failure(...)`, e.g. src/cited-analysis/guidance-handler.js:245: `olog.failure('guidance_complete', 'Error processing cited analysis', { ...errorField(error) }, error)`. `createOffsiteLogger`'s `emit` forwards that as a real second arg to `log.error` purely for stack capture, while the Splunk taxonomy stays in the single sanitized string. `return badRequest(...)` / the ack behavior is unchanged — logging only, per the PR's stated scope. Same pattern in reddit/youtube `guidance-handler.js`.

src/offsite-brand-presence/handler.js
rarescheseli: Fixed — restored to `olog.failure` (error level, `outcome=failure`) at src/offsite-brand-presence/handler.js:603. The downgrade wasn't intentional; the sibling "no imsOrgId" case a few lines below is intentionally a warn/skip since that's a per-org config gap rather than an operational misconfiguration, and stays as-is.

src/offsite-brand-presence/drs-status-handler.js
rarescheseli: Fixed. Renamed the field key to `drsJobId` (src/offsite-brand-presence/drs-status-handler.js:343: `drsJobId: job.jobId`), and the same rename on the "DRS job created" log in `offsite-brand-presence/handler.js`. Only the log field key changed — the underlying `job.jobId`/`result.job_id` variables are untouched. `drsJobId` is in the helper's canonical `FIELD_ORDER` (src/utils/offsite-logging.js) so it renders in the same position as before.

test/utils/offsite-audit-utils.test.js
rarescheseli: Fixed. Replaced every `sinon.match.object` in that describe block with content-specific matchers against the actual fields each call site passes — e.g. test/utils/offsite-audit-utils.test.js:189/210/239 assert `peer: PEER.DRS` plus `removed`/`remaining`/`datasetId`, and the two `DRS lookup failed`/`All DRS lookups failed` assertions (lines 279-280) now assert `reason` and, where applicable, `errorName`/`errorMessage`. Same tightening applied to the `url_limit_resolve` debug assertion.

src/wikipedia-analysis/handler.js
```diff
@@ -95,7 +101,7 @@ function resolveWikipediaUrlOverride(auditContext, log) {
   const md = auditContext?.messageData;
```
grigoreadobe, line 102: issue (blocking): `resolveWikipediaUrlOverride` calls the raw `log?.info?.()` directly - not `createOffsiteLogger`/`olog` - and interpolates `wikiVal`, `wikipediaVal`, `rawOverride`, and `normalized` straight into template strings with zero sanitization (no `sanitizeForLog`, no `appendFields`). These values come from `auditContext.messageData.wikiUrl`/`wikipediaUrl`, i.e. Slack-supplied custom arguments on the audit-trigger message. A crafted value containing a real `\n` (e.g. `wikiUrl = "x\nFAKE [offsite:wikipedia] event=audit_start outcome=success siteId=<victim>"`) will split into two literal log lines, the second a fully forged entry - exactly the more severe half of the original Critical finding, unpatched, in brand-new code.

Fix: route this through the bound `olog` (already constructed at `runWikipediaAnalysisAudit` just after this function is called), or at minimum wrap with `sanitizeForLog`:

```js
import { sanitizeForLog } from '../utils/offsite-logging.js';
log?.info?.(sanitizeForLog(`${HUMAN_PREFIX} Wikipedia URL override: wikiUrl=${wikiVal} wikipediaUrl=${wikipediaVal}`));
```

Better: pass `olog` down and emit as `olog.debug('url_override', 'Wikipedia URL override', { wikiUrl: wikiVal, wikipediaUrl: wikipediaVal })` so the values are quoted via `fields`.

src/utils/offsite-audit-utils.js
```diff
@@ -497,27 +504,32 @@ export async function filterUrlsByDrsStatus(urls, datasetIds, siteId, drsClient,
...
 }
 
-export function resolveMystiqueUrlLimit(auditContext, log, logPrefix) {
-  const prefix = logPrefix ?? '';
+export function resolveMystiqueUrlLimit(auditContext, olog) {
   const ctx = auditContext ?? {};
   const raw = ctx.messageData?.urlLimit;
   if (raw === undefined || raw === null || raw === '') {
     return MYSTIQUE_URLS_LIMIT;
   }
```
grigoreadobe, line 519: issue (blocking): This shared `makeResolveOverride` helper (used by every offsite audit family for `enableBrandProfile`/`enableSemrush` resolution) takes the raw platform `log` here, not `olog`, and interpolates `JSON.stringify(raw).slice(0, 100)` directly into the message with no sanitization. `JSON.stringify` happens to escape literal newlines, accidentally mitigating line-splitting, but `=`/`"` pass through unescaped, so same-line token forgery remains possible, and there's no defense-in-depth if a future caller passes an already-stringified value. This is called from every one of `cited-analysis/handler.js`, `reddit-analysis/handler.js`, `youtube-analysis/handler.js`, and `offsite-brand-presence/handler.js` on Slack-custom-arg-controlled input.

Separately, `companyName` is still interpolated raw into free-text messages (not routed through `fields`) across all four analysis families, e.g. `cited-analysis/handler.js:304` (`Config: companyName=${citedConfig.companyName}, ...}`, where the same value is also safely passed via `fields` two lines later - the raw copy is simply redundant and unsafe) and `cited-analysis/handler.js:610`, plus the equivalent `guidance_receive`/`guidance_complete`/`mystique_dispatch` messages in all four `guidance-handler.js`/`handler.js` pairs. `wikipedia-analysis/guidance-handler.js`'s `company` value is sourced from Mystique's echoed analysis payload rather than site config directly - one hop closer to attacker-influenced content than a site-config value.

Fix: route through `fields` exclusively (delete the redundant raw message copy where a field duplicate already exists, e.g. `cited-analysis/handler.js:304-306`), and fix `makeResolveOverride` once at this single chokepoint rather than patching each of its four call sites separately - either route through the caller's `olog` or wrap with `sanitizeForLog` at minimum.

src/offsite-brand-presence/drs-status-handler.js
```diff
@@ -275,9 +333,12 @@ export default async function offsiteBrandPresenceDrsStatusHandler(message, cont
     enableBrandProfile, urlLimit, enableSemrush,
   } = auditContext;
   const { channelId, threadTs } = slackContext;
+  const olog = createOffsiteLogger(log, { audit: AUDIT.BRAND_PRESENCE, siteId });
 
   if (jobs.length === 0) {
-    log.warn(`${LOG_PREFIX} No jobs to poll, skipping (site ${siteId})`);
+    olog.warn('drs_poll', `No jobs to poll, skipping (site ${siteId})`, {
```
grigoreadobe: issue (blocking): The ADR's new "## Conventions" section explicitly names "no jobs to poll" as the canonical example for `.skip()` (info level, `outcome=skip`), but this line uses `.warn()` (warn level) for exactly that scenario, only setting `outcome: SKIP` via the extra fields rather than the log level itself. Compare to `offsite-brand-presence/handler.js`'s `scheduleDrsStatusPoll`, which correctly uses `.skip()` for the analogous "no successfully submitted DRS jobs" condition. Separately, `offsite-brand-presence/handler.js`'s `drs_poll_schedule` failure path converts a previously-swallowed, non-fatal SQS-scheduling failure into `olog.failure(...)` (error/paging level) with no comment explaining the escalation - the audit run continues regardless, so this will now alert on every transient SQS hiccup in a path that was deliberately tolerant of failure.

Also, `src/wikipedia-analysis/handler.js:271` (`olog.success('config_resolve', configMsg)`) passes no third `fields` argument at all - unlike its three siblings (e.g. `cited-analysis/handler.js:304`, which passes `{ companyName, competitors }`), so `companyName`/`website`/`wikipediaUrl` exist only as unstructured prose for this event, and any Splunk query keyed on those fields silently returns nothing for wikipedia.

Fix: align the two skip/severity call sites with the ADR's own convention (or add a comment explaining the intentional escalation), and add the missing `fields` argument to wikipedia's `config_resolve` call.

docs/decisions/002-offsite-structured-logging-taxonomy.md
```diff
@@ -0,0 +1,92 @@
...
+DB write that is acked without retry, etc.).
+
+We wanted one consistent, machine-parseable taxonomy that serves human
+debugging, Splunk queries/dashboards/alerts, and agents — and to close the
+silent-event gaps.
+
+## Decision
+
+1. **A single taxonomy, emitted as `key=value` tokens in the message string.**
+   Every offsite log line carries `domain=offsite`, `audit=<cited|reddit|youtube|brand-presence>`,
```
grigoreadobe: issue (blocking): Both this ADR and `docs/specs/2026-07-31-offsite-structured-logging.md` still enumerate `audit=<cited|reddit|youtube|brand-presence>` and `peer=<drs|mystique|url_store|s3|postgres|sharepoint|slack|sqs|spacecat>` (lines 28/30 here), omitting `wikipedia` (now `AUDIT.WIKIPEDIA` in `src/utils/offsite-logging.js`) and `semrush` (now `PEER.SEMRUSH`, actively used in the brand-presence data-source routing). Both docs also still say `jobId` (line 31 here) where the code now uses `drsJobId` per the prior review's fix. Both docs carry `Status: Accepted`/`Status: Implemented` - exactly the state this repo's CLAUDE.md targets with "Specs should reflect the implemented state once work is complete." A reader building a Splunk query off either doc today would not know `wikipedia`/`semrush` are valid values, or that the DRS job-id field is `drsJobId` not `jobId`.

Fix: update both docs' enum lists and the `jobId`->`drsJobId` reference in the same PR that introduced the drift.

test/audits/wikipedia-analysis/guidance-handler.test.js
```diff
@@ -465,6 +465,31 @@ describe('Wikipedia Analysis Guidance Handler', function () {
       );
```
grigoreadobe, line 470: issue (blocking): The Mystique-error test only asserts a message substring (`/Mystique returned an error/`), where the equivalent cited/reddit/youtube tests assert the full taxonomy (`event=guidance_receive`, `outcome=failure`, `peer=mystique`, and that the upstream error text is correctly quoted into `mystiqueError="..."`) - wikipedia's version doesn't verify the `mystiqueError` field is emitted or quoted at all. The generic outer-catch test similarly asserts only a message substring plus `sinon.match.any` for the second logger argument, where cited's equivalent asserts `event=guidance_complete`, `outcome=failure`, `errorName=Error`, and that the second arg is a real Error instance. These are exactly the code paths this PR is hardening, so a regression in either would currently pass CI. Separately, the new Semrush routing/failure logging in `test/audits/offsite-brand-presence.test.js` (the `describe('Semrush source ...')` block) has zero log-content assertions on the new `data_source_select`/`brand_data_load source=semrush` events - it checks only the business-outcome result, not `event=`/`outcome=`/`peer=semrush`/`reason=`, unlike this same PR's DRS logging tests in the same file, which were upgraded with exactly such assertions.

Fix: tighten the two wikipedia assertions to match the sibling taxonomy-checking pattern, and add `event=`/`outcome=`/`peer=semrush`/`reason=` assertions to the existing Semrush hard-stop/fallback/degradation tests (the behavioral scenarios already exist; they just need the log-content assertion added).

grigoreadobe (changes requested):
Hey @rarescheseli,

**Verdict: Request changes** - the newline-splitting log-injection vector from the last round is genuinely closed for the shared logging path, but the same bug class reappears unmitigated in brand-new code, plus several residual gaps and a scope-driven documentation drift.
**Complexity: HIGH** - large diff (2382+ lines across 31 files, up from 27), single service, no elevated-risk domains touched but crosses the size threshold on its own.
_Changes: Extends the offsite structured-logging taxonomy to a new `wikipedia-analysis` audit family and Semrush data-source routing, and addresses the prior review's five findings (31 files)._

## Must fix before merge
1. [Critical] `resolveWikipediaUrlOverride` bypasses the sanitizer entirely and reintroduces newline-based log-line-splitting on Slack-controlled input, in brand-new code - `src/wikipedia-analysis/handler.js:100` (details inline)
2. [Important] The injection-safety contract from the prior fix isn't uniformly applied yet: `companyName` is still interpolated raw into free-text messages across all four analysis families, and a shared chokepoint used by all five audit families bypasses the sanitizer for Slack-controlled input - `src/utils/offsite-audit-utils.js:567` (details inline)
3. [Important] Two conventions this PR itself just established in the ADR aren't fully honored yet: the new `wikipedia` family's `config_resolve` event carries zero structured fields, and a couple of severity/outcome choices contradict the ADR's own worked examples - `src/offsite-brand-presence/drs-status-handler.js:339` (details inline)
4. [Important] The ADR and spec docs are stale against the code they now describe: both still enumerate the pre-wikipedia/pre-Semrush taxonomy and still reference `jobId` instead of the renamed `drsJobId` - `docs/decisions/002-offsite-structured-logging-taxonomy.md:28` (details inline)
5. [Important] Test coverage for the newest additions is thinner than the established sibling standard - two `wikipedia-analysis` failure-path tests skip taxonomy-field verification, and the new Semrush routing/failure logging has zero log-content assertions despite equivalent DRS logging being upgraded in the same round - `test/audits/wikipedia-analysis/guidance-handler.test.js` (details inline)

<details>
<summary>Non-blocking (8): minor issues and suggestions</summary>

- nit: `AUDIT_SLUG_BY_TYPE` in `src/common/offsite-refresh.js` has no fallback for an unmapped `auditType`; the `audit` field would silently disappear from the emitted line rather than surfacing the gap - consider `?? 'unknown'`
- suggestion: `requestOffsiteScrape`'s post-merge signature is now 8 positional parameters, several of them booleans (`enableBrandProfile`, `enableSemrush`) at fixed ordinal positions - a well-known source of call-site transposition bugs; consider an options object in a follow-up
- nit: the doc comment header in `src/utils/offsite-logging.js` overstates the guarantee ("free-text message... cannot forge a second key=value token") - true only for the line-splitting sub-case, not same-line token forgery; correct once the Critical/Important findings above are fixed
- suggestion: five near-identical analysis-handler families (cited/reddit/youtube/wikipedia, soon a shared Semrush parameter too) with byte-for-byte-duplicated helpers (`classifyFetchFailure`, `syncSuggestions` wrapping) is now a real, recurring source of the same bugs being introduced independently at each site - file a concretely-scoped follow-up ticket for a shared wrapper rather than deferring again
- nit: PR title ("...for offsite-brand-presence") still undersells the diff's actual scope (all five audit families) - worth retitling before merge for changelog clarity
- nit: Wikipedia's Slack outcome message interpolates `data.errorMessage`/`company` directly into Slack mrkdwn text - a different sink than the Splunk log stream and out of this finding's scope, but the same "don't interpolate externally-sourced text raw" principle applies; worth a follow-up
- nit: `resolveMystiqueUrlLimit` embeds `JSON.stringify(raw)` in the message rather than a field; low risk since `raw` is expected numeric, but inconsistent with the "route external values through fields" convention
- suggestion: add a lint rule or `no-restricted-syntax` check flagging template-literal interpolation of a non-allowlisted identifier inside an `olog.*`/raw `log.*` message argument in `src/*-analysis/**` and `src/offsite-brand-presence/**` - this would have caught every injection-contract gap found in this round mechanically rather than relying on manual review a third time

</details>

- Stack trace loss in the three (now four) guidance-handler catch-alls is fixed - the raw `Error` is now passed as a genuine 4th argument to `failure(...)` for CloudWatch stack capture.
- The DRS_API_URL/DRS_API_KEY severity downgrade (error to warn) has been reverted to error level.
- The DRS-poll `jobId` field has been renamed to `drsJobId` in code, resolving the collision with the platform-wide async-queue `jobId` (though the ADR/spec docs weren't updated to reflect the rename - see finding 4 above).
- The loose `sinon.match.object` test assertions in `offsite-audit-utils.test.js` now use content-specific matchers.

src/wikipedia-analysis/handler.js
```diff
@@ -231,23 +232,25 @@ function getWikipediaConfig(site) {
...
+  olog.start('audit_start', `Starting Wikipedia analysis audit for site: ${siteId}`);
 
   try {
     const wikipediaConfig = getWikipediaConfig(site);
 
-    const wikipediaUrlOverride = resolveWikipediaUrlOverride(auditContext, log);
+    const wikipediaUrlOverride = resolveWikipediaUrlOverride(auditContext, olog);
     if (wikipediaUrlOverride?.invalid) {
-      log.warn(`${LOG_PREFIX} Ignoring invalid wikipedia URL override: ${wikipediaUrlOverride.value}`);
+      olog.warn('config_resolve', `Ignoring invalid wikipedia URL override: ${wikipediaUrlOverride.value}`, { reason: 'invalid_url_override' });
```
grigoreadobe: **issue (blocking):** `wikipediaUrlOverride.value` is Slack-controlled input (normalized via `unwrapSlackMrkdwnLink` but not URL-valid), interpolated directly into the message string. Per the updated doc comment in `offsite-logging.js`: message-string interpolation strips control characters but does NOT quote `=`/`"`, so a crafted value can still forge a same-line `key=value` token — the same class of gap this PR was opened to fix.

Also worth noting: `resolveWikipediaUrlOverride` already logs this value safely as a structured field (`reason: 'invalid_url', value: normalized`) at the debug level before returning, so this warn is a second emission of the same data via an unsafe path.

Fix — route through a field:
```js
olog.warn('config_resolve', 'Ignoring invalid wikipedia URL override', {
  reason: 'invalid_url_override',
  value: wikipediaUrlOverride.value,
});
```

grigoreadobe (changes requested):
Hey @rarescheseli,

**Verdict: Request changes** — one message-string interpolation of Slack-controlled input remains; every other Round 2 finding is closed.
**Complexity: HIGH** — large diff; security surface (structured log-injection fix across five audit families).
_Changes: extends offsite structured logging taxonomy to wikipedia-analysis and Semrush, closes log-injection gaps, syncs ADR and spec docs (32 files)._

> Note: Recommend a human read before merge — this change amends an architectural decision record in `docs/decisions/`. The bot review complements, not replaces, a human read here.

## Must fix before merge

1. **[Important]** `olog.warn` message still interpolates Slack-controlled `value` raw — `src/wikipedia-analysis/handler.js:245` (details inline)

<details>
<summary>Non-blocking (1)</summary>

- nit: `wikipediaUrlOverride.url` is URL-validated so the forgery risk is lower, but it is still raw in the debug message — consider routing it as `{ url: wikipediaUrlOverride.url }` for consistency — `src/wikipedia-analysis/handler.js:248`
</details>

- ✅ **[Critical]** `resolveWikipediaUrlOverride` fully rewritten — all Slack-controlled values now route through `olog.debug` with structured fields; `HUMAN_PREFIX` removed.
- ✅ **[Important]** `companyName` removed from message strings across all four analysis families (`cited`, `reddit`, `youtube`, `wikipedia`) in both `config_resolve` and `mystique_dispatch` logs; moved to structured fields.
- ✅ **[Important]** `config_resolve` for wikipedia now emits `{ companyName, website, wikipediaUrl }` as structured fields.
- ✅ **[Important]** `makeResolveOverride` in `offsite-audit-utils.js` now uses `appendFields` to route Slack-controlled `raw` as a quoted field.
- ✅ **[Important]** `drs_poll` corrected from `olog.warn` to `olog.skip`; ADR-convention comment added.
- ✅ **[Important]** ADR 002 taxonomy updated: `wikipedia` added to the audit enum, `semrush` added to peers, `drsJobId` replaces `jobId`.
- ✅ **[Important]** Spec doc (`2026-07-31-offsite-structured-logging.md`) synced with the same taxonomy updates.

grigoreadobe (approved):
Hey @rarescheseli,

**Verdict: Approve** - all blocking findings are resolved; this is clean to merge.
**Complexity: HIGH** - large diff; security surface (log-injection fix across five audit families).
_Changes: fixes the two remaining wikipedia URL-override log calls to route Slack-controlled values as structured fields, matching the rule established in offsite-logging.js (1 fix commit + main merges)._

Note: Recommend a human read before merge - this change amends an architectural decision record in `docs/decisions/`. The bot review is a complement to, not a replacement for, a human read here.

- [Important] `olog.warn` at `src/wikipedia-analysis/handler.js` now emits a static message with `value: wikipediaUrlOverride.value` as a structured field - Slack-controlled input no longer interpolated into the message string.
- nit: `olog.debug` at the same location also corrected to route `url: wikipediaUrlOverride.url` as a field.

PR #2839, opened 2026-07-31
fix(preflight): stop form-accessibility identify/suggest racing on the shared S3 result key (SITES-49003)

clotton:
## Problem (SITES-49003)
Preflight runs the `form-accessibility` check in **both** the `identify` and `suggest` steps, and the MFE submits those two steps concurrently. Both invocations poll the **same, non-run-scoped** S3 result key `form-accessibility-preflight/{siteId}/{filename}` and then unconditionally `DeleteObjectsCommand` it after reading. Whichever step reads first **deletes the shared file**, starving the other, which polls a now-missing key until its ~600s timeout — the UI's form-accessibility card hangs. Verified live in prod (identify stalled while suggest completed) and reproduced deterministically.

## Fix
- **Remove the destructive after-read delete** — the sole source of the starvation.
- **Freshness-gate the poll**: accept a result file only if its S3 `LastModified >= scrapeStartTime − 60s` (skew tolerance). Missing `LastModified` **fails open** (accept), so it can never introduce a new hang. This prevents reading a stale leftover now that we no longer delete.
- The result key is **overwritten in place** on the next run for the same site+URL+selector, so files don't grow unbounded (a prefix lifecycle rule is a nice-to-have infra follow-up).

## Blast radius (contained)
- `getObjectMetadataUsingPrefix` is **added** to `s3-utils.js`; `getObjectKeysUsingPrefix`/`getObjectFromKey` are untouched, so other audits that use them are unaffected.
- Only `preflight/handler.js` imports `form-accessibility.js`; the change is internal to that check and its own S3 prefix. Its outer try/catch is preserved (stays non-throwing in the sequential runner).

## Tests
- New deterministic **concurrency regression** test: two runs (identify + suggest) share an in-memory S3 store; both produce opportunities (no starvation) and **`DeleteObjectsCommand` is never sent** (fails on the old code).
- New **freshness-gate** tests: stale file rejected then fresh accepted; missing-`LastModified` fail-open.
- `getObjectMetadataUsingPrefix` unit tests. 100% line/branch/statement/function coverage on both changed source files; lint clean.

## Scope / follow-ups (not in this PR)
- `accessibility.js` has the identical latent race (never observed to hang — its result is written by the fast content-scraper, not the slow Mystique round-trip). Tracked separately.
- Mystique-side `opportunities/{siteId}/opportunity.json` last-writer-wins clobber (race #2) — Mystique-only; this PR greatly reduces failures but doesn't eliminate that mode.

## Verification after merge (deploys to prod)
Via the `run-preflight` skill, fire `--step identify` and `--step suggest` concurrently on a form page; confirm both steps COMPLETE (no ~600s hang) and Splunk shows both "processing complete" with no shared-file delete.

### DIFF
--- a/src/preflight/form-accessibility.js
+++ b/src/preflight/form-accessibility.js
@@ -12,16 +12,22 @@
 
 import { createHash } from 'crypto';
 import { isNonEmptyArray } from '@adobe/spacecat-shared-utils';
-import { DeleteObjectsCommand } from '@aws-sdk/client-s3';
 import { load as cheerioLoad } from 'cheerio';
 
 import { saveIntermediateResults } from './utils.js';
 import { sleep } from '../support/utils.js';
-import { getObjectFromKey, getObjectKeysUsingPrefix } from '../utils/s3-utils.js';
+import { getObjectFromKey, getObjectMetadataUsingPrefix } from '../utils/s3-utils.js';
 import { generateAccessibilityFilename } from './accessibility.js';
 import { getDomElementSelector } from './utils/dom-selector.js';
 
 export const PREFLIGHT_FORM_ACCESSIBILITY = 'form-accessibility';
+
+// Only accept a Mystique-written result file whose S3 LastModified is at least this old
+// relative to the run start, to avoid reading a stale leftover file from a previous run on the
+// same site+URL+selector (the result key is not run-scoped, and we no longer delete after read).
+// Set far larger than plausible audit-worker↔S3 clock skew and far smaller than a typical
+// inter-run gap; missing LastModified fails open (accept) so it can never introduce a new hang.
+const CLOCK_SKEW_TOLERANCE_MS = 60 * 1000;
 
 /**
  * Generates a stable, unique filename for a single form's result file.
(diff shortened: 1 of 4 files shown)


MysticatBot (approved):
Hey @clotton,

**Verdict: Approve** - well-scoped race condition fix with sound freshness gate design and strong test coverage.
**Complexity: MEDIUM** - medium diff, single-service bug fix with new utility function.
_Changes: Removes the destructive S3 delete that caused identify/suggest starvation and adds a LastModified freshness gate to reject stale leftover files (4 files)._

<details>

- nit: `new Date(LastModified).getTime()` is redundant - AWS SDK v3 already returns `LastModified` as a `Date` object, so `LastModified.getTime()` suffices without the extra allocation - `src/preflight/form-accessibility.js:421`
- suggestion: `getObjectMetadataUsingPrefix` duplicates the pagination/error structure of `getObjectKeysUsingPrefix` - a shared internal helper parameterized by a mapper callback would prevent drift if pagination logic changes - `src/utils/s3-utils.js:64`
- suggestion: Log when the freshness gate rejects a file as stale (the delta between LastModified and threshold) - currently only the aggregate "Found 0 out of N" is logged, which is less actionable during incident triage - `src/preflight/form-accessibility.js:418`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 56s | <b>Cost:</b> $5.11 | <b>Commit:</b> <code>0903decac252620505642adff4f278203baaeec8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2840, opened 2026-07-31
fix(preflight): scope form-accessibility opportunityId per step + gate stale result reads (SITES-49003)

clotton:
## Problem (SITES-49003, race #2 + a stale-read gap)
The MFE submits the form-accessibility **identify** and **suggest** steps concurrently as two separate AsyncJobs. Both sent `opportunityId: siteId` to Mystique, which keys its shared `opportunities/{opportunityId}/opportunity.json` on that value and read-modify-writes it during detection. When the two runs' detection windows overlap — reliably in the **UI** (they land ~seconds apart); only *intermittently* from the CLI (where the two Mystique writes happened to land 20–30 s apart) — they clobber that one file (last-writer-wins, a hazard the Mystique code itself documents), and a detection task can **stall past the ~600 s poll**. Result: no fresh result file is written and the UI's form-accessibility card hangs.

Prod evidence (UI trace `e6ecd5a3…`): both Mystique tasks registered `opportunityId == siteId`, both assembled issues, **neither** reached `Detect opportunity for…`/result-write/SQS-delete, both SQS messages **redelivered at ~600 s**, and the audit-worker poll timed out (`completed in 600.99 s`).

## Fix B — root cause, audit-worker-only (no Mystique release)
Send `opportunityId: jobId` (the per-step AsyncJob id, already unique per run) instead of `siteId`, so identify and suggest no longer share `opportunity.json`. Verified safe end-to-end:
- Mystique reads `opportunityId` **from the SQS message** (`Using opportunity ID from SQS message data`), so this is entirely audit-worker-controlled.
- The S3 **result key** is keyed on `siteId + URL + formSource` (**not** `opportunityId`) on both sides and must match byte-for-byte — **unchanged**, so no Mystique change and no filename coordination.
- Mystique's opportunity store is **S3-only** (`opportunities/{id}/opportunity.json`) — a unique id just isolates that file: no SpaceCat DB writes, no orphans; `get_opportunity` now reads each step's own file.

## Fix A — correctness (stale-read gap from race #1's no-delete)
Since we no longer delete result files, a stale leftover lingers on the shared key. The poll already freshness-gated, but the **post-poll processor read did not** — so a timeout could serve a previous run's file as if fresh (invisible on an unchanged page; wrong data if the form changed). The processor now only reads files written at/after this run started; a timeout yields **empty**, never stale.

## Tests
- Assert `opportunityId === jobId` in the Mystique message (contract for Fix B).
- New freshness-gate test: a stale result file is **skipped** — never read, no opportunities produced.
- Full suite green (9364 passing); **100% line/branch/statement coverage** on `form-accessibility.js`; lint clean.

## Verification after deploy (the meaningful check)
A green CLI run is **necessary but not sufficient** — it can pass by timing luck. Reproduce a real **overlap** (UI/bookmarklet, or several forced concurrent CLI pairs) and confirm in Splunk that both Mystique tasks use **distinct** `opportunities/{jobId}/opportunity.json`, both reach `Detect opportunity for…` → result write → SQS delete, and there is **no 600 s redelivery/stall**.

## Scope / residuals (not in this PR)
- If the stall has a *second* cause inside Mystique (CrewAI flow / worker-pool contention), it could persist even with unique ids — the overlap test settles this; Fix A still prevents stale data meanwhile.
- The result file is still shared across concurrent runs (benign for identical content; run-scoped result key would need Mystique).
- form-a11y detection tasks still lack an SQS visibility heartbeat (robustness follow-up).

### DIFF
--- a/src/preflight/form-accessibility.js
+++ b/src/preflight/form-accessibility.js
@@ -183,7 +183,16 @@ export async function detectFormAccessibility(context, auditContext) {
         time: new Date().toISOString(),
         data: {
           url: previewUrls[0], // M expects url in the data object for forms opportunity
-          opportunityId: siteId,
+          // Scope the opportunity to THIS preflight step+run, not the site. Mystique keys
+          // its shared `opportunities/{opportunityId}/opportunity.json` on this value
+          // (read-modify-written during detection). The MFE submits the identify and
+          // suggest steps concurrently as two separate jobs; when both sent
+          // opportunityId=siteId they collided on that one file (last-writer-wins), and a
+          // detection task could stall past the ~600s poll — SITES-49003 (race #2). The
+          // per-step jobId is unique per run, so each step gets its own opportunity file.
+          // The S3 result key we poll is unaffected (it is keyed on siteId+URL+formSource,
+          // not opportunityId), so this needs no Mystique change.
+          opportunityId: jobId,
           a11y: urlsToDetect,
         },
         options: {
@@ -218,7 +227,12 @@ export async function detectFormAccessibility(context, auditContext) {
 /**
  * Step 2: Process detected form accessibility issues and create opportunities
  */
-export async function processFormAccessibilityOpportunities(context, auditContext, formEntries) {
+export async function processFormAccessibilityOpportunities(
+  context,
+  auditContext,
+  formEntries,
+  freshnessThreshold,
+) {
   const {
     site, job, log, env, s3Client,
   } = context;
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @clotton,

**Verdict: Approve** - well-scoped race-condition fix with defense-in-depth freshness gating.
**Complexity: LOW** - small, single-service bug fix.
_Changes: Scopes the Mystique opportunityId to the per-step jobId (preventing concurrent identify/suggest collisions) and adds a freshness gate so the processor never serves stale result files from a previous run (2 files)._

<details>

- suggestion: The freshness gate in `processFormAccessibilityOpportunities` re-lists the same S3 prefix the poller just confirmed fresh. In the happy path this is redundant (the poller only exits once all expected files pass). Consider passing the poller's confirmed-fresh key set directly into the processor to eliminate the extra `ListObjectsV2` call on every successful run - `src/preflight/form-accessibility.js:265`
- nit: The new freshness-gate test sets up stale timestamps for both page1 and page2 but only asserts on page1's warn log and opportunities. Adding a page2 assertion would catch an early-return or off-by-one in the iteration - `test/audits/preflight-form-accessibility.test.js:633`
- suggestion: Add a positive-path test where `freshnessThreshold` is set and `LastModified` is newer than the threshold, confirming the file is read and opportunities are produced normally. The existing tests pass `undefined` for the threshold, so the `freshKeys != null` acceptance branch is only exercised indirectly - `test/audits/preflight-form-accessibility.test.js:591`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 46s | <b>Cost:</b> $4.37 | <b>Commit:</b> <code>ab7b675fd6163c532faede1496f78c1b6bf57dd3</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2843, opened 2026-08-03
fix(cdn-logs): use site id for CDN logs Glue database and table names

akshaymagapu:
Names the CDN logs Glue databases and tables by site id instead of a URL-derived key, for stable, unique naming. No data migration - S3 layout is unchanged and tables recreate on the next run. Deploys in lockstep with a matching spacecat-auth-service change.

LLMO-6692

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


src/utils/cdn-utils.js
```diff
@@ -407,16 +407,24 @@ export function resolveConsolidatedBucketName(context, region) {
...
   const region = resolveSiteCdnRegion(site, context);
   const siteKey = extractSiteKeyFromBaseURL(site);
   const siteKeyParts = siteKey.split(/[._]/);
   const siteName = siteKeyParts[0] === 'www' && siteKeyParts.length > 1 ? siteKeyParts[1] : siteKeyParts[0];
   const bucket = resolveConsolidatedBucketName(context, region);
   const siteId = site?.getId?.();
-  const databaseName = `cdn_logs_${siteKey}`;
-  const tableName = `aggregated_logs_${siteKey}_consolidated`;
-  const referralTableName = `aggregated_referral_logs_${siteKey}_consolidated`;
+  const catalogKey = siteCatalogKey(siteId);
```
MysticatBot, line 424: issue (blocking): `getS3Config` uses `site?.getId?.()` (optional chaining, suggesting it tolerates a missing ID) but then immediately passes the result to `siteCatalogKey(siteId)` which throws on falsy input. The old code handled a missing siteId gracefully (returned `undefined` for `aggregatedLocation`). The new contract is stricter, but the optional chaining obscures it.

**Why it matters:** Callers that previously received a degraded config now get an unhandled throw from a utility function with a generic message. The defensive posture of `?.` is now deceptive.

**How to fix:** Remove the optional chaining to make the hard requirement explicit:
```js
const siteId = site.getId();
```
This produces a clearer stack trace at the real failure site and communicates the contract to future readers.

src/llm-error-pages/handler.js
```diff
@@ -388,6 +388,17 @@ export async function runAuditAndSendToMystique(context) {
   log.info(`[LLM-ERROR-PAGES] Starting audit for ${url}`);
 
   try {
+    const configuration = await context.dataAccess.Configuration.findLatest();
+    if (!configuration?.isHandlerEnabledForSite('cdn-logs-analysis', site)) {
```
MysticatBot: issue (blocking): If `Configuration.findLatest()` returns null (no configuration record exists in the DB), `configuration?.isHandlerEnabledForSite(...)` evaluates to `undefined` (falsy). Then `!undefined` is `true`, causing the early return - silently skipping the audit for ALL sites when no configuration document is present.

**Why it matters:** A missing configuration record (bootstrap state, accidental deletion, data-access failure returning null) would disable this audit globally with only an INFO log line. The behavior should be explicit: either proceed when configuration is absent (safe default) or fail loudly.

**How to fix:**
```js
const configuration = await context.dataAccess.Configuration.findLatest();
if (configuration && !configuration.isHandlerEnabledForSite('cdn-logs-analysis', site)) {
  log.info(...);
  return { ... };
}
```
This treats missing configuration as "not disabled" (proceed with audit), which matches the pre-PR behavior where no such gate existed.

MysticatBot (changes requested):
Hey @akshaymagapu,

**Verdict: Request changes** - two error-handling regressions introduced alongside the naming fix.
**Complexity: MEDIUM** - medium diff across 12 files, tests-heavy.
_Changes: Switches CDN logs Glue database and table naming from URL-derived keys to site-id-derived keys for stable, unique identifiers (12 files)._
_Scope note: Title says "use site id for Glue names" but the diff also adds a `cdn-logs-analysis` configuration gate in `llm-error-pages/handler.js`._

## Must fix before merge
1. [Important] Optional chaining inconsistent with new hard-throw contract - `src/utils/cdn-utils.js:424` (details inline)
2. [Important] Null configuration silently disables all audits - `src/llm-error-pages/handler.js:392` (details inline)

<details>

- nit: The `_consolidated` suffix was dropped from table names in addition to the key source change - verify no external consumers (dashboards, saved Athena queries) depend on the old suffix pattern - `src/cdn-logs-report/constants/report-configs.js:20`
- suggestion: `getS3Config` still computes and returns `siteKey` via `extractSiteKeyFromBaseURL` - if no remaining callers use it after this PR, remove it as dead code - `src/utils/cdn-utils.js:419`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 44s | <b>Cost:</b> $4.86 | <b>Commit:</b> <code>c01e0b5fc88206a0ca69a46d94571d8920a6a03b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/utils/cdn-utils.js
akshaymagapu: Fixed. `getS3Config` now uses `const siteId = site.getId();` (no optional chaining), and `siteCatalogKey` throws on a missing/falsy id - so the hard requirement is explicit at the call site with a clear message. Pushed in d778ada2.

src/llm-error-pages/handler.js
akshaymagapu: Addressed by splitting the cdn-logs-analysis gate out of this PR into #2844. In that PR the gate now treats a missing configuration as "not disabled" and proceeds: `if (configuration && !configuration.isHandlerEnabledForSite('cdn-logs-analysis', site))`, with a unit test for the null-configuration case. This PR (#2843) no longer contains the gate - it is the naming change only.

MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - prior findings addressed, no new blocking issues.
**Complexity: MEDIUM** - medium diff across 10 files, tests-heavy.
_Changes: Switches CDN logs Glue database and table naming from URL-derived keys to site-id-derived keys for stable, unique identifiers (10 files)._

<details>

- suggestion: `siteCatalogKey` only replaces hyphens - consider replacing all non-alphanumeric/underscore characters (`/[^a-zA-Z0-9_]/g`) for defense-in-depth, so the function is safe-by-construction regardless of future callers - `src/utils/cdn-utils.js:413`
</details>

- Optional chaining inconsistency with hard-throw contract - `getS3Config` now calls `site.getId()` directly
- Null configuration silently disabling audits - configuration gate reverted (extracted to separate PR)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 19s | <b>Cost:</b> $6.11 | <b>Commit:</b> <code>d7b6a32aa5a84d55a0d2984dce2a4b2f96018ee2</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2846, opened 2026-08-04
fix(sync): split OUTDATED logic for isEdited + readability scrapedUrlsSet gap (LLMO-6761)

baranovskyalexandr:
## Summary

Aligns edited-suggestion behavior with the two scenarios defined by ESE team:

- **Scenario 1** — same issue re-detected after customer edit: edited suggestions (`data.isEdited`) are **kept as-is** (centralized hard-skip guard on the matched-key path, replacing per-handler `mergeDataFunction` delegation)
- **Scenario 2** — issue no longer detected: edited suggestions are **cleared** (OUTDATED / FIXED) like any other — the customer edit is preserved only while the issue persists
- `protectEditedFromOutdated` opt-in flag for handlers where `buildKey` includes editable content (e.g. FAQs), preventing key-drift from falsely aging out a suggestion
- Readability handler: scopes OUTDATED eligibility to `finalUrl` of pages scraped this run

## Test plan

- [x] Edited suggestion skipped on matched-key path (Scenario 1)
- [x] Non-edited suggestion overwritten on matched-key path (Scenario 1)
- [x] Edited suggestion marked OUTDATED when issue disappears (Scenario 2, default)
- [x] Edited suggestion protected from OUTDATED when `protectEditedFromOutdated=true` (FAQs key-drift)
- [x] Edited NEW suggestion reconciled to FIXED when issue resolved (Scenario 2)
- [x] Readability OUTDATED scoped to scraped URLs + `finalUrl`
- [x] 151 tests passing

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Duy Nguyen", "Dominique Jäggi"]
```


baranovskyalexandr:
Pushed a follow-up cleanup commit: `syncSuggestions`' centralized `isEdited` hard-skip guard (matched-key path) means the per-handler `mergeDataFunction`/`mergeStatusFunction` `isEdited` branches in FAQ, Summarization, and Readability are now unreachable dead code — they can never execute since the guard returns before either function is called. Removed the dead branches (kept `edgeDeployed`/`edgeOptimizeStatus` guards, which are still live) and the tests that only covered those dead paths. TOC/headings/internal-links are out of this PR's scope, left untouched.

All affected suites green: FAQ guidance-handler (53), summarization guidance-handler (23), readability guidance-handler + opportunities-handler (53), data-access (151).

duynguyen:
Nice work untangling the two isEdited scenarios here — the Scenario 1 / Scenario 2 split is much clearer than the old per-handler `mergeDataFunction` branches, and the readability `scrapedUrlsSet`/`finalUrl` redirect fix is a solid catch for the sampling gap.

A couple of things I think need a look before merge:

**1. The new centralized `isEdited` hard-skip in `syncSuggestions` silently changes behavior for `headings` and `toc` (High)**

Both handlers call the shared `syncSuggestions` and pass their own `isEdited`-aware `mergeDataFunction` (headings preserves only `recommendedAction`; toc preserves only `transformRules.value`, refreshing everything else). The new guard:

```js
if (existingData?.isEdited === true) {
  log.debug(`Skipping edited suggestion ${existingKey} - preserving customer edit`);
  return;
}
```

fires before `mergeDataFunction` is ever called on the matched-key path, so those `isEdited` branches are now dead code and edited headings/toc suggestions get completely frozen (no field refresh at all) instead of the previous partial merge. That contradicts the comments in both handlers claiming they're "out of scope" and prior behavior is preserved via `protectEditedFromOutdated` — that flag only affects the OUTDATED sweep, not this hard-skip.

Worth noting the existing headings/toc tests still pass because they call the extracted `mergeDataFunction` directly, bypassing `syncSuggestions`, so the now-unreachable branch still shows as covered.

Could we gate this hard-skip behind an opt-in flag (mirroring `protectEditedFromOutdated`) so headings/toc keep their existing per-field merge, or if freezing them entirely is actually intended, update those handlers to drop the now-dead branches and add an integration test through `syncSuggestions` itself?

**2. Scenario 2 reconcile change also affects `backlinks`, which isn't mentioned in scope (Medium)**

`reconcileDisappearedSuggestions` is currently only called from `src/backlinks/handler.js`. Dropping the `!isEdited` exemption for NEW suggestions means customer-edited backlink/redirect suggestions that disappear can now auto-reconcile to FIXED. Plausibly fine given the function's own docstring (isEdited on a redirect suggestion = high-confidence customer pick), but it's a behavior change outside the stated FAQ/Summarization/Readability scope and there's no backlinks-specific test covering it. Worth a quick confirmation + regression test.

duynguyen (approved):
Both review findings addressed — `skipEditedOnMatch` opt-in keeps headings/toc's per-field merge behavior intact, and the backlinks reconcile change was correctly reverted/scoped out. LGTM.

PR #2848, opened 2026-08-05
feat(gsc-search-analytics): tracking-only GSC before/after audit for fixed URLs

inkypark623:
## What this change intends

Adds a new **tracking-only** SpaceCat audit, `gsc-search-analytics`. For each URL that ASO fixed (supplied as `auditContext.fixedUrls = [{ url, fixType, fixDate }]`), it records that URL's Google Search Console clicks / impressions / CTR / position for the 84 days **before** and **after** its own fix date, and stores one row in the `audits` table.

This is the "Measured" layer of an outcome-proof effort: it **records raw before/after figures only**. It makes **no causal claim, no attribution, no dollar value** — the stored row carries an explicit `interpretation` label saying so.

## How it works

- Runner audit via `AuditBuilder`; registered in `src/index.js` `HANDLERS` as `gsc-search-analytics`.
- Reuses the existing `@adobe/spacecat-shared-google-client` `getOrganicSearchData` (page dimension), paginated with a 50-page safety cap.
- URLs sharing a fix date share one pair of window pulls; fixed URLs are matched client-side (normalized) against the site's per-page rows.
- Result envelope: `{ schemaVersion, interpretation, connected, status, fixCount, measuredCount, fixes[] }`. Each fix has a `status` of `measured | not_found | incomplete | invalid_date | failed`, plus a `dataQuality` marker.
- **Data-integrity gating:** a delta is stored only when both windows are fully elapsed (GSC ~3-day freshness lag) and within retention (~16 months); otherwise `status: incomplete`, `delta: null`. Completeness is checked before presence, so a not-yet-elapsed window reads `incomplete`, not `not_found`.

## What it changes / breaks

- Additive only: one new handler package + one `HANDLERS` entry. No existing behavior changed.
- `spacecat-shared` is **not** touched — plain-string audit type, no shared release needed.

## Testing

- 34 unit tests for the new module; **100% coverage** (lines / branches / statements) on `src/gsc-search-analytics/**`.
- Full local suite green: 9,424 passing, 0 failing; repo coverage gate met.
- Went through multiple code-review rounds plus a plan/scope conformance review.

## Two decisions to confirm before enabling

1. **Failure handling:** any `GoogleClient.createFrom` failure is recorded as `status: not_connected` with the raw `reason` (matches repo convention, never crashes the run). A transient throttle/5xx therefore reads as "not connected" — confirm that's acceptable.
2. **Oversized list:** `> 500` fixedUrls fails closed (`too_many_fixed_urls`) rather than processing the first 500. Fine for the manual pilot; revisit if the fixed-URL feed is automated.

> Draft: opened for CI / online-testing only. Do not merge — pending review + prod registration for the pilot.

Issue: _TBD_

### DIFF
--- a/src/gsc-search-analytics/fetch.js
+++ b/src/gsc-search-analytics/fetch.js
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+const PAGE = 1000; // GSC returns at most 1000 rows per call
+const MAX_PAGES = 50; // safety cap: up to 50k rows per window
+
+/**
+ * Fetch all page-dimension rows for one window, paginating until Google returns
+ * fewer than a full page. Bounded by MAX_PAGES so a very large site cannot run away.
+ *
+ * @param {object} google - GoogleClient instance exposing getOrganicSearchData.
+ * @param {Date} startDate - window start.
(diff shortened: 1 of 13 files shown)


inkypark623:
Superseded by #2849, opened from an upstream branch. The fork PR's ci/build failed only because fork PRs don't receive the repo secret needed to install the private mysticat-data-service dependency; the upstream-branch PR runs CI with secrets.

PR #2849, opened 2026-08-05
feat(gsc-search-analytics): tracking-only GSC before/after audit for fixed URLs

inkypark623:
## What this change intends

Adds a new **tracking-only** SpaceCat audit, `gsc-search-analytics`. For each URL that ASO fixed (supplied as `auditContext.fixedUrls = [{ url, fixType, fixDate }]`), it records that URL's Google Search Console clicks / impressions / CTR / position for the 84 days **before** and **after** its own fix date, and stores one row in the `audits` table.

This is the "Measured" layer of an outcome-proof effort: it **records raw before/after figures only**. It makes **no causal claim, no attribution, no dollar value** — the stored row carries an explicit `interpretation` label saying so.

## How it works

- Runner audit via `AuditBuilder`; registered in `src/index.js` `HANDLERS` as `gsc-search-analytics`.
- Reuses the existing `@adobe/spacecat-shared-google-client` `getOrganicSearchData` (page dimension), paginated with a 50-page safety cap.
- URLs sharing a fix date share one pair of window pulls; fixed URLs are matched client-side (normalized) against the site's per-page rows.
- Result envelope: `{ schemaVersion, interpretation, connected, status, fixCount, measuredCount, fixes[] }`. Each fix has a `status` of `measured | not_found | incomplete | invalid_date | failed`, plus a `dataQuality` marker.
- **Data-integrity gating:** a delta is stored only when both windows are fully elapsed (GSC ~3-day freshness lag) and within retention (~16 months); otherwise `status: incomplete`, `delta: null`. Completeness is checked before presence, so a not-yet-elapsed window reads `incomplete`, not `not_found`.

## What it changes / breaks

- Additive only: one new handler package + one `HANDLERS` entry. No existing behavior changed.
- `spacecat-shared` is **not** touched — plain-string audit type, no shared release needed.

## Testing

- 34 unit tests for the new module; **100% coverage** (lines / branches / statements) on `src/gsc-search-analytics/**`.
- Full local suite green: 9,424 passing, 0 failing; repo coverage gate met.
- Went through multiple code-review rounds plus a plan/scope conformance review.

## Two decisions to confirm before enabling

1. **Failure handling:** any `GoogleClient.createFrom` failure is recorded as `status: not_connected` with the raw `reason` (matches repo convention, never crashes the run). A transient throttle/5xx therefore reads as "not connected" — confirm that's acceptable.
2. **Oversized list:** `> 500` fixedUrls fails closed (`too_many_fixed_urls`) rather than processing the first 500. Fine for the manual pilot; revisit if the fixed-URL feed is automated.

## Spec

`docs/specs/gsc-search-analytics.md` (in this PR). Implementation plan lives in a private basecamp repo.

> Do not merge yet — pending review + prod registration for the pilot.

Issue: _none (tracking-only pilot audit)_

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Daniel Huser", "Dominique Jäggi"]
```

### DIFF
--- a/src/gsc-search-analytics/lib.js
+++ b/src/gsc-search-analytics/lib.js
@@ -1,0 +1,16 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+// Stub — replaced in Task 6 with the real runner.
+export async function runGscSearchAnalytics(finalUrl) {
+  return { auditResult: { stub: true }, fullAuditRef: finalUrl };
+}
--- a/src/gsc-search-analytics/lib.js
+++ b/src/gsc-search-analytics/lib.js
@@ -18,7 +18,11 @@ import { buildDelta } from './summarize.js';
 
 const SCHEMA_VERSION = 1;
 const MAX_FIXED_URLS = 500; // guard against an unbounded, Lambda-timeout-risking run
+// Each distinct fix date is one date-group = up to two paginated window pulls run
+// sequentially, so cap the number of groups to stay well inside the Lambda timeout.
+const MAX_DATE_GROUPS = 30;
 const toDate = (s) => new Date(`${s}T00:00:00Z`);
+const clip = (s) => String(s).slice(0, 300); // bound stored error text
 
 // A fix entry with the stable shape every consumer can rely on.
 function baseFix(f, status, extra) {
@@ -86,6 +90,18 @@ export async function runGscSearchAnalytics(finalUrl, context, site, auditContex
     }, finalUrl);
   }
 
+  // Each distinct valid fix date becomes one sequentially-processed date-group; bound the
+  // count so a spread-out list can't exhaust the Lambda timeout.
+  const distinctDateCount = new Set(
+    fixedUrls.filter((f) => isValidFixDate(f.fixDate)).map((f) => f.fixDate),
+  ).size;
+  if (distinctDateCount > MAX_DATE_GROUPS) {
+    log.warn(`gsc-search-analytics: ${distinctDateCount} distinct fix dates exceeds cap ${MAX_DATE_GROUPS} for ${finalUrl}`);
+    return envelope({
+      connected: null, status: 'too_many_date_groups', fixCount: 0, measuredCount: 0, fixes: [],
+    }, finalUrl);
+  }
+
   let google;
   try {
     google = await GoogleClient.createFrom(context, finalUrl);
@@ -94,7 +110,7 @@ export async function runGscSearchAnalytics(finalUrl, context, site, auditContex
     // any createFrom failure means the site is not connected to GSC. Record it, don't crash.
     log.info(`gsc-search-analytics: GSC not connected for ${finalUrl}: ${e.message}`);
     return envelope({
-      connected: false, status: 'not_connected', reason: e.message, fixCount: 0, measuredCount: 0, fixes: [],
+      connected: false, status: 'not_connected', reason: clip(e.message), fixCount: 0, measuredCount: 0, fixes: [],
     }, finalUrl);
   }
 
@@ -162,7 +178,7 @@ export async function runGscSearchAnalytics(finalUrl, context, site, auditContex
       // A failure on one date-group must not wipe the others; leave a diagnostic entry.
       log.error(`gsc-search-analytics: fetch failed for ${fixDate} / ${finalUrl}: ${e.message}`);
       for (const f of group) {
-        fixes.push(baseFix(f, 'failed', { windows, error: e.message }));
+        fixes.push(baseFix(f, 'failed', { windows, error: clip(e.message) }));
       }
     }
   }
--- a/src/gsc-search-analytics/lib.js
+++ b/src/gsc-search-analytics/lib.js
@@ -11,21 +11,48 @@
  */
 
 import GoogleClient from '@adobe/spacecat-shared-google-client';
-import { computeWindows } from './windows.js';
+import { computeWindows, assessCompleteness, isValidFixDate } from './windows.js';
 import { fetchWindow } from './fetch.js';
 import { indexRows, lookup } from './match.js';
 import { buildDelta } from './summarize.js';
 
-// Only a genuinely-absent GSC connection is an expected clean stop; anything
-// else (throttling, IAM, parse errors) is a real failure we must surface.
-const isNotConnected = (e) => /no secrets|not connected|no google|not onboarded/i.test(e.message || '');
+const SCHEMA_VERSION = 1;
+const MAX_FIXED_URLS = 500; // guard against an unbounded, Lambda-timeout-risking run
 const toDate = (s) => new Date(`${s}T00:00:00Z`);
+
+// A fix entry with the stable shape every consumer can rely on.
+function baseFix(f, status, extra) {
+  return {
+    url: f.url,
+    fixType: f.fixType,
+    fixDate: f.fixDate,
+    status,
+    before: null,
+    after: null,
+    delta: null,
+    found: { before: false, after: false },
+    ...extra,
+  };
+}
+
+function envelope(fields, finalUrl) {
+  return {
+    auditResult: { schemaVersion: SCHEMA_VERSION, ...fields },
+    fullAuditRef: finalUrl,
+  };
+}
 
 /**
  * Tracking-only audit: for each URL ASO fixed, record its GSC clicks/impressions/
  * ctr/position for the 84 days before and after the URL's own fix date. No causal
  * claim — this is the "Measured" layer. Input is a manually-supplied list of
  * { url, fixType, fixDate } in auditContext.fixedUrls.
+ *
+ * Result envelope (audit_result JSON): { schemaVersion, connected, status, fixCount,
+ * measuredCount, fixes[] }. Each fix carries a `status` of one of:
+ *   measured | not_found | incomplete | invalid_date | failed
+ * plus before/after/delta/found and a dataQuality marker so a later reader can tell a
+ * real signal from a data gap.
  *
  * @param {string} finalUrl - resolved site base URL.
  * @param {object} context - audit context ({ log, ... }).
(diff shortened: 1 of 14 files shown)


src/gsc-search-analytics/lib.js
```diff
@@ -0,0 +1,174 @@
...
+
+  // Split invalid dates out (recorded, never fetched); group the rest by fix date so
+  // URLs sharing a date share one pair of pulls.
+  for (const f of fixedUrls) {
+    if (!isValidFixDate(f.fixDate)) {
+      fixes.push(baseFix(f, 'invalid_date', { error: `Invalid fix date: ${f.fixDate}` }));
+    } else if (!byDate.has(f.fixDate)) {
+      byDate.set(f.fixDate, [f]);
+    } else {
+      byDate.get(f.fixDate).push(f);
```
MysticatBot, line 129: issue (blocking): **Unbounded date-group iteration can exhaust Lambda timeout.**

`MAX_FIXED_URLS = 500` caps the total URL count, but does not prevent 500 URLs with 500 distinct `fixDate` values. Each date group runs sequentially, issuing up to 2 * 50 = 100 paginated API calls. With many distinct dates the audit trivially exceeds any reasonable Lambda timeout with no graceful degradation (no partial result, no diagnostic).

**Why it matters:** a well-formed input within the stated cap can cause a silent timeout and wasted compute.

**Fix:** Add a `MAX_DATE_GROUPS` constant (e.g., 50) and fail closed or warn when exceeded, mirroring the `MAX_FIXED_URLS` pattern:
```js
const MAX_DATE_GROUPS = 50;
if (byDate.size > MAX_DATE_GROUPS) {
  log.warn(`gsc-search-analytics: ${byDate.size} unique fix dates exceeds cap ${MAX_DATE_GROUPS}`);
  return envelope({ connected: true, status: 'too_many_date_groups', ... }, finalUrl);
}
```

src/gsc-search-analytics/lib.js
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  for (const [fixDate, group] of byDate) {
+    const windows = computeWindows(fixDate);
```
MysticatBot, line 134: issue (blocking): **Runner return-value field name mismatch with CLAUDE.md.**

The repo CLAUDE.md documents RunnerAudit runners as returning `{ auditData, fullAuditRef }`, but the `envelope()` helper here produces `{ auditResult, fullAuditRef }`. Since CI passes and AuditBuilder accepts this, `auditResult` is likely the current convention and CLAUDE.md is stale.

**Fix:** Either rename `auditResult` to `auditData` here to match docs, or update CLAUDE.md to reflect the current field name. The latter is probably correct given the framework accepts it.

MysticatBot (changes requested):
Hey @inkypark623,

**Verdict: Request changes** - well-structured audit that follows repo patterns, but a date-group cap is needed to prevent timeout and CLAUDE.md convention gaps should be addressed.
**Complexity: HIGH** - large diff (931 lines, new audit module across 13 files).
_Changes: Adds a new `gsc-search-analytics` tracking-only audit that records GSC clicks/impressions/CTR/position for 84-day before/after windows around each fixed URL's fix date (13 files)._

## Must fix before merge

1. [Important] Unbounded date-group iteration can exhaust Lambda timeout - `src/gsc-search-analytics/lib.js:113` (details inline)
2. [Important] Missing spec document per CLAUDE.md requirement - `docs/specs/` (details inline)
3. [Important] CLAUDE.md documents runner returning `{ auditData, fullAuditRef }` but code returns `{ auditResult, fullAuditRef }` - either update CLAUDE.md or align the field name - `src/gsc-search-analytics/lib.js:118` (details inline)

<details>

- nit: `buildDelta` position field - negative = improvement is non-obvious and undocumented for consumers - `src/gsc-search-analytics/summarize.js:6`
- suggestion: Sort query params in `normalizeUrl` (`url.searchParams.sort()`) to prevent false `not_found` when fixedUrls have reordered params vs GSC canonical form - `src/gsc-search-analytics/match.js:4`
- nit: `indexRows` silently overwrites when two GSC rows normalize to the same URL (last-writer-wins, no aggregation or warning) - `src/gsc-search-analytics/match.js:14`
- nit: Raw `e.message` from GoogleClient/fetch failures persisted in audit result `reason`/`error` fields could leak internal details if results are exposed externally - `src/gsc-search-analytics/lib.js:175`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 10s | <b>Cost:</b> $4.51 | <b>Commit:</b> <code>d240b90bbdae8ebbc1f16bd498b75d6fd7bfa470</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

inkypark623:
Thanks @MysticatBot — addressed the review in fc06db4d:

- **[Important] date-group timeout** — added a distinct-fix-date cap (MAX_DATE_GROUPS=30 -> status too_many_date_groups) bounding the sequential window pulls.
- **[Important] missing spec** — added docs/specs/gsc-search-analytics.md.
- **Nits** — normalizeUrl now sorts query params; stored reason/error text is clipped to 300 chars.

On **[Important] auditData vs auditResult**: this one is a stale-doc false positive, not a code change. base-audit.js:131 destructures `{ auditResult, fullAuditRef }` from the runner and then builds its internal `auditData` wrapper from it, so the runner MUST return `auditResult` (returning `auditData` would be undefined). The repo CLAUDE.md wording ("runner returns { auditData }") is the stale side; leaving the code as-is. Re-requesting review for a fresh pass.

MysticatBot (approved):
Hey @inkypark623,

**Verdict: Approve** - prior findings addressed, no new blocking issues; well-structured additive audit with solid guards and full coverage.
**Complexity: HIGH** - large diff (1017 lines, new audit module across 14 files).
_Changes: Adds a new `gsc-search-analytics` tracking-only audit that records GSC clicks/impressions/CTR/position for 84-day before/after windows around each fixed URL's fix date (14 files)._

<details>

- nit: `indexRows` silently overwrites when two GSC rows normalize to the same URL (last-writer-wins, no aggregation or warning) - unlikely with page dimension but worth a defensive check - `src/gsc-search-analytics/match.js:14`
- suggestion: Confirm that `@adobe/spacecat-shared-google-client` `getOrganicSearchData` accepts `Date` objects - `lib.js` converts ISO strings to Dates via `toDate()` before passing to `fetchWindow`; if the shared client expects strings, pass `windows.before.start`/`.end` directly instead - `src/gsc-search-analytics/lib.js:92`
- suggestion: CLAUDE.md documents runner returning `{ auditData, fullAuditRef }` but the framework (`base-audit.js:131`) destructures `{ auditResult, fullAuditRef }` - fix the three stale occurrences in a follow-up commit - `CLAUDE.md`

</details>

- Date-group timeout cap added (MAX_DATE_GROUPS=30 with fail-closed status)
- Spec document added at `docs/specs/gsc-search-analytics.md`
- `auditResult` naming confirmed correct per framework contract; CLAUDE.md is the stale side
- Error text now clipped to 300 chars via `clip()` helper
- `normalizeUrl` now sorts query params for stable matching

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 50s | <b>Cost:</b> $5.28 | <b>Commit:</b> <code>fc06db4ded54056c6d02bfc6620ccd3823a8ea94</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

inkypark623:
Thanks for the approval @MysticatBot. Dispositions on the 3 non-blocking notes:

1. **`indexRows` last-writer-wins on duplicate normalized URLs** — acknowledged. With the `page` dimension GSC returns one row per page URL, so a post-normalization collision only happens if two source URLs normalize identically (e.g. trailing-slash variants both present), which GSC won't emit for a single property. Leaving as-is for the pilot; a defensive de-dupe/warn is a reasonable later hardening.

2. **Confirm `getOrganicSearchData` accepts `Date` objects** — confirmed, and it in fact *requires* them: the shared client validates `isValidDate(startDate)` and calls `startDate.toISOString()` (`spacecat-shared-google-client/src/index.js:101,129`). Passing ISO strings would fail validation, so the `toDate()` conversion is correct and necessary — no change.

3. **Stale `auditData` wording in CLAUDE.md** — agreed it's the doc that's wrong (framework destructures `auditResult` at `base-audit.js:131`). Deferring to a separate doc-only follow-up to keep this feature PR focused, per your own suggestion.

No code changes from these; holding merge pending prod registration for the pilot.

inkypark623:
Hi @danieljchuser! Could you take a look when you get a chance? This is a small tracking-only audit — it records GSC clicks/impressions/CTR/position for the 84 days before/after each fixed URL's deploy date, for the outcome-proof work. No causal claim, additive only (new handler + one HANDLERS entry).

MysticatBot has approved it and it's CLEAN, with a spec at docs/specs/gsc-search-analytics.md. When you're happy with it, could you also handle the prod registration + a run for 1-2 pilot sites? Happy to walk through any of it. Thanks!

danieljchuser (changes requested):
✨ **Review — `gsc-search-analytics`**

Solid, self-aware audit — spec-first, 100% coverage, and the tracking-only / no-causal-claim discipline is exactly right. I traced the full path against the installed deps and it holds up: the runner contract matches `RunnerAudit`, `getOrganicSearchData`'s signature (Date args, `rowLimit ≤ 1000`, `startRow` pagination) lines up, and `Audit.create` accepts a plain-string `auditType` with no shared-schema change. Nice work.

One change I'd like before this is enabled for the pilot, plus non-blocking notes.

### 🔴 Blocking — matching silently fails when the site's base-URL host ≠ its GSC property host

`normalizeUrl` lowercases the host but doesn't reconcile `www.` vs apex. The matcher compares the supplied `fixedUrls` against GSC page rows, and those rows always carry the **GSC property host** — while the fixed-URL list will naturally be built from the site's *registered base-URL* host. When those differ (apex vs `www` is the common case), every URL falls through to `not_found` — healthy connection, data present. Silent-wrong, and it defeats the audit's purpose for exactly that kind of site. This isn't hypothetical for the intended pilot target — its base URL and its Search Console property sit on different hosts.

The fetch itself is fine: the client's `composeAuditURL()` follows redirects, so the page filter resolves to the property host either way. The gap is only in the client-side match — and the audit already *has* the right host (it's in every row it fetched), so it can reconcile with no new input.

Suggested fix:
- In `normalizeUrl`, strip a leading `www.` (shared-utils exports `stripWWW`) so apex/www match either way. A run only queries one property, so this can't collide across properties. Worth a comment that it assumes `www` and apex serve the same content (true for essentially every real site).
- Add a diagnostic: if a window returned rows but **zero** fixed URLs matched, `log.warn` a likely host mismatch — turns this from silent into visible.

### Non-blocking

- **Measurable window is narrow — worth a spec line.** A fix yields a `measured` delta only once its after-window elapses past the ~3-day lag (≈ fix + 87 days) and its before-window is still in retention (fix newer than ~13 months). So only fixes aged ~3–13 months produce a delta; more-recent ones are `incomplete` and need a re-run later.
- **`latest_audits` is a snapshot, not cumulative** — each run replaces the previous `fixedUrls`. Fine for the pilot; worth noting for `query-audits` consumers.
- **Transient GSC failures read as `not_connected`** (decision #1) — fine for a manual, re-runnable pilot; if this automates, split `not_connected` (secret missing) from `fetch_failed` (throttle/5xx).
- **Rollout:** the type still needs registering in prod `global-config` + per-site enablement, or `isAuditEnabledForSite` short-circuits before the runner runs. On-demand only (no jobs-dispatcher entry) — consistent with the pilot framing.

Happy to pair on the `match.js` tweak — it's small.

inkypark623:
Thanks for the thorough review @danieljchuser - great catch on the host mismatch. Addressed in 7f4d90d4:

**Blocking (www/apex host match):** normalizeUrl now strips a leading www. via shared-utils stripWWW, so fixedUrls built from the site base-URL host reconcile with GSC rows on the property host. Added the diagnostic too: a window that returns rows but matches zero fixed URLs now log.warns a likely host mismatch instead of silently reporting not_found.

**Non-blocking, now in the spec:** the ~3-13mo measurable window (recent fixes come back incomplete and need a re-run) and that latest_audits is a per-run snapshot, not cumulative. The not_connected vs fetch_failed split I left for the automation phase as you suggested; the prod global-config registration + per-site enablement is yours to drive for the pilot.

39 tests, 100% coverage, lint clean. Re-requesting for another pass - happy to pair if anything's still off.

MysticatBot (approved):
Hey @inkypark623,

**Verdict: Approve** - prior blocking findings addressed, www/apex reconciliation is correct, no new issues introduced.
**Complexity: HIGH** - large diff (1073 lines, new audit module across 14 files).
_Changes: Adds a new `gsc-search-analytics` tracking-only audit that records GSC clicks/impressions/CTR/position for 84-day before/after windows around each fixed URL's fix date (14 files)._
Note: Recommend a human read before merge - this change includes a new spec document under `docs/specs/`. The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: `url.host` includes the port when non-default (e.g. `:8080`); using `url.hostname` instead would be more defensive against edge cases where fixed URLs carry explicit ports, though GSC page rows never do in practice - `src/gsc-search-analytics/match.js:34`
- nit: Host-mismatch warning fires even for date-groups where `!completeness.afterComplete` (window not yet elapsed), which could emit a misleading "likely host mismatch" when the real explanation is sparse data in a barely-started window. Consider gating on `completeness.beforeComplete && completeness.afterComplete` - `src/gsc-search-analytics/lib.js:178`

</details>

- Date-group timeout cap added (MAX_DATE_GROUPS=30 with fail-closed status)
- Spec document added at `docs/specs/gsc-search-analytics.md`
- `auditResult` naming confirmed correct per framework contract; CLAUDE.md is the stale side
- `normalizeUrl` now sorts query params for stable matching
- Error text clipped to 300 chars via `clip()` helper
- www/apex host reconciliation via `stripWWW` from shared-utils (human reviewer blocking - addressed)

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 28s | <b>Cost:</b> $4.80 | <b>Commit:</b> <code>7f4d90d4322216008f9bd2283455151a0cbec13e</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

danieljchuser (approved):
✨ **Re-review — blocking item resolved, approving**

Thanks for the fast turnaround — the www/apex host reconciliation is handled cleanly:

- `normalizeUrl` now applies `stripWWW(url.host.toLowerCase())`, so an apex fixed URL matches a `www` GSC property row (and vice versa) without over-collapsing other subdomains or ports. Verified against the installed `@adobe/spacecat-shared-utils`.
- The zero-match diagnostic (`log.warn` when a window returns rows but nothing matched) turns the old silent-`not_found` failure mode into a visible signal.
- Tests cover both the apex↔www match and the warning path; full suite + 100% coverage gate and codecov are green.

Nice touch folding the timing window (~3–13 months) and the `latest_audits`-is-a-snapshot caveat into the spec too.

Approving. The only things left are the ones we already agreed are non-blocking / operational: the accepted transient-failure-reads-as-`not_connected` behavior, and your own "do not merge yet" gate — prod `global-config` registration + per-site enablement before the pilot can actually run.

PR #2847, opened 2026-08-05
feat(offsite): source cited URLs from the Semrush URL-Inspector API

AndreiAlexandruParaschiv:
## What

Adds a **flag-gated** data source that fetches the offsite cited URLs (YouTube, Reddit) from the **Semrush-backed Serenity URL-Inspector** endpoints exposed by **spacecat-api-service** (the Elements proxy, `src/controllers/elements.js`) and feeds them into the existing offsite pipeline **unchanged**.

Epic: [LLMO-6488](https://jira.corp.adobe.com/browse/LLMO-6488) · Story: [LLMO-6709](https://jira.corp.adobe.com/browse/LLMO-6709) · Plan: #2779

**Spec (PR template §4):** `docs/specs/2026-08-05-offsite-cited-urls-semrush-source.md` · **ADR:** `docs/decisions/002-offsite-cited-urls-semrush-source.md` (both added in this PR).

## Why a loader

The downstream pipeline (`selectTopUrls` → `addUrlsToUrlStore` → DRS → cited/youtube/reddit-analysis) only understands one input contract: `allUrls: Map<url, { count, domain }>`. The Semrush API returns a shape-incompatible response (structured JSON, per-hostname, requires auth + brand/org resolution + week→date mapping, carries exact citation counts). The loader is the single, isolated adapter that turns that into the exact map the pipeline already consumes — so **everything downstream is reused untouched**.

## How

- **New** `src/utils/offsite-brand-presence-semrush.js` → `loadCitedUrlsFromSemrush()`:
  - Calls `GET {SPACECAT_API_URI}/v2/orgs/:spaceCatId/brands/:brandId/serenity/brand-presence/url-inspector/domain-urls?hostname=…` for each offsite hostname (`youtube.com`, `reddit.com`).
  - Builds `allUrls` with **`count = row.citations`** (exact; not recounted by repetition), reusing the existing `classifyAndNormalize` for URL normalization, owned-URL filtering and domain bucketing → **the same map shape** as today.
  - `spaceCatId` from `site.getOrganizationId()`, `brandId` via `resolveBrandForSite`, date window via `getDateWindowForPreviousWeeks`.
[...]

PR #2852, opened 2026-08-06
fix(offsite): v2 IMS getServiceAccessToken + hard-stop on enableSemrush:true failure

AndreiAlexandruParaschiv:
## What

Follow-up to #2847. Fixes the runtime `IMS getServiceAccessTokenV3 request failed with status: 400 - unauthorized_client`.

## 1. Auth: v3 → v2 IMS token

The loader minted the service token with **v3 `getServiceAccessTokenV3()`** = `client_credentials` grant, using the worker's **default IMS client**. That client is only provisioned for the **`authorization_code`** grant → IMS returns `400 unauthorized_client`.

Switched to **v2 `getServiceAccessToken()`** (`authorization_code`, default `IMS_*` creds) — the exact S2S path `commerce-product-enrichments`, `vulnerabilities` and `permissions` already use successfully. (v3/`client_credentials` needs a *dedicated* integration, like content-ai's `CONTENTAI_*`; that isn't provisioned for LLMO/Semrush.)

## 2. Hard-stop vs fallback, by how Semrush was enabled

| Enabled by | Semrush **fails** (loader `null`) | Semrush returns 0 URLs (success) |
|---|---|---|
| `enableSemrush:true` (Slack override) | **HARD STOP** — `success:false`, `dataSource:'semrush'`, `fallbackReason`, **no legacy** | zero-URL semrush run |
| env var (or override `false`/absent) | **fall back** to legacy PostgREST → SharePoint | zero-URL semrush run |
| flag off | — (legacy only) | — |

Rationale: an operator explicitly testing Semrush on one run (`enableSemrush:true`) wants a failure **visible**, not masked by legacy; but an **env-enabled** production run must never be silently zeroed out, so it keeps the fallback.

## Auth still gated (LLMO-6709)

This makes the **token mint** succeed. Whether Semrush accepts the worker's *service* identity (the proxy forwards it unchanged and is built around *user* tokens) is still the open gate — verify with one canary run via `enableSemrush:true` and read `auditResult.dataSource` / the `Service token rejected` log.

## Testing
- v2 stub; hard-stop (override) vs env-fallback vs genuinely-empty; diagnostics `fallbackReason` on hard stop; fixed the poll-forwarding test (an `enableSemrush:true` run now hard-stops on the default null mock, so it seeds a success).
- Loader/constants **100% coverage**; full suite **9460 passing**.
- Flag off = unchanged; safe to merge.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Dominique Jäggi", "TB ADBE"]
```

### DIFF
--- a/docs/decisions/002-offsite-cited-urls-semrush-source.md
+++ b/docs/decisions/002-offsite-cited-urls-semrush-source.md
@@ -15,12 +15,21 @@ involves several non-obvious trade-offs, so it warrants an ADR alongside the spe
 
 ## Decision
 
-1. **Source swap behind a flag, Semrush-first with automatic legacy fallback.**
-   When `OFFSITE_BRAND_PRESENCE_SEMRUSH_ENABLED=true`, the runner tries the Semrush
-   loader first; if it yields **no usable result** it falls back to the legacy
-   `loadBrandPresenceData` (PostgREST → SharePoint). Flag off = legacy only.
-2. **"No usable result" = fall back, at the SURFACE granularity.** The loader
-   returns `null` — and the handler falls back — on: no org/brand, transient
+1. **Source swap behind a flag, Semrush-first — failure handling depends on HOW it
+   was enabled.** When Semrush is enabled (env var
+   `OFFSITE_BRAND_PRESENCE_SEMRUSH_ENABLED=true`, or the per-run override in Decision 6),
+   the runner uses the Semrush loader. On a Semrush **failure** (loader returns `null`):
+   - **Env-enabled, or override `false`/absent →** fall back to the legacy
+     `loadBrandPresenceData` (PostgREST → SharePoint), so production is never silently
+     zeroed out.
+   - **Explicitly forced on via `enableSemrush:true` (Slack override) →** **hard stop**:
+     return `success:false` with `dataSource:'semrush'` + `fallbackReason`, and do **NOT**
+     run legacy. An operator testing Semrush on one run wants the failure **visible**, not
+     masked by legacy.
+   A genuinely-empty-but-successful result (Map size 0) is **not** a failure — it is used as
+   a normal zero-URL Semrush run in both modes. Flag off (no override) = legacy only.
+2. **What counts as a Semrush FAILURE (`null`), at the SURFACE granularity.** The loader
+   returns `null` on: no org/brand, transient
    brand-resolution failure, no date window, IMS-token failure, or when a **whole
    surface (hostname) produced zero successful responses**. Rationale: a dropped
    *surface* (all YouTube or all Reddit engines failed) is indistinguishable from a
@@ -53,11 +62,16 @@ involves several non-obvious trade-offs, so it warrants an ADR alongside the spe
    `search-gpt`, `microsoft-copilot`, `gemini-2.5-flash`, `google-ai-overview`,
    `perplexity`) — and **sum** citations per URL. Validating this map upstream is
    LLMO-6710.
-4. **Auth = service IMS bearer, forwarded unchanged to Semrush.** No
-   `x-promise-token` for a service caller (that header is the non-IMS/browser path).
-   **Open risk:** the proxy forwards the token unchanged and is designed around a
-   real *user* IMS token; whether Semrush accepts the worker's *service* token is
-   unverified — the flag stays off until confirmed end-to-end (LLMO-6709).
+4. **Auth = IMS service bearer via v2 `getServiceAccessToken()` (`authorization_code`
+   grant), default IMS client.** This is the same S2S path `commerce-product-enrichments`,
+   `vulnerabilities` and `permissions` use. v3 `getServiceAccessTokenV3()`
+   (`client_credentials`) was tried first but returns IMS **`400 unauthorized_client`** —
+   the worker's default IMS client is only provisioned for `authorization_code`; only a
+   dedicated integration (content-ai's `CONTENTAI_*`) is registered for `client_credentials`.
+   The token is forwarded unchanged to Semrush; no `x-promise-token` for a service caller.
+   **Open risk (LLMO-6709):** the proxy is designed around a real *user* IMS token, so whether
+   Semrush accepts the worker's *service* token is unverified — the flag stays off until
+   confirmed end-to-end (use `enableSemrush:true` on one canary run, per Decision 6, to test).
 5. **Format parity.** The loader re-applies `YOUTUBE_URL_REGEX` / `REDDIT_URL_REGEX`
    (matching the legacy `handler.js` classify) so non-thread Reddit and lookalike
    YouTube hosts are dropped identically.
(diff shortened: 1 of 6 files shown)


MysticatBot (approved):

**Verdict: Approve** - well-scoped fix with clear justification; non-blocking notes below.
**Complexity: MEDIUM** - medium diff; ADR/spec update + auth change.
_Changes: Switches IMS auth from v3 (client_credentials) to v2 (authorization_code) and adds hard-stop semantics when Semrush is explicitly enabled via Slack override and fails (6 files)._
Note: Recommend a human read before merge - this change amends an architectural decision record (`docs/decisions/002-offsite-cited-urls-semrush-source.md`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: Broken comment wrap at `src/offsite-brand-presence/handler.js:998-1001` reads "there is / no legacy fallback on the Semrush path" which implies no fallback exists on the entire Semrush path (contradicts the env-enabled fallback behavior). Suggest: "An enableSemrush:true run that failed already returned success:false above - it never reaches this point."
- nit: ADR `docs/decisions/002-offsite-cited-urls-semrush-source.md` Decision 6 paragraph has a runon line after the insertion ("anything else (absent, empty, invalid) falls through to the env var unchanged. This is the intended mechanism...") - was not reflowed.
- suggestion: Add a test for `enableSemrush:true` + empty Map (size 0) to prove the hard-stop only triggers on `null`, not on a successful zero-URL result - `src/offsite-brand-presence/handler.js`
- suggestion: `dataSource: 'semrush'` on the hard-stop `success: false` return means "attempted source" rather than "served by source" - a one-line comment would prevent LLMO-6711 shadow-run parity queries from including failed runs in the join - `src/offsite-brand-presence/handler.js:965`
- nit: The env-fallback test ("falls back to legacy when an ENV-enabled Semrush run fails") dropped the `log.warn` assertion that existed before; re-adding it verifies the Splunk observability contract - `test/audits/offsite-brand-presence.test.js`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 47s | <b>Cost:</b> $5.75 | <b>Commit:</b> <code>67fe802008ad0cbc0998cd34bf45375e7b1f455b</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

tb-adbe (commented):

## Strengths
- The v3 to v2 IMS token switch matches an established pattern already used successfully by three other S2S integrations in this codebase (commerce-product-enrichments, vulnerabilities, and permissions all already call the v2 `getServiceAccessToken()` authorization_code path), so this aligns with proven precedent rather than introducing something new.
- The null-vs-empty-Map distinction in handler.js is now correct: a `null` loader return is treated as failure (routed through the hard-stop or fallback branch), while an empty `Map` (size 0) is treated as a legitimate zero-URL Semrush success with no fallback. This closes a real gap in the previous logic, which mis-classified an empty-but-successful map as a fallback case.
- The hard-stop branch (`enableSemrush:true` forcing an explicit failure with no legacy fallback) is correctly scoped to the Slack-triggered per-run override only, not reachable from any external or end-user input, and returns before any DRS scraping or URL-store writes happen, so the abort is side-effect-free.
- Every `diagnostics.fallbackReason` code the loader can produce (`no-organization-id`, `no-active-brand`, `brand-resolution-failed`, `no-date-window`, `ims-token-failed`, `surface-failed:<host>`) is set before the loader returns null, so the generic `?? 'semrush-failed'` default in the handler is a defensive fallback, not a path that fires in practice.
- New log lines and Slack messages on the hard-stop path only interpolate the diagnostic reason code, siteId, and baseURL, never the token or client secret.
- Test coverage maps cleanly onto the new branches (hard-stop with default reason, hard-stop with a loader-supplied reason, env-enabled fallback, and empty-but-successful with no fallback), and the assertions check actual result shape rather than just mock call counts. The poll-forwarding test was correctly adapted to stub a real Semrush result, since `enableSemrush:true` now hard-stops before reaching DRS poll scheduling on failure - the fix preserves what that test's name claims to verify.
- The ADR and spec are updated in the same diff with a clear rationale for both the auth change and the hard-stop-vs-fallback split.

## Issues

### Minor (Nice to Have)
- `src/offsite-brand-presence/handler.js:994` - the comment above the legacy-source block has an awkward mid-sentence line wrap ("...already hard-stopped / above -- there is / no legacy fallback on the Semrush path.") that splits "there is" and "no legacy fallback" across a bare line. Cosmetic only, worth squashing into one line on your next touch of this block.

## Recommendations
- The generic fallback string changed from `semrush-no-usable-urls` to `semrush-failed`. Since the loader always sets a specific reason before returning null, this default is effectively unreachable in production, but if any Splunk saved search or LLMO-6711 dashboard query keys on the literal old string, it would silently stop matching. Worth a quick check outside this repo.
- The ADR frames the hard-stop-on-override behavior primarily around the LLMO-6709 verification window, but doesn't say whether "enableSemrush:true means strict mode, no fallback" is meant to be a permanent convention for this override going forward, or scaffolding to remove once LLMO-6709 closes. One sentence in the ADR's Consequences section would remove that ambiguity for future maintainers.
- Since `enableSemrush:true` failures now produce `success:false` instead of a masked fallback, it's worth confirming before the first canary run that no external dashboard or on-call alert keyed on this audit's success rate would page or flag on an intentional, single-site, manually-triggered failure.
- When the canary run happens, worth keeping in mind that `getServiceAccessToken()` (v2) doesn't take a scope parameter the way `getServiceAccessTokenV3()` did - the minted token carries whatever the shared default client's static profile grants, with no per-call narrowing. That's consistent with how the other three integrations already use it, so it isn't a regression, just context for reading the canary's "Service token rejected" diagnostics if they show up.
- The `fallbackReason` codes flowing into the hard-stop log and Slack message are fixed, enum-like values today. Worth keeping that invariant in mind if the loader (not touched in this diff) is edited later - assigning `fallbackReason` from a raw `error.message` could start surfacing IMS response internals into a Slack channel, since the IMS client truncates but doesn't scrub error bodies.
- Consider adding a test for the hard-stop path with a real `slackContext` supplied, asserting `postMessageOptional` is called with the expected message/channel/threadTs. The two existing hard-stop tests omit `slackContext`, so the actual Slack-notify mechanism the ADR describes as "how the operator sees the failure" is never exercised end to end.

- Reusing the single default IMS Technical Account credential across a fourth, heterogeneous downstream (Semrush's proxy, which the ADR notes is designed around a real user IMS token) increases the blast radius of that shared credential once LLMO-6709 goes live - a leak or misconfiguration on the Semrush path would now also touch commerce-product-enrichments, vulnerabilities, and permissions token issuance. This is the concrete protocol-level shape of the identity risk the author already gates behind LLMO-6709, not a finding against this PR.

## Assessment
Ready to merge? Yes
Reasoning: The auth fix is protocol-correct and matches proven precedent elsewhere in the codebase, the hard-stop path is side-effect-free and reachable only via an operator-only override, and the null-vs-empty-Map logic is correctly implemented with tests that check real behavior. Only one cosmetic Minor finding surfaced.

Note: the `ci / build` check was still pending as of this review - worth confirming it's green before merge.

tb-adbe (approved):

PR #2850, opened 2026-08-05
fix(opportunity): stamp scopeType='site' + scopeId on new opps (SITES-49175)

absarasw:
## Summary

**DRAFT — blocked on [`spacecat-shared#1866`](https://github.com/adobe/spacecat-shared/pull/1866) releasing before this can merge.**

Stamps `scopeType='site'` and `scopeId=<siteId>` on every new site-scoped opportunity written by the audit-worker via `convertToOpportunity` in `src/common/opportunity.js`. Repairs the divergence from the V2 Mystique projector shape that was causing duplicate active opportunities per `(siteId, type)` on customer sites (Walmart, KPMG) — see [SITES-49175](https://jira.corp.adobe.com/browse/SITES-49175) for the full diagnosis.

## Root cause (recap)

- **V1 audit-worker writes** (THIS repo, today) → `scopeType=NULL, scopeId=NULL, siteId=<site>`
- **V2 Mystique projector writes** → `scopeType='site', scopeId=<site>, siteId=<site>`

Postgres unique-index NULL semantics treat these as distinct tuples, so both survive as `NEW` rows for the same site+type. Customer sees two `broken-backlinks` opportunities per site; suggestion grants split across them; "Edit target URL" modal shows "No AI suggestion available" because it resolves by opp id and lands on the empty duplicate.

## What changes

Two lines in `src/common/opportunity.js`:

```diff
       const opportunityData = {
         siteId: auditData.siteId,
         auditId: auditData.id,
         ...
         data: opportunityInstance.data,
+        scopeType: Oppty.SCOPE_TYPES.SITE,
+        scopeId: auditData.siteId,
       };
```

`Oppty` (aliased `Opportunity`) is already imported at line 12. The schema attribute validator at `opportunity.schema.js:70-73` on the data-access side gates `scopeType` against `Object.values(SCOPE_TYPES).includes(value)`, so once **spacecat-shared#1866** adds `SITE: 'site'`, this write starts landing successfully.

## Ordering

1. **`spacecat-shared#1866` merges + releases** (adds `SITE` to `SCOPE_TYPES`).
2. Renovate (or a manual bump) updates `@adobe/spacecat-shared-data-access` here.
3. **This PR** un-drafts + merges.

If this PR merges BEFORE step 1, every `Opportunity.create(...)` call in the audit-worker will fail the schema validator with "`site` not in SCOPE_TYPES" — hard breakage on every audit. That's why this PR is a **draft** and I'll leave it that way until the dep is available.

## Tests

[...]

PR #2853, opened 2026-08-06
fix(opportunity): reopen RESOLVED opportunity instead of creating duplicate

jsanchez07:
## Summary

- Adds a RESOLVED-status fallback in `convertToOpportunity`: when no NEW opportunity is found for a site/type, also checks RESOLVED opportunities and reopens the most recent matching one (sets it back to NEW) instead of creating a duplicate
- This prevents duplicate broken-backlinks opportunities from being created on sites whose original opportunity was resolved by the v2 audit's domain_score filter, and then found new broken backlinks on a subsequent audit run

## Root cause this addresses

`convertToOpportunity` only searched `STATUSES.NEW` when looking for an existing opportunity. If the original was RESOLVED (e.g. v2 returned 0 results due to domain_score >= 50 filtering), it missed it and created a new opportunity — leaving two active NEW opportunities for the same site and type.

## Relationship to PR #2850

PR #2850 (Abhinav's scope-type NULL fix) addresses the V1/V2 uniqueness gap at the DB index level. This change covers the separate case where the original opportunity is RESOLVED at the time of the audit — the unique index cannot prevent that scenario since RESOLVED rows are excluded from the partial index. The two changes are complementary and do not conflict. Consider folding together or reviewing side-by-side.

## Jira

Closes SITES-49355

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["@rpapani", "Dominique Jäggi"]
```

### DIFF
--- a/src/common/opportunity.js
+++ b/src/common/opportunity.js
@@ -51,6 +51,24 @@ export async function convertToOpportunity(auditUrl, auditData, context, createO
         }
         return false;
       });
+
+      if (!opportunity) {
+        // eslint-disable-next-line max-len
+        const resolvedOpportunities = await Opportunity.allBySiteIdAndStatus(auditData.siteId, Oppty.STATUSES.RESOLVED);
+        const resolvedOpportunity = resolvedOpportunities.find((oppty) => {
+          if (oppty.getType() === auditType) {
+            if (comparisonFn && typeof comparisonFn === 'function') {
+              return comparisonFn(oppty, opportunityInstance);
+            }
+            return true;
+          }
+          return false;
+        });
+        if (resolvedOpportunity) {
+          await resolvedOpportunity.setStatus(Oppty.STATUSES.NEW);
+          opportunity = resolvedOpportunity;
+        }
+      }
     } catch (e) {
       log.error(`Fetching opportunities for siteId ${auditData.siteId} failed with error: ${e.message}`);
       throw new Error(`Failed to fetch opportunities for siteId ${auditData.siteId}: ${e.message}`);
--- a/src/common/opportunity.js
+++ b/src/common/opportunity.js
@@ -55,15 +55,17 @@ export async function convertToOpportunity(auditUrl, auditData, context, createO
       if (!opportunity) {
         // eslint-disable-next-line max-len
         const resolvedOpportunities = await Opportunity.allBySiteIdAndStatus(auditData.siteId, Oppty.STATUSES.RESOLVED);
-        const resolvedOpportunity = resolvedOpportunities.find((oppty) => {
-          if (oppty.getType() === auditType) {
-            if (comparisonFn && typeof comparisonFn === 'function') {
-              return comparisonFn(oppty, opportunityInstance);
+        const resolvedOpportunity = [...resolvedOpportunities]
+          .sort((a, b) => new Date(b.getUpdatedAt()) - new Date(a.getUpdatedAt()))
+          .find((oppty) => {
+            if (oppty.getType() === auditType) {
+              if (comparisonFn && typeof comparisonFn === 'function') {
+                return comparisonFn(oppty, opportunityInstance);
+              }
+              return true;
             }
-            return true;
-          }
-          return false;
-        });
+            return false;
+          });
         if (resolvedOpportunity) {
           await resolvedOpportunity.setStatus(Oppty.STATUSES.NEW);
           opportunity = resolvedOpportunity;
--- a/src/common/opportunity.js
+++ b/src/common/opportunity.js
@@ -56,7 +56,7 @@ export async function convertToOpportunity(auditUrl, auditData, context, createO
         // eslint-disable-next-line max-len
         const resolvedOpportunities = await Opportunity.allBySiteIdAndStatus(auditData.siteId, Oppty.STATUSES.RESOLVED);
         const resolvedOpportunity = [...resolvedOpportunities]
-          .sort((a, b) => new Date(b.getUpdatedAt()) - new Date(a.getUpdatedAt()))
+          .sort((a, b) => new Date(b.getUpdatedAt?.() ?? 0) - new Date(a.getUpdatedAt?.() ?? 0))
           .find((oppty) => {
             if (oppty.getType() === auditType) {
               if (comparisonFn && typeof comparisonFn === 'function') {
--- a/test/common/opportunity.test.js
+++ b/test/common/opportunity.test.js
@@ -19,6 +19,165 @@ import chaiAsPromised from 'chai-as-promised';
 use(sinonChai);
 use(chaiAsPromised);
 
+describe('convertToOpportunity — RESOLVED fallback', () => {
+  let sandbox;
+  let convertToOpportunity;
+  let checkGoogleConnectionStub;
+
+  const siteId = 'site-id';
+  const auditId = 'audit-id';
+  const auditType = 'broken-backlinks';
+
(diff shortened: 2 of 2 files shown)


rpapani:
## Code review

### Overview
Adds a RESOLVED-status fallback to `convertToOpportunity`: when no NEW opportunity matches, it looks up RESOLVED opportunities for the site, reopens a matching one (`setStatus(NEW)`) and reuses it instead of creating a duplicate — fixing the case where the original was RESOLVED by v2's `domain_score` filter and a later audit finds new broken backlinks. The mechanism is sound and I confirmed it wires up correctly, but there are a couple of things to address before merge.

### Verified / good
- **The reopen persists.** Once `opportunity = resolvedOpportunity`, the flow falls into the existing `else` (update) branch, which calls `setAuditId` / `setData` / `setUpdatedBy('system')` / `await opportunity.save()`. So the in-memory `setStatus(NEW)` is persisted by that `save()` — no missing-save bug, and no `create()` is called (so the duplicate is actually prevented). ✓
- Respects the `high-organic-low-ctr` bypass (the new block is inside that same `if` guard) and reuses `comparisonFn` consistently with the NEW-search. ✓

### Findings

**1. (blocking — CI coverage) No test for the new branch; the repo enforces 100% coverage.** `.nycrc.json` sets `check-coverage: true` with `lines/branches/statements: 100`. The added block introduces several new branches (`if (!opportunity)`, the `comparisonFn && typeof … === 'function'` true/false, `if (resolvedOpportunity)` found/not-found) with no test — so `npm test` will fail the coverage gate. Add cases for: RESOLVED match found → reopened to NEW + reused (no `create`); RESOLVED present but no type match → still creates; and the `comparisonFn` true/false paths.

**2. (correctness vs. stated intent) "most recent" isn't guaranteed.** The PR summary says it reopens the *most recent* matching RESOLVED opp, but `resolvedOpportunities.find(...)` returns the **first** element in whatever order the query yields — there's no explicit sort. If a site has more than one RESOLVED opportunity of the same type (likely, since each resolve leaves a row), this can reopen an arbitrary/older one and carry its stale opportunity-level `data` forward. Either sort by recency before selecting (e.g. `[...resolvedOpportunities].sort((a, b) => new Date(b.getUpdatedAt()) - new Date(a.getUpdatedAt())).find(...)`) or confirm `allBySiteIdAndStatus` documents a newest-first ordering.

**3. (semantics — worth confirming) The reopen is indiscriminate.** It reopens **any** RESOLVED opportunity of that type — including one an operator **intentionally resolved/dismissed** in the UI, not just ones auto-resolved by the `domain_score` filter. That would silently override the operator's decision the next time an audit finds backlinks. If that's not intended, scope the fallback (e.g. only reopen when `getUpdatedBy() === 'system'`, or gate on the resolution reason). At minimum, call out in the PR whether un-resolving an operator-dismissed opportunity is acceptable.

**4. (minor) Stale suggestions on the reopened opportunity.** A RESOLVED opp's suggestions were resolved/outdated; reopening the opp to NEW doesn't reset them. The audit's downstream `syncSuggestions` should reconcile current findings, but please confirm old resolved suggestions don't linger on the reopened opportunity.

### Verdict
The fix is small and correct for its goal (reuse-and-reopen instead of duplicate), and it persists properly. Before merge: (1) add the tests the 100% coverage gate requires, and (2) make the "most recent" selection explicit rather than relying on `.find()` order. (3) is a product-intent check worth a sentence in the PR. Also — as you noted re #2850 — worth confirming the two changes are reviewed together, since they cover complementary halves of the same duplicate-opportunity problem.

jsanchez07:
**1 & 2 addressed** — tests added covering all new branches (found/not-found, comparisonFn true/false, wrong type), and the sort is now explicit: RESOLVED candidates are sorted by `updatedAt` descending before `.find()`, so the most recent is always picked regardless of query order.

**3 (indiscriminate reopen)** — intentional. If new broken backlinks are detected for a site, reopening a previously-resolved opportunity is the right behavior even if an operator dismissed it, because the underlying broken links are genuinely back. An operator who doesn't want to act on them can resolve again after reviewing. Silently not-creating would mean the data is never surfaced. That said, if the team prefers to gate on `getUpdatedBy() === 'system'`, happy to add that — just want to confirm either way.

**4 (stale suggestions)** — confirmed `syncSuggestionsWithPublishDetection` downstream reconciles this: it diffs current findings against stored suggestions and removes resolved ones, so stale suggestions on the reopened opportunity don't linger.

rpapani:
## Follow-up review — thanks for the tests + sort, but CI is red (6 failing)

Good progress: the RESOLVED-fallback tests are added, the `updatedAt` desc sort addresses the "most recent" concern, and **`opportunity.js` coverage is back to 100%**. But the latest build (`933c9b9`) fails with **6 failing tests**, 5 of which trace directly to this change.

### 1. The new tests fail — incomplete mocks (tests 4, 5, 6)
```
TypeError: opportunity.setScopeType is not a function
  at convertToOpportunity (src/common/opportunity.js:128)
```
When a RESOLVED opp is reopened, `opportunity` is set to it and then flows into the existing `else`/update branch, which drives the full SITES-49175 self-heal path: `setAuditId()`, `setScopeType('site')`, `setScopeId()`, `setData()` (for CWV/META_TAGS/etc.), and `save()`. The mock RESOLVED opportunity in the new tests only stubs `getType`/`getUpdatedAt`/`setStatus`, so it blows up on `setScopeType`. The mock needs the full setter surface the update branch calls (`setAuditId`, `setScopeType`, `setScopeId`, `setData`, `save`) — and the tests should assert `save()` was called (that's what actually persists the reopen).

### 2. Two existing tests regressed — unguarded sort (tests 1 & 2)
```
Error: Failed to fetch opportunities for siteId ...: b.getUpdatedAt is not a function
  (canonical.test.js, hreflang.test.js)
```
The sort runs unconditionally over every RESOLVED candidate:
```js
[...resolvedOpportunities].sort((a, b) => new Date(b.getUpdatedAt()) - new Date(a.getUpdatedAt()))
```
Existing tests stub `allBySiteIdAndStatus(…RESOLVED)` with mock opps that have no `getUpdatedAt()`, so the sort throws, gets caught, and rethrows as "Failed to fetch opportunities" — failing the whole audit. Real `Opportunity` model objects do have `getUpdatedAt()`, so this isn't a prod crash, but it's a real backward-compat break: those fixtures now need `getUpdatedAt`, and the sort has no fallback if a row ever lacks a parseable timestamp.

### 3. One failure looks unrelated (test 3)
```
TypeError: Cannot read properties of undefined (reading 'url')
  at prerender/handler.test.js:511
```
Pure prerender test-file error, nothing touching `opportunity.js` — likely pre-existing/flaky or pulled in via the `main` merge. Worth confirming it's also red on `main` so it isn't mistaken for this PR.

### Still open from the first review
Semantics (non-blocking, product-intent): the reopen fires for *any* RESOLVED opp of that type, including one an operator intentionally dismissed — that silently un-resolves their decision. Consider gating on auto-resolved only (e.g. `getUpdatedBy() === 'system'`), or confirm that reopening operator-dismissed opps is intended.

### Net
On the right track and coverage is satisfied, but not mergeable until the 5 opportunity-related failures are green — fix the new-test mocks (full setter surface + assert `save()`), update the two regressed fixtures to provide `getUpdatedAt`, and rule out the prerender failure against `main`.

rpapani (approved):

PR #2854, opened 2026-08-06
feat(offsite): log Semrush non-2xx response body to pinpoint the 401 rejecter (LLMO-6709)

AndreiAlexandruParaschiv:
## What

The offsite Semrush canary (`enableSemrush:true` on lovesac.com) returns **`HTTP 401 — Service token rejected`** for all requests, confirming the LLMO-6709 auth gate: the worker's IMS **service** token is rejected (the same endpoints/engines return data with a **user** token).

The worker currently logs only the **status** — not the body. But the 401 body identifies **who** is rejecting, which decides who owns the fix:
- **api-service `requireImsBearer`** → body like *"Elements proxy requires IMS authentication; send the x-promise-token header instead"* (the proxy never reached Semrush).
- **Semrush upstream** → a Semrush error payload (the proxy forwarded the token and Semrush rejected the identity).

## Change

- Add `readErrorBodySnippet(response)` — safe (never throws), capped at 500 chars.
- Include `responseBody` in the non-2xx error log context (`[offsite-brand-presence][semrush]` … `status`, `platform`, `hostname`, `responseBody`).

No behavior change — logging only. Flag-off path untouched.

## Why now

It's the fastest way to settle the "api-service vs Semrush" ownership question for LLMO-6709 without waiting on api-service logs — the next canary run will show the rejecter in the worker logs directly.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["@amitruluimadalina", "Dominique Jäggi"]
```


MysticatBot (approved):

**Verdict: Approve** - clean, well-scoped diagnostic logging addition with intentional coverage of all three paths.
**Complexity: LOW** - small diff, single utility, logging-only.
_Changes: Adds response body capture to non-2xx error logs for the Semrush offsite loader to identify the 401 rejecter (2 files)._
Note: CI checks are currently pending - confirm they pass before merge.

<details>

- nit: JSDoc says "trimmed + capped" but implementation only caps (no `.trim()` call) - `src/utils/offsite-brand-presence-semrush.js:65`. Either add `.trim()` before `.slice()` or remove "trimmed +" from the doc.
- suggestion: Add `text: async () => '...'` to the existing 403 stub (~line 304 in test file) so it explicitly exercises the body-read path rather than implicitly relying on the catch branch. The coverage is intentional per the PR description, but an assertion on `responseBody` would make the contract obvious to future readers.
- suggestion: A test with a body exceeding 500 chars would document the truncation cap contract directly.
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 8s | <b>Cost:</b> $4.04 | <b>Commit:</b> <code>50d205c139dd4d081a7e1cf49aadd9db90834f94</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

amitruluimadalina (approved):

PR #2855, opened 2026-08-06
fix(preflight): loosen accessibility scrape nav wait to avoid 45s timeout (SITES-49365)

clotton:
## What

Preflight's accessibility check scrapes via the content-scraper accessibility handler, which navigates with `waitUntil: ['networkidle2', 'load']` (both events required). On heavy author/marketing pages whose network never goes idle, `page.goto` aborts at the ~45s nav timeout, the scrape fails, and the audit polls ~10 min and returns an **empty (false-clean)** result.

This sets `accessibilityScrapingParams: { waitUntil: 'domcontentloaded', additionalTimeout: 3000 }` on the Preflight accessibility scrape message so navigation completes reliably (axe needs the DOM built, not an idle network), with a short fixed settle for late/SPA content.

## Scope — Preflight only

- The `accessibilityScrapingParams.waitUntil` / `additionalTimeout` hook **already exists** in the content-scraper accessibility handler, so **no content-scraper change** is needed.
- ASO scheduled accessibility scans are sent from a **different** file (`src/accessibility/handler.js`, config-driven `scrapingParams`) and are **unaffected**. The fleet-wide ASO timeout is tracked separately in **SITES-49369**.

## Provability

Adds an info-level `[preflight-audit]` marker so prod Splunk can confirm the tuning is live (the existing full-message log is debug-only). The query plus the two content-scraper signals are documented in SITES-49365.

## Testing

- Extended `test/audits/preflight.test.js` to assert the scrape message carries `waitUntil: 'domcontentloaded'` + `additionalTimeout: 3000` and that the info marker is emitted.
- `preflight.test.js`: 149 passing; eslint clean.

Fixes SITES-49365. Sibling: SITES-49369 (ASO). Related: SITES-49360 (error surfacing).

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Guy Pelletier", "Dominique Jäggi"]
```

### DIFF
--- a/src/preflight/accessibility.js
+++ b/src/preflight/accessibility.js
@@ -21,6 +21,25 @@ import { formatWcagRule } from '../accessibility/utils/generate-individual-oppor
 
 export const PREFLIGHT_ACCESSIBILITY = 'accessibility';
 
+/*
+ * Navigation tuning for the Preflight accessibility scrape.
+ *
+ * The content-scraper accessibility handler defaults to
+ * waitUntil ['networkidle2', 'load'], which requires the network to fully
+ * settle before axe runs. Heavy author/marketing pages with persistent
+ * background traffic never reach that state, so page.goto aborts at the ~45s
+ * nav timeout, the scrape fails, and the audit returns an empty (false-clean)
+ * result. axe only needs the DOM built, not an idle network, so we navigate on
+ * 'domcontentloaded' (which always fires) and add a short fixed settle for
+ * late/SPA content.
+ *
+ * Scoped to Preflight only: ASO scheduled accessibility scans are sent from
+ * src/accessibility/handler.js (config-driven scrapingParams) and are NOT
+ * affected by this. See SITES-49365.
+ */
+export const PREFLIGHT_A11Y_SCRAPE_WAIT_UNTIL = 'domcontentloaded';
+export const PREFLIGHT_A11Y_SCRAPE_SETTLE_MS = 3000;
+
 /**
  * Generate normalized filename from URL
  */
@@ -105,9 +124,21 @@ export async function scrapeAccessibilityData(context, auditContext) {
         options: {
           enableAuthentication,
           a11yPreflight: true,
+          // Loosen navigation so heavy pages don't 45s-timeout into empty
+          // results; Preflight-scoped, does not affect ASO scans (SITES-49365).
+          accessibilityScrapingParams: {
+            waitUntil: PREFLIGHT_A11Y_SCRAPE_WAIT_UNTIL,
+            additionalTimeout: PREFLIGHT_A11Y_SCRAPE_SETTLE_MS,
+          },
           ...(context.promiseToken ? { promiseToken: context.promiseToken } : {}),
         },
       };
+
+      // Info-level marker so prod Splunk can confirm the nav tuning is live
+      // (the full-message log below is debug-only and not emitted in prod).
+      log.info(`[preflight-audit] site: ${siteId}, job: ${jobId}, step: ${step}. `
+        + `Accessibility scrape nav tuned: waitUntil=${PREFLIGHT_A11Y_SCRAPE_WAIT_UNTIL} `
+        + `settleMs=${PREFLIGHT_A11Y_SCRAPE_SETTLE_MS} (SITES-49365)`);
 
       log.debug(`[preflight-audit] Scrape message being sent: ${JSON.stringify(scrapeMessage, null, 2)}`);
       log.debug(`[preflight-audit] Processing type: ${scrapeMessage.processingType}`);
--- a/test/audits/preflight.test.js
+++ b/test/audits/preflight.test.js
@@ -2359,6 +2359,18 @@ describe('Preflight Audit', () => {
           promiseToken: 'test-token',
         });
 
+        // Preflight-scoped nav tuning: loosen waitUntil so heavy pages don't
+        // 45s-timeout into empty results (SITES-49365). ASO scans are unaffected.
+        expect(message.options.accessibilityScrapingParams).to.deep.equal({
+          waitUntil: 'domcontentloaded',
+          additionalTimeout: 3000,
+        });
+
+        // Info-level marker must be emitted so prod Splunk can prove it is live.
+        expect(log.info).to.have.been.calledWithMatch(
+          'Accessibility scrape nav tuned: waitUntil=domcontentloaded settleMs=3000',
+        );
+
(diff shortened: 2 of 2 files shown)


MysticatBot (approved):
Hey @clotton,

**Verdict: Approve** - clean, well-scoped fix with good test coverage and clear documentation.
**Complexity: LOW** - small targeted fix, single service, 2 files.
_Changes: Passes `accessibilityScrapingParams` with `waitUntil: 'domcontentloaded'` and a 3s settle to the Preflight accessibility scrape message, avoiding the 45s navigation timeout on heavy pages (2 files)._

<details>

- nit: The info-level log at `src/preflight/accessibility.js:138` is described as a rollout-verification marker but will emit on every single Preflight accessibility scrape permanently. Consider demoting to `debug` after confirming the tuning is live, or adding a follow-up to do so.
- suggestion: `PREFLIGHT_A11Y_SCRAPE_WAIT_UNTIL` and `PREFLIGHT_A11Y_SCRAPE_SETTLE_MS` are exported but only consumed within the same file. If the export exists solely for test assertions, the test already validates the outbound message shape directly (which is the stronger behavioral assertion). Keeping them unexported would narrow the module's public contract.
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 0m 42s | <b>Cost:</b> $2.91 | <b>Commit:</b> <code>677d8a3d62a42d43782969ba7a797e7906897398</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

clotton:
## Local validation — before/after on the three worst offenders

Ran the accessibility handler's navigation semantics (same puppeteer-extra + stealth stack, real Chrome, real sites, same 45s `pageLoadTimeout`) with the OLD `waitUntil` vs this PR's NEW one:

| Site (prod 14d timeouts) | OLD `['networkidle2','load']` | NEW `domcontentloaded` +3s settle | axe violations |
|---|---|---|---|
| T-Mobile (18) | ❌ **45.0s TIMEOUT** — empty (`Navigation timeout of 45000 ms exceeded`) | ✅ 0.53s, HTTP 200 | — → **3** |
| AT&T (5) | ✅ 9.0s | ✅ 1.5s | 2 → **2** |
| Walmart (4) | ✅ 18.8s | ✅ 0.5s | 3 → **3** |

**Takeaways:**
1. **Fix eliminates the timeout.** T-Mobile reproduced the exact prod error under the old wait (empty result); the new wait navigated in 0.53s and returned 3 real axe violations — i.e. empty false-clean → populated.
2. **No axe regression.** Every site returned the identical violation count under the new wait (the 3s settle covers late/SPA content). This was the main risk with `domcontentloaded`.
3. **6×–38× faster navigation** even when the old wait didn't time out, which also reduces preflight latency against the MFE polling budget.

Scope: reproduces the `waitUntil` navigation behaviour faithfully; not the full Lambda/S3/SQS path. T-Mobile matching the exact prod error string confirms it's not a local artifact.

GeezerPelletier (approved):
## Review: loosen accessibility scrape nav wait to avoid 45s timeout (SITES-49365)

Ran the tests and lint against the PR branch locally — **149 passing** (matches the PR's claim exactly), clean eslint on both changed files (verified against the committed blob content directly, bypassing a local Windows CRLF checkout artifact that otherwise produces noisy false positives).

### What it does
Preflight's accessibility scrape was navigating with the content-scraper's default `waitUntil: ['networkidle2', 'load']` — requiring the network to fully settle before axe runs. Heavy author/marketing pages with persistent background traffic (analytics beacons, polling, etc.) never reach network-idle, so `page.goto` aborts at the accessibility profile's 45s nav timeout, the scrape fails, and the audit silently returns an empty (false-clean) result after ~10 minutes of polling. This sets `waitUntil: 'domcontentloaded'` (always fires) + a fixed 3s settle for late/SPA content, scoped to Preflight's own scrape message.

### Verified, not just trusted
Checked the two load-bearing claims against the actual dependent code rather than taking the PR description at face value:
- **The content-scraper hook already exists exactly as described** (`spacecat-content-scraper/src/handlers/accessibility-handler.js:183-196`): `navWaitUntil = accessibilityScrapingParams?.waitUntil || waitUntil || ['networkidle2', 'load']`, then a plain `setTimeout` for `additionalTimeout` after `page.goto` resolves. So this PR genuinely needs no companion content-scraper change.
- **The "~45s nav timeout" is exact**, not rounded for the PR description — `handler-profiles.js:46`, the `accessibility` profile's `pageLoadTimeout: 45000`.
- **The ASO-isolation claim is real.** `src/accessibility/handler.js` (the scheduled ASO path) independently derives its own `accessibilityScrapingParams` from `scrapingConfig?.state?.handlers?.accessibility?.scrapingParams` — a completely separate, config-driven source, with zero shared code path or constant with the Preflight file being changed here. The two files' only shared import is an unrelated WCAG-formatting helper.

### Test quality
The new assertions are exactly the two things worth pinning: the scrape message actually carries the tuned params (`deep.equal` on the whole `accessibilityScrapingParams` object, not just a loose property check), and the new info-level Splunk marker fires with the right values — which matters here specifically because the PR's own stated goal is "provability in prod," so a test that didn't check the log line would leave that half of the fix unverified.

### Minor, non-blocking observation
`domcontentloaded` + a flat 3s settle is a reasonable, conservative tradeoff (axe needs DOM structure, not idle network) but is inherently a heuristic — a sufficiently slow SPA could still render its accessibility-relevant DOM after the 3s window closes, trading "always empty" for "occasionally still incomplete." Given the PR explicitly frames this as fixing a **100%-failure** case (nav timeout abort) rather than claiming perfect coverage, and ships a Splunk marker to make the tuning's effect observable in prod, that seems like the right call — just flagging it as the honest limit of this fix, not a defect.

Small, well-scoped, well-verified fix with accurate isolation claims. Approving.

clotton:
Follow-up: **SITES-49491** — adobe/spacecat-audit-worker#2861 + adobe/spacecat-content-scraper#1009.

The fixed 3s settle introduced here races async/lazy content and causes intermittent accessibility **under-reporting** (verified in prod: the same Casio page returned 4 a11y issues in one run and 0 in another minutes apart — jobIds 019fdc8c vs 019fdc8e). The follow-up adds a **bounded** `page.waitForNetworkIdle()` settle so axe waits for async content and returns consistent results, while keeping this PR's no-45s-hang property (the wait is capped + non-fatal).

PR #2857, opened 2026-08-07
fix(prerender): restore unrestricted top-pages fallback when baseURL is empty

clotton:
## Problem

`@adobe/spacecat-shared-utils` 1.124.1 (picked up on `main` via the recent `fix(deps): update adobe fixes` Renovate PR) changed `isWithinSiteScope`'s handling of an empty `siteBaseUrl` from `return true` (unrestricted) to `return false` (fail closed). That's the right default for a scope *check*, but it broke `src/prerender/handler.js`'s `submitForScraping`: sites with no configured baseURL now get **zero** scraping URLs from `filterBySiteScope` instead of falling back to the unfiltered top-pages list — a real behavior regression, not just a test artifact, since all three `filterBySiteScope` call sites in this file share the same `siteBaseUrl`.

This is currently failing on `main` (`test/audits/prerender/handler.test.js` — "should fall back to top pages when baseUrl is empty"), blocking `semantic-release`/deploy for everything merged since.

## Fix

Added a local `scopeToBaseUrl(urls, siteBaseUrl)` wrapper that skips `filterBySiteScope` entirely when there's no baseURL to scope to, restoring the previous, intended behavior explicitly at the call site rather than relying on the shared library's default. Replaced all three `filterBySiteScope` call sites in `src/prerender/handler.js` with it.

## Testing

- `test/audits/prerender/handler.test.js`: 291 passing (was 290 passing / 1 failing)
- `eslint src/prerender/handler.js`: clean

## Related Issues

None yet — this is a live production-blocking regression discovered while verifying an unrelated fix's deploy status, not pre-planned work. Happy to file a Jira ticket for tracking if the team wants one; flagging here in the meantime so it isn't lost.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -59,6 +59,14 @@ function rebaseUrl(url, preferredBase, log) {
   }
 }
 
+// Sites with no configured baseURL have no scope to restrict to, so nothing should be
+// filtered out. filterBySiteScope's own empty-siteBaseUrl handling is fail-closed (it
+// rejects everything), which is the right default for a scope check but not what
+// prerender wants here: an unconfigured baseURL should fall back to unfiltered top pages.
+function scopeToBaseUrl(urls, siteBaseUrl) {
+  return siteBaseUrl ? filterBySiteScope(urls, siteBaseUrl) : urls;
+}
+
 const LOG_PREFIX = 'Prerender -';
 const AUDIT_TYPE = Audit.AUDIT_TYPES.PRERENDER;
 const { AUDIT_STEP_DESTINATIONS } = Audit;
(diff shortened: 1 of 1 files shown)


MysticatBot (approved):
Hey @clotton,

**Verdict: Approve** - clean, minimal fix that correctly restores the intended fallback behavior.
**Complexity: TRIVIAL** - single-file, 8-line wrapper function replacing 3 call sites.
_Changes: Adds a local `scopeToBaseUrl` wrapper that bypasses `filterBySiteScope` when no baseURL is configured, restoring the prerender handler's unfiltered top-pages fallback after the upstream library changed empty-baseURL handling to fail-closed (1 file)._
Note: CI checks are currently pending - confirm build passes before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 27s | <b>Cost:</b> $2.76 | <b>Commit:</b> <code>cbb1d7709571219375b0b3fa45c14f55d8f01cb8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2863, opened 2026-08-07
fix(opportunity): prefer scoped row to avoid unique constraint collision (SITES-49175)

jsanchez07:
## Summary

PR #2850 added scope-stamping to `convertToOpportunity` to self-heal legacy null-scope opportunity rows. What it didn't account for: when Mystique's projector has already created a scoped row (`scope_type='site'`) for a site, the legacy null-scope row still exists alongside it. When audit-worker picks the null-scope row and tries to stamp it, `save()` violates the partial unique index `idx_opportunities_unique_active_scope` — a scoped row for the same `(site_id, type)` already exists. That exception propagates uncaught through `generateSuggestionData`, aborting it entirely: Bright Data never runs, no new suggestions are created.

**Confirmed in prod: 161 sites currently have this two-row pattern (one `scope_type=null`, one `scope_type=site`).**

## Fix

Switch from `.find()` to `.filter()` and explicitly prefer the already-scoped row when both exist:

```js
const typeMatches = opportunities.filter((oppty) => {
  if (oppty.getType() === auditType) {
    if (comparisonFn && typeof comparisonFn === 'function') {
      return comparisonFn(oppty, opportunityInstance);
    }
    return true;
  }
  return false;
});
// Prefer a scoped row over a legacy null-scope row: if both exist for the same
// (siteId, type), stamping scope on the null-scope row would violate
// idx_opportunities_unique_active_scope and abort the audit step.
opportunity = typeMatches.find((oppty) => oppty.getScopeType() !== null)
  ?? typeMatches[0];
```

## Behavioral impact

This change is in `convertToOpportunity`, which is shared by all audits. The behavior differs by case:

- **One matching row (the vast majority of sites and audits):** `typeMatches[0]` is identical to what `.find()` returned. No change.
- **Two matching rows, one null-scope and one scoped (the 161-site scenario):** old behavior was to pick the null-scope row and crash with a unique constraint violation. New behavior picks the already-scoped row, skips stamping, and proceeds correctly. This applies to any audit type that lands in the two-row situation, not just broken-backlinks.
- **Two matching rows, same scope type:** falls back to `typeMatches[0]`, same result as `.find()`. No change.

## Why 40 files changed

Only **one production file** changed (`src/common/opportunity.js`). The other 39 are test files. The fix calls `oppty.getScopeType()` on opportunity objects — real DB objects have this method, but mock objects in tests don't unless explicitly added. Each test file gets `getScopeType: () => null` and `getScopeId: () => null` added to mock opportunity objects — one or two lines per file.

## Test plan

- [ ] All 9,465 existing tests pass at 100% coverage (verified locally)
- [ ] After merge + deploy: re-trigger broken-backlinks audit for one of the 161 affected sites and confirm `generateSuggestionData` completes, Bright Data runs, and new suggestions are created

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Abhinav Saraswat", "Dominique Jäggi"]
```

### DIFF
--- a/src/common/opportunity.js
+++ b/src/common/opportunity.js
@@ -40,17 +40,20 @@ export async function convertToOpportunity(auditUrl, auditData, context, createO
     try {
       // eslint-disable-next-line max-len
       const opportunities = await Opportunity.allBySiteIdAndStatus(auditData.siteId, Oppty.STATUSES.NEW);
-      opportunity = opportunities.find((oppty) => {
+      const typeMatches = opportunities.filter((oppty) => {
         if (oppty.getType() === auditType) {
-          // If comparison function is provided, use it to determine if opportunities are the same
           if (comparisonFn && typeof comparisonFn === 'function') {
             return comparisonFn(oppty, opportunityInstance);
           }
-          // Default behavior: just match by type
           return true;
         }
         return false;
       });
+      // Prefer a scoped row over a legacy null-scope row: if both exist for the same
+      // (siteId, type), stamping scope on the null-scope row would violate
+      // idx_opportunities_unique_active_scope and abort the audit step.
+      opportunity = typeMatches.find((oppty) => oppty.getScopeType() !== null)
+        ?? typeMatches[0];
 
       if (!opportunity) {
         // eslint-disable-next-line max-len
--- a/src/common/opportunity.js
+++ b/src/common/opportunity.js
@@ -50,9 +50,9 @@ export async function convertToOpportunity(auditUrl, auditData, context, createO
         return false;
       });
       // Prefer a scoped row over a legacy null-scope row: if both exist for the same
-      // (siteId, type), stamping scope on the null-scope row would violate
+      // (siteId, type), stamping scope on the null/undefined-scope row would violate
       // idx_opportunities_unique_active_scope and abort the audit step.
-      opportunity = typeMatches.find((oppty) => oppty.getScopeType() !== null)
+      opportunity = typeMatches.find((oppty) => oppty.getScopeType())
         ?? typeMatches[0];
 
       if (!opportunity) {
--- a/test/audits/accessibility/generate-individual-opportunities.test.js
+++ b/test/audits/accessibility/generate-individual-opportunities.test.js
@@ -2184,6 +2184,8 @@ describe('createAccessibilityIndividualOpportunities', () => {
       // SITES-49175 — self-heal legacy NULL-scope rows on every audit touch
       setScopeType: sandbox.stub(),
       setScopeId: sandbox.stub(),
+      getScopeType: () => null,
+      getScopeId: () => null,
       setUpdatedBy: sandbox.stub(),
       save: sandbox.stub().rejects(new Error('Create Error')),
     };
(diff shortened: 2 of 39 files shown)


absarasw (approved):
LGTM — the fix is correct and well-targeted. Approving; a few observations below, only the first is a nit worth addressing before merge.

## Why the fix works (rephrased for the PR description)

The description says "skips the stamp entirely" but the code actually stamps the already-scoped row — that's still safe because it's a no-op UPDATE on the same tuple. Worth aligning the wording:

> Prefer the already-scoped row so `setScopeType('site')` becomes a no-op UPDATE (same tuple, no constraint violation) instead of a NULL→'site' UPDATE on the legacy row that would collide with the existing scoped row.

## Nit — truthy check instead of `!== null`

`opportunity = typeMatches.find((oppty) => oppty.getScopeType() !== null) ?? typeMatches[0];`

`!== null` misses `undefined`. If the model layer ever returns `undefined` for an unset attribute (some PostgREST/ElectroDB paths do), a null-scope row would masquerade as scoped and we'd fall through to `typeMatches[0]`, defeating the preference.

```js
// truthy check — treats null/undefined uniformly
opportunity = typeMatches.find((oppty) => oppty.getScopeType()) ?? typeMatches[0];
// or explicit loose-null check
opportunity = typeMatches.find((oppty) => oppty.getScopeType() != null) ?? typeMatches[0];
```

Cheap guard against a future surprise.

## Observation — legacy null-scope rows stay orphaned

When both rows exist, we prefer the scoped one and never touch the null-scope one. It stays NEW forever, invisible to the audit path, but still a real DB row with `status='NEW'` and `scope_type=NULL`. Correct call for this PR's scope, but flagging for follow-up:

- ~161 sites on the two-row pattern per your DB query → same number of orphaned null-scope rows after audits run.
- Options:
  1. One-off backfill script: PATCH the null-scope row → `IGNORED` for each affected site.
  2. Have this fix also flip the null-scope row to `IGNORED` when both exist:

```js
const scopedMatch = typeMatches.find((oppty) => oppty.getScopeType() != null);
const nullScopeMatch = typeMatches.find((oppty) => oppty.getScopeType() == null);
if (scopedMatch && nullScopeMatch) {
  await nullScopeMatch.setStatus(Oppty.STATUSES.IGNORED);
  await nullScopeMatch.save();
}
opportunity = scopedMatch ?? typeMatches[0];
```

Fine either way. If we don't do it here, worth a follow-up ticket so the orphans don't linger indefinitely.

## Observation — pre-existing CREATE-branch race (not for this PR)

Latent race that predates this PR:

- T1: audit-worker's `allBySiteIdAndStatus` returns empty.
- T2: projector-service creates a scoped row via the RPC.
- T3: audit-worker CREATE branch → `Opportunity.create` INSERT → collides.

`Opportunity.create` doesn't do ON CONFLICT. Rare but non-zero. Worth a small follow-up: either use an RPC-with-ON-CONFLICT flavor for audit-worker's CREATE, or retry on 409. Not blocking this PR.

## Tests — LGTM

- ✅ Both-rows-prefer-scoped
- ✅ Only-null-scope-uses-first
- ✅ Central `MockContextBuilder` in `test/shared.js` returns `null` for `getScopeType`/`getScopeId` — avoids per-test churn going forward.

Optional: one defensive test for `two rows both scoped` (shouldn't happen under the partial unique index but tests the code path if it ever does — `.find(scoped)` returns the first).

PR #2866, opened 2026-08-08
fix(cwv): reclaim stale stuck IN_PROGRESS suggestions on re-audit

tkotthakota-adobe:
## Problem

Mystique flips a CWV `CODE_CHANGE` suggestion to **`IN_PROGRESS`** the moment it finishes generating a code-fix patch — that status is the *deploy hand-off* signal and the sole trigger consumed by the autofix-worker. On the large majority of sites `cwv-auto-fix` is **off**, so nothing ever consumes the hand-off and the suggestion sits `IN_PROGRESS` **forever**.

The CWV re-audit preserves `IN_PROGRESS` on every run, so these never self-heal — they **accumulate every weekly audit** and clog the customer-facing "Deployed" tab even though the patch is ready and no deploy is (or ever will be) in flight. A transient upstream outage — e.g. the **Aug 2026 clientlib-builder outage** that stranded ~2,048 suggestions across ~203 sites — produces the same end state at scale.

Blackboard V2 (the durable fix) is far off; this is the near-term, self-healing stopgap.

## Fix

Add a CWV-specific `mergeStatusFunction` (`createMergeCwvStatus`) to the `syncSuggestions` call in `syncOpportunitiesAndSuggestions`. On re-audit it reclaims a matched suggestion **`IN_PROGRESS → NEW` only when all** hold:

1. the suggestion is `IN_PROGRESS`;
2. it has **no active fix entity** (`PENDING`/`DEPLOYED`/`PUBLISHED`, keyed by `changeDetails.suggestionId`); and
3. it is **stale** (> `STALE_IN_PROGRESS_MS` = **24h** since `updatedAt`).

Every other case delegates to the existing `defaultMergeStatusFunction` unchanged (preserving `OUTDATED`-regression and `ERROR → NEW`). The fix-entity fetch is wrapped in try/catch — **any failure fails safe** (skips reclaim for that run).

## Why it is safe

- **No data loss** — the patch lives in `suggestion.data`; only the status flips. Reclaimed suggestions keep `isCodeChangeAvailable === true`, so they re-surface as actionable and are **not** re-dispatched for regeneration.
- **Never touches a live deploy** — active-fix guard + 24h window (far longer than a real hand-off) mean a freshly-set `IN_PROGRESS` is never reclaimed mid-flight.
- **Reversible** — plain status change via the same API.
- **Fail-safe** — fix-entity fetch error disables reclaim for that run.

## Consequences

- First re-audit after ship flips each site's accumulated stale `IN_PROGRESS` to `NEW` in one bounded pass (one-time cleanup). Customer-visible: stuck "Deployed" items reappear as actionable "Current" items — the intended outcome.
- Suggestions with a **stuck `PENDING`** fix (deploy PR opened, never merged) are deliberately **not** reclaimed here — handled via `delete-cwv-fixes`.
- The 24h threshold is the one tuning knob (`STALE_IN_PROGRESS_MS`).

See **ADR-003** (`docs/decisions/003-reclaim-stale-inprogress-cwv-suggestions.md`) for the full decision record and alternatives.

## Testing

11 new tests (3 integration through `syncOpportunityAndSuggestionsStep`, 8 unit on `createMergeCwvStatus`): stale-reclaim, active-fix-preserved, fresh-not-stale, fetch-failure-no-reclaim, non-`IN_PROGRESS` unchanged, `OUTDATED`-regression (both branches), `ERROR → NEW`, 24h constant. Full `npm test` green with **100% line/branch/statement** coverage; `npm run lint` clean.

## Relationship to other work

Complements autofix-worker **#686** (no-PR/`CM_STANDARD` structural strand). This PR covers the strand #686 cannot reach — the transient-outage / auto-deploy-off backlog — because the re-audit is the one place that runs periodically on every site regardless of deploy channel.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Julien Ramboz", "Dominique Jäggi"]
```

### DIFF
--- a/src/cwv/opportunity-sync.js
+++ b/src/cwv/opportunity-sync.js
@@ -265,6 +265,14 @@ export async function syncOpportunitiesAndSuggestions(context) {
     const fixEntities = await opportunity.getFixEntities();
     (fixEntities || []).forEach((fixEntity) => {
       if (ACTIVE_FIX_STATUSES.has(fixEntity.getStatus())) {
+        // Contract (cross-repo): the CWV fix→suggestion link is the denormalized
+        // changeDetails.suggestionId that spacecat-autofix-worker owns — its
+        // cwv/handler.js writes changeDetails = { suggestionId, issueId, issueTitle }
+        // and code-repo-manager.js reads it back. We deliberately key on that
+        // rather than the canonical FixEntitySuggestion join for an O(1) reverse
+        // lookup. If a future CWV fix writer ever lands without suggestionId, this
+        // guard misses it — but the 24h staleness window + fail-safe (fixFetchFailed)
+        // bound the blast radius, so a genuinely-live deploy is never reclaimed here.
         const suggestionId = fixEntity.getChangeDetails?.()?.suggestionId;
         if (suggestionId) {
           activeFixSuggestionIds.add(suggestionId);
--- a/src/cwv/opportunity-sync.js
+++ b/src/cwv/opportunity-sync.js
@@ -10,8 +10,8 @@
  * governing permissions and limitations under the License.
  */
 
-import { Audit, Suggestion } from '@adobe/spacecat-shared-data-access';
-import { syncSuggestions } from '../utils/data-access.js';
+import { Audit, Suggestion, FixEntity } from '@adobe/spacecat-shared-data-access';
+import { syncSuggestions, defaultMergeStatusFunction } from '../utils/data-access.js';
 import { createOpportunityData } from './opportunity-data-mapper.js';
 import { convertToOpportunity } from '../common/opportunity.js';
 import calculateKpiDeltasForAudit, { THRESHOLDS, METRICS, calculateConfidenceScore } from './kpi-metrics.js';
@@ -144,6 +144,60 @@ export function mergeCwvData(existingData, newDataItem) {
 }
 
 /**
+ * How long a CWV suggestion may sit IN_PROGRESS before a re-audit treats it as
+ * stale and reclaims it back to NEW. Mystique flips a suggestion to IN_PROGRESS
+ * the moment it hands a generated patch off for deploy; on an auto-deploy-off
+ * site nothing ever consumes that hand-off, so without this the suggestion is
+ * stuck forever. 24h is comfortably longer than the transient deploy window, so
+ * a genuinely in-flight IN_PROGRESS is never reclaimed mid-deploy.
+ */
+export const STALE_IN_PROGRESS_MS = 24 * 60 * 60 * 1000;
+
+/**
+ * FixEntity statuses that represent a real in-flight or live deploy. A suggestion
+ * covered by a fix in one of these states must never be reclaimed — the deploy
+ * hand-off is legitimately still owned downstream.
+ */
+const ACTIVE_FIX_STATUSES = new Set([
+  FixEntity.STATUSES.PENDING,
+  FixEntity.STATUSES.DEPLOYED,
+  FixEntity.STATUSES.PUBLISHED,
+]);
+
+/**
+ * Builds the CWV mergeStatusFunction for syncSuggestions. On re-audit it reclaims
+ * STALE stuck IN_PROGRESS suggestions back to NEW so they self-heal instead of
+ * accumulating every weekly run (Mystique sets IN_PROGRESS as the deploy hand-off;
+ * on auto-deploy-off sites nothing ever deploys it). A suggestion is reclaimed ONLY
+ * when ALL of the following hold:
+ *   - it is IN_PROGRESS,
+ *   - it has NO active fix entity (PENDING/DEPLOYED/PUBLISHED) — never touch a real
+ *     in-flight/live deploy, and
+ *   - it is stale (> STALE_IN_PROGRESS_MS since updatedAt) — never touch a freshly-set
+ *     IN_PROGRESS still inside the transient deploy window.
+ * Every other case delegates to defaultMergeStatusFunction, preserving its behaviour
+ * (incl. OUTDATED-regression and ERROR→NEW handling).
+ *
+ * @param {Set<string>} activeFixSuggestionIds - suggestion ids covered by an active fix.
+ * @param {boolean} fixFetchFailed - true if the fix-entity fetch failed; when true the
+ *   function behaves exactly like defaultMergeStatusFunction (fail safe: no reclaim).
+ * @returns {Function} mergeStatusFunction(existing, newDataItem, context)
+ */
+export function createMergeCwvStatus(activeFixSuggestionIds, fixFetchFailed) {
+  return (existing, newDataItem, context) => {
+    if (
+      !fixFetchFailed
+      && existing.getStatus() === Suggestion.STATUSES.IN_PROGRESS
+      && !activeFixSuggestionIds.has(existing.getId())
+      && Date.now() - new Date(existing.getUpdatedAt()).getTime() > STALE_IN_PROGRESS_MS
+    ) {
+      return Suggestion.STATUSES.NEW;
+    }
+    return defaultMergeStatusFunction(existing, newDataItem, context);
+  };
+}
+
+/**
  * Synchronizes opportunities and suggestions for a CWV audit
  * Creates or updates opportunity and syncs suggestions
  * @param {Object} context - Context object containing site, audit, finalUrl, log, dataAccess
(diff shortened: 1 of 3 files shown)


src/cwv/opportunity-sync.js
```diff
@@ -200,6 +254,29 @@ export async function syncOpportunitiesAndSuggestions(context) {
...
+  // stuck IN_PROGRESS suggestions that have no real in-flight/live deploy. On any
+  // fetch failure we fail safe — skip reclaim entirely this run (fixFetchFailed) —
+  // rather than risk reclaiming a suggestion whose deploy is genuinely in flight.
+  const activeFixSuggestionIds = new Set();
+  let fixFetchFailed = false;
+  try {
+    const fixEntities = await opportunity.getFixEntities();
+    (fixEntities || []).forEach((fixEntity) => {
+      if (ACTIVE_FIX_STATUSES.has(fixEntity.getStatus())) {
+        const suggestionId = fixEntity.getChangeDetails?.()?.suggestionId;
```
ramboz, line 276: **[LOW / robustness]** The active-fix guard keys only on the denormalized `changeDetails.suggestionId`. I verified this matches the CWV fix-entity contract the autofix-worker owns — `spacecat-autofix-worker/src/cwv/handler.js` builds `changeDetails = { suggestionId, issueId, issueTitle }` and `code-repo-manager.js` reads it back — so this is **correct today** and the guard is not a no-op.

It is cross-repo coupling, though: the *canonical* fix→suggestion link is the `FixEntitySuggestion` join (`fixEntity.getSuggestions()`), not this denormalized field. If a fix entity ever lands without `changeDetails.suggestionId` (e.g. a `reporting`-origin fix, or a future writer), this guard silently misses it and a suggestion with a genuinely live deploy could be reclaimed. The 24h staleness window + fail-safe bound the blast radius, so this is non-blocking — but consider either falling back to the join, or pinning the contract with a short comment here so the dependency is explicit.

*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

src/cwv/opportunity-sync.js
```diff
@@ -200,6 +254,29 @@ export async function syncOpportunitiesAndSuggestions(context) {
...
 
+  // Build the set of suggestion ids that are covered by an ACTIVE fix entity
+  // (PENDING/DEPLOYED/PUBLISHED). mergeCwvStatus uses this to reclaim ONLY the
+  // stuck IN_PROGRESS suggestions that have no real in-flight/live deploy. On any
+  // fetch failure we fail safe — skip reclaim entirely this run (fixFetchFailed) —
+  // rather than risk reclaiming a suggestion whose deploy is genuinely in flight.
+  const activeFixSuggestionIds = new Set();
+  let fixFetchFailed = false;
+  try {
+    const fixEntities = await opportunity.getFixEntities();
```
ramboz, line 265: **[LOW / efficiency]** `getFixEntities()` runs on every CWV re-audit, even though its result can only matter when the opportunity has at least one `IN_PROGRESS` suggestion. `existingSuggestions` aren't fetched here (`syncSuggestions` fetches them internally at the call below), so this is one extra query per run — negligible in isolation, but it fires for every site on every weekly audit regardless of whether anything is stuck. Optional: fetch `existingSuggestions` once here, short-circuit when none are `IN_PROGRESS`, and pass them through to `syncSuggestions` to avoid the double fetch. Non-blocking.

*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

ramboz (approved):
## PR Review: fix(cwv): reclaim stale stuck IN_PROGRESS suggestions on re-audit

### Summary
Adds a CWV-specific `mergeStatusFunction` (`createMergeCwvStatus`) to `syncOpportunitiesAndSuggestions` that, on re-audit, reclaims a matched suggestion `IN_PROGRESS → NEW` **only when** it is `IN_PROGRESS`, has **no active fix entity** (`PENDING`/`DEPLOYED`/`PUBLISHED`), and is **stale** (>24h since `updatedAt`). Every other case delegates to `defaultMergeStatusFunction` unchanged. Scope matches the description exactly (1 commit, +367/−2: source, ADR-003, tests). Overall risk: **low** — narrow, fail-safe, reversible, well-tested.

### Change Surface
- New: `STALE_IN_PROGRESS_MS` (export), `ACTIVE_FIX_STATUSES` (private), `createMergeCwvStatus` (export); new imports `FixEntity`, `defaultMergeStatusFunction`. Modified: `syncOpportunitiesAndSuggestions`.
- Surfaces touched: CWV opportunity/suggestion sync only. **Scope: match** — no creep.

### Blast Radius
- `syncOpportunitiesAndSuggestions` has a single caller — `src/cwv/handler.js:64` — so the behaviour change is contained to the CWV audit.
- Wiring verified against `syncSuggestions` (`src/utils/data-access.js:461`, invoked at `:578` with the matching `(existing, newDataItem, {...context, isTBYB})` signature); `defaultMergeStatusFunction` is exported (`:376`) and correctly delegated to.
- The merge fn only runs on **matched** suggestions (still present in the new audit data). Disappeared `IN_PROGRESS` suggestions are untouched (`IN_PROGRESS` is in the outdate skip-set) — consistent with the ADR's "matched suggestion" scope. No dangling refs, no new dead code.

### Contract verification (the load-bearing external dependency)
- The guard keys on `changeDetails.suggestionId`. I confirmed this **is** the canonical CWV fix-entity contract: `spacecat-autofix-worker/src/cwv/handler.js` builds `changeDetails = { suggestionId, issueId, issueTitle }` (denormalized for O(1) reverse lookup) and `code-repo-manager.js:401` reads it back — so the active-fix guard is **not** a no-op. (Note: the local `backlinks` handler links via the `suggestions[]` join and its `changeDetails` has no `suggestionId`; CWV has its own convention.) See the inline nit on line 268 re: the cross-repo coupling.
- `FixEntity.STATUSES` = PENDING / DEPLOYED / PUBLISHED / **FAILED** / **ROLLED_BACK**; the active set excludes FAILED/ROLLED_BACK, so a failed or rolled-back fix correctly does **not** shield its suggestion. ✓
- "Not re-dispatched after reclaim" holds: `src/cwv/auto-suggest.js:131` short-circuits dispatch when `data.isCodeChangeAvailable === true`, which reclaimed suggestions retain. ✓

### Correctness
- Fail-safe try/catch around `getFixEntities()` → `fixFetchFailed` disables reclaim for the run. ✓
- Invalid/missing `updatedAt` → `new Date(undefined).getTime()` is `NaN` → `Date.now() - NaN > STALE` is `false` → falls through to default (no reclaim, no throw). Safe by construction. ✓
- Reclaim returns `NEW`; `syncSuggestions` stamps `updatedBy='system'` and bumps `updatedAt`, so the next run sees `NEW`/fresh — no reclaim loop. ✓
- Minor: on `requiresValidation` sites the reclaim returns `NEW` directly (not `PENDING_VALIDATION`). Defensible — the suggestion already passed validation to reach `IN_PROGRESS` — but a one-line justification would help. Non-blocking.

### Test Gaps
- Strong coverage: 3 integration tests (reclaim / active-fix-preserved / FAILED-ignored / no-`suggestionId`-skip / fetch-failure / null-result) + 8 unit tests (24h constant, stale-reclaim, active-fix, fresh, fail-safe, NEW-delegate, both OUTDATED branches, ERROR→NEW). Assertions check concrete values, not just truthiness.
- Minor: no explicit test for the exact 24h boundary (`>` vs `>=`) or the invalid/missing `updatedAt` (NaN) path — the latter is covered branch-wise. Edge nits only.

### Affected-Test Run
- CI `ci / build` is **green** (runs the full `npm test` + the repo's enforced 100% line/branch/statement gate) and `codecov/patch` passed.
- I could **not** reproduce locally: this review worktree shares the main repo's `node_modules`, which has `@adobe/spacecat-shared-data-access@3.27.1`, while the branch requires `4.17.0` (no `Opportunity.SCOPE_TYPES`) — so pre-existing opportunity-creation tests error *before* the PR's code is reached. That is a local-environment artifact, not a PR issue; I verified the PR's logic by reading and rely on green CI for the execution signal.

### Security
- No secrets, no injection surface, no new deps, no installers/hooks/telemetry/endpoints. `log.warn` interpolates only `e.message` + `site.getId()`. Single `getFixEntities()` fetch (not N+1); the reclaim persists through the existing `saveMany` batch in `syncSuggestions` — aligns with the repo's N+1 guidance. **No findings.**

### Docs
- ADR-003 added, sequentially numbered (001/002/003), status Accepted, and it accurately describes the implementation, the safety argument, and the rejected alternatives.

### Verdict
**Approve.** No blockers. Well-scoped, fail-safe, reversible, thoroughly tested, and documented with an ADR. Two optional LOW nits inline (contract coupling on the denormalized `suggestionId`; unconditional fix-entity fetch), plus the minor `requiresValidation`/boundary-test notes above — none hold up the merge.

---
*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

src/cwv/opportunity-sync.js
tkotthakota-adobe: Pinned the contract in `622992d5` — added a comment at this line documenting that `changeDetails.suggestionId` is the denormalized CWV fix->suggestion link owned by spacecat-autofix-worker (cwv/handler.js writes it, code-repo-manager.js reads it), why we key on it over the FixEntitySuggestion join (O(1) reverse lookup), and that the 24h staleness window + fail-safe bound the blast radius if a future writer omits it. Went with the comment option over the join fallback since the case is fail-safe-bounded.

src/cwv/opportunity-sync.js
tkotthakota-adobe: Leaving this as-is. The version that actually saves a query needs `syncSuggestions` to accept pre-fetched `existingSuggestions` and return/expose them — but `syncSuggestions` is shared across callers (data-access.js), so changing its signature to plumb suggestions up-and-back is broader than this CWV fix should reach. A CWV-local short-circuit that fetches `getSuggestions()` here to gate `getFixEntities()` would just trade one query for another (and add a second when something IS in progress, since syncSuggestions still fetches its own). Given the fetch is once-per-run and fail-safe-wrapped, the cost is negligible — noting it as a fair follow-up if the sync boundary gets a pre-fetch param later. Non-blocking, as flagged.

PR #2868, opened 2026-08-10
fix(offsite-opportunities): Populate youtube, reddit and top cited buckets using only one domain-urls request

amitruluimadalina:
**What**
Populate the youtube.com, reddit.com, and third-party "cited" buckets from a single
domain-urls request per audit run, instead of the previous per-hostname × per-engine fan-out
(and a two-hop cited-domains discovery walk for the third-party bucket).

**Why**
[LLMO-6844](https://jira.corp.adobe.com/browse/LLMO-6844) made hostname optional on
domain-urls (returns URLs across every source host instead of one) and
[LLMO-6818](https://jira.corp.adobe.com/browse/LLMO-6818) added platform=all (aggregates
citations across every AI engine server-side). Together they make the old fan-out
unnecessary — one request now returns a single citations-sorted page spanning every host and
engine, and the loader classifies each row client-side into a bucket. Request volume drops
from ~12–78 requests to exactly 1 per audit run.

**How**
- src/utils/offsite-brand-presence-semrush.js: loadCitedUrlsFromSemrush() now issues one
domain-urls request (no hostname, platform=all); classifyRow() buckets each row into
youtube.com / reddit.com (enforcing YOUTUBE_URL_REGEX / REDDIT_URL_REGEX for parity
with the legacy classify) or third-party "cited" (domain: null, filtered for
contentType: "Owned", TOP_CITED_EXCLUDED_DOMAINS, and isExcludedCitedHost).
- PAGE_SIZE is now a fixed constant (1000), matching domain-urls's server-side pageSize
clamp — not an env override, since there's no fan-out cost to a large page when it's one
request either way.
- Removed as dead code: the per-engine platform list/env override, the cited-domains
discovery hop, and the per-candidate-domain walk.
- Docs: ADR docs/decisions/002-offsite-cited-urls-semrush-source.md and spec
docs/specs/2026-08-05-offsite-cited-urls-semrush-source.md updated to describe the
single-request design.
- Testing: test/utils/offsite-brand-presence-semrush.test.js rewritten — 100% statement/branch/
function/line coverage on the loader, lint clean.

- [ ] If data sources for any opportunity has been updated/added, please update the [wiki (→ wiki.corp.adobe.com)](https://wiki.corp.adobe.com/display/AEMSites/Data+Sources+for+Opportunities) for same opportunity.
If data sources for any opportunity has been updated/added, please update the [wiki](https://wiki.corp.adobe.com/display/AEMSites/Data+Sources+for+Opportunities) for same opportunity.

Related Issues
[LLMO-6709](https://jira.corp.adobe.com/browse/LLMO-6709) · depends on [LLMO-6844](https://jira.corp.adobe.com/browse/LLMO-6844) and [LLMO-6818](https://jira.corp.adobe.com/browse/LLMO-6818) (spacecat-api-service)

Thanks for contributing!

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Andrei Paraschiv", "Dominique Jäggi", "TB ADBE"]
```

### DIFF
--- a/src/utils/offsite-brand-presence-semrush.js
+++ b/src/utils/offsite-brand-presence-semrush.js
@@ -95,9 +95,11 @@ export function buildDomainUrlsUrl({
 /**
  * Fetches the single `domain-urls` page (all hosts, all platforms).
  *
- * @returns {Promise<{ rows: object[], ok: boolean, authFailure: boolean }>} `ok` is false
- *   on network error / timeout / non-2xx / unparseable body. `authFailure` distinguishes a
- *   401/403 (auth issue) from other failures.
+ * @returns {Promise<{ rows: object[], ok: boolean, authFailure: boolean, truncated: boolean }>}
+ *   `ok` is false on network error / timeout / non-2xx / unparseable body. `authFailure`
+ *   distinguishes a 401/403 (auth issue) from other failures. `truncated` is true when a
+ *   full page came back (LLMO-6711 shadow-run parity signal — a starved run is visible on
+ *   `diagnostics` without grepping logs).
  */
 async function fetchDomainUrls(url, headers, log, siteId, pageSize) {
   const ctx = { siteId };
@@ -109,7 +111,9 @@ async function fetchDomainUrls(url, headers, log, siteId, pageSize) {
     log.error(`${LOG_PREFIX} Fetch failed for domain-urls: ${error.message}`, {
       ...ctx, error: error.message, durationMs: Date.now() - startedAt,
     });
-    return { rows: [], ok: false, authFailure: false };
+    return {
+      rows: [], ok: false, authFailure: false, truncated: false,
+    };
   }
 
   if (!response.ok) {
@@ -123,7 +127,9 @@ async function fetchDomainUrls(url, headers, log, siteId, pageSize) {
         ...ctx, status: response.status, durationMs: Date.now() - startedAt,
       });
     }
-    return { rows: [], ok: false, authFailure };
+    return {
+      rows: [], ok: false, authFailure, truncated: false,
+    };
   }
 
   let body;
@@ -133,21 +139,32 @@ async function fetchDomainUrls(url, headers, log, siteId, pageSize) {
     log.error(`${LOG_PREFIX} Could not parse domain-urls response: ${error.message}`, {
       ...ctx, error: error.message, durationMs: Date.now() - startedAt,
     });
-    return { rows: [], ok: false, authFailure: false };
+    return {
+      rows: [], ok: false, authFailure: false, truncated: false,
+    };
   }
 
   const raw = Array.isArray(body?.urls) ? body.urls : [];
-  if (raw.length >= pageSize) {
+  const truncated = raw.length >= pageSize;
+  if (truncated) {
     log.warn(`${LOG_PREFIX} domain-urls returned a full page (${raw.length} >= ${pageSize}); response may be truncated`, {
       ...ctx, rowCount: raw.length, pageSize,
     });
   }
-  return { rows: raw.slice(0, pageSize), ok: true, authFailure: false };
+  return {
+    rows: raw.slice(0, pageSize), ok: true, authFailure: false, truncated,
+  };
 }
 
 /**
  * Classifies one `domain-urls` row into a bucket, or drops it.
  *
+ * - Non-`http(s)` schemes (`mailto:`, `tel:`, `data:`, `javascript:`, `ftp:`, `ws:`, ...) are
+ *   dropped upfront via an explicit allowlist on the raw `row.url`, rather than relying on
+ *   `classifyAndNormalize` happening to produce an unparseable value for some of them
+ *   (opaque schemes serialize `origin` to the literal string `"null"`) — that's incidental
+ *   for opaque schemes and doesn't catch a scheme that reparses cleanly (`ftp:`, `ws:`) but
+ *   is never a real citation source.
  * - `youtube.com` / `reddit.com` — matched via `classifyAndNormalize`, then the strict
  *   format regexes (drops non-thread Reddit URLs and lookalike YouTube hosts).
  * - Everything else is the third-party "cited" bucket (`domain: null`), unless it's
@@ -157,7 +174,20 @@ async function fetchDomainUrls(url, headers, log, siteId, pageSize) {
  * @returns {{url: string, domain: string|null}|null} `null` when the row is dropped.
  */
 function classifyRow(row, siteHostname, brandTokens) {
-  const classified = row?.url ? classifyAndNormalize(row.url, siteHostname) : null;
+  if (!row?.url) {
+    return null;
+  }
+  let scheme;
+  try {
+    scheme = new URL(row.url).protocol;
+  } catch {
+    return null;
+  }
+  if (scheme !== 'http:' && scheme !== 'https:') {
+    return null;
+  }
+
+  const classified = classifyAndNormalize(row.url, siteHostname);
   if (!classified) {
     return null;
   }
@@ -167,17 +197,11 @@ function classifyRow(row, siteHostname, brandTokens) {
   if (classified.domain === 'reddit.com') {
     return REDDIT_URL_REGEX.test(row.url) ? { url: classified.url, domain: 'reddit.com' } : null;
   }
-  if (row?.contentType === 'Owned') {
+  if (row.contentType === 'Owned') {
     return null;
   }
-  let host;
-  try {
-    host = new URL(classified.url).hostname.toLowerCase().replace(/^www\./, '');
-  } catch {
-    // classified.url already passed through classifyAndNormalize's own URL parse, so this
-    // should be unreachable — defensive only, in case that contract ever changes.
-    return null;
-  }
+  // The allowlist above guarantees classified.url is always a reparseable http(s) URL here.
+  const host = new URL(classified.url).hostname.toLowerCase().replace(/^www\./, '');
   if (TOP_CITED_EXCLUDED_DOMAINS.some((d) => host === d || host.endsWith(`.${d}`))) {
     return null;
   }
@@ -213,8 +237,9 @@ function classifyRow(row, siteHostname, brandTokens) {
  * @param {object} [params.diagnostics] - Optional out-param, mutated in place. On a null
  *   return, set to `{ fallbackReason }` with a specific code (`no-organization-id`,
  *   `no-active-brand`, `brand-resolution-failed`, `no-date-window`, `ims-token-failed`,
- *   `domain-urls-auth-failed`, or `domain-urls-failed`). Left unset on a successful return —
- *   with a single request there's no per-engine degradation to report.
+ *   `domain-urls-auth-failed`, or `domain-urls-failed`). On a successful return, set to
+ *   `{ truncated }` — true when the response came back at `PAGE_SIZE`, so a bucket may be
+ *   starved (LLMO-6711 shadow-run parity signal).
  * @returns {Promise<Map<string, {count:number, domain:string|null}> | null>}
  */
 export async function loadCitedUrlsFromSemrush({
@@ -331,6 +356,7 @@ export async function loadCitedUrlsFromSemrush({
     setDiagnostics({ fallbackReason: result.authFailure ? 'domain-urls-auth-failed' : 'domain-urls-failed' });
     return null;
   }
+  setDiagnostics({ truncated: result.truncated });
 
   const brandKeywords = site.getConfig?.()?.getBrandKeywords?.() || [];
   const brandTokens = computeBrandTokens(siteHostname, brandKeywords);
--- a/src/utils/offsite-brand-presence-semrush.js
+++ b/src/utils/offsite-brand-presence-semrush.js
@@ -15,11 +15,11 @@ import { tracingFetch as fetch } from '@adobe/spacecat-shared-utils';
 import { resolveBrandResultForSite } from './brand-resolver.js';
 import { getDateWindowForPreviousWeeks } from './offsite-brand-presence-postgrest.js';
 import { classifyAndNormalize } from './offsite-brand-presence-enrichment.js';
+import { computeBrandTokens, isExcludedCitedHost } from './offsite-audit-utils.js';
 import {
-  OFFSITE_DOMAINS,
+  TOP_CITED_EXCLUDED_DOMAINS,
   YOUTUBE_URL_REGEX,
   REDDIT_URL_REGEX,
-  SEMRUSH_PLATFORM_BY_PROVIDER,
 } from '../offsite-brand-presence/constants.js';
 
 const LOG_PREFIX = '[offsite-brand-presence][semrush]';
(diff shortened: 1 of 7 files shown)


MysticatBot (changes requested):
Hey @amitruluimadalina,

**Verdict: Request changes** - one stale JSDoc block references removed diagnostics fields that will mislead callers.
**Complexity: HIGH** - large diff (1344 lines across source + tests + docs).
_Changes: Replaces the per-hostname x per-engine fan-out (~12-78 requests) with a single `domain-urls` request per audit run, classifying rows client-side into youtube/reddit/cited buckets (4 files)._
Note: Recommend a human read before merge - this change amends an architectural decision record (`docs/decisions/002`). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] Stale JSDoc on `diagnostics` out-param references removed fields - `src/utils/offsite-brand-presence-semrush.js:206` (details inline)

<details>

- nit: ADR and spec status remain "Proposed" after implementation is complete - consider updating to "Accepted" once the auth path is validated and the flag is enabled - `docs/decisions/002-offsite-cited-urls-semrush-source.md:3`
- nit: Spec Related field still references only "PR #2847" but not this PR (2868) - `docs/specs/2026-08-05-offsite-cited-urls-semrush-source.md:7`
- suggestion: Consider a defensive `try/catch` around `new URL(classified.url)` in `classifyRow` (cited-bucket path) as belt-and-suspenders against future `classifyAndNormalize` refactors - `src/utils/offsite-brand-presence-semrush.js:173`
- suggestion: Log the dropped-row count at the end of the classification loop (e.g. `droppedCount: result.rows.length - allUrls.size`) to make it easier to diagnose classification issues during LLMO-6711 shadow-run parity validation - `src/utils/offsite-brand-presence-semrush.js:345`
</details>


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 58s | <b>Cost:</b> $8.23 | <b>Commit:</b> <code>32ae5fd5cb4f4a3adf245834e67ac1d284cbc8bc</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -110,111 +74,131 @@ async function getAuthorizationHeader(context) {
...
+  if (classified.domain === 'youtube.com') {
+    return YOUTUBE_URL_REGEX.test(row.url) ? { url: classified.url, domain: 'youtube.com' } : null;
+  }
+  if (classified.domain === 'reddit.com') {
+    return REDDIT_URL_REGEX.test(row.url) ? { url: classified.url, domain: 'reddit.com' } : null;
+  }
+  if (row?.contentType === 'Owned') {
+    return null;
+  }
+  let host;
```
tb-adbe: `new URL(classified.url)` here throws uncaught on opaque-scheme values (`mailto:`, `tel:`, `data:`, `javascript:`, etc.). `classifyAndNormalize`'s `domain: null` branch builds `${parsed.origin}${parsed.pathname}`, and `origin` serializes to the literal string "null" for these schemes, producing a non-reparseable string that throws here - verified directly against the current implementation. Neither this loop nor the handler's call site catches it, so a single non-HTTP row in the Semrush response turns the loader's designed `return null` -> legacy fallback into an unhandled exception that fails the audit instead of falling back. A prior automated review flagged this line as a hypothetical "belt-and-suspenders" suggestion; it is not hypothetical. Wrap this in a try/catch and drop the row (`return null`) on parse failure.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -333,144 +314,63 @@ export async function loadCitedUrlsFromSemrush({
...
-    }
-  }
-  for (const [hostname, entry] of byHost) {
-    // One summary line per hostname (not per-request) — enough to see partial engine
-    // degradation in Splunk without a log line per one of the ~12 concurrent requests.
-    log.info(`${LOG_PREFIX} ${hostname}: ${entry.okCount}/${entry.okCount + entry.failCount} engine requests succeeded`, {
-      siteId, hostname, okCount: entry.okCount, failCount: entry.failCount,
+  const result = await fetchDomainUrls(url, headers, log, siteId, PAGE_SIZE);
+  if (!result.ok) {
+    log.warn(`${LOG_PREFIX} domain-urls request failed; using legacy fallback`, {
```
tb-adbe, line 380: The `diagnostics` out-param contract changed here (only `fallbackReason` is set now; the loader no longer sets `engineFailureCount`/`degradedHosts`/`authFailureDetected` on success), but two things weren't updated to match: the JSDoc on this function still documents those removed fields plus a `surface-failed:<host>` reason that no longer exists (already flagged by a prior automated review, still unaddressed), and `handler.js` (same repo/branch, not touched by this diff) still reads all three fields - the reads are null-guarded so nothing crashes, but the auth-rejection WARN and the `semrushEngineFailureCount`/`semrushDegradedHosts` audit-result fields are now permanently dead code, and the handler test that stubs `engineFailureCount: 1` still passes even though it validates a state the real loader can never produce. Please correct the JSDoc and either remove the dead handler branches (+ their test) or make an explicit call to keep them as a no-op contract.

tb-adbe (commented):
Hey @amitruluimadalina,

## Strengths

The core redesign is a real simplification. Collapsing the per-hostname x per-engine fan-out (up to 78 requests) plus the separate `cited-domains` discovery hop into one `domain-urls` request (`src/utils/offsite-brand-presence-semrush.js`) removes a lot of accidental complexity: the `byHost` grouping, `getSemrushPlatforms`/`OFFSITE_SEMRUSH_PLATFORMS`, and the per-engine degradation bookkeeping are all gone with zero dangling references elsewhere in the repo. `fetchDomainUrls` has clean, consistent error handling (fetch throw, non-2xx with a distinct 401/403 branch, unparseable body all collapse to a uniform `{ rows, ok, authFailure }`), and path segments stay `encodeURIComponent`-wrapped so there's no injection risk in the request URL. The rewritten test suite is strong on the new request-level fallback surface (network error, 500, 401, 403, bad JSON each assert the right `fallbackReason`) and on the clamping/dedup logic. The ADR is unusually candid about the design's own tradeoffs, which made this review easier.

## Issues

### Important (Should Fix)

**1. `classifyRow`'s `new URL(classified.url)` call throws uncaught on non-HTTP-scheme URLs, which defeats the loader's own fallback safety net.** `src/utils/offsite-brand-presence-semrush.js:173`

`classified.url` comes from `classifyAndNormalize`, whose `domain: null` branch builds `${parsed.origin}${parsed.pathname}`. For opaque-scheme values in the Semrush response (`mailto:`, `tel:`, `data:`, `javascript:`, etc.), `origin` serializes to the literal string `"null"`, producing garbage like `"nullfoo@bar.com"` that throws when re-parsed by `new URL()` at line 173. This was verified directly against the current `classifyAndNormalize` implementation for `mailto:`, `tel:`, `data:`, and `javascript:` inputs - all reach this line and all throw. Neither the classification loop nor the handler's call site wraps this in a try/catch, so a single non-HTTP row in the Semrush response turns the loader's designed `return null` -> graceful legacy fallback into an unhandled exception that fails the audit outright instead of falling back - the exact "worse than legacy-only" outcome the module's own comments warn against elsewhere. The prior automated review flagged this same line as a hypothetical "belt-and-suspenders" suggestion; it isn't hypothetical, the current `classifyAndNormalize` already produces non-reparseable output for opaque schemes, so this stands as a real, unaddressed bug. Fix: wrap the host derivation in a try/catch and drop the row (return `null`) on parse failure, and add a test that feeds an opaque-scheme row and asserts the loader still returns a Map rather than throwing.

**2. The `diagnostics` contract was changed but only half-migrated, leaving a stale JSDoc and dead branches in `handler.js` guarded by a now-vacuous test.** `src/utils/offsite-brand-presence-semrush.js:327` (and `src/offsite-brand-presence/handler.js`, unchanged by this PR)

The loader no longer sets `engineFailureCount` / `degradedHosts` / `authFailureDetected` on a successful return, only `fallbackReason` on failure. Two things weren't updated to match: the JSDoc on `loadCitedUrlsFromSemrush`'s `diagnostics` param still documents those three fields plus a `surface-failed:<host>` reason that no longer exists (this is the "Important" item the prior automated review already flagged and remains unaddressed). More significantly, `handler.js` (same repo, same branch, just not touched by this diff) still reads all three fields to decide whether to log an auth-rejection warning and whether to emit `semrushEngineFailureCount`/`semrushDegradedHosts` on the audit result. Those reads are null-guarded so nothing crashes, but the branches are now permanently dead: the WARN can never fire and the two audit-result fields can never be populated again. Worse, the existing handler test that stubs a diagnostics object with `engineFailureCount: 1` and asserts the handler surfaces it still passes, so it now validates a state the real loader can never produce, masking the regression. Since a single-request design has no "succeeded but one engine failed" state, removing this is defensible. Fix in this PR: correct the JSDoc, and either delete the dead `handler.js` branches (and their now-misleading test) or make an explicit decision to keep them as an intentional no-op contract.

### Minor (Nice to Have)

**3. `SEMRUSH_PLATFORM_BY_PROVIDER` in `src/offsite-brand-presence/constants.js` is now dead code with a factually wrong comment.** This file isn't touched by this PR, but this PR removed its only consumer (the old per-engine query logic). Its comment still says it's "used by the Semrush URL-Inspector loader to query each engine explicitly... it does NOT aggregate" - the opposite of the new `platform=all` design. The comment itself ties this map to LLMO-6710, so it may be intentionally retained for that follow-up; either delete it now or correct the comment so LLMO-6710 doesn't inherit stale guidance.

**4. The cited (third-party) bucket's count and dedup logic is untested.** `test/utils/offsite-brand-presence-semrush.test.js` asserts youtube.com and reddit.com bucket counts at 1 and duplicate-URL summing for youtube.com, but the cited bucket is only ever asserted at a count of 0 - a mis-keyed cited counter or broken dedup on that path would pass today. Add a case with a positive cited count and one with duplicate cited URLs.

## Recommendations

- Once the try/catch guard from finding 1 lands, consider an explicit `http`/`https` scheme allowlist on the cited bucket as defense-in-depth, rather than relying on the parser's throw to incidentally reject non-web targets before they reach the downstream scrape.
- The single global-citations-sorted page (`PAGE_SIZE = 1000`) feeds three buckets that `selectTopUrls` sizes independently (youtube, reddit, and top-cited each get their own top-70). A site where third-party press citations dominate could starve the reddit or youtube bucket entirely within one page, with no pagination available to compensate. The ADR already discloses this as an accepted tradeoff, but its alternatives section never weighs the middle ground: three requests (youtube with `hostname`, reddit with `hostname`, and one hostname-less third-party request, all `platform=all`) would preserve independent per-bucket sizing while still cutting request volume roughly 25x. Worth adding to the ADR's alternatives with an explicit rationale either way before the flag is considered for enablement.
- There are now three separate implementations of "classify and filter an offsite URL": the enrichment module's `classifyAndNormalize`, `handler.js`'s own stricter variant, and this PR's `classifyRow`. They already show small behavioral drift (for example, `handler.js` tags `wikipedia.org` rows and excludes them downstream, while `classifyRow` drops them outright). The ADR itself flags this duplication as debt to resolve "before the 3rd caller" - this PR is that third caller. Consolidating into one shared classify function before the Semrush flag is enabled would make future parity work (LLMO-6711) structural rather than something re-verified by hand on every regex or exclusion-list edit.
- Consider surfacing a full-page truncation as a structured `diagnostics` signal (not just a log warning), so shadow-run parity validation (LLMO-6711) can detect a starved run from the audit result itself.

## Assessment

Ready to merge? With fixes.
Reasoning: the redesign's direction is sound and the flag stays off by default, so nothing here is unsafe to merge as-is, but finding 1 is a verified bug that bypasses the PR's own fallback guarantee, and finding 2 leaves a stale contract with a test that no longer tests anything real. Both are small, contained fixes.

## Next Steps

1. Fix the unguarded `new URL()` call in `classifyRow` (finding 1) and add a regression test for it.
2. Resolve the orphaned diagnostics contract in `handler.js` and its test, and correct the stale JSDoc (finding 2).
3. Minor items and recommendations are optional but worth a look before the Semrush flag is ever enabled.

MysticatBot (approved):
Hey @amitruluimadalina,

**Verdict: Approve** - all prior blocking findings are resolved; only minor nits remain.
**Complexity: HIGH** - large diff (1559 lines across source + tests + docs).
_Changes: Replaces the per-hostname x per-engine fan-out (~12-78 requests) with a single `domain-urls` request per audit run, classifying rows client-side into youtube/reddit/cited buckets (7 files)._
Note: Recommend a human read before merge - this change amends an architectural decision record (`docs/decisions/002`). The bot review is a complement to, not a replacement for, a human read here.

<details>

- nit: ADR and spec status still "Proposed" despite implementation being complete - update to "Accepted" or "Implemented" per your lifecycle convention - `docs/decisions/002-offsite-cited-urls-semrush-source.md:7`
- suggestion: `droppedCount` metric (`result.rows.length - allUrls.size`) conflates filtered rows with deduplicated/merged rows, making Splunk investigation harder - consider splitting into `filteredCount` + `mergedCount` for actionability - `src/utils/offsite-brand-presence-semrush.js:345`
</details>

- Stale JSDoc on `diagnostics` out-param referencing removed fields - updated to document `fallbackReason` and `truncated`
- `new URL(classified.url)` throwing on opaque schemes (mailto/tel/data/javascript) - fixed via explicit http(s) scheme allowlist in `classifyRow`
- Dead diagnostics branches in handler.js (`authFailureDetected`, `degradedHosts`, `engineFailureCount`) - removed
- Spec Related field not referencing this PR - updated to include PR 2868
- Dropped-row count not logged - now emitted as structured `droppedCount` field
- `SEMRUSH_PLATFORM_BY_PROVIDER` comment was stale - corrected to "Currently unused, kept for LLMO-6710"
- Cited bucket count and dedup logic untested - dedicated tests added

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 3s | <b>Cost:</b> $5.69 | <b>Commit:</b> <code>cc7c49b4aa75d2998ccc7a92ff08b1d6a7bb3346</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

tb-adbe (approved):
Hey @amitruluimadalina,

## Strengths

Previously flagged issues now addressed, and thoroughly:

- **Uncaught throw on non-HTTP-scheme URLs** (`classifyRow`) - fixed with an explicit scheme allowlist checked upfront (`new URL(row.url).protocol`, wrapped in try/catch, rejecting anything but `http:`/`https:`), which is stronger than a minimal patch: it also catches schemes like `ftp:`/`ws:` that reparse cleanly and wouldn't have thrown, and it makes the later `new URL(classified.url)` call provably safe rather than incidentally safe. Backed by three new targeted tests (opaque schemes, non-throwing non-http schemes, and a plain invalid URL string).
- **Orphaned diagnostics contract** - `handler.js`'s dead `authFailureDetected`/`degradedHosts`/`engineFailureCount` branches and the two tests that were validating an unreachable state are both removed cleanly. The JSDoc on `loadCitedUrlsFromSemrush` now accurately describes the current `fallbackReason` codes. Verified no dangling references to the removed fields remain anywhere in the repo.
- **Cited-bucket test coverage gap** - closed with a duplicate-summing test and a cited-only-bucket count assertion that checks the `onProgress` message text, not just the Map size.
- **`SEMRUSH_PLATFORM_BY_PROVIDER` stale comment** - corrected to state it's currently unused and kept for LLMO-6710.
- Beyond the required fixes, the bucket-starvation tradeoff from the earlier review is now explicitly recorded in the ADR's alternatives section (the three-request-per-bucket option, with a stated rationale for rejecting it and a note to revisit if LLMO-6711 shows real starvation) - good-faith engagement with that recommendation even though the simpler single-request design was kept.
- A new `truncated` diagnostics signal was added on top of what was asked, addressing the earlier suggestion to make a full-page truncation visible on the result rather than only in logs.

## Recommendations

- The new `diagnostics.truncated` flag is set by the loader but `handler.js` never reads it, so it isn't yet persisted onto `auditResult` the way `fallbackReason` is. As it stands it's only usable by a caller that inspects the `diagnostics` out-param directly. Worth wiring it into `auditResult` alongside `dataSource`/`fallbackReason` once there's a concrete need for it in the LLMO-6711 shadow-run parity work, so it doesn't sit unused.

## Assessment

Ready to merge? Yes.
Reasoning: both prior Important findings are fixed with solid, targeted test coverage, and no new issues surfaced in this round.

AndreiAlexandruParaschiv (approved):

PR #2870, opened 2026-08-10
fix: resolve the organization's own brandalf flag row, not a brand's override

### LINKED ISSUE #2869: isBrandalfEnabled depends on the feature_flags unique key having no brand_id
## Context

`feature_flags` in mysticat-data-service is becoming **optionally brand-aware**. A nullable `brand_id` joins the unique key, and a row naming a brand *overrides* the organization's value for that brand alone:

```
resolve(org, brand) := row (org, product, flag_name, brand_id = brand)
                       else row (org, product, flag_name, brand_id IS NULL)
                       else the flag does not exist for that brand
```

Migration: https://github.com/adobe/mysticat-data-service/pull/891
Design: https://github.com/adobe/serenity-docs/issues/317

This exists so an organization with many brands migrates onto the Semrush-backed read path in waves rather than one cutover.

## What this repo does

`src/utils/brandalf-utils.js` reads the table **directly**, not through the api-service feature-flags endpoint:

```js
const { data, error } = await postgrestClient
  .from('feature_flags')
  .select('flag_value')
  .eq('organization_id', organizationId)
  .eq('product', FEATURE_FLAG_PRODUCT)
  .eq('flag_name', BRANDALF_FLAG_NAME)
  .maybeSingle();
```

Callers: `src/utils/offsite-brand-presence-enrichment.js` and `src/llmo-customer-analysis/handler.js`.

## Why this is not urgent, and why it is still worth fixing

`.maybeSingle()` raises when more than one row matches. After the migration, more than one row matches as soon as an organization holds a **per-brand `brandalf` override**.

Nothing writes one today. The Serenity migration CLI writes per-brand rows only for `serenity`; `brandalf` is written at organization scope by `spacecat-api-service` alone. So this code is correct on the day the migration lands — **by convention, not by construction**. Nothing in the schema prevents a `brandalf` override, and the whole point of the change is that per-brand rows become an ordinary thing to write.

The failure mode is also quiet rather than loud. On error the helper logs a warning and returns `null`:

```js
if (error) {
  log?.warn(`Failed to read brandalf flag for org ${organizationId}: ${error.message}`);
  return null;
}
```

`null` means "unknown", so the first per-brand `brandalf` row degrades brandalf resolution for that organization into an unknown that both call sites then treat as not-enabled — with nothing but a warning line to explain it. That is the kind of thing that gets diagnosed twice.

## Suggested fix

Resolve the organization's own row explicitly, using the pattern adopted in `spacecat-api-service` (https://github.com/adobe/spacecat-api-service/pull/3021). It is correct against **both** the old and the new schema, so it can ship before, during, or after the migration with no coordination:

```js
// The organization's own row, not a brand's override of it. `brand_id` is absent
// from every row before the brand-scope migration and NULL on the organization's
// row after it, so this selects correctly under both schemas.
const isOrgRow = (row) => (row.brand_id ?? null) === null;

const { data, error } = await postgrestClient
  .from('feature_flags')
  .select('*')              // wildcard is required — see below
  .eq('organization_id', organizationId)
  .eq('product', FEATURE_FLAG_PRODUCT)
  .eq('flag_name', BRANDALF_FLAG_NAME);

const row = (data ?? []).find(isOrgRow);
return row?.flag_value === true;
```

Two traps worth stating in a comment at the call site, because both look like tidy-ups:

- **Do not name `brand_id` in the projection.** Against the current schema the column does not exist and PostgREST returns an error, so every read fails.
- **Do not narrow to `.select('flag_value')` either.** Post-migration `row.brand_id` is then `undefined`, `?? null` makes every override look like the organization's row, and the predicate silently stops working.
- **Do not use a `brand_id IS NULL` filter instead.** Same problem as naming it in a projection — it cannot be evaluated before the column exists.

The wildcard is the only shape correct under both schemas.

## Scope

One helper, plus a test that a brand-scoped row alongside the organization's row still resolves the organization's value. No behaviour change today: with no override present, `.find(isOrgRow)` returns exactly what `.maybeSingle()` returned.

## Ordering

Independent of the migration in both directions — it does not block https://github.com/adobe/mysticat-data-service/pull/891 and is not blocked by it.

rainer-friederich:
## 1. Abstract

`isBrandalfEnabled` resolves the organization's own `feature_flags` row explicitly instead of assuming its query matches at most one row. One helper in `src/utils/brandalf-utils.js`, plus test coverage for a brand-scoped row sitting alongside the organization's.

Closes https://github.com/adobe/spacecat-audit-worker/issues/2869

## 2. Reasoning

`feature_flags` in mysticat-data-service is becoming optionally brand-aware. A nullable `brand_id` joins the unique key, and a row naming a brand overrides the organization's value for that brand alone: resolution for a given organization and brand takes the brand's own row if one exists, else the organization's row, else the flag does not exist for that brand.

This repo reads that table directly rather than through the api-service feature-flags endpoint, and the read terminated in `.maybeSingle()` — which raises as soon as more than one row matches. After the migration, more than one row matches the moment an organization holds a per-brand `brandalf` override.

Nothing writes one today: the Serenity migration CLI writes per-brand rows only for `serenity`, and `brandalf` is written at organization scope by spacecat-api-service alone. So this code is correct on the day the migration lands — by convention, not by construction. Nothing in the schema prevents a `brandalf` override, and the whole point of the migration is that per-brand rows become an ordinary thing to write.

The failure mode is quiet rather than loud. On a query error the helper logs a warning and returns `null`. `null` means "unknown", and both call sites treat unknown as not-enabled — so the first per-brand `brandalf` row would silently degrade brandalf resolution for that organization, with a single warning line to explain it. That is the kind of thing that gets diagnosed twice.

Before: the read selected the `flag_value` column and terminated in `.maybeSingle()`, so it depended on the `(organization_id, product, flag_name)` tuple matching at most one row.

After: the read selects every matching row and picks the organization's own — the row whose `brand_id` is NULL, or absent entirely on the pre-migration schema. A brand's override of the same flag is ignored.

Behaviour delta today: none. With no override row present, the new selection returns exactly what `.maybeSingle()` returned — enabled, disabled, and absent-row all resolve as before, and an error still resolves to `null`. What changes is the behaviour on the day an override exists: the organization's value continues to resolve instead of collapsing into an unknown.

The change is correct against both the old and the new schema, so it can ship before, during, or after the migration with no coordination.

One thing worth flagging for review, because it looks like a tidy-up: **the wildcard projection is load-bearing.** Naming `brand_id` in the projection, or filtering on it, fails against the current schema where the column does not exist yet — every read would error. Narrowing the projection back to `flag_value` once the column does exist makes `brand_id` arrive as `undefined` on every row, which makes every brand override read as the organization's own and silently disables the predicate. The wildcard is the only shape correct under both schemas. The diff carries comments saying so at both the predicate and the call site.

## 4. Required information

- Jira / issue: https://github.com/adobe/spacecat-audit-worker/issues/2869
- Spec: https://github.com/adobe/serenity-docs/issues/317
- Other: pattern already merged in spacecat-api-service — https://github.com/adobe/spacecat-api-service/pull/3021 — same `isOrgRow` predicate and the same wildcard-projection requirement

- **mysticat-data-service** — contract. Owns the `feature_flags` table and the brand-scope migration this read must survive. Not changed here.
- **spacecat-api-service** — contract. Sole writer of `brandalf` at organization scope, and the repo the resolution pattern is borrowed from. Not changed here.

The changed read was exercised against the real `@supabase/postgrest-js` client rather than only against test doubles, by pointing it at a throwaway local HTTP server that served `feature_flags` rows and recorded the request line the builder emitted.

Scenarios driven, covering both schema shapes:

- Pre-migration rows carrying no `brand_id` at all — flag on, flag off, and no row.
- Post-migration organization row alone.
- Post-migration organization row plus a brand override, in both orders and with the two rows disagreeing in both directions — the organization's value resolved in every case.
- A brand override with no organization row — resolved as not-enabled.

The emitted query never names or filters `brand_id` under either schema, confirming it is valid against the table as it exists today.

The same harness confirmed the premise of the change: with two matching rows the previous `.maybeSingle()` call fails with `JSON object requested, multiple (or no) rows returned`, which the helper converts into exactly the quiet `null` this PR removes.

A review finding raised during development was declined: it claimed the pre-existing optional-chaining on the logger and site accessors in this file would trip the repo's branch-coverage gate and needed either new tests or `c8` ignore comments. It does not reproduce, and those lines are untouched by this change.

## 7. Test plan

Locally, beyond the automated tests, the helper was driven against a real PostgREST client as described in section 6 — that is the meaningful end-to-end check here, since the change is a query-shape change and the behaviour that matters is what the client emits and how it handles a multi-row response.

Both call sites (`src/utils/offsite-brand-presence-enrichment.js` and `src/llmo-customer-analysis/handler.js`) stub this helper, so their behaviour is unchanged by construction; their suites were run to confirm that.

To verify on **dev** after deploy:

- Pick an organization with `brandalf` enabled and run an offsite brand-presence audit or an llmo-customer-analysis run for one of its sites. Brand resolution should proceed exactly as before, with no `Failed to read brandalf flag` warning in the logs.
- Confirm the negative case too: an organization without the flag still skips brand resolution.

Once the brand-scope migration has landed on an environment, the case this PR exists for can be verified directly: insert a per-brand `brandalf` row alongside the organization's row for a test organization, then run the audit and confirm the organization's value still resolves and no warning is logged. On the pre-migration schema there is no way to construct that state, which is why it is covered by tests rather than by an environment check.

- Related, with no ordering constraint: https://github.com/adobe/mysticat-data-service/pull/891 — the brand-scope migration this read is being hardened against.

This PR is independent of that migration in both directions. It does not block it and is not blocked by it, because the new read is correct under both the pre- and post-migration schema. Either can merge and deploy first.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/utils/brandalf-utils.js
+++ b/src/utils/brandalf-utils.js
@@ -14,9 +14,27 @@ const BRANDALF_FLAG_NAME = 'brandalf';
 const FEATURE_FLAG_PRODUCT = 'LLMO';
 
 /**
+ * The organization's own row, not a brand's override of it. `brand_id` is absent
+ * from every row before the brand-scope migration and NULL on the organization's
+ * row after it, so this selects correctly under both schemas.
+ *
+ * The query feeding this predicate must keep selecting the full row. Naming
+ * `brand_id` in a projection — or filtering on it — fails against the current
+ * schema, where the column does not exist yet; narrowing the projection once it
+ * does exist makes every override row arrive with `brand_id: undefined` and read
+ * as the organization's own.
+ *
+ * @param {object} row - Raw PostgREST `feature_flags` row.
+ * @returns {boolean} `true` for the organization-level row.
+ */
+const isOrgRow = (row) => (row.brand_id ?? null) === null;
+
+/**
  * Checks whether the brandalf feature flag is enabled for an organization by
  * reading the `feature_flags` table directly via the mysticat PostgREST client
  * (available at `context.dataAccess.services.postgrestClient`).
+ *
+ * Resolves the organization's own value, ignoring any brand's override of it.
  *
  * @param {string} organizationId - SpaceCat org UUID
  * @param {object} postgrestClient - mysticat PostgREST client (dataAccess.services.postgrestClient)
@@ -33,13 +51,13 @@ export async function isBrandalfEnabled(organizationId, postgrestClient, log) {
   }
 
   try {
+    // Wildcard projection is required — see `isOrgRow`.
     const { data, error } = await postgrestClient
       .from('feature_flags')
-      .select('flag_value')
+      .select('*')
       .eq('organization_id', organizationId)
       .eq('product', FEATURE_FLAG_PRODUCT)
-      .eq('flag_name', BRANDALF_FLAG_NAME)
-      .maybeSingle();
+      .eq('flag_name', BRANDALF_FLAG_NAME);
 
     if (error) {
       log?.warn(`Failed to read brandalf flag for org ${organizationId}: ${error.message}`);
@@ -48,7 +66,7 @@ export async function isBrandalfEnabled(organizationId, postgrestClient, log) {
 
     // Absent row => flag not set => disabled, matching the previous behaviour
     // where a missing flag resolved to `false`.
-    return data?.flag_value === true;
+    return (data ?? []).find(isOrgRow)?.flag_value === true;
   } catch (error) {
     log?.warn(`Error checking brandalf flag for org ${organizationId}: ${error.message}`);
     return null;
--- a/test/utils/brandalf-utils.test.js
+++ b/test/utils/brandalf-utils.test.js
@@ -42,14 +42,17 @@ describe('brandalf-utils', () => {
   });
 
   /**
-   * Builds a chainable PostgREST client stub whose terminal `maybeSingle()`
-   * resolves to `{ data, error }`.
+   * Builds a chainable PostgREST client stub. The builder is awaited directly
+   * rather than through a terminal method, so the query object is a thenable
+   * resolving to `{ data, error }` — or rejecting with `rejectsWith`.
    */
-  function stubPostgrestClient({ data = null, error = null } = {}) {
+  function stubPostgrestClient({ data = null, error = null, rejectsWith = null } = {}) {
     const query = {
       select: sandbox.stub().returnsThis(),
       eq: sandbox.stub().returnsThis(),
-      maybeSingle: sandbox.stub().resolves({ data, error }),
+      then: (resolve, reject) => (
+        rejectsWith ? reject(rejectsWith) : resolve({ data, error })
+      ),
     };
     const from = sandbox.stub().returns(query);
     return { client: { from }, query, from };
@@ -57,20 +60,24 @@ describe('brandalf-utils', () => {
 
   describe('isBrandalfEnabled', () => {
     it('returns true when the brandalf flag is enabled', async () => {
-      const { client, from, query } = stubPostgrestClient({ data: { flag_value: true } });
+      const { client, from, query } = stubPostgrestClient({
+        data: [{ flag_value: true, brand_id: null }],
+      });
 
       const result = await isBrandalfEnabled('org-123', client, log);
 
       expect(result).to.equal(true);
       expect(from).to.have.been.calledWith('feature_flags');
-      expect(query.select).to.have.been.calledWith('flag_value');
+      // Wildcard projection is load-bearing — see `isOrgRow` for why.
+      expect(query.select).to.have.been.calledWith('*');
+      expect(query.eq).to.not.have.been.calledWithMatch('brand_id');
       expect(query.eq).to.have.been.calledWith('organization_id', 'org-123');
       expect(query.eq).to.have.been.calledWith('product', 'LLMO');
       expect(query.eq).to.have.been.calledWith('flag_name', 'brandalf');
     });
 
     it('returns false without querying when organizationId is missing', async () => {
-      const { client, from } = stubPostgrestClient({ data: { flag_value: true } });
+      const { client, from } = stubPostgrestClient({ data: [{ flag_value: true }] });
 
       const result = await isBrandalfEnabled(null, client, log);
 
@@ -94,6 +101,14 @@ describe('brandalf-utils', () => {
     });
 
     it('returns false when no flag row exists', async () => {
+      const { client } = stubPostgrestClient({ data: [] });
+
+      const result = await isBrandalfEnabled('org-123', client, log);
+
+      expect(result).to.equal(false);
+    });
+
+    it('returns false when the query yields no rows at all', async () => {
       const { client } = stubPostgrestClient({ data: null });
 
       const result = await isBrandalfEnabled('org-123', client, log);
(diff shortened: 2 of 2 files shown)


MysticatBot (approved):
Hey @rainer-friederich,

**Verdict: Approve** - well-scoped fix with correct dual-schema logic and thorough test coverage.
**Complexity: LOW** - small diff, single utility, single-service.
_Changes: Resolves the organization's own `brandalf` feature-flag row explicitly via an `isOrgRow` predicate instead of relying on `.maybeSingle()`, preventing a future brand-scoped override from silently degrading flag resolution (2 files)._

<details>

- suggestion: `select('*')` is load-bearing while the `brand_id` column does not yet exist, but should be narrowed to `select('flag_value, brand_id')` post-migration to avoid pulling unnecessary payload - add a TODO or follow-up issue to track - `src/utils/brandalf-utils.js:54`
- suggestion: Consider logging a warning when `data.filter(isOrgRow).length > 1` to surface data-integrity issues (duplicate org-level rows) rather than silently picking the first match - `src/utils/brandalf-utils.js:69`

</details>

Note: CI build is still pending - confirm it passes before merge.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 5s | <b>Cost:</b> $3.99 | <b>Commit:</b> <code>021bf8b831f1e8c15e783c33f05ed90092503997</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

rainer-friederich:
Thanks — both non-blocking suggestions triaged. No code changes; declining both, with reasoning.

- **suggestion: narrow `select('*')` to `select('flag_value, brand_id')` post-migration, tracked by a TODO** — declined. The payload argument is real but negligible here: the query is keyed on `(organization_id, product, flag_name)` and matches at most the organization's row plus any per-brand overrides — a handful of narrow rows. Set against that, a TODO to re-narrow this projection is the one change most likely to reintroduce the bug: narrowing it to `flag_value` alone makes `brand_id` arrive as `undefined` on every row, which makes every override read as the organization's own and silently disables the predicate — no error, no failing test. The wildcard is also not a stopgap unique to this repo; spacecat-api-service carries the same predicate and the same wildcard, so if narrowing ever becomes worthwhile it is a coordinated cleanup across both readers belonging with the migration's own tidy-up, not a dangling TODO here. Leaving the comments in place as the guard instead.

- **suggestion: warn when more than one row satisfies `isOrgRow`, to surface duplicate org-level rows** — declined; the state it would report cannot occur. Duplicate organization-level rows are excluded by the unique constraint under both schemas: today by `feature_flags_org_product_flag_name_key` on `(organization_id, product, flag_name)`, and after the migration by `UNIQUE NULLS NOT DISTINCT (organization_id, product, flag_name, brand_id)`. The `NULLS NOT DISTINCT` is deliberate and load-bearing — without it an organization could accumulate unlimited `brand_id IS NULL` rows, which is exactly the case this suggestion anticipates. With it, at most one row can satisfy the predicate. A warning branch for that state would be unreachable, and unreachable branches are a poor fit for a repo that gates on full branch coverage.

On the CI note: all checks are green now, `ci / build` and `codecov/patch` included.

rainer-friederich:
Runtime validation.

The change is a query-shape change, so what matters is the request the PostgREST client actually emits and how it handles a multi-row response. That was exercised directly against the real `@supabase/postgrest-js` client — the version this repo resolves — driving the changed helper against a local server serving `feature_flags` rows and recording each emitted request:

- Pre-migration row shapes (no `brand_id` on any row), flag on / flag off / no row → resolved as before in every case.
- Post-migration organization row alone → resolved.
- Organization row plus a brand override, both orderings, with the two rows disagreeing in both directions → the organization's value resolved every time.
- Brand override with no organization row → resolved as not-enabled.
- The emitted request never names or filters `brand_id` under either shape.
- Control: the previous `.maybeSingle()` call against two matching rows fails with `JSON object requested, multiple (or no) rows returned` — the quiet `null` this PR removes.

Live checks against the dev mysticat database, confirming which regime the deployed code runs under today:

- `feature_flags` columns → no `brand_id` yet; dev is still on the pre-migration schema, so the wildcard projection is currently load-bearing rather than merely future-proof.
- Unique constraint → `feature_flags_org_product_flag_name_key UNIQUE (organization_id, product, flag_name)`, so at most one row can match the helper's filter today.
- `brandalf` rows → 8 rows across 8 organizations, never more than one per `(organization_id, product, flag_name)`. With no override present anywhere, the new selection returns exactly what `.maybeSingle()` returned, which is what makes this a no-op on today's data.

Not exercised: a live audit run against an organization holding a per-brand `brandalf` override. That state cannot be constructed on the pre-migration schema — the column does not exist — which is why it is covered by tests and by the client-level check above rather than by an environment check. It becomes verifiable on an environment once the migration has landed there; the steps are in the test plan.

rainer-friederich:
Post-deploy check, prod.

Deployed: release v1.516.5 — `spacecat-services--audit-worker` code updated, version 1487 published and the `latest` alias repointed at 12:40:09 UTC. Prod log streams carry the `[1487]` prefix, confirming the running code is this build.

Health, 30 minutes post-deploy (~155k log records): no `Task timed out`, `Unhandled`, `Runtime.ImportModuleError`, and none of `Failed to read brandalf flag` / `Error checking brandalf flag` / `PostgREST client not available; cannot check brandalf flag`.

What this does not show: the changed path did not execute in the observation window. `brandalf` and `[BrandPresenceEnrichment]` return zero matches over 24 hours across ~6M records. `isBrandalfEnabled` is reached only from offsite brand-presence enrichment and from `llmo-customer-analysis` on first-time onboarding; the parent audits ran 478 times in 24h without reaching that branch. So this is absence of regression, not positive confirmation of the new read.

That gap is expected rather than concerning: on prod's current schema the change is a verified no-op — no `brand_id` column, the original three-column unique key, and 1106 `brandalf` rows across 1106 organizations with at most one per key, so `.find(isOrgRow)` necessarily returns the row `.maybeSingle()` returned. The override case only becomes reachable once the brand-scope migration lands.

PR #2871, opened 2026-08-10
feat(semrush-offsite): implement entitlement check for Semrush data access and update fallback logic for offsite audits (LLMO-6841)

tb-adbe:

## Related Issues
 - LLMO-6841

Thanks for contributing!

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Andrei Paraschiv", "@amitruluimadalina", "Dominique Jäggi"]
```

### DIFF
--- a/src/offsite-brand-presence/handler.js
+++ b/src/offsite-brand-presence/handler.js
@@ -18,6 +18,7 @@ import { AuditBuilder } from '../common/audit-builder.js';
 import { noopUrlResolver } from '../common/index.js';
 import { getPreviousWeeks, loadBrandPresenceData } from '../utils/offsite-brand-presence-enrichment.js';
 import { loadCitedUrlsFromSemrush } from '../utils/offsite-brand-presence-semrush.js';
+import { SEMRUSH_ENTITLEMENT_SKIP_REASONS } from '../utils/semrush-entitlement.js';
 import { postMessageOptional } from '../utils/slack-utils.js';
 import {
   computeBrandTokens,
@@ -107,12 +108,6 @@ const LOG_PREFIX = '[OffsiteBrandPresence]';
 const TOP_CITED_BUCKET = 'top-cited';
 // Valid values for messageData.domainScope on granular single-audit runs.
 const VALID_DOMAIN_SCOPES = new Set([...Object.keys(OFFSITE_DOMAINS), TOP_CITED_BUCKET]);
-
-// fallbackReason codes for a deliberate entitlement-based skip (brand not provisioned
-// in Semrush) — never a hard-stop, even under an explicit `enableSemrush:true` canary
-// run: the "no errors for non-entitled brands" contract must hold regardless of how
-// Semrush was enabled for this run.
-const ENTITLEMENT_SKIP_REASONS = new Set(['not-entitled', 'entitlement-check-failed']);
 
 const DOMAIN_ALIASES = Object.freeze({
   'youtu.be': 'youtube.com',
@@ -959,7 +954,7 @@ export async function offsiteBrandPresenceRunner(finalUrl, context, site, auditC
       // errors" for a non-entitled brand must hold regardless of how Semrush was
       // enabled. Hard-stop stays reserved for genuine technical failures (auth,
       // outage, etc.) that a canary run wants surfaced, not for expected scoping.
-      const isEntitlementSkip = ENTITLEMENT_SKIP_REASONS.has(reason);
+      const isEntitlementSkip = SEMRUSH_ENTITLEMENT_SKIP_REASONS.has(reason);
       if (hardStopOnFailure && !isEntitlementSkip) {
         // enableSemrush:true forced this run — surface the failure, no fallback.
         log.error(`${LOG_PREFIX} Semrush source failed (${reason}); hard stop — no legacy fallback (enableSemrush:true)`, {
@@ -1013,6 +1008,12 @@ export async function offsiteBrandPresenceRunner(finalUrl, context, site, auditC
   }
   const dataSource = usedSemrush ? 'semrush' : 'legacy';
   const semrushDegraded = usedSemrush && (semrushDiagnostics?.engineFailureCount ?? 0) > 0;
+  // Granular cause behind an entitlement-based `fallbackReason` (`flag-disabled` |
+  // `no-workspace` | `no-client` | `check-failed`) — set only on the two entitlement
+  // skip reasons (see the loader). Kept separate from `fallbackReason` so a wiring
+  // bug (`no-client`) stays distinguishable from a one-off transient blip
+  // (`check-failed`) without changing the coarse-grained hard-stop-exemption contract.
+  const entitlementReason = semrushDiagnostics?.entitlementReason;
 
   // Legacy source: runs when the flag is off, OR when Semrush was env-enabled but
   // failed (fallback). An enableSemrush:true run that failed already hard-stopped
@@ -1056,6 +1057,7 @@ export async function offsiteBrandPresenceRunner(finalUrl, context, site, auditC
         weeks: previousWeeks,
         dataSource,
         ...(fallbackReason ? { fallbackReason } : {}),
+        ...(entitlementReason ? { entitlementReason } : {}),
       },
       fullAuditRef: finalUrl,
     };
@@ -1124,6 +1126,7 @@ export async function offsiteBrandPresenceRunner(finalUrl, context, site, auditC
       weeks: previousWeeks,
       dataSource,
       ...(fallbackReason ? { fallbackReason } : {}),
+      ...(entitlementReason ? { entitlementReason } : {}),
       ...(semrushDegraded ? {
         semrushEngineFailureCount: semrushDiagnostics.engineFailureCount,
         semrushDegradedHosts: semrushDiagnostics.degradedHosts,
--- a/src/offsite-brand-presence/handler.js
+++ b/src/offsite-brand-presence/handler.js
@@ -108,6 +108,12 @@ const TOP_CITED_BUCKET = 'top-cited';
 // Valid values for messageData.domainScope on granular single-audit runs.
 const VALID_DOMAIN_SCOPES = new Set([...Object.keys(OFFSITE_DOMAINS), TOP_CITED_BUCKET]);
 
+// fallbackReason codes for a deliberate entitlement-based skip (brand not provisioned
+// in Semrush) — never a hard-stop, even under an explicit `enableSemrush:true` canary
+// run: the "no errors for non-entitled brands" contract must hold regardless of how
+// Semrush was enabled for this run.
+const ENTITLEMENT_SKIP_REASONS = new Set(['not-entitled', 'entitlement-check-failed']);
+
 const DOMAIN_ALIASES = Object.freeze({
   'youtu.be': 'youtube.com',
 });
@@ -948,7 +954,13 @@ export async function offsiteBrandPresenceRunner(finalUrl, context, site, auditC
     // with size 0) is NOT a failure and continues as a normal zero-URL run.
     if (semrushUrls === null) {
       const reason = semrushDiagnostics.fallbackReason ?? 'semrush-failed';
-      if (hardStopOnFailure) {
+      // A deliberate entitlement-based skip is never a hard stop, even when this run
+      // explicitly forced Semrush on via enableSemrush:true — "no wasted calls, no
+      // errors" for a non-entitled brand must hold regardless of how Semrush was
+      // enabled. Hard-stop stays reserved for genuine technical failures (auth,
+      // outage, etc.) that a canary run wants surfaced, not for expected scoping.
+      const isEntitlementSkip = ENTITLEMENT_SKIP_REASONS.has(reason);
+      if (hardStopOnFailure && !isEntitlementSkip) {
         // enableSemrush:true forced this run — surface the failure, no fallback.
         log.error(`${LOG_PREFIX} Semrush source failed (${reason}); hard stop — no legacy fallback (enableSemrush:true)`, {
           siteId, fallbackReason: reason,
@@ -969,12 +981,18 @@ export async function offsiteBrandPresenceRunner(finalUrl, context, site, auditC
           fullAuditRef: finalUrl,
         };
       }
-      // Enabled by the env var (or override not forced) — fall back to legacy so a
-      // Semrush problem never silently zeroes out offsite.
+      // Enabled by the env var (or override not forced, or an entitlement skip) —
+      // fall back to legacy so a Semrush problem never silently zeroes out offsite.
       fallbackReason = reason;
-      log.warn(`${LOG_PREFIX} Semrush source failed (${reason}); falling back to PostgREST/SharePoint`, {
-        siteId, fallbackReason: reason,
-      });
+      if (isEntitlementSkip) {
+        log.info(`${LOG_PREFIX} Semrush skipped (${reason}); falling back to PostgREST/SharePoint`, {
+          siteId, fallbackReason: reason,
+        });
+      } else {
+        log.warn(`${LOG_PREFIX} Semrush source failed (${reason}); falling back to PostgREST/SharePoint`, {
+          siteId, fallbackReason: reason,
+        });
+      }
     } else {
       for (const [url, info] of semrushUrls) {
         allUrls.set(url, info);
--- a/src/utils/brandalf-utils.js
+++ b/src/utils/brandalf-utils.js
@@ -14,9 +14,30 @@ const BRANDALF_FLAG_NAME = 'brandalf';
 const FEATURE_FLAG_PRODUCT = 'LLMO';
 
 /**
+ * The organization's own row, not a brand's override of it. `brand_id` is absent
+ * from every row before the brand-scope migration and NULL on the organization's
+ * row after it, so this selects correctly under both schemas.
+ *
+ * The query feeding this predicate must keep selecting the full row. Naming
+ * `brand_id` in a projection — or filtering on it — fails against the current
+ * schema, where the column does not exist yet; narrowing the projection once it
+ * does exist makes every override row arrive with `brand_id: undefined` and read
+ * as the organization's own.
+ *
+ * Exported so other `feature_flags` readers (e.g. `semrush-entitlement.js`) share
+ * this exact predicate instead of re-deriving it.
+ *
+ * @param {object} row - Raw PostgREST `feature_flags` row.
+ * @returns {boolean} `true` for the organization-level row.
+ */
+export const isOrgRow = (row) => (row.brand_id ?? null) === null;
+
+/**
  * Checks whether the brandalf feature flag is enabled for an organization by
  * reading the `feature_flags` table directly via the mysticat PostgREST client
  * (available at `context.dataAccess.services.postgrestClient`).
+ *
+ * Resolves the organization's own value, ignoring any brand's override of it.
  *
  * @param {string} organizationId - SpaceCat org UUID
  * @param {object} postgrestClient - mysticat PostgREST client (dataAccess.services.postgrestClient)
@@ -33,13 +54,13 @@ export async function isBrandalfEnabled(organizationId, postgrestClient, log) {
   }
 
   try {
+    // Wildcard projection is required — see `isOrgRow`.
     const { data, error } = await postgrestClient
       .from('feature_flags')
-      .select('flag_value')
+      .select('*')
       .eq('organization_id', organizationId)
       .eq('product', FEATURE_FLAG_PRODUCT)
-      .eq('flag_name', BRANDALF_FLAG_NAME)
-      .maybeSingle();
+      .eq('flag_name', BRANDALF_FLAG_NAME);
 
     if (error) {
       log?.warn(`Failed to read brandalf flag for org ${organizationId}: ${error.message}`);
@@ -48,7 +69,7 @@ export async function isBrandalfEnabled(organizationId, postgrestClient, log) {
 
     // Absent row => flag not set => disabled, matching the previous behaviour
     // where a missing flag resolved to `false`.
-    return data?.flag_value === true;
+    return (data ?? []).find(isOrgRow)?.flag_value === true;
   } catch (error) {
     log?.warn(`Error checking brandalf flag for org ${organizationId}: ${error.message}`);
     return null;
--- a/src/utils/brandalf-utils.js
+++ b/src/utils/brandalf-utils.js
@@ -10,34 +10,19 @@
  * governing permissions and limitations under the License.
  */
 
+import { readOrgFeatureFlag } from './feature-flags-utils.js';
+
 const BRANDALF_FLAG_NAME = 'brandalf';
 const FEATURE_FLAG_PRODUCT = 'LLMO';
-
-/**
- * The organization's own row, not a brand's override of it. `brand_id` is absent
- * from every row before the brand-scope migration and NULL on the organization's
- * row after it, so this selects correctly under both schemas.
- *
- * The query feeding this predicate must keep selecting the full row. Naming
- * `brand_id` in a projection — or filtering on it — fails against the current
- * schema, where the column does not exist yet; narrowing the projection once it
- * does exist makes every override row arrive with `brand_id: undefined` and read
- * as the organization's own.
- *
- * Exported so other `feature_flags` readers (e.g. `semrush-entitlement.js`) share
- * this exact predicate instead of re-deriving it.
- *
- * @param {object} row - Raw PostgREST `feature_flags` row.
- * @returns {boolean} `true` for the organization-level row.
- */
-export const isOrgRow = (row) => (row.brand_id ?? null) === null;
 
 /**
  * Checks whether the brandalf feature flag is enabled for an organization by
  * reading the `feature_flags` table directly via the mysticat PostgREST client
  * (available at `context.dataAccess.services.postgrestClient`).
  *
- * Resolves the organization's own value, ignoring any brand's override of it.
+ * Resolves the organization's own value, ignoring any brand's override of it —
+ * see `readOrgFeatureFlag` (`feature-flags-utils.js`) for the shared query/
+ * fail-closed shape every `feature_flags` reader in this worker uses.
  *
  * @param {string} organizationId - SpaceCat org UUID
  * @param {object} postgrestClient - mysticat PostgREST client (dataAccess.services.postgrestClient)
@@ -45,35 +30,9 @@ export const isOrgRow = (row) => (row.brand_id ?? null) === null;
  * @returns {Promise<boolean|null>} true/false when the flag state is known, null when unknown
  */
 export async function isBrandalfEnabled(organizationId, postgrestClient, log) {
-  if (!organizationId) {
-    return false;
-  }
-  if (!postgrestClient?.from) {
-    log?.warn('PostgREST client not available; cannot check brandalf flag');
-    return null;
-  }
-
-  try {
-    // Wildcard projection is required — see `isOrgRow`.
-    const { data, error } = await postgrestClient
-      .from('feature_flags')
-      .select('*')
-      .eq('organization_id', organizationId)
-      .eq('product', FEATURE_FLAG_PRODUCT)
-      .eq('flag_name', BRANDALF_FLAG_NAME);
-
-    if (error) {
-      log?.warn(`Failed to read brandalf flag for org ${organizationId}: ${error.message}`);
-      return null;
-    }
-
-    // Absent row => flag not set => disabled, matching the previous behaviour
-    // where a missing flag resolved to `false`.
-    return (data ?? []).find(isOrgRow)?.flag_value === true;
-  } catch (error) {
-    log?.warn(`Error checking brandalf flag for org ${organizationId}: ${error.message}`);
-    return null;
-  }
+  return readOrgFeatureFlag(postgrestClient, {
+    organizationId, product: FEATURE_FLAG_PRODUCT, flagName: BRANDALF_FLAG_NAME, log,
+  });
 }
 
 /**
--- a/src/utils/offsite-brand-presence-semrush.js
+++ b/src/utils/offsite-brand-presence-semrush.js
@@ -13,7 +13,11 @@
 import { ImsClient } from '@adobe/spacecat-shared-ims-client';
 import { tracingFetch as fetch } from '@adobe/spacecat-shared-utils';
 import { resolveBrandResultForSite } from './brand-resolver.js';
-import { resolveSemrushEntitlement } from './semrush-entitlement.js';
+import {
+  resolveSemrushEntitlement,
+  SEMRUSH_NOT_ENTITLED_REASON,
+  SEMRUSH_ENTITLEMENT_CHECK_FAILED_REASON,
+} from './semrush-entitlement.js';
 import { getDateWindowForPreviousWeeks } from './offsite-brand-presence-postgrest.js';
 import { classifyAndNormalize } from './offsite-brand-presence-enrichment.js';
 import {
@@ -231,7 +235,11 @@ async function fetchRows({ hostname, platform, url }, headers, log, siteId) {
  *   set to `{ fallbackReason }` with a specific code (`no-organization-id`, `no-active-brand`,
  *   `brand-resolution-failed`, `not-entitled`, `entitlement-check-failed`, `no-date-window`,
  *   `ims-token-failed`, or `surface-failed:<host>`) instead of the single generic reason the
- *   handler used to report for every case. On a
+ *   handler used to report for every case. The two entitlement reasons additionally set
+ *   `entitlementReason` to the granular cause from `resolveSemrushEntitlement`
+ *   (`flag-disabled` | `no-workspace` | `no-client` | `check-failed`) — `fallbackReason` alone
+ *   cannot distinguish a confirmed non-entitlement from a wiring bug (`no-client`) vs a
+ *   transient blip (`check-failed`). On a
  *   successful (non-null) return, set to
  *   `{ engineFailureCount, degradedHosts, authFailureDetected }` so a surface that tolerated
  *   partial engine failures is distinguishable from a clean run — both `dataSource` and
@@ -315,13 +323,24 @@ export async function loadCitedUrlsFromSemrush({
         durationMs: elapsed(),
       });
       await notify(':information_source: Brand is not entitled for Semrush — falling back to the legacy source.');
-      setDiagnostics({ fallbackReason: 'not-entitled' });
+      // fallbackReason is the coarse, contract-level signal the handler's hard-stop
+      // exemption keys off (SEMRUSH_ENTITLEMENT_SKIP_REASONS); entitlementReason keeps
+      // the granular cause (`flag-disabled` | `no-workspace` | `no-client` |
+      // `check-failed`) visible in diagnostics/auditResult without changing that
+      // contract — see ADR 002, Decision 7.
+      setDiagnostics({
+        fallbackReason: SEMRUSH_NOT_ENTITLED_REASON,
+        entitlementReason: entitlement.reason,
+      });
     } else {
       log.warn(`${LOG_PREFIX} Semrush entitlement check failed (transient) for org ${spaceCatId}; using legacy fallback`, {
         siteId, orgId: spaceCatId, brandId: brand.brandId, durationMs: elapsed(),
       });
       await notify(':warning: Could not verify Semrush entitlement (transient) — falling back to the legacy source.');
-      setDiagnostics({ fallbackReason: 'entitlement-check-failed' });
+      setDiagnostics({
+        fallbackReason: SEMRUSH_ENTITLEMENT_CHECK_FAILED_REASON,
+        entitlementReason: entitlement.reason,
+      });
     }
     return null;
   }
--- a/src/utils/offsite-brand-presence-semrush.js
+++ b/src/utils/offsite-brand-presence-semrush.js
@@ -13,6 +13,7 @@
 import { ImsClient } from '@adobe/spacecat-shared-ims-client';
 import { tracingFetch as fetch } from '@adobe/spacecat-shared-utils';
 import { resolveBrandResultForSite } from './brand-resolver.js';
+import { resolveSemrushEntitlement } from './semrush-entitlement.js';
 import { getDateWindowForPreviousWeeks } from './offsite-brand-presence-postgrest.js';
 import { classifyAndNormalize } from './offsite-brand-presence-enrichment.js';
 import {
@@ -228,8 +229,9 @@ async function fetchRows({ hostname, platform, url }, headers, log, siteId) {
  *   outage must not affect the Semrush attempt itself.
  * @param {object} [params.diagnostics] - Optional out-param, mutated in place. On a null return,
  *   set to `{ fallbackReason }` with a specific code (`no-organization-id`, `no-active-brand`,
- *   `brand-resolution-failed`, `no-date-window`, `ims-token-failed`, or `surface-failed:<host>`)
- *   instead of the single generic reason the handler used to report for every case. On a
+ *   `brand-resolution-failed`, `not-entitled`, `entitlement-check-failed`, `no-date-window`,
+ *   `ims-token-failed`, or `surface-failed:<host>`) instead of the single generic reason the
+ *   handler used to report for every case. On a
  *   successful (non-null) return, set to
  *   `{ engineFailureCount, degradedHosts, authFailureDetected }` so a surface that tolerated
  *   partial engine failures is distinguishable from a clean run — both `dataSource` and
@@ -290,6 +292,36 @@ export async function loadCitedUrlsFromSemrush({
       });
       await notify(':warning: Brand resolution failed (transient) — falling back to the legacy source.');
       setDiagnostics({ fallbackReason: 'brand-resolution-failed' });
+    }
+    return null;
+  }
+
+  // Gate on entitlement BEFORE any Semrush HTTP call (or minting an IMS token for
+  // one): Semrush data only exists for brands provisioned in Semrush (serenity flag
+  // on AND a resolvable workspace — same "flag AND workspace" gate api-service uses
+  // to serve any Serenity route). Calling it for a non-entitled brand is a wasted
+  // request on a paid, rate-limited product, and reliably yields an error/empty
+  // response that would just fall back anyway.
+  const entitlement = await resolveSemrushEntitlement(context, {
+    orgId: spaceCatId, brandId: brand.brandId,
+  });
+  if (!entitlement.entitled) {
+    if (entitlement.resolved) {
+      log.info(`${LOG_PREFIX} Brand not entitled for Semrush (${entitlement.reason}) for org ${spaceCatId}; skipping Semrush source`, {
+        siteId,
+        orgId: spaceCatId,
+        brandId: brand.brandId,
+        entitlementReason: entitlement.reason,
+        durationMs: elapsed(),
+      });
+      await notify(':information_source: Brand is not entitled for Semrush — falling back to the legacy source.');
+      setDiagnostics({ fallbackReason: 'not-entitled' });
+    } else {
+      log.warn(`${LOG_PREFIX} Semrush entitlement check failed (transient) for org ${spaceCatId}; using legacy fallback`, {
+        siteId, orgId: spaceCatId, brandId: brand.brandId, durationMs: elapsed(),
+      });
+      await notify(':warning: Could not verify Semrush entitlement (transient) — falling back to the legacy source.');
+      setDiagnostics({ fallbackReason: 'entitlement-check-failed' });
     }
     return null;
   }
--- a/src/utils/semrush-entitlement.js
+++ b/src/utils/semrush-entitlement.js
@@ -38,22 +38,59 @@
  * from a confirmed non-entitlement so the caller can log/diagnose accordingly.
  */
 
+import { isOrgRow } from './brandalf-utils.js';
+
 const LOG_PREFIX = '[semrush-entitlement]';
 
 const SERENITY_FLAG_PRODUCT = 'LLMO';
 const SERENITY_FLAG_NAME = 'serenity';
 
 /**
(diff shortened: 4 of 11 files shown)


src/utils/semrush-entitlement.js
```diff
@@ -0,0 +1,190 @@
...
+ */
+async function isSerenityEnabledForOrg(organizationId, postgrestClient, log) {
+  try {
+    const { data, error } = await postgrestClient
+      .from('feature_flags')
+      .select('flag_value')
+      .eq('organization_id', organizationId)
+      .eq('product', SERENITY_FLAG_PRODUCT)
+      .eq('flag_name', SERENITY_FLAG_NAME)
+      .maybeSingle();
```
amitruluimadalina: The `feature_flags` query here uses `.maybeSingle()`, which throws if PostgREST returns more than one row. `brandalf-utils.js`'s `isBrandalfEnabled` reads the identical table (`organization_id` + `product` + `flag_name`) but deliberately does a wildcard select plus a client-side `isOrgRow` filter, because its own comment documents that `brand_id` is moving from "absent on every row" to "NULL on the org row, set on brand-override rows" - i.e. more than one row can match this predicate once brand-scoped overrides exist. The moment a brand-level `serenity` override row appears for any org, this query throws, gets caught, and permanently returns `check-failed` for that org with no distinguishing alarm from an ordinary transient blip. Recommend mirroring the `isOrgRow` filter pattern (or `.is('brand_id', null)` server-side) and adding a two-row regression test.

src/offsite-brand-presence/handler.js
```diff
@@ -108,6 +108,12 @@ const TOP_CITED_BUCKET = 'top-cited';
 // Valid values for messageData.domainScope on granular single-audit runs.
 const VALID_DOMAIN_SCOPES = new Set([...Object.keys(OFFSITE_DOMAINS), TOP_CITED_BUCKET]);
 
+// fallbackReason codes for a deliberate entitlement-based skip (brand not provisioned
+// in Semrush) — never a hard-stop, even under an explicit `enableSemrush:true` canary
+// run: the "no errors for non-entitled brands" contract must hold regardless of how
+// Semrush was enabled for this run.
+const ENTITLEMENT_SKIP_REASONS = new Set(['not-entitled', 'entitlement-check-failed']);
```
amitruluimadalina: This `Set` is a second, hand-maintained copy of the reason strings the loader produces (`offsite-brand-presence-semrush.js:309,319`). Nothing ties them together at compile time or test time beyond the currently-passing tests. Consider exporting the reason strings as named constants from `semrush-entitlement.js` (or a small shared module) and importing them here, so a future rename can't silently widen or break the exemption.

src/utils/semrush-entitlement.js
```diff
@@ -0,0 +1,190 @@
...
+  try {
+    const timeoutPromise = new Promise((_, reject) => {
+      timeoutHandle = setTimeout(() => {
+        const e = new Error(`Semrush entitlement check timed out after ${SEMRUSH_ENTITLEMENT_TIMEOUT_MS}ms`);
+        e.name = 'TimeoutError';
+        reject(e);
+      }, SEMRUSH_ENTITLEMENT_TIMEOUT_MS);
+    });
+
+    const [flagEnabled, workspace] = await Promise.race([
```
amitruluimadalina, line 217: When the timeout wins this race, `isSerenityEnabledForOrg`/`resolveSemrushWorkspace` are still in flight with nothing attached to their eventual settlement. If that inner `Promise.all` rejects after the timeout has already returned `check-failed`, the rejection is effectively orphaned. The existing timeout test uses promises that never resolve or reject, so it doesn't actually exercise this. Recommend attaching a no-op `.catch()` to the raced `Promise.all` and adding a test where the losing promise rejects after the timeout fires.

src/utils/semrush-entitlement.js
```diff
@@ -0,0 +1,190 @@
...
+
+const SERENITY_FLAG_PRODUCT = 'LLMO';
+const SERENITY_FLAG_NAME = 'serenity';
+
+/**
+ * Maximum milliseconds to wait for the combined flag + workspace PostgREST lookup
+ * before failing closed. Mirrors `BRAND_RESOLUTION_TIMEOUT_MS` in `brand-resolver.js`
+ * — kept short so a PostgREST outage never amplifies offsite-audit latency.
+ */
+export const SEMRUSH_ENTITLEMENT_TIMEOUT_MS = 300;
```
amitruluimadalina, line 103: This value is copied from `brand-resolver.js`'s `BRAND_RESOLUTION_TIMEOUT_MS`, which was budgeted for a single PostgREST round-trip. Here it has to cover a PostgREST flag read running in parallel with two `dataAccess.Organization`/`Brand.findById()` calls. Worth a comment justifying why 300ms still holds for this three-call workload, or a plan to validate against real latency post-rollout - a budget that's too tight under real load silently converts this into a mostly-dead fallback path with no visibility.

src/utils/semrush-entitlement.js
```diff
@@ -0,0 +1,190 @@
...
+ * spacecat-api-service's `workspace-resolver.js` reads (minus its caching; see the
+ * file-level doc comment). Dual-mode: a brand's own sub-workspace takes precedence;
+ * the org's flat workspace is the fallback.
+ *
+ * @param {object} dataAccess - `context.dataAccess` (already validated by caller)
+ * @param {{orgId: string, brandId: string}} params
+ * @returns {Promise<{workspaceId: string, mode: 'subworkspace'|'flat'} | null>}
+ * @throws when either underlying lookup throws (caller decides how to treat it)
+ */
+async function resolveSemrushWorkspace(dataAccess, { orgId, brandId }) {
```
amitruluimadalina, line 153: This function never checks that the resolved `Brand` actually belongs to `orgId` - it's safe today only because the sole caller sources `brandId` from a lookup that's already scoped to `orgId` upstream. Since this is exported as a general-purpose entitlement primitive, recommend enforcing that invariant here directly (compare the brand's own org id to `orgId` before trusting its workspace) rather than relying on caller discipline.

src/offsite-brand-presence/handler.js
```diff
@@ -948,7 +954,13 @@ export async function offsiteBrandPresenceRunner(finalUrl, context, site, auditC
     // with size 0) is NOT a failure and continues as a normal zero-URL run.
     if (semrushUrls === null) {
       const reason = semrushDiagnostics.fallbackReason ?? 'semrush-failed';
-      if (hardStopOnFailure) {
+      // A deliberate entitlement-based skip is never a hard stop, even when this run
+      // explicitly forced Semrush on via enableSemrush:true — "no wasted calls, no
+      // errors" for a non-entitled brand must hold regardless of how Semrush was
+      // enabled. Hard-stop stays reserved for genuine technical failures (auth,
+      // outage, etc.) that a canary run wants surfaced, not for expected scoping.
+      const isEntitlementSkip = ENTITLEMENT_SKIP_REASONS.has(reason);
```
amitruluimadalina: `entitlement-check-failed` (a genuine PostgREST/data-access outage) and `not-entitled` (expected scoping) are both exempted from the `enableSemrush:true` hard-stop here. The entitlement decision itself is correctly fail-closed either way, but `enableSemrush:true` is specifically the canary override built to surface technical failures loudly - folding a real infrastructure outage into the same silent-skip bucket as expected scoping removes that signal, and a sustained outage would now look identical to normal operation in logs/dashboards. Worth a decision: keep `entitlement-check-failed` in the hard-stop path (matching `ims-token-failed`), or add a distinct alert on its volume outside the canary path.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -294,6 +296,36 @@ export async function loadCitedUrlsFromSemrush({
...
+  if (!entitlement.entitled) {
+    if (entitlement.resolved) {
+      log.info(`${LOG_PREFIX} Brand not entitled for Semrush (${entitlement.reason}) for org ${spaceCatId}; skipping Semrush source`, {
+        siteId,
+        orgId: spaceCatId,
+        brandId: brand.brandId,
+        entitlementReason: entitlement.reason,
+        durationMs: elapsed(),
+      });
+      await notify(':information_source: Brand is not entitled for Semrush — falling back to the legacy source.');
```
amitruluimadalina, line 358: This posts a Slack notification for every non-entitled brand, not just the transient-failure branch. Once Semrush is enabled fleet-wide with only a subset of brands entitled, this becomes routine, expected noise in a channel meant for anomalies. Consider keeping this at log/metric level and reserving the Slack ping for the `entitlement-check-failed` branch below, which is the actually anomalous case.

amitruluimadalina (changes requested):
Hey @tb-adbe,

## Strengths

- The entitlement gate is placed correctly: after brand resolution succeeds, before minting an IMS token, and before any Semrush HTTP call (`src/utils/offsite-brand-presence-semrush.js:304`). A non-entitled brand never triggers a token mint or a paid outbound request.
- Fail-closed semantics are applied consistently everywhere: missing input, missing client, query error, thrown exception, and timeout all resolve to `entitled: false`. There is no path where an exception or partial failure reads as "entitled."
- The `entitled`/`resolved` two-axis result shape cleanly separates "confirmed not entitled" from "check was inconclusive," and the caller threads that into different log levels and Slack messages accordingly.
- The `enableSemrush:true` canary override cannot force a call against a non-entitled brand. It only changes error-surfacing behavior after the entitlement gate has already said no, never re-enables Semrush for a rejected brand.
- The actual outbound Semrush request uses `spaceCatId`/`brand.brandId` directly, never the resolved `workspaceId`. The workspace resolution is purely a gate signal, never a routing parameter, which closes off a class of cross-tenant redirect concerns immediately.
- Unit test coverage on `resolveSemrushEntitlement` itself is thorough: flag-on/off, workspace present/absent, sub-workspace-vs-flat precedence, missing client, missing collections, and errors thrown from each of the three underlying calls.
- Regression coverage for the pre-existing hard-stop behavior is intact (`test/audits/offsite-brand-presence.test.js:323-337`, `:434-449` for `semrush-failed`/`ims-token-failed`), confirming the new `ENTITLEMENT_SKIP_REASONS` exemption is scoped to exactly the two intended reason strings.
- The ADR update (Decision 7) is unusually well reasoned: it separates genuine reuse (the shared `Organization`/`Brand` data-access models) from acknowledged duplication (the flag read, the gate shape, no caching), and records rejected alternatives with reasons.

## Issues

### Important (Should Fix)

1. `src/utils/semrush-entitlement.js:71` - the `feature_flags` query drops the multi-row safeguard that a sibling file in this same PR's dependency graph deliberately carries for the identical table. `isSerenityEnabledForOrg` does `.select('flag_value').eq('organization_id', ...).eq('product', ...).eq('flag_name', ...).maybeSingle()`. Compare with `brandalf-utils.js`'s `isBrandalfEnabled`, which does a wildcard `.select('*')` and a client-side `isOrgRow` filter specifically because its own comment documents that `brand_id` is transitioning from "absent on every row" to "NULL on the org row, set on brand-override rows" - i.e. more than one row can match `organization_id + product + flag_name` once brand-scoped overrides exist. `maybeSingle()` throws if PostgREST returns more than one row. The moment a brand-level override row for `serenity` appears for any org, every subsequent entitlement check for that org throws, is caught, and permanently returns `check-failed`, silently and invisibly disabling Semrush for that org with no alarm (it looks identical to an ordinary transient blip). Fix: filter server-side with `.is('brand_id', null)` if the column exists for this flag, or mirror the wildcard-select-plus-`isOrgRow`-filter pattern from `brandalf-utils.js`, and add a regression test with two matching rows.

2. `src/offsite-brand-presence/handler.js:962` and `src/utils/semrush-entitlement.js` (reason strings) - the `'not-entitled'`/`'entitlement-check-failed'` contract between the loader and the handler is unenforced. These two literals are produced in `offsite-brand-presence-semrush.js:310,320` and independently duplicated as a hardcoded `Set` in `handler.js:115` (`ENTITLEMENT_SKIP_REASONS`), with no shared constant and no test tying them together. A future rename of either string breaks the "never hard-stop on entitlement scoping" contract silently. Export the reason strings as named constants from `semrush-entitlement.js` and import them into both call sites, and add a test that imports both and asserts membership.

3. `src/utils/semrush-entitlement.js:159-166` (`Promise.race` timeout) - the losing branch (the flag query and the two `findById` calls) is never cancelled or given a `.catch()`. If the timeout wins the race, `isSerenityEnabledForOrg`/`resolveSemrushWorkspace` are still in flight; if that `Promise.all` later rejects after the timeout already returned `check-failed`, the rejection has nothing attached to it downstream of the timeout path. The existing timeout test (`test/utils/semrush-entitlement.test.js:791-808`) uses promises that never resolve or reject, so it proves the function returns the right shape after the timer fires but does not exercise a late rejection - the exact scenario that would surface this as an unhandled-rejection warning. Attach a no-op `.catch()` to the raced `Promise.all`, and add a test where the losing promise rejects after the timeout has already fired.

4. `src/utils/semrush-entitlement.js:51` - `SEMRUSH_ENTITLEMENT_TIMEOUT_MS = 300` is copied from `brand-resolver.js`'s `BRAND_RESOLUTION_TIMEOUT_MS` without re-deriving the budget for a materially different workload. That constant was sized for a single PostgREST round-trip; here it has to cover a PostgREST flag read running in parallel with two `dataAccess.Organization`/`Brand.findById()` calls. This looks fine against mocks and can flake under real connection-pool contention or a cold start, silently converting "genuine reuse" into a mostly-dead fallback path with no visibility (see issue 6). Add a comment justifying why 300ms still holds for this specific three-call workload, or give it its own derived value plus a plan to validate against real latency post-rollout.

5. `src/utils/semrush-entitlement.js:98` (`resolveSemrushWorkspace`) - no assertion that the resolved `Brand` actually belongs to the given `orgId`. Today this is safe only because the sole caller obtains `brand.brandId` from `resolveBrandResultForSite`, which is itself already scoped by `organization_id = orgId` upstream. But this module is written and exported as a general-purpose entitlement primitive with no invariant enforced at its own boundary. A future caller that resolves `orgId` and `brandId` from different sources could get a false `entitled: true` for the wrong org's provisioning state. Add an explicit check that the resolved brand's own org id matches `orgId` before trusting its workspace, and a regression test for the mismatch case.

6. `src/offsite-brand-presence/handler.js:962` - `entitlement-check-failed` (a genuine PostgREST/data-access outage, `resolved:false`) is treated identically to `not-entitled` (expected scoping, `resolved:true`) for purposes of exempting the `enableSemrush:true` hard-stop. There are two views on this. One: the entitlement decision itself is correctly fail-closed either way, so this is fine as designed, and the ADR explicitly frames both as expected non-errors. Two: `enableSemrush:true` is specifically the canary override built to surface technical failures loudly, and folding a real infrastructure outage into the same silent-skip bucket as expected scoping removes the one loud signal (hard stop plus `log.error`) that a sustained PostgREST outage, credential rotation break, or wiring regression would otherwise produce - it would now degrade every entitled org to legacy with just an info-level Slack note, indistinguishable in dashboards from normal operation. Worth a decision: either keep `entitlement-check-failed` in the hard-stop path (matching how `ims-token-failed` and `surface-failed:<host>` already hard-stop) while keeping `not-entitled` exempted, or explicitly commit to a separate metric/alert on `entitlement-check-failed` volume so this is still alertable outside the canary path.

7. `src/utils/offsite-brand-presence-semrush.js:317` - `notify()` fires a Slack message on every non-entitled brand, not just on the transient-failure branch. Once Semrush is enabled fleet-wide but only a subset of orgs/brands are entitled, every non-entitled audit run posts to Slack. Depending on audit cadence and fleet size this could be a lot of expected, routine noise in a channel meant for anomalies. Consider downgrading the `not-entitled` notify to log/metric only, and reserving the Slack ping for `entitlement-check-failed` (the actually anomalous case).

8. Test gap: the handler's `log.info` vs `log.warn` branch (`handler.js:981`) is not directly asserted. The two new tests in `test/audits/offsite-brand-presence.test.js:451-490` only assert `log.error` was not called; they never assert `log.info` was called for the skip message or that `log.warn` was not. A regression that swaps the branches or always logs via `log.warn` would pass both new tests untouched.

9. Test gap: `notify()` on the entitlement-skip paths is never exercised. In `test/utils/offsite-brand-presence-semrush.test.js:454-477`, both new tests call `run()` with `onProgress` undefined, so `notify()` is a guaranteed no-op. Nothing exercises the actual Slack message text added at `offsite-brand-presence-semrush.js:317,323`. Add at least one test per branch with a real `onProgress` stub asserting the expected text.

10. Test gap: no observability signal distinguishes a systemic entitlement-check failure (e.g. `no-client`, meaning the middleware wiring itself broke) from a one-off transient blip. Both currently log at `warn` with similar phrasing. If `dataAccess.Organization`/`.Brand`/`postgrestClient` wiring ever regresses in a deploy, every audit run for every entitled brand silently falls back to legacy with no hard stop and no distinguishable signal from expected scoping - which is exactly what this feature's contract makes hard to detect from the outside. Recommend a distinct log level or counter for `no-client` specifically, since that reason indicates a wiring bug rather than expected scoping.

### Minor (Nice to Have)

- `src/utils/semrush-entitlement.js` JSDoc return type lists all six `reason` values as one flat union; grouping them into "confirmed" (`flag-disabled`, `no-workspace`) vs "inconclusive" (`missing-input`, `no-client`, `check-failed`) in the doc comment itself would make the `resolved` distinction impossible to miss on a skim.
- `src/offsite-brand-presence/handler.js:981` - the `if (isEntitlementSkip) { log.info(...) } else { log.warn(...) }` branch duplicates the whole multi-line log call rather than varying just the level. Consider `const logFn = isEntitlementSkip ? log.info : log.warn;` bound appropriately, so the message text can't drift between the two branches.
- `test/utils/semrush-entitlement.test.js` - one assertion line in the "Brand.findById throws" test appears to lose its indentation in the diff (flush against the left margin). Likely a formatting slip lint should catch once CI resolves.
- `src/utils/semrush-entitlement.js:253` - the strict `flag_value === true` comparison is brittle if the column/serialization ever returns a truthy-but-non-boolean value. Low risk since it mirrors the existing `isBrandalfEnabled` pattern, but worth a one-line comment noting the assumption.
- `docs/decisions/002-offsite-cited-urls-semrush-source.md` - the "Considered and rejected" note on the `serenity/brand-presence/access` endpoint would be clearer stating the service-IMS-token mismatch is currently unverified (LLMO-6709) rather than reading as categorically ruled out, in case it's revisited once that ticket closes.
- The loader collapses five internal reason codes down to two `fallbackReason` values surfaced in `auditResult`. The finer-grained reason is still in loader-level structured logs, but if `not-entitled` volume ever needs a `flag-disabled` vs `no-workspace` breakdown for reporting, that data doesn't reach the audit result today. Cheap to add now, harder to backfill later.

## Recommendations

- File a follow-up ticket for extracting a shared entitlement-resolution package once a third consumer of this "flag AND workspace" pattern appears, or once api-service's own semantics change, and reference it from ADR Decision 7 so the duplication has a tracked exit condition rather than only living in prose.
- Add a short note (even in the ADR) on what a silent systemic fallback looks like operationally, since this feature is designed to fail invisibly by intent for the expected-scoping case, which is exactly what makes a genuine regression hard to distinguish from normal behavior.
- Consider one integration-style test that exercises `loadCitedUrlsFromSemrush` through a real (non-stubbed) `resolveSemrushEntitlement` with a fake `dataAccess`, since the two files are currently glued together purely through per-file mocking and a signature mismatch between them wouldn't be caught by either test file alone.

## Assessment

Ready to merge? With fixes.
Reasoning: the design is sound (fail-closed throughout, gate placed before any side effect, no entitlement bypass), but the `feature_flags` multi-row query hazard (issue 1) is a concrete bug waiting for a normal data migration to trigger it, and the timeout/observability/test-gap issues around it are cheap to fix now and expensive to debug once this is live against fleet traffic.
Note: CI checks were still pending at review time - recommend confirming build/lint pass before merge.

## Next Steps

1. Fix the `feature_flags` multi-row query hazard first - it is the one issue here that will eventually break silently in production.
2. Resolve the `entitlement-check-failed` hard-stop question (issue 6) and close the notify/log-level test gaps (issues 8 through 10).
3. The remaining important items (shared reason constants, timeout budget, org-ownership assertion, Slack noise) are worth doing before merge but are lower urgency than issue 1.
4. Minor items are optional.

src/utils/semrush-entitlement.js
tb-adbe: Checked the actual shared Brand data-access model (@adobe/spacecat-shared-data-access) — it has no organizationId/getOrganizationId() at all; it's deliberately minimal, scoped to only the fields the serenity provisioning flow reads/writes. There's no getter to check the suggested invariant against. Enforcing it would require a raw PostgREST query against brands.organization_id, reintroducing the table-level dependency this PR deliberately moved away from, for a scenario the sole caller (resolveBrandResultForSite, already server-side scoped by organization_id) cannot hit today. Documented the contract explicitly in both functions' JSDoc instead: a future caller resolving orgId/brandId from independent sources must guarantee the pairing itself.

src/utils/semrush-entitlement.js
tb-adbe: Verified this empirically rather than assuming — wrote a repro and confirmed Promise.race/Promise.all already attach a handler to every input promise at construction time (per spec), so a losing promise rejecting after the race has settled is never actually "unhandled" — it's a no-op against an already-settled race. I also tested what a naive .catch() chained directly onto the raced promise would do in the realistic fast-failure case: it swallows the real error and destructuring undefined throws a new TypeError, so the log ends up saying (intermediate value) is not iterable instead of the actual cause (e.g. brand lookup failed: connection refused). So this fix is either fully redundant or actively worse for debugging depending on how it's wired — leaving as-is.

src/offsite-brand-presence/handler.js
tb-adbe: Decided to keep entitlement-check-failed exempted from the enableSemrush:true hard-stop (gating a fleet-wide outage signal behind a per-run canary override only helps if someone happens to be testing that exact site at that moment). Considered and prototyped a dedicated, unconditional ops-channel Slack alert for this specific failure reason, then intentionally dropped it in favor of the simpler existing path: the loader already posts into the Slack thread that triggered the audit when this happens (unrelated to this PR). Trade-off accepted: channelId/threadTs are only populated on manually-triggered Slack runs — scheduled/automatic runs (the actual fleet-scale traffic) have no thread — so a sustained outage during normal operation produces no Slack signal, only log.info/log.warn and the new auditResult.entitlementReason field for anyone watching logs/dashboards. If fleet-wide alerting on this becomes a real operational need, worth revisiting with a dedicated channel then.

src/utils/offsite-brand-presence-semrush.js
tb-adbe: Checked postMessageOptional — it's a hard no-op (no Slack API call) whenever channelId/threadTs are absent, which is exactly the case for every scheduled/automatic run (only manually-triggered Slack runs carry those). So this call is fully inert for the fleet-wide scenario once the flag is on — it only ever fires on an attended, manually-triggered run, where every other stage of this same loader already posts an equivalent progress line regardless of entitlement outcome. Leaving as-is.

src/utils/brandalf-utils.js
```diff
@@ -24,10 +24,13 @@ const FEATURE_FLAG_PRODUCT = 'LLMO';
...
  * as the organization's own.
  *
+ * Exported so other `feature_flags` readers (e.g. `semrush-entitlement.js`) share
+ * this exact predicate instead of re-deriving it.
+ *
  * @param {object} row - Raw PostgREST `feature_flags` row.
  * @returns {boolean} `true` for the organization-level row.
  */
-const isOrgRow = (row) => (row.brand_id ?? null) === null;
+export const isOrgRow = (row) => (row.brand_id ?? null) === null;
```
amitruluimadalina: isOrgRow is a generic feature_flags schema fact, not a brandalf concept - it is now imported into the unrelated semrush-entitlement.js. Worth relocating to a neutral module before a third feature_flags reader (already anticipated in the ADR) cements this dependency.

src/utils/semrush-entitlement.js
```diff
@@ -0,0 +1,245 @@
...
+ * `brandId` from independent sources MUST enforce this invariant itself before calling
+ * `resolveSemrushEntitlement` — passing a mismatched pair here can attribute a
+ * different org's Semrush provisioning to this one.
+ *
+ * @param {object} dataAccess - `context.dataAccess` (already validated by caller)
+ * @param {{orgId: string, brandId: string}} params
+ * @returns {Promise<{workspaceId: string, mode: 'subworkspace'|'flat'} | null>}
+ * @throws when either underlying lookup throws (caller decides how to treat it)
+ */
+async function resolveSemrushWorkspace(dataAccess, { orgId, brandId }) {
```
amitruluimadalina, line 153: This gate reads only the org-level row and ignores any brand-scoped override. Asserted to mirror api-service's own org-wide flag read, but unverified here for the revoke direction - if a brand override can ever disable serenity for one brand while the org flag stays on, this would over-grant that brand. Worth confirming with api-service owners and recording the answer in ADR 002.

amitruluimadalina (commented):
Hey @tb-adbe,

Re-reviewed 55c301b against my prior findings, and this time actually read your inline replies (my mistake for not doing that first pass) — several of them changed my assessment materially.

## Strengths

- The `feature_flags` multi-row hazard is fixed at the root in both `isBrandalfEnabled` and `isSerenityEnabledForOrg`, with real two-row regression tests, not just a happy-path patch.
- The reason-string contract is now structurally enforced via frozen, shared, exported constants consumed by both the loader and the handler — drift is no longer possible by construction.
- Both explicit test gaps (log.info vs log.warn branch, notify() on skip paths) are closed with meaningful assertions verified directly against the current source.
- Your reply on the `Promise.race` thread is the strongest kind of response to a review comment: you didn't just assert it was fine, you reproduced it and additionally showed that the naive fix I implicitly suggested (`.catch()`) would make debugging worse (swallowed error, misleading `TypeError` from destructuring). I independently verified the "no unhandled rejection" half myself before seeing your reply and reached the same conclusion — leaving as-is is correct.
- Your reply on the Slack-noise thread checked `postMessageOptional`'s actual no-op behavior for scheduled/automatic runs — that's exactly the kind of "verify against the real runtime shape, not the worst-case reading of the diff" that resolves a review comment properly. Same for the `entitlement-check-failed` hard-stop trade-off reply: you prototyped the dedicated alert, found the mitigating fact (thread-only posting), and made a reasoned call with a stated revisit trigger. I'm treating both as resolved, not re-asking.
- The `Brand` model reply (no `organizationId` getter) confirms the CONTRACT doc's reasoning is accurate, not just plausible-sounding.

## Issues

### Important (Should Fix)

1. `src/utils/brandalf-utils.js:33` - `isOrgRow` is a generic `feature_flags` schema fact (org row vs. brand-override row), not a brandalf-specific concept, but it's exported from `brandalf-utils.js` and imported into the unrelated `semrush-entitlement.js` - the module's own JSDoc admits this ("Exported so other `feature_flags` readers... share this"). This is the classic second-consumer moment to relocate a shared primitive before a third `feature_flags` reader (already anticipated in the ADR's debt section) cements the wrong dependency graph. Move `isOrgRow` into a neutral module (e.g. `feature-flags-utils.js`) that both files import from. The two readers also now duplicate the more fragile invariant - wildcard projection + org-row filter + absent-row-means-false - verbatim; extracting a shared `readOrgFeatureFlag(postgrestClient, { product, flagName, orgId, log })` helper would be the higher-leverage version of the same fix.

2. `src/utils/semrush-entitlement.js` / `src/utils/brandalf-utils.js` - the gate reads only the org-level row and ignores any brand-scoped override for entitlement purposes; this is asserted to mirror spacecat-api-service's own org-wide flag read, but that's unverified here specifically for the revoke direction: if a brand override can ever disable serenity for one brand while the org flag stays on, this gate would over-grant that brand. Worth a quick confirmation with the api-service owners and a line in ADR 002 recording the answer either way, plus a test pinning "org=false + brand-override=true -> still not entitled" (or the reverse, if overrides can revoke) once confirmed.

### Minor (Nice to Have)

- The granular `entitlementReason` values (`flag-disabled|no-workspace|no-client|check-failed`) are bare literals, not exported constants, even though they now persist on `auditResult` - the same class of drift the two `fallbackReason` constants were promoted to prevent, one layer down. Export them if anything downstream will ever match on them.
- Missing the specific conflict-case test "org row `flag_value:false` + brand-override row `flag_value:true`" in both readers - the strongest proof a brand's `true` override can't flip an org-disabled verdict.
- The new behavioral tests assert `fallbackReason === 'not-entitled'` as a bare string rather than importing `SEMRUSH_NOT_ENTITLED_REASON` - low risk, but importing the constant would make the single-source-of-truth enforcement end-to-end.

## Recommendations

- Your reply on `entitlement-check-failed` names a real future trigger ("if fleet-wide alerting on this becomes a real operational need, worth revisiting"). Worth a one-line note in ADR 002 recording that decision + trigger so it isn't re-litigated from scratch next time someone reads the hard-stop logic cold.
- Same for the `resolveSemrushWorkspace` org-membership question - the CONTRACT doc plus your reply are a solid explanation for today, but since the sole caller assumption is load-bearing, a tracking ticket for "add enforcement if a second, independently-scoped caller appears" would make that assumption revisit itself automatically instead of relying on the next reader noticing the JSDoc.

## Assessment

Ready to merge? With fixes.
Reasoning: most of the substantive concerns from the first review are genuinely closed - either fixed in code (multi-row hazard, reason-string contract) or resolved with verified, well-reasoned replies (Promise.race, Slack noise, hard-stop trade-off, Brand model limits). What's left is smaller: a shared primitive (`isOrgRow`) landed in the wrong module on its second use, and one unverified assumption (brand-override revoke semantics) worth a quick confirmation before this is relied on at fleet scale. Neither blocks correctness of the flag-off default path.

## Next Steps

1. Relocate `isOrgRow` out of `brandalf-utils.js` before a third `feature_flags` reader lands.
2. Confirm brand-override revoke semantics with api-service owners and record the answer in the ADR.
3. Minor items are optional.

src/utils/brandalf-utils.js
tb-adbe: Agreed — this was the right call. Moved `isOrgRow`, and the whole duplicated query/error-handling block it sat inside (not just the predicate), into a new `src/utils/feature-flags-utils.js`. Both `isBrandalfEnabled` and `isSerenityEnabledForOrg` are now thin wrappers around a shared `readOrgFeatureFlag(postgrestClient, { organizationId, product, flagName, log })`, so the wildcard-select + `isOrgRow`-filter + fail-closed error handling lives in exactly one place instead of two copies that could drift independently. Added `test/utils/feature-flags-utils.test.js` to cover the shared helper directly, on top of the existing coverage through both consumers. Documented as Decision 8e in ADR 002.

src/utils/semrush-entitlement.js
tb-adbe: Checked this directly against spacecat-api-service rather than leaving it open. `feature-flags-storage.js`'s `readFeatureFlag()` — which backs `isSerenityActiveForOrg` in `serenity-active.js`, the exact org-wide gate this module mirrors — queries only `organization_id`/`product`/`flag_name`, with zero `brand_id` awareness. I grepped the entire api-service repo for any combination of `brand_id` with the `feature_flags` table: zero hits, anywhere. Brand-scoped feature-flag overrides — grant or revoke direction — aren't a feature that exists in production today in either codebase, so "can a brand override ever revoke an org-enabled flag" is moot rather than unconfirmed: there's no such override being written anywhere for this to happen with. `isOrgRow` reading only the org's own row and ignoring any override is therefore correct under the current schema by construction. Recorded this in ADR 002 (Decision 8e), and added the regression test you separately flagged as missing — org row `flag_value:false` + brand-override row `flag_value:true` still resolves not-entitled — to both `isBrandalfEnabled` and `isSerenityEnabledForOrg`, closing the gap where we only had the inverse case covered.

tb-adbe:
Addressed the minor items too: `resolveSemrushEntitlement`'s internal reason values are now a frozen `SEMRUSH_ENTITLEMENT_REASONS` lookup (`ENTITLED`/`FLAG_DISABLED`/`NO_WORKSPACE`/`MISSING_INPUT`/`NO_CLIENT`/`CHECK_FAILED`) instead of bare literals, and both `offsite-brand-presence.test.js` and `offsite-brand-presence-semrush.test.js` now assert against `SEMRUSH_NOT_ENTITLED_REASON`/`SEMRUSH_ENTITLEMENT_CHECK_FAILED_REASON`/`SEMRUSH_ENTITLEMENT_REASONS` instead of copying the string values. Skipped filing a separate tracking ticket for the `resolveSemrushWorkspace` org-membership enforcement — it's a documented JSDoc contract (Decision 8d) and the sole caller can't hit the scenario today; will revisit if a second, independently-scoped caller actually shows up.

src/utils/semrush-entitlement.js
```diff
@@ -0,0 +1,246 @@
...
+ * the organization's own row and ignores any brand-scoped override for the same
+ * `organization_id`/`product`/`flag_name` — see that module for why a narrower,
+ * single-row-assuming query is unsafe against this table.
+ *
+ * @param {string} organizationId - SpaceCat org UUID
+ * @param {object} postgrestClient - mysticat PostgREST client (already validated by caller)
+ * @param {object} log - Logger
+ * @returns {Promise<boolean|null>} true/false when known, null when the query itself failed
+ */
+async function isSerenityEnabledForOrg(organizationId, postgrestClient, log) {
```
amitruluimadalina: This mirrors isSerenityActiveForOrg in spacecat-api-service, but that function's own doc comment says it's reserved for the no-brand-yet case and that "everything that acts on an existing brand must use isSerenityActiveForBrand instead." Since resolveSemrushEntitlement operates per-brand throughout, it should mirror isSerenityActiveForBrand/resolveFlagRowForBrand (brand override takes precedence over the org row), not the org-only predicate. See the review body for the full evidence against the current api-service source.

amitruluimadalina (changes requested):
Hey @tb-adbe,

Round 3 reviewed. The isOrgRow relocation and most round-2 minor items are cleanly closed, but I found a concrete correctness gap in the brand-override closure that I verified directly against the actual spacecat-api-service source, not just against the ADR's characterization of it.

## Strengths

- `isOrgRow` and the read logic are now in a genuinely neutral home. `src/utils/feature-flags-utils.js`'s `readOrgFeatureFlag(postgrestClient, { organizationId, product, flagName, log })` is a real extraction, not a cosmetic one: both `isBrandalfEnabled` (brandalf-utils.js) and `isSerenityEnabledForOrg` (semrush-entitlement.js) are now thin delegates, and neither feature's file imports the other's anymore. That fully resolves the wrong-module finding from round two.
- The `SEMRUSH_ENTITLEMENT_REASONS` frozen lookup replaces every bare reason-string literal in `resolveSemrushEntitlement`, and the behavioral tests in `offsite-brand-presence.test.js` / `offsite-brand-presence-semrush.test.js` now import and assert against the constants instead of copying string literals. All three round-two Minor items (granular reason constants, the missing conflict-case test, tests pinning literals) are genuinely closed.
- `feature-flags-utils.test.js` exercises the shared helper's full fail-closed contract (missing org id, missing client, query error, throw, no-match, happy path) with warn messages pinned by regex against the actual interpolated strings.
- Fail-closed semantics hold throughout: every non-entitled path still returns `entitled:false` before any IMS token mint or Semrush HTTP call.

## Issues

### Important (Should Fix)

1. `src/utils/semrush-entitlement.js:115` (`isSerenityEnabledForOrg`) mirrors the wrong gate in spacecat-api-service. I saw your reply on this thread reporting that you grepped the entire api-service repo for any combination of `brand_id` with the `feature_flags` table and found zero hits, and that ADR 002 Decision 8e records brand-scoped overrides as "not a feature that exists in production today in either codebase." I checked this directly against the current api-service source rather than taking that at face value, and it doesn't hold up:

   - `src/support/feature-flags-storage.js` exports `resolveFlagRowForBrand(scopes, brandId)`, returning `scopes.brandRows.get(brandId) ?? scopes.orgRow ?? null` - a brand override takes precedence over the org row. Its doc comment: "A brand row of `false` likewise holds one brand back from an organization that is on."
   - `src/support/serenity/serenity-active.js` exports `isSerenityActiveForOrg` (org row only, the one this module mirrors) and `isSerenityActiveForBrand` (resolves via `resolveFlagRowForBrand`) for the exact `serenity` flag in question. The doc comment on `isSerenityActiveForOrg` says explicitly: "Use this only where there is no brand to resolve against... Everything that acts on an existing brand must use `isSerenityActiveForBrand` instead."
   - `resolveSemrushEntitlement` here operates per-brand throughout (takes both `orgId` and `brandId`, resolves a per-brand Semrush sub-workspace) - exactly the "acts on an existing brand" case api-service's own docs say requires the brand-aware predicate.
   - The regression test added this round ("org row `false` + brand-override row `true` -> still not entitled") pins the opposite of what `resolveFlagRowForBrand` actually resolves for that input.

   I'm not sure why the grep didn't surface these - `brand_id` and `feature_flags` are combined in the function names themselves - but the functions and their doc comments are there on `main` today. Given `isSerenityActiveForBrand`'s own doc frames brand-scoped overrides as the intended mechanism for a brand-by-brand rollout (not a hypothetical), this worker will diverge from api-service in both directions the day any org uses one. Recommend mirroring `resolveFlagRowForBrand`/`isSerenityActiveForBrand` instead - since `readOrgFeatureFlag`'s query already wildcard-selects, this is a resolution change (`brandRows.get(brandId) ?? orgRow`), not an added round trip. At minimum, correct ADR 8e's wording so it doesn't assert a safety property that api-service's own source contradicts, and flip the new regression test to assert the intended (brand-overrides-org) behavior once decided.

### Minor (Nice to Have)

- `docs/decisions/002-offsite-cited-urls-semrush-source.md` (Decision 8e) and the JSDoc in `feature-flags-utils.js` / `semrush-entitlement.js` restate api-service's internals inline in prose. Once the finding above is resolved, consider anchoring the cross-repo claim in the ADR only and having source comments point at it ("mirrors api-service's per-brand gate; see ADR 002 section 8e"), so two files don't go stale independently of the real source.
- `src/utils/feature-flags-utils.js` - a future caller of the now-shared, more discoverable `readOrgFeatureFlag(client, { organizationId })` could pass an `organizationId` from a less-trusted source without realizing the function only filters by the id it's given; a one-line contract note (mirroring the CONTRACT comments already on `resolveSemrushWorkspace`) would carry that forward. Both current callers pass an authoritative, already-scoped org id, so no live exposure today.
- `test/utils/feature-flags-utils.test.js` - the shared helper's own test file doesn't assert the "wildcard projection, not narrowed to exclude `brand_id`" negative (`query.eq` never called with `'brand_id'`) that `brandalf-utils.test.js` still carries for its delegate. Since this file is now the canonical home for that query, the load-bearing negative belongs here too. Also missing from this file specifically: a `data: null` case (vs. `data: []`) and a client-without-`.from` case (vs. `undefined` client) - both covered at the delegate layer, but the shared helper should own its own branches.
- No test anywhere pins the strict `flag_value === true` comparison against a truthy-but-non-boolean value (e.g. `'true'`, `1`) resolving to `false`. Low-priority latent gap.
- `brandalf-utils.test.js` now substantially duplicates `feature-flags-utils.test.js`'s branch coverage through the delegate; worth trimming to a delegation contract (correct params threaded, one happy path, the `brand_id` negative) once the shared file is fully self-sufficient. Not blocking.

## Recommendations

- Before deciding the fix shape for the Important item, it would help to know whether any organization currently has a brand-scoped row for the `serenity` flag specifically (as opposed to the mechanism existing but being unused so far for this particular flag) - that changes whether this is a live divergence or a dormant one, though either way the mirrored predicate should be the correct one going forward.
- If a full fix needs more design time, a cheap interim mitigation: log a warning if `readOrgFeatureFlag`'s underlying query ever returns a row with a non-null `brand_id` for a flag this reader treats as org-only, so the day a brand-scoped `serenity` row appears anywhere, it's an observable signal rather than a silent divergence.

## Assessment

Ready to merge? With fixes.
Reasoning: the round-two architectural finding (isOrgRow placement) is fully and cleanly resolved, and the accompanying test/constant cleanup is thorough. But the closure of the brand-override question rests on an ADR claim that doesn't match the actual mirrored codebase, and the underlying mismatch (mirroring `isSerenityActiveForOrg` instead of `isSerenityActiveForBrand` for a check that resolves per-brand) is a real, findable divergence with a fail-open direction (over-granting Semrush access) once brand-scoped `serenity` rows exist anywhere. This is the one item I'd want resolved, or explicitly and accurately documented as a known deferred gap, before merge.

## Next Steps

1. Resolve the wrong-gate-mirrored finding: either switch to a brand-aware read matching `resolveFlagRowForBrand`/`isSerenityActiveForBrand`, or correct ADR 8e to state the accurate risk and track it explicitly rather than asserting it's moot.
2. Flip the new regression test to match whichever resolution is chosen.
3. Minor items are optional.

src/utils/semrush-entitlement.js
tb-adbe: Confirmed — this was a real bug, not a re-hash of the settled question. My previous check was against a local api-service checkout that turned out to be 24 commits stale; `origin/main` had since shipped `#3024` ("resolve the LLMO/serenity flag per brand, so an org migrates in waves"), adding exactly `resolveFlagRowForBrand`/`isSerenityActiveForBrand`. Re-fetched and re-verified directly against current `origin/main` — your read of both functions and their doc comments is accurate.

Fixed: added `resolveFeatureFlagForBrand(postgrestClient, { organizationId, brandId, product, flagName, log })` to `feature-flags-utils.js`, mirroring `resolveFlagRowForBrand`'s exact precedence (brand row wins, else org row, else unset). Renamed the internal `isSerenityEnabledForOrg` to `isSerenityEnabledForBrand`, now threading `brandId` through. `isBrandalfEnabled` stays on the org-only `readOrgFeatureFlag` — re-checked against current `origin/main` that `brandalf`/`brandalf_migration` are read org-only everywhere in api-service too, unrelated to the per-brand serenity mechanism, so that one's correct as-is.

One correction to your note on the regression test: the test I added last round used a *different* brand's override id (`'some-other-brand'`), not the brand actually under test — so it wasn't pinning the wrong direction, it just wasn't exercising the matching-brand case at all either way. Added two new tests using the real brand id under test, pinning both directions: a `true` brand override entitles despite a `false` org row, and (the higher-severity one) a `false` brand override denies despite a `true` org row.

ADR 002: struck through the incorrect claim in Decision 8e rather than rewriting history, and added Decision 8f with the full correction, including a note-to-self about dating cross-repo "this doesn't exist" claims since a sibling repo can ship the exact thing being ruled out before a review round closes.

src/utils/semrush-entitlement.js
tb-adbe: Confirmed — this was a real bug, not a re-hash of the settled question. My previous check was against a local api-service checkout that turned out to be 24 commits stale; `origin/main` had since shipped `#3024` ("resolve the LLMO/serenity flag per brand, so an org migrates in waves"), adding exactly `resolveFlagRowForBrand`/`isSerenityActiveForBrand`. Re-fetched and re-verified directly against current `origin/main` — your read of both functions and their doc comments is accurate.

Fixed: added `resolveFeatureFlagForBrand(postgrestClient, { organizationId, brandId, product, flagName, log })` to `feature-flags-utils.js`, mirroring `resolveFlagRowForBrand`'s exact precedence (brand row wins, else org row, else unset). Renamed the internal `isSerenityEnabledForOrg` to `isSerenityEnabledForBrand`, now threading `brandId` through. `isBrandalfEnabled` stays on the org-only `readOrgFeatureFlag` — re-checked against current `origin/main` that `brandalf`/`brandalf_migration` are read org-only everywhere in api-service too, unrelated to the per-brand serenity mechanism, so that one's correct as-is.

One correction to the regression-test note: the test I added last round used a *different* brand's override id (`'some-other-brand'`), not the brand actually under test — so it wasn't pinning the wrong direction, it just wasn't exercising the matching-brand case at all either way. Added two new tests using the real brand id under test, pinning both directions: a `true` brand override entitles despite a `false` org row, and (the higher-severity one) a `false` brand override denies despite a `true` org row.

ADR 002: struck through the incorrect claim in Decision 8e rather than rewriting history, and added Decision 8f with the full correction, including a note-to-self about dating cross-repo "this doesn't exist" claims since a sibling repo can ship the exact thing being ruled out before a review round closes.

amitruluimadalina (approved):
Hey @tb-adbe,

Round 4 reviewed. The serenity per-brand fix closes the last open item cleanly.

## Strengths

- `resolveFeatureFlagForBrand` (`feature-flags-utils.js`) mirrors api-service's `resolveFlagRowForBrand` precedence exactly - brand override wins, falls back to the org row - and is built on the same row-fetch (`fetchFeatureFlagRows`) that `readOrgFeatureFlag` now shares, so there is exactly one place the query itself is defined.
- `semrush-entitlement.js` now calls `isSerenityEnabledForBrand`, correctly mirroring api-service's `isSerenityActiveForBrand` instead of the org-only predicate, matching the fact that `resolveSemrushEntitlement` operates per-brand throughout.
- The new regression tests use the actual brand under test (not a different brand's id) and pin both directions: a `true` brand override entitles despite a `false` org row, and a `false` brand override denies despite a `true` org row - the higher-severity direction, since it's the one that would have over-granted the paid Semrush call.
- Verified `isBrandalfEnabled` correctly stays on the org-only `readOrgFeatureFlag` - that flag has no per-brand override mechanism in api-service, so no change was needed there.
- The ADR correction is exemplary: Decision 8e's incorrect claim is struck through rather than deleted, Decision 8f records the full story (including that the original research was against a 24-commits-stale local checkout of api-service, not a fabricated claim), and closes with a genuinely useful process note for future cross-repo "this doesn't exist" claims - that they have a shelf life and should record what commit they were checked against.
- Picked up several round-three minor items along the way: the shared test file now asserts the `brand_id`-not-filtered negative, and gained the `data: null` and missing-`.from` branch cases it was missing.
- CI is green.

## Assessment

Ready to merge? Yes.
Reasoning: the substantive finding from this review (mirroring the wrong api-service gate for a per-brand check) is fixed correctly, with test coverage that actually exercises both directions against the real brand under test, and the ADR is corrected honestly rather than just patched over. Nothing outstanding blocks merge.

AndreiAlexandruParaschiv (approved):

PR #2880, opened 2026-08-11
fix(metatags): emit absolute suggestion/page URLs for sub-path sites (metatags + product-metatags)

rpapani:
## What
Meta-tags and product-meta-tags built the suggestion `data.url`, the Mystique `pageUrls`, and the SSR-validator fetch URL by **string-concatenating** the page's absolute pathname onto the full base/final URL:

```js
`${site.getBaseURL()}${endpoint}`        // pageUrls
getBaseUrl(finalUrl, useHostnameOnly) + endpoint   // suggestion data.url
`${baseUrl}${endpoint}`                  // SSR validator
```

For a **sub-path site** (baseURL e.g. `https://example.com/foo`) the endpoint already carries the `/foo` prefix, so concatenation produced a malformed `https://example.com/foo/foo/...` URL.

## Fix
- **metatags** (`handler.js`, `ssr-meta-validator.js`): resolve each endpoint (an absolute path) against the base URL's **origin** — `new URL(endpoint, baseUrl)` for `pageUrls`/SSR, and default `useHostnameOnly` to **true** (both the initial value and the `?? true` fallback) so `getBaseUrl(finalUrl)` returns the origin for the suggestion `data.url`. Now that endpoints are always absolute pathnames, hostname-only is the correct default.
- **product-metatags** (`handler.js`): identical defect — same `useHostnameOnly` default flip (initial value + `?? true`).

All spots are **no-ops for root-domain sites** (their base URL has no path).

## Re-keying note
Existing sub-path sites' malformed suggestions will be re-keyed (**OUTDATED → new**) on the next run, since `buildKey` includes `data.url`.

## Tests
- metatags: SSR validator and `opportunityAndSuggestions` sub-path `data.url` assertions; a **wired** `runAuditAndGenerateSuggestions` assertion on the `newData` handed to `syncSuggestions` (data.url is `https://example.com/foo/<page>`, not `/foo/foo/`); the four `useHostnameOnly`-default tests updated to the hostname-only result.
- product-metatags: sub-path `data.url` assertion on the wired `syncSuggestions` → `addSuggestions` path; existing default-branch assertions corrected.
- All affected suites green; lint clean; coverage bar held.

## Context
Bug 2 of **SITES-49656** (multi-site-per-domain / sub-path onboarding).

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/metatags/handler.js
+++ b/src/metatags/handler.js
@@ -52,10 +52,9 @@ export async function opportunityAndSuggestions(finalUrl, auditData, context) {
   const { log } = context;
   const { detectedTags } = auditData.auditResult;
   log.debug(`started to audit metatags for site url: ${auditData.auditResult.finalUrl}`);
-  // Default to hostname-only: the endpoint is the page's absolute pathname, so it must
-  // be appended to the origin. Appending to a base that already carries a path (a
-  // sub-path site, e.g. https://oklahoma.gov/omes) duplicated it (/omes/omes/...).
-  // For root-domain sites the base has no path, so this is a no-op.
+  // Default hostname-only: the endpoint is an absolute pathname appended to the origin.
+  // A base that already carries a path (sub-path site, e.g. example.com/foo) duplicated
+  // it (/foo/foo/...). No-op for root-domain sites.
   let useHostnameOnly = true;
   try {
     const siteId = opportunity.getSiteId();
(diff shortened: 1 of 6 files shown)


FentPams (approved):
✅ Approve — `new URL(endpoint, base)` + the hostname-only default correctly kills the `/omes/omes` duplication, and root-domain sites are a provable no-op.

One thing to keep in mind (not blocking): flipping the `useHostnameOnly` default `false → true` is **fleet-wide** (every site with the config unset), and correctness depends on detected endpoints being absolute paths that already carry the sub-path. That holds today because endpoints are derived as `new URL(finalUrl).pathname` (`handler.js:152-153`), but it's unguarded — if detection ever emits a path relative to the sub-path, `new URL(endpoint, base)` silently drops the base path. A short assert/comment at the URL-build site would make that invariant explicit. Explicit `useHostnameOnly:false` still overrides, so the blast radius is bounded.

Nit: the suggestion `url` is still `getBaseUrl(...) + endpoint` (`handler.js:368`) while `pageUrls` moved to `new URL(...)` — same result under the new default, but unifying on `new URL` avoids future drift.

PR #2881, opened 2026-08-11
fix(audits): query RUM by hostname for sub-path sites (cwv, internal-links, forms, alt-text)

rpapani:
## What
cwv, broken-internal-links and forms-opportunities passed the resolved audit URL **directly** as the RUM `domain`. For a **sub-path site** the resolved URL carries a path (e.g. `example.com/foo`) which has **no RUM domainkey**, so the domainkey exchange 404s and the audit **fails outright**:
```
Query 'cwv' failed. Opts:{"domain":"example.com/foo",...}
Error during fetching domainkey for domain 'example.com/foo' using admin key. Status: 404
```
(Verified in prod on a sub-path site — both cwv and broken-internal-links errored at the initial step.)

## Fix
- New shared `getRUMDomain(url)` (`src/support/utils.js`) — strips a resolved audit/final URL to its **hostname** (no-op for root/bare-hostname sites).
- Use it for the RUM `domain` in **cwv** (`cwv-audit-result.js`), **broken-internal-links** (`internal-links/rum-detection.js`), and **forms-opportunities** (`forms-opportunities/handler.js`).
- RUM is domain-wide, so results are scoped to the site base path: cwv & internal-links already had `isWithinAuditScope` filters; **forms** now filters `formVitals` via `filterByAuditScope`, and **image-alt-text** scopes its RUM `traffic-acquisition` fallback URLs (its domain was already hostname-only via `getRUMUrl`).

All no-ops for root-domain sites.

## Landing-page widening
The shared subpath filter (`isWithinAuditScope`) used a strict `${basePath}/` boundary, which silently dropped the section's own landing page (`${basePath}.html`) for a sub-path site. The filter now also accepts `${basePath}.html` in both the absolute- and relative-URL branches, while preserving directory-boundary safety (siblings like `/foobar` / `/foo-x` are still dropped).

## Why it complements #2873
The merged CWV per-URL base-path filter (#2873) only runs **after** a successful RUM query. On a sub-path site the query itself 404s first — this PR fixes that earlier failure so the filter can do its job.

## Deferred call sites (found in review)
Not addressed here to avoid a whole-domain leak (each needs both the hostname-strip and a base-path filter together):
- `experimentation-ess/common.js` (`processAudit`) — path-carrying `domain`, no try/catch, would 404 the domainkey exchange for sub-path sites.
- `metatags/handler.js` + `product-metatags/handler.js` `calculateProjectedTraffic` — wwwUrlResolver-sourced RUM `traffic-acquisition` domain; fails soft to zero projected traffic for sub-path sites.

## Tests
- `getRUMDomain` unit tests (hostname strip / root no-op / unparseable fallback).
- cwv, internal-links, and forms: assert the RUM query `domain` is the bare hostname, not the sub-path URL.
- subpath-filter: asserts the `${basePath}.html` landing page is kept while sibling prefixes are still dropped.
- All affected suites green; lint clean.

## Context
Bug 3 of **SITES-49656** (multi-site-per-domain / sub-path onboarding).

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Xinyi Feng", "Julien Ramboz", "Dominique Jäggi"]
```

### DIFF
--- a/src/forms-opportunities/handler.js
+++ b/src/forms-opportunities/handler.js
@@ -39,9 +39,8 @@ export async function formsAuditRunner(auditUrl, context) {
   const { site, log } = context;
   const rumAPIClient = RUMAPIClient.createFrom(context);
   const options = {
-    // RUM is keyed per hostname; a sub-path site's auditUrl carries a path
-    // (e.g. oklahoma.gov/omes) with no domainkey. Query by hostname, then scope
-    // the returned per-URL form vitals to the site's base path.
+    // RUM is keyed per hostname; a sub-path auditUrl (e.g. example.com/foo) has no
+    // domainkey. Query by hostname; form vitals are scoped to the base path below.
     domain: getRUMDomain(auditUrl),
     interval: FORMS_AUDIT_INTERVAL,
     granularity: 'hourly',
--- a/src/forms-opportunities/handler.js
+++ b/src/forms-opportunities/handler.js
@@ -21,7 +21,8 @@ import {
   getUrlsDataForAccessibilityAudit,
   sendMessageToMystiqueForGuidance,
 } from './utils.js';
-import { getScrapedDataForSiteId } from '../support/utils.js';
+import { getScrapedDataForSiteId, getRUMDomain } from '../support/utils.js';
+import { filterByAuditScope } from '../internal-links/subpath-filter.js';
 import createLowConversionOpportunities from './oppty-handlers/low-conversion-handler.js';
 import createLowNavigationOpportunities from './oppty-handlers/low-navigation-handler.js';
 import createLowViewsOpportunities from './oppty-handlers/low-views-handler.js';
@@ -35,16 +36,20 @@ const FORMS_OPPTY_QUERIES = [
 ];
 
 export async function formsAuditRunner(auditUrl, context) {
+  const { site, log } = context;
   const rumAPIClient = RUMAPIClient.createFrom(context);
   const options = {
-    domain: auditUrl,
+    // RUM is keyed per hostname; a sub-path site's auditUrl carries a path
+    // (e.g. oklahoma.gov/omes) with no domainkey. Query by hostname, then scope
+    // the returned per-URL form vitals to the site's base path.
+    domain: getRUMDomain(auditUrl),
     interval: FORMS_AUDIT_INTERVAL,
     granularity: 'hourly',
   };
 
   const queryResults = await rumAPIClient.queryMulti(FORMS_OPPTY_QUERIES, options);
   const auditResult = {
-    formVitals: queryResults['form-vitals'],
+    formVitals: filterByAuditScope(queryResults['form-vitals'], site?.getBaseURL?.(), {}, log),
     auditContext: {
       interval: FORMS_AUDIT_INTERVAL,
     },
--- a/src/cwv/cwv-audit-result.js
+++ b/src/cwv/cwv-audit-result.js
@@ -110,9 +110,8 @@ export async function buildCWVAuditResult(context) {
   const rumApiClient = RUMAPIClient.createFrom(context);
   const groupedURLs = site.getConfig().getGroupedURLs(Audit.AUDIT_TYPES.CWV);
   const options = {
-    // RUM is keyed per hostname; a sub-path site's auditUrl carries a path
-    // (e.g. oklahoma.gov/omes) with no domainkey. Query by hostname, then the
-    // per-URL results below are scoped to the site's base path.
+    // RUM is keyed per hostname; a sub-path auditUrl (e.g. example.com/foo) has no
+    // domainkey. Query by hostname; per-URL results are scoped to the base path below.
     domain: getRUMDomain(auditUrl),
     interval: INTERVAL,
     granularity: 'hourly',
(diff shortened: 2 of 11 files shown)


FentPams (approved):
✅ Approve — verified the sub-path RUM scoping end-to-end against the PR head.

- **CWV**: the now domain-wide RUM query is scoped back to the base path — `scopedCwvData = cwvData.filter((item) => item.type !== 'url' || isWithinAuditScope(item.url, baseURL))` in `cwv-audit-result.js` — so a sub-path site (`oklahoma.gov/omes`) doesn't ingest other agencies' high-traffic pages. Nice touch keeping non-`url` group aggregates.
- **forms / image-alt-text / internal-links** all scope their RUM results too; `form-vitals` items carry `.url`, so `filterByAuditScope` (default `urlProperty:'url'`) doesn't silently drop everything.
- `getRUMDomain` keeps `www` and falls back to the raw input on parse failure — consistent with the existing `getRUMUrl`. The www-vs-apex domainkey mismatch is pre-existing, not introduced here.

Clean and well-tested. LGTM.

src/forms-opportunities/handler.js
```diff
@@ -35,16 +36,19 @@ const FORMS_OPPTY_QUERIES = [
...
+    // domainkey. Query by hostname; form vitals are scoped to the base path below.
+    domain: getRUMDomain(auditUrl),
     interval: FORMS_AUDIT_INTERVAL,
     granularity: 'hourly',
   };
 
   const queryResults = await rumAPIClient.queryMulti(FORMS_OPPTY_QUERIES, options);
   const auditResult = {
-    formVitals: queryResults['form-vitals'],
+    formVitals: filterByAuditScope(queryResults['form-vitals'], site?.getBaseURL?.(), {}, log),
```
ramboz, line 51: **[LOW – test gap]** The hostname strip (`domain`) is now well tested, but this `formVitals` base-path filter still isn't exercised on a real sub-path site.

The new forms test calls `formsAuditRunner('https://example.com/foo', context, subpathSite)`, but the runner reads `site` from **`context`** (line 39: `const { site, log } = context;`) — not that 3rd argument — and the harness `context.site` is root-domain (`https://example.com`). So `filterByAuditScope` no-ops and the actual scoping/removal path goes unasserted.

Consider a case where `context.site.getBaseURL()` returns a sub-path URL and asserting that out-of-scope form-vitals are dropped (the CWV suite's `subpath-site base-path scoping` block is a good template).

---
*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

ramboz (approved):
## PR Review: fix(audits): query RUM by hostname for sub-path sites

### Summary
Correct, well-scoped bug fix. A new `getRUMDomain(url)` helper strips a resolved audit/final URL to its **hostname** so the RUM domainkey exchange stops 404-ing for sub-path sites, applied to **cwv**, **broken-internal-links**, and **forms**; **forms** and **image-alt-text** additionally scope their per-URL RUM results to the site base path via `filterByAuditScope`. I verified every call site has the URL/`site` in scope, `form-vitals` items expose a `url` property (so `filterByAuditScope` won't silently drop everything), and `context.site` is populated for `formsAuditRunner` (forms is a StepAudit, so the step context carries `site`). No regressions — sub-path sites previously failed these audits outright, so any scoped result set is strictly better. Low blast radius: one new symbol (`getRUMDomain`), three callers, no deletions.

> _I reviewed the full two-commit range. The second commit (`review fixes + scrub example refs`) landed while I was reviewing and resolved the two gaps I had against the first commit — it adds the `isWithinAuditScope` `${basePath}.html` landing-page widening (with an absolute+relative test and sibling-boundary safety), adds hostname assertions for the forms + internal-links RUM queries, and scrubs real customer references to `example.com/foo`._ 👍

### Change Surface
- 6 src files, 5 test files; +155/−33; 2 commits.
- New public symbol: `getRUMDomain` (`src/support/utils.js`) — unit-tested. No deleted/starred symbols.
- Surfaces: cwv, internal-links, forms, image-alt-text (RUM query `domain` + base-path scoping) and the shared `subpath-filter`.

### Correctness (verified)
- `getRUMDomain` → `new URL(prependSchema(url)).hostname` with a safe fallback to the input on parse failure. Correct for sub-path, no-op for root.
- `isWithinAuditScope` widening keeps `${basePath}` and `${basePath}.html` while still dropping siblings (`/foobar`, `/foo-x`) in both the absolute and relative branches — matches the new test.
- forms `filterByAuditScope(form-vitals, …)` is safe: form-vitals items carry `url`, and `context.site` is present (StepAudit wiring).
- image-alt-text: the RUM fallback already used a hostname-only domain via `getRUMUrl`; the added `filterByAuditScope` correctly scopes the domain-wide results and gracefully degrades to `includedURLs` if it empties the list.

### Blast Radius
- `getRUMDomain`: 3 callers, all introduced here; each checked. No unupdated callers, no dangling references, no new orphans.

### Test Coverage
- New: `getRUMDomain` unit tests; CWV hostname assertion; forms + internal-links hostname assertions; subpath-filter landing-page test. Good coverage of the fix.
- Minor gap (non-blocking): the forms **formVitals base-path filtering** itself isn't directly asserted — see the inline note on `handler.js:51`.

### Additional un-deferred call site (informational)
Your “Deferred call sites” list is a nice touch. One more path-carrying RUM `domain` isn't in it: `src/experimentation-opportunities/handler.js:166` (`experimentOpportunitiesAuditRunner` → `domain: auditUrl`) — same latent 404 for sub-path sites. (`src/image-alt-text/opportunityHandler.js:35` is fine — its `finalUrl` comes from `getRUMUrl`, already hostname-only.) Worth a follow-up ticket.

### Affected-Test Run
- Not executed locally — this isolated review worktree has no `node_modules`, and a multi-minute install wasn't warranted for a ~120-line change. All added tests are verified correct by inspection (e.g. the CWV test reuses the existing `subpathSite`/`fetchStub` harness that already reaches `.query`). Relying on CI for the suite.

### Security
- No issues. Pure URL→hostname parsing with a safe fallback; no secrets, injection, or new network/permission/telemetry surface. Agent-integration posture: not applicable to this diff.

### Verdict
**Approve.** No blockers. Optional follow-ups: (1) assert the forms formVitals sub-path filtering (inline), (2) track the extra `experimentation-opportunities` call site.

---
*Reviewed with ❤️ by [Claude Code](https://claude.ai) 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

src/forms-opportunities/handler.js
rpapani: Good catch — fixed in 22e12c63. Added a dedicated test where `context.site.getBaseURL()` returns a sub-path (`https://example.com/foo`) and `queryMulti` resolves mixed in/out-of-scope `form-vitals`; it asserts only the `/foo/*` entries survive (sibling `/other/contact` and boundary `/foobar` dropped). Also dropped the misleading (ignored) 3rd arg from the domain test.

PR #2882, opened 2026-08-12
fix(backlinks): scope broken-backlink results to the site sub-path

rpapani:
## What
Broken-backlink results from Semrush are only host-filtered (`isOnDifferentSubdomain`), never base-path filtered, so a **sub-path site** still receives whole-domain target URLs. The companion server-side `target_url LIKE '%host/path%'` filter (seo-client) is a coarse substring match — e.g. `%example.com/foo%` also matches `/foo-budget`.

## Fix
Add a directory-boundary-safe base-path filter on each broken backlink's target URL (`url_to`), reusing `isWithinAuditScope` from `internal-links/subpath-filter.js`, keyed off the effective base (`overrideBaseURL || baseURL`, matching the subdomain filter). No-op for root sites; the host-only subdomain filter is unchanged and runs alongside it.

Semrush `target_url` can be **scheme-less** (`example.com/foo/page`), which `isWithinAuditScope` treats as a relative path and silently drops even when in-scope. `url_to` is wrapped with `prependSchema` (already imported from `@adobe/spacecat-shared-utils`) — a no-op when a scheme is present, matching this file's existing `isOnDifferentSubdomain` / `normalizeUrl` convention.

## Composition with the seo-client PR
This composes with the seo-client change: the coarse server-side `LIKE` pre-filter is refined here to a precise directory boundary, and this now also handles scheme-less target URLs.

## Tests
- sub-path scoping keeps `/foo/*`, drops same-host siblings and the `/foo-budget` boundary path;
- honours the configured `overrideBaseURL` sub-path;
- no-op for root-domain sites;
- **scheme-less** in-scope `url_to` is kept while a scheme-less out-of-scope one is dropped.
- Suite green; lint clean.

## Context
Part of **SITES-49721** (multi-site-per-domain / sub-path onboarding).

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "@adobe/helix-shared-wrap": "2.0.2",
         "@adobe/helix-status": "10.1.5",
         "@adobe/helix-universal-logger": "3.0.30",
-        "@adobe/mysticat-shared-seo-client": "1.8.0",
+        "@adobe/mysticat-shared-seo-client": "1.8.1",
         "@adobe/spacecat-helix-content-sdk": "1.4.33",
         "@adobe/spacecat-shared-athena-client": "1.11.0",
         "@adobe/spacecat-shared-data-access": "4.19.0",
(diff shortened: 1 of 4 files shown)


FentPams (approved):
✅ Approve — `isWithinAuditScope` is genuinely directory-boundary-safe (`startsWith(basePath + '/')` + exact match + host compare), so `/omes` correctly excludes `/omes-budget`. This is the layer that enforces the real boundary on top of the coarse server-side `target_url LIKE` in spacecat-shared#1875 — good pairing; ship the two together.

Nit (pre-existing in `subpath-filter.js:60`, but newly relevant here since backlink target URLs are third-party-authored): the strict `parsedUrl.port !== parsedBaseURL.port` compare treats an explicit `:443`/`:80` on a target URL as a mismatch against an implicit-port base, which could drop a legitimately in-scope backlink. Worth normalizing default ports.

PR #2886, opened 2026-08-12
fix(audits): scope domain-keyed top pages + experimentation opportunities to site sub-path

rpapani:
## Problem
Sub-path sites (a `baseURL` that carries a path, e.g. `example.com/foo`) inherited **whole-domain** data in two remaining V1 audit paths, because RUM/top-pages are keyed per hostname:

1. **Top pages** — `getMergedAuditInputUrls` returned the domain's entire top-pages set, so audits built on it (structured-data, hreflang, toc, canonical, llm-blocked) scraped/audited pages outside the site's sub-path.
2. **Experimentation opportunities** — `experimentation-opportunities` returned whole-domain RUM opportunities, attributing other sub-paths' pages to the site.

Follow-up hardening for the sub-path work in #2873 / #2880 / #2881 / #2882.

## Fix
Scope to the site base path, reusing the existing directory-boundary-safe `isWithinAuditScope` / `filterByAuditScope` helper (host-normalized, `${basePath}.html` landing-aware). **No-op for root-domain sites.**

- **`src/utils/audit-input-urls.js`** — new opt-in `scopeTopPagesToBasePath` option (default **off**) that filters the domain-keyed top pages to the base path. Operator-**included** and **agentic** URLs are always left explicit/unfiltered. Enabled for the audits that source URLs solely from this helper: **structured-data, hreflang, toc, canonical, llm-blocked**. Left **off** for **broken-backlinks**, which does its own sub-path filtering and emits a precise "all top pages filtered out by audit scope" error — so its behavior is unchanged.
- **`src/experimentation-opportunities/handler.js`** — filter the whole-domain RUM opportunities by base path. `domain` here already resolves to a bare hostname via `wwwUrlResolver`, so no additional hostname strip is needed.

### structured-data
No structured-data-specific change: it sources every URL (and its GSC per-URL inspection list) from `getMergedAuditInputUrls`, so opting the helper in scopes it automatically. Verified by tracing `getIssuesFromGSC`, which iterates the already-scoped page list.

### Verified non-issues (traced, no change)
- `llmo-customer-analysis` `checkOptelData` — `getRUMUrl` already returns a bare hostname; it's a boolean pageviews existence gate (no per-URL leak).
- `experimentation-ess` and `metatags` / `product-metatags` `calculateProjectedTraffic` — `domain` comes from `wwwUrlResolver` (bare hostname), so the domainkey lookup does not 404 for sub-path sites.

## Tests
- New sub-path scoping test in `test/utils/audit-input-urls.test.js` (out-of-scope top pages dropped; included/agentic preserved).
- New `sub-path scoping` cases in `test/audits/experimentation-opportunities.test.js` (filtering + no-baseURL pass-through).
- Full suite green (9581 passing) with 100% coverage; all consumer suites (backlinks, structured-data, canonical, hreflang, toc, llm-blocked) pass unchanged.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/canonical/handler.js
+++ b/src/canonical/handler.js
@@ -89,6 +89,7 @@ export async function submitForScraping(context) {
     dataAccess,
     auditType,
     getAgenticUrls: () => Promise.resolve([]),
+    scopeTopPagesToBasePath: true,
     log,
   });
 
--- a/src/experimentation-opportunities/handler.js
+++ b/src/experimentation-opportunities/handler.js
@@ -17,6 +17,7 @@ import { isNonEmptyArray } from '@adobe/spacecat-shared-utils';
 import { AuditBuilder } from '../common/audit-builder.js';
 import { wwwUrlResolver } from '../common/index.js';
 import { getCountryCodeFromLang, parseCustomUrls } from '../utils/url-utils.js';
+import { filterByAuditScope } from '../internal-links/subpath-filter.js';
 import {
   getObjectFromKey,
 } from '../utils/s3-utils.js';
@@ -159,16 +160,22 @@ export async function processCustomUrls(customUrls, rumAPIClient, options, conte
  * @returns
  */
 export async function experimentOpportunitiesAuditRunner(auditUrl, context) {
-  const { log } = context;
+  const { log, site } = context;
 
   const rumAPIClient = RUMAPIClient.createFrom(context);
   const options = {
+    // auditUrl resolves via wwwUrlResolver (already a bare hostname), so no strip needed here.
     domain: auditUrl,
     interval: DAYS,
     granularity: 'hourly',
   };
   const queryResults = await rumAPIClient.queryMulti(OPPTY_QUERIES, options);
-  const experimentationOpportunities = Object.values(queryResults).flatMap((oppty) => oppty);
+  const rawOpportunities = Object.values(queryResults).flatMap((oppty) => oppty);
+  // Scope whole-domain RUM opportunities to the site sub-path (no-op for root-domain sites).
+  const baseURL = site?.getBaseURL?.();
+  const experimentationOpportunities = baseURL
+    ? filterByAuditScope(rawOpportunities, baseURL, { urlProperty: 'page' }, log)
+    : rawOpportunities;
   processRageClickOpportunities(experimentationOpportunities);
   log.debug(`Found ${experimentationOpportunities.length} experimentation opportunites for ${auditUrl}`);
 
--- a/src/headings/handler.js
+++ b/src/headings/handler.js
@@ -388,6 +388,7 @@ export async function headingsAuditRunner(baseURL, context, site) {
       auditType,
       getAgenticUrls: () => getTopAgenticUrlsFromAthena(site, context),
       topOrganicLimit: 200,
+      scopeTopPagesToBasePath: true,
       log,
     });
     const topPages = mergedInput.urls.map((url) => ({ url }));
--- a/src/hreflang/handler.js
+++ b/src/hreflang/handler.js
@@ -176,6 +176,7 @@ export async function hreflangAuditRunner(baseURL, context, site) {
       auditType,
       getAgenticUrls: () => Promise.resolve([]),
       topOrganicLimit: 200,
+      scopeTopPagesToBasePath: true,
       log,
     });
     const topPages = mergedInput.urls.map((url) => ({ url }));
--- a/src/internal-links/subpath-filter.js
+++ b/src/internal-links/subpath-filter.js
@@ -97,8 +97,9 @@ export function filterByAuditScope(items, baseURL, options = {}, log = console) 
   try {
     const baseURLWithSchema = prependSchema(baseURL);
     const parsedBaseURL = new URL(baseURLWithSchema);
-    const basePath = parsedBaseURL.pathname;
-    const hasBasePath = basePath && basePath !== '/';
+    // Trim trailing slash(es) to stay consistent with isWithinAuditScope.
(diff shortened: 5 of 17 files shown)


rpapani:
## Incorporated independent-review feedback

- **Ordering bug (correctness):** `getMergedAuditInputUrls` scoped top pages *after* the domain-wide `topOrganicLimit` slice, so a low-traffic sub-path (e.g. `oklahoma.gov/omes` within a much larger state site) could be truncated away before scoping ran — defeating the fix for `hreflang`/`toc` (which pass `topOrganicLimit: 200`). Now scopes **before** the slice, so the "top N" is computed within the sub-path. No-op for root-domain sites (identical output). New test `scopes to the sub-path before applying topOrganicLimit`.
- **Completeness:** opted in the six remaining `getMergedAuditInputUrls` consumers that still sourced whole-domain top pages — `metatags`, `product-metatags`, `headings`, `readability`, `summarization`, `semantic-value-visibility`. No-op for root sites; scopes for sub-path sites.

## Deferred (documented, not silently forgotten)

- **Misleading "No top pages found" error** in `structured-data`/`llm-blocked` when scoping empties the set for a legitimate sub-path site (pre-existing behavior, unchanged by this PR, but more reachable now). Fix is to adopt the `broken-backlinks` pattern that distinguishes "no data in DB" vs "all filtered out by scope". Tracking as a follow-up rather than expanding this PR's blast radius further.

592 tests pass across all affected + newly-opted-in suites; lint clean.

FentPams (approved):
Reviewed + independent second pass (repo cloned, call chain traced). Clean, low-risk — reuses the proven `filterByAuditScope`/`isWithinAuditScope` rather than new logic. Approving, with two non-blocking notes.

**Verified:**
- **Reorder is safe.** `defaultTopPagesToUrls` is a strict 1:1 `.map`, and a repo-wide grep confirms **no caller overrides `topPagesToUrls`** — so moving the `topOrganicLimit` slice to after the map (`map → [scope] → slice`) is byte-identical for every existing caller when `scopeTopPagesToBasePath` is off. No regression. Scope-before-limit is intentional (top-N within the sub-path), as the comment says.
- **Handler set is complete + correct.** Of 14 `getMergedAuditInputUrls` callers: 11 opted in (this diff), and 3 correctly left off because they do their own sub-path filtering — `backlinks` (broken-backlinks, the JSDoc example) + `internal-links` scrape-submission & opportunity-suggestions. None missed, none wrongly included.
- **Experimentation handler is safe.** `context.site` is reliably populated on the step-audit path (`StepAudit.run` builds `stepContext` with `site`); `urlProperty: 'page'` matches how RUM opportunity objects carry their URL elsewhere in the file; and `baseURL ?` + the helper's own root-domain no-op is double-safe (no behavior change for root-domain sites, exercised by the existing suite).
- **Helper correctness:** root-domain no-op, fail-open catch (a parse error returns all items, never wipes the set), trailing-slash guard against `/fr/`↔`/french`, host+port protocol-agnostic. No async/N+1/mutation concerns.

**Non-blocking notes:**
1. **[LOW, pre-existing — quick pre-merge check] Trailing-slash blast radius.** `isWithinAuditScope` builds `basePathWithSlash = ${basePath}/` straight off `parsedBaseURL.pathname` with no trailing-slash trim. If `Site.getBaseURL()` ever returns a sub-path *with* a trailing slash (`…/foo/`), that becomes `/foo//` and legit in-scope URLs (`/foo/bar`) get dropped. Not introduced by this PR, but it adds 11 new call sites so the blast radius grows — worth confirming `getBaseURL()` is normalized (no trailing slash) fleet-wide, or trimming inside the helper.
2. **[LOW] Test gap.** No explicit test that `scopeTopPagesToBasePath: false` (default) leaves an out-of-scope URL untouched on a *sub-path* site — a guard that would catch someone flipping the default to `true` later. One-liner follow-up.

Approving.

rpapani:
Thanks @FentPams — both notes addressed in 7bd417fc:

1. **Trailing-slash blast radius:** hardened `isWithinAuditScope` to trim trailing slash(es) from `basePath` (`pathname.replace(/\/+$/, '')`), so a `getBaseURL()` returning `.../foo/` no longer becomes `/foo//` and drops in-scope URLs. Added a test covering `bulk.com/uk/` (behaves like `bulk.com/uk`) and `bulk.com/` (root scope). This also independently resolves the same note raised on the earlier review pass.

2. **Test gap:** added a test asserting the default (`scopeTopPagesToBasePath` off) leaves an out-of-scope top page untouched on a sub-path site.

140 tests pass across subpath-filter + audit-input-urls + backlinks + internal-links; lint clean.

rpapani:
Second independent regression review — no regressions found (reorder, trailing-slash fix, and 11 opt-ins verified no-op for root-domain sites). Addressed both notes in ebe27023:

- **Empty-after-scope observability (Important):** `getMergedAuditInputUrls` now logs a distinct message when scoping empties a *previously non-empty* top-pages set (i.e. top pages exist but none under the sub-path — a newly-onboarded site or a lagging top-pages import), so it is distinguishable from a genuine "no data" import failure before a downstream empty-set throw. Added a test. Kept the existing throw behavior (fail-loud is correct — better than auditing the whole domain).
- **basePath drift (Minor):** aligned `filterByAuditScope` to trim trailing slash(es) exactly like `isWithinAuditScope`, so the two no longer express "has sub-path?" differently (and the debug log prints the normalized path).

Full suite green: 9584 passing (2 unrelated `cdn-detector` DNS-timeout flakes that pass standalone), 100% coverage on the touched files, lint clean, CI green.

PR #2891, opened 2026-08-14
fix(preflight): [SITES-42093] remove duplicate h1-count audit 

radhikagpt1208:
## Summary
- H1 issues were flagged by three overlapping Preflight checks — `h1-count` (dedicated), `headings` (`heading-multiple-h1`/`heading-missing-h1`), and `metatags` (`h1` tagName check) — causing authors to see the same finding duplicated in results.
- Removes the standalone `h1-count` audit from `src/preflight/handler.js` (constant, enablement flag, opportunity generation, structured logging entry), since `headings` already reports `missing-h1`/`multiple-h1` with equivalent (and more complete) detail.
- Updates tests/fixtures (`test/audits/preflight.test.js`, `test/fixtures/preflight/*`) and docs (`src/preflight/README.md`, `docs/preflight-response-schema.md`, `docs/specs/preflight-structured-audit-logging.md`) to drop `h1-count` references.
- The remaining `metatags` vs. `headings` H1 overlap is out of scope here (tracked separately, see linked [SITES-42024](https://jira.corp.adobe.com/browse/SITES-42024)).

## Test plan
- [x] `npm run test:spec -- test/audits/preflight.test.js` — 149 passing, 0 failing
- [x] `npm run lint` on touched files — clean

Jira: https://jira.corp.adobe.com/browse/SITES-42093


clotton (approved):
Verified this end-to-end:

- Confirmed `headings.js` (`heading-missing-h1` / `heading-multiple-h1` in `../headings/handler.js`) already covers both cases `h1-count` reported, with equivalent-or-better detail (issue title, description, explanation, AI suggestion for the missing-h1 case) — so this is a safe dedup, not a coverage loss.
- Grepped the full branch for leftover `h1-count`/`AUDIT_H1_COUNT`/`h1CountEnabled` references — none remain outside one intentional historical note in `docs/specs/preflight-structured-audit-logging.md`.
- Checked out the branch, `npm ci`, ran `test/audits/preflight.test.js` — 149 passing, 0 failing, matches the PR's test plan.
- `eslint` on the touched source file is clean.

Clean, well-scoped removal with matching test/doc updates. Approving.

src/preflight/handler.js
```diff
@@ -59,7 +59,6 @@ export const AUDIT_LINKS = 'links';
 export const AUDIT_METATAGS = 'metatags';
 export const AUDIT_BODY_SIZE = 'body-size';
 export const AUDIT_LOREM_IPSUM = 'lorem-ipsum';
```
GeezerPelletier, line 61: **Cross-repo dead reference (not blocking, but should be tracked):** `aem-sites-optimizer-preflight-mfe` still has `h1-count` wired as a first-class audit type — `AUDIT_NAME.H1_COUNT` in `src/types/audits.ts`, an entry in `auditCategoryRegistry.tsx`'s SEO `auditNames`, a dedicated localized title string in all 8 `resources/__localization__/*.json` files, and `docs/audits/h1-count.json`. None of that is a required-presence check (it's a lookup map keyed by audit name), so nothing breaks — but it all becomes permanently dead once this audit stops being emitted, and neither this PR nor the linked SITES-42093/SITES-42024 (both `Franklin - Spacecat` only) tracks cleaning it up. Worth a follow-up ticket so the MFE-side registration doesn't just quietly rot.

src/preflight/handler.js
```diff
@@ -71,7 +70,6 @@ const AVAILABLE_CHECKS = [
   AUDIT_METATAGS,
   AUDIT_BODY_SIZE,
   AUDIT_LOREM_IPSUM,
```
GeezerPelletier, line 72: **Please confirm before merge:** each check's enablement is resolved independently here (`isAuditEnabledForSite(`${audit}-preflight`, ...)` per audit name). The test fixtures always toggle `h1-count-preflight` and `headings-preflight` together, implying they're normally set as a pair at the product default — but if any live site's Configuration has `h1-count-preflight` enabled **without** `headings-preflight`, that site silently loses all H1 checking after this merge (not just the intended dedup). A quick Configuration query for that combination before shipping would close this out.

GeezerPelletier (changes requested):
Clean, well-tested removal within this repo — the dead branches (h1CountEnabled, the DOM-block push/finally arrays, the whole detection block) are fully excised with no leftover references, docs are updated in the same PR, and the test/fixture updates match. No correctness bugs in the diff itself.

Before merging I'd like two things checked/addressed, since both are real consequences of this change that aren't covered by this PR or the linked tickets (SITES-42093/SITES-42024 are both `Franklin - Spacecat` only — neither mentions the MFE side):

1. **Cross-repo dead reference in `aem-sites-optimizer-preflight-mfe`** — that repo has `h1-count` registered as a first-class, independent audit type: `AUDIT_NAME.H1_COUNT` (`src/types/audits.ts`), its own category-registry entry (`auditCategoryRegistry.tsx`), a dedicated localized title in all 8 locale files (`resources/__localization__/*.json`), and its own mock/doc fixture (`docs/audits/h1-count.json`). It's a graceful no-op once this audit stops being emitted (the registry is a lookup map, not a required-presence check, and existing tests confirmed it won't crash) — but all of that becomes permanently dead weight with nothing tracking its removal. Worth a follow-up ticket (or a note in SITES-42093) so it doesn't get forgotten.
2. **Possible silent coverage loss via per-site Configuration** — `enabledChecks` is resolved per check name independently (`isAuditEnabledForSite('h1-count-preflight', ...)` vs `isAuditEnabledForSite('headings-preflight', ...)`). If any live site's Configuration has `h1-count-preflight` enabled without `headings-preflight` also enabled (rather than both toggled together at the product default, as the test fixtures assume), that site now silently loses **all** H1 checking after this merge, not just the intended dedup. Worth a quick prod Configuration query to confirm no such site exists before this ships.

Marking as needs-updates for #2 in particular — it's cheap to verify and would be a quiet regression for the wrong site if skipped.

src/preflight/handler.js
radhikagpt1208: MFE clean-up: https://github.com/OneAdobe/aem-sites-optimizer-preflight-mfe/pull/363

src/preflight/handler.js
radhikagpt1208: Confirmed from the prod config. There is no live site that has `h1-count-preflight` enabled without `headings-preflight`, since both are enabled by default so no site will silently lose H1 checks after this merge.
```

		"h1-count-preflight": {
			"disabled": {
				"orgs": [],
				"sites": []
			},
			"enabled": {
				"orgs": [],
				"sites": []
			},
			"enabledByDefault": true,
			"productCodes": [
				"ASO"
			]
		},
		"headings-preflight": {
			"disabled": {
				"orgs": [],
				"sites": []
			},
			"enabled": {
				"orgs": [],
				"sites": []
			},
			"enabledByDefault": true,
			"productCodes": [
				"ASO"
			]
		},
```

radhikagpt1208:
Closing as part of the team's decision to defer V1 and put full effort into V2. V2 will only carry the Metatags audit, so resolving the H1-Count/Headings/Metatags duplicacy in V1 wouldn't carry forward — not an investment we're making in V1 at this point.

PR #2894, opened 2026-08-15
fix(internal-links)[v1]: parse srcset per spec so Scene7 image URLs don't create phantom broken links (SITES-49919)

rpapani:
## Problem

The **broken-internal-links** audit (v1) surfaces false-positive broken links whose target (`url_to`) is a mangled Scene7/AEM image-rendition fragment — e.g. `…/food-security-update/0&resMode=sharp&fmt=webp`. These aren't real hyperlinks; the underlying images return **200**, and the fragment is **never present in the page HTML**.

## Root cause

`resolveUrlCandidates` (`src/internal-links/crawl-detection.js`) split a page's `srcset` on a raw comma:

```js
String(rawValue).split(',')...  // BUG
```

Scene7 image URLs contain a comma **inside the query string** — the quality param `qlt=85,0`. Real value on the page:

```
srcset="https://worldbank.scene7.com/is/image/worldbankprod/WB_flags_716x402?wid=780&hei=439&qlt=85,0&resMode=sharp"
```

The naive split broke that single URL into:
- `…/WB_flags_716x402?wid=780&hei=439&qlt=85` (external host → correctly dropped), and
- `0&resMode=sharp` (a **relative** URL).

`0&resMode=sharp` resolved against the page's parent dir → `…/en/topic/agriculture/brief/0&resMode=sharp`, passed the internal-host check, validated as a genuine 404, and was surfaced as a broken internal link. Per the HTML `srcset` spec, a URL is a run of non-whitespace characters (commas included) — the comma-split is simply wrong.

## Fix

Replace `.split(',')` with `splitSrcsetCandidates`, a spec-correct parser: collect the URL as a run of non-whitespace chars (commas preserved), then skip an optional descriptor up to the next comma; commas separate candidates only at candidate boundaries. The Scene7 URL now stays intact (single candidate → dropped as external host) and **no orphan fragment is produced**.

## Scope: v1 only

`resolveUrlCandidates` lives in `spacecat-audit-worker`'s crawl detector — the **v1** broken-internal-links path. **This fix is v1 only.** v2 broken-internal-links runs entirely on Mystique (no `spacecat-audit-worker` code) and is out of scope here (and may not be GA yet).

## Trace (worldbank.org prod — SITES-49919)

Input `srcset` `…qlt=85,0&resMode=sharp`:
- **Before:** `split(',')` → `[…qlt=85, 0&resMode=sharp]` → `0&resMode=sharp` resolves to `…/brief/0&resMode=sharp` → internal host → 404 → **phantom broken link** (6 such NEW suggestions on the site).
- **After:** `splitSrcsetCandidates` → `[…qlt=85,0&resMode=sharp]` (one URL) → external `scene7.com` host → dropped → **no suggestion**.

This eliminates all 6 false positives on the site (opportunity `a588a8be-021c-467a-b19d-bff3a7ce68fd`), verified: the real Scene7 images are 200 and the fragments were absent from the page HTML.

## Tests

- New `splitSrcsetCandidates` unit suite: single URL with a query comma stays intact; no `0&resMode=sharp` orphan; multi-candidate split + descriptor strip; commas-in-URL preserved alongside real splits; trailing/multiple commas; empty/whitespace-only input; leading whitespace/commas.
- Full suite: **9663 passing**, `crawl-detection.js` **100%** coverage (lines/branches/functions), lint clean.

Fixes SITES-49919.

### DIFF
--- a/src/internal-links/crawl-detection.js
+++ b/src/internal-links/crawl-detection.js
@@ -166,12 +166,64 @@ function extractMetaRefreshUrl(content = '') {
   return match?.[1]?.trim() || '';
 }
 
+/**
+ * Splits an HTML `srcset` attribute value into its candidate URLs following the
+ * WHATWG rules: a candidate URL is a run of non-whitespace characters (which MAY
+ * contain commas — e.g. Scene7's `qlt=85,0` quality parameter), optionally
+ * followed by whitespace and a descriptor, with candidates separated by commas.
+ * A naive `split(',')` wrongly breaks URLs whose query string contains a comma,
+ * orphaning fragments like `0&resMode=sharp` that then resolve against the page
+ * into phantom internal links. Collecting the URL as a non-whitespace run avoids
+ * that.
+ *
+ * @param {string} rawValue - The raw `srcset` attribute value.
+ * @returns {string[]} Candidate URL strings, descriptors stripped.
+ */
+export function splitSrcsetCandidates(rawValue) {
+  const input = String(rawValue);
+  const { length } = input;
+  const isWhitespace = (ch) => /\s/.test(ch);
+  const urls = [];
+  let pos = 0;
+
+  while (pos < length) {
+    // Skip whitespace and the commas that separate candidates.
+    while (pos < length && (isWhitespace(input[pos]) || input[pos] === ',')) {
+      pos += 1;
+    }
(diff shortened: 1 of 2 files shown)


mikeweis (approved):
LGTM 👍

PR #2897, opened 2026-08-17
docs: add ADR for dedicated preflight-to-Mystique SQS queue

radhikagpt1208:
## Summary

Adds an ADR proposing a dedicated SQS queue + Mystique consumer lane for Preflight, so interactive Preflight runs stop starving behind bulk audit traffic on the shared `spacecat-to-mystique` queue.

- **Incident:** [SITES-49801](https://jira.corp.adobe.com/browse/SITES-49801) — Preflight suggest jobs hung 30-60+ min and timed out client-side (`AuditTimeoutError`), while backend jobs eventually completed. Traced to consumer-pickup latency on Mystique's shared queue, not LLM compute (confirmed via Langfuse: ~14s/paragraph, healthy).
- **Decision:** isolate Preflight's Mystique traffic across three repos:
  1. `spacecat-infrastructure` — new `spacecat-to-mystique-preflight` queue + DLQ
  2. `spacecat-audit-worker` — `sendPreflightMessages` routes to the new queue for `mode === 'preflight'`
  3. `mystique` — a second independent consumer/thread-pool, modeled on the existing `BpEventConsumer` pattern
- Alternatives considered and rejected/deferred: bumping shared worker concurrency, batching-only, full pod-level isolation, migrating V1 readability to the V2 direct-HTTP path.

This PR is docs-only (ADR). Implementation PRs for infra/audit-worker/mystique will follow once the team reviews.

## Test plan

- [ ] N/A — documentation only, no code changes
- [ ] Team review of the ADR (posted to `#cq-dev` thread) before implementation starts

### DIFF
--- a/docs/decisions/004-dedicated-preflight-mystique-queue.md
+++ b/docs/decisions/004-dedicated-preflight-mystique-queue.md
@@ -1,0 +1,167 @@
+# 004 — Dedicated SQS Queue + Consumer Pool for Preflight-to-Mystique Traffic
+
+- **Status:** Proposed
+- **Date:** 2026-08-17
+- **Repos affected:** spacecat-infrastructure, spacecat-audit-worker, mystique
+
+## Context
+
+Preflight's readability "suggest" step sends one SQS message per flagged
+paragraph to Mystique on the shared `spacecat-to-mystique` queue, and the
+job (`readability-suggest` guidance handler) only completes once **all**
+per-paragraph answers return — there is no timeout or partial-completion
+path.
+
+In the incident that prompted this ADR (SITES-49801), a customer
+(Dover, via CS) ran Preflight on `motor-bearing-protection.html`
+(`inpro-seal.com`) and the suggest step spun for 30+ minutes before the
+MFE's `AuditTimeoutError` fired (`MAX_POLLING_TIME = 1800s`). The job
+actually completed ~62 minutes in — long after the UI had stopped
+listening. Traced root cause:
+
+1. **`spacecat-to-mystique` is a single shared queue** for ~40 audit
+   types' outbound Mystique traffic, bulk/scheduled and interactive
+   Preflight alike. `MessageGroupId`-based fairness in audit-worker's
+   `src/support/sqs.js` is a no-op today because the queue is a
+   *standard*, not FIFO, queue.
+2. On the Mystique side, the consumer is a single in-process
+   `ThreadPoolExecutor` (`TASK_WORKERS=2`, `MAX_TASK_QUEUE_SIZE=4`) shared
+   across all message types. `is_queue_at_capacity()` halts **all**
+   receiving — interactive included — once more than 2 opportunities are
+   already tracked. No heartbeat/visibility-extension happens while a
+   message waits for a free worker; only after a worker starts does it
+   extend the visibility timeout. Combined with the default 3600s
+   visibility timeout, a message can sit invisible for up to an hour
+   before redelivery.
+3. Confirmed via Langfuse that the AI compute itself is fast and healthy
+   (~14s/paragraph, no retries, no errors) — the entire delay is
+   consumer pickup/scheduling latency, not model compute. 7-day queue
+   stats show this isn't a one-off: oldest-message age p50 15.4h, p90
+   52.1h, backlog spikes to ~13,000, with two other slow-run traceIds
+   (`bed87e2d…`, `c63fe808…`) showing the same pattern.
+4. Preflight's all-or-nothing completion makes this worse than for bulk
+   audits: it just takes one straggler paragraph, caught behind a bulk
+   burst, to hold an entire user-facing run hostage.
+
+Because Preflight is an on-demand feature with a user watching the
+screen, and Walmart is ramping up to ~300 authors on legacy V1 Preflight
+imminently, we need to decouple Preflight's Mystique traffic from bulk
+audit traffic rather than just tuning shared capacity.
+
+## Decision
+
+Introduce an isolated request lane for Preflight's SQS traffic to
+Mystique, spanning three repos:
+
+1. **spacecat-infrastructure** — provision a new standard SQS queue
+   `spacecat-to-mystique-preflight` (`modules/sqs/queues.tf`), mirroring
+   the existing `spacecat_queue_spacecat_to_mystique` resource, with its
+   own DLQ so preflight-specific backlog/age metrics are distinguishable
+   from bulk's. Output the queue URL for downstream env wiring.
+2. **spacecat-audit-worker** — in
+   `src/readability/shared/async-mystique.js`, `sendPreflightMessages`
+   branches on `mode === 'preflight'` to publish to a new
+   `env.QUEUE_SPACECAT_TO_MYSTIQUE_PREFLIGHT` instead of
+   `env.QUEUE_SPACECAT_TO_MYSTIQUE`. `sendOpportunityBatch` (bulk path)
+   is untouched. Scope is intentionally limited to readability-preflight
+   — the proven offender — not all ~40 handlers that currently share
+   `QUEUE_SPACECAT_TO_MYSTIQUE`.
+3. **mystique** — add a second, independent consumer modeled directly
+   on the existing `BpEventConsumer` precedent
+   (`app/services/claims/bp_event_consumer.py`, which already mirrors
+   `opportunity_service.py`'s transport/poll shape with its own queue
+   URL and its own capacity gate). The new consumer reads a new
+   `SQS_SPACECAT_TO_MYSTIQUE_PREFLIGHT_QUEUE_URL`, owns its own
+   `TaskManager` sized by its own env vars (independent of the bulk
+   pool's `TASK_WORKERS`/`MAX_TASK_QUEUE_SIZE`), and reuses the existing
+   `guidance:readability` task/routing logic unchanged (routing is
+   already keyed by message `type`/`mode`, not by queue). Started/stopped
+   in `server.py`'s lifespan alongside `opportunity_service` and
+   `bp_event_consumer`; no-ops when the env var is unset, per existing
+   convention.
+
+## Alternatives Considered
+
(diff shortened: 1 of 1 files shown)


johobot (commented):
I fact checked the technical claims in this ADR against the current code in spacecat-audit-worker, mystique, and spacecat-infrastructure. It's a well researched ADR and most of the details hold up, but I left five inline comments on claims that are inaccurate in ways that could mislead whoever implements this.

One more thing worth raising here since it's not tied to a single line: the ADR's stated reason for needing a new queue is that AWS MessageGroupId based fairness is a no-op on a standard, non-FIFO queue. That's not correct. It's a real AWS feature called SQS Fair Queues that works specifically on standard queues, and audit-worker already has the code for it from PR #2001, just not turned on in infrastructure yet. If that's right, finishing that rollout could solve the fairness problem in the current repo instead of building a second queue and consumer across three repos. Worth a quick gut check with the author before this merges.

docs/decisions/004-dedicated-preflight-mystique-queue.md
```diff
@@ -0,0 +1,167 @@
...
+listening. Traced root cause:
+
+1. **`spacecat-to-mystique` is a single shared queue** for ~40 audit
+   types' outbound Mystique traffic, bulk/scheduled and interactive
+   Preflight alike. `MessageGroupId`-based fairness in audit-worker's
+   `src/support/sqs.js` is a no-op today because the queue is a
+   *standard*, not FIFO, queue.
+2. On the Mystique side, the consumer is a single in-process
+   `ThreadPoolExecutor` (`TASK_WORKERS=2`, `MAX_TASK_QUEUE_SIZE=4`) shared
+   across all message types. `is_queue_at_capacity()` halts **all**
```
johobot: `is_queue_at_capacity()` only gates the SQS poll loop (`opportunity_service.py:647`). Mystique's interactive/HTTP path never checks it. For Preflight this doesn't change the conclusion since Preflight traffic goes through the gated path anyway, but the phrase "halts all receiving" overstates what the gate actually covers, and could mislead someone designing back pressure for the new consumer.

docs/decisions/004-dedicated-preflight-mystique-queue.md
```diff
@@ -0,0 +1,167 @@
...
+
+## Decision
+
+Introduce an isolated request lane for Preflight's SQS traffic to
+Mystique, spanning three repos:
+
+1. **spacecat-infrastructure** — provision a new standard SQS queue
+   `spacecat-to-mystique-preflight` (`modules/sqs/queues.tf`), mirroring
+   the existing `spacecat_queue_spacecat_to_mystique` resource, with its
+   own DLQ so preflight-specific backlog/age metrics are distinguishable
```
johobot: The existing `spacecat-to-mystique` queue doesn't actually have its own DLQ. It shares one generic dead letter queue with 13 other queues (`spacecat-infrastructure/modules/sqs/queues.tf:277-292` and `:423`). It also only outputs an ARN today, not a URL. If the plan is for the new queue to have its own DLQ and a URL output, that's a fine choice, but it should be called out as a deliberate change from the existing pattern rather than described as mirroring it.

docs/decisions/004-dedicated-preflight-mystique-queue.md
```diff
@@ -0,0 +1,167 @@
...
+   from bulk's. Output the queue URL for downstream env wiring.
+2. **spacecat-audit-worker** — in
+   `src/readability/shared/async-mystique.js`, `sendPreflightMessages`
+   branches on `mode === 'preflight'` to publish to a new
```
johobot: This points at the wrong function. `sendPreflightMessages` doesn't have a `mode` parameter and doesn't branch on it. The `mode === 'preflight'` check actually lives one level up, in the caller `sendReadabilityToMystique` (`async-mystique.js:221`), which is what decides whether to call `sendPreflightMessages` or `sendOpportunityBatch`. Both of those currently send to the same hardcoded queue. Worth updating this section so the implementer knows to change the queue selection in the caller, not add a new branch inside `sendPreflightMessages`.

docs/decisions/004-dedicated-preflight-mystique-queue.md
```diff
@@ -0,0 +1,167 @@
...
+3. **mystique** — add a second, independent consumer modeled directly
+   on the existing `BpEventConsumer` precedent
+   (`app/services/claims/bp_event_consumer.py`, which already mirrors
+   `opportunity_service.py`'s transport/poll shape with its own queue
+   URL and its own capacity gate). The new consumer reads a new
+   `SQS_SPACECAT_TO_MYSTIQUE_PREFLIGHT_QUEUE_URL`, owns its own
+   `TaskManager` sized by its own env vars (independent of the bulk
+   pool's `TASK_WORKERS`/`MAX_TASK_QUEUE_SIZE`), and reuses the existing
+   `guidance:readability` task/routing logic unchanged (routing is
+   already keyed by message `type`/`mode`, not by queue). Started/stopped
```
johobot: Small correction: routing in mystique is keyed by `type` only. `mode` is a real field on the message, but it's never used in any routing or dispatch logic. Suggest dropping mode from this description so it doesn't imply mode-based isolation that doesn't exist.

docs/decisions/004-dedicated-preflight-mystique-queue.md
```diff
@@ -0,0 +1,167 @@
...
+  (Preflight V2 already bypasses SQS/Mystique-queue entirely via
+  in-process `PreflightExecutor`). Rejected for now: not all legacy V1
+  crews (including `guidance:readability`) are ported to V2 yet, and the
+  migration timeline doesn't fit before the Walmart rollout. Revisit
+  once readability's V2 migration lands — this queue may become
+  obsolete at that point.
+
+## Consequences
+
+- New env vars needed across three repos:
```
johobot, line 191: This section is missing a step for audit-worker. The existing `QUEUE_SPACECAT_TO_MYSTIQUE` env var reaches the Lambda through a CloudFormation Parameter declared in `template.yml` (lines 39 and 121-123). The new `_PREFLIGHT` variant will need the same treatment there, not just a code change. Separately, nothing in the infrastructure repo actually sets this parameter's value today (the IAM policies reference the queue by a hardcoded ARN string, not a Terraform output), so it would help to spell out how the new queue's URL is meant to get from Terraform into the Lambda's environment.

docs/decisions/004-dedicated-preflight-mystique-queue.md
radhikagpt1208: Confirmed — verified directly in mystique. `is_queue_at_capacity()` is only referenced at `opportunity_service.py:647`, gating the SQS poll loop. Three HTTP routes (`meta_tag_benchmark_routes.py`, `org_detector_routes.py`, `variation_routes.py`) call `task_manager.add_task()` directly and bypass it entirely. Fixed in 9b4a4a1 — scoped the claim to the SQS-received path and flagged that it doesn't gate `TaskManager` intake as a whole.

docs/decisions/004-dedicated-preflight-mystique-queue.md
radhikagpt1208: Confirmed — `spacecat-to-mystique` shares the single `spacecat_queue_dead_letter` resource with 9 other queues (`modules/sqs/queues.tf`), and only outputs an ARN (`spacecat_queue_spacecat_to_mystique_arn`, `outputs.tf:73-74`); only one queue in the whole module outputs a URL today. Fixed in 9b4a4a1 — called this out as a deliberate deviation from the existing pattern rather than a mirror of it.

docs/decisions/004-dedicated-preflight-mystique-queue.md
radhikagpt1208: Confirmed — `sendPreflightMessages` has no `mode` parameter. The `isPreflight = mode === 'preflight'` branch lives in the caller `sendReadabilityToMystique` (`async-mystique.js:221`), which decides between `sendPreflightMessages` and `sendOpportunityBatch`; both currently hardcode `env.QUEUE_SPACECAT_TO_MYSTIQUE`. Fixed in 9b4a4a1 — Decision #2 now points at the actual branch location and describes threading the queue URL down from there instead of adding a branch inside `sendPreflightMessages`. Good catch, thanks — this one would have sent an implementer to the wrong place.

docs/decisions/004-dedicated-preflight-mystique-queue.md
radhikagpt1208: Confirmed — `_route_message`'s own docstring says "Route SQS message based on type field," and it only reads `body_json.get("type", "")`. `mode` is passed through as data (`opportunity_service.py:1411`) but never read for dispatch. Dropped `mode` from the routing description in 9b4a4a1.

docs/decisions/004-dedicated-preflight-mystique-queue.md
radhikagpt1208: Confirmed on both counts — `QUEUE_SPACECAT_TO_MYSTIQUE` reaches the Lambda via the CFN Parameter at `template.yml:39,121-123`, and nothing in spacecat-infrastructure sets that parameter's value (it's Helix Deploy params, outside all four repos this ADR touches). Added the `_PREFLIGHT` parameter touch point to Decision #1 in 9b4a4a1. Also found a bonus while checking: IAM likely needs no change at all — `modules/iam/policies.tf:460` already grants access via a `spacecat-*` wildcard ARN pattern, which already covers a queue named `spacecat-to-mystique-preflight`. Noted that too.

radhikagpt1208:
Closing as part of the team's decision to defer V1 and put full effort into V2. This parallel queue was a V1-side stopgap; since we're not investing further in V1, it's not worth carrying forward.
More details in this [thread](https://cq-dev.slack.com/archives/C0A91S5UKRC/p1787122201668449?thread_ts=1786561838.978239&cid=C0A91S5UKRC).

cc: @bhellema

PR #2898, opened 2026-08-17
fix(internal-links): scope-guard OUTDATED against crawl coverage gaps (SITES-49911)

alinarublea:
## Related Issues

Fixes SITES-49911 (child of SITES-49909 — the "verify-before-resolve, scope-guarded" thread).

## Problem

Broken-internal-links marked suggestions `OUTDATED` whenever a link pair was absent from the newest crawl, **without confirming the source page was actually re-examined this run**. A page that simply fell outside the run's crawl scope (or a run that scraped nothing) was therefore mistaken for a genuine fix — the "absence ≠ resolution" bug the ESE suggestion-status review flagged.

## Changes

- **Generalized coverage guard** — `handleOutdatedSuggestions` gains an optional `isWithinCoverage(data)` predicate that subsumes the single-URL `scrapedUrlsSet`. No-op when neither is supplied, so all existing callers are unchanged.
- **BIL sync path** — a broken-link pair is only eligible for `OUTDATED` when its source page (`urlFrom`) was actually crawled this run. The crawl-coverage set is reconstructed from the `scrape-result-paths` manifest (`loadScrapeResultPaths`). A non-empty manifest ⟺ a crawl run; RUM/LinkChecker-only runs (no manifest) keep prior absence-only behavior.
- **RESOLVED path (zero broken links) coverage-gated** — on a crawl run, outdate only crawled suggestions and flip the opportunity to `RESOLVED` **only when every non-frozen suggestion was crawled**; otherwise hold it open (unconfirmed links aren't treated as fixed). RUM-only runs retain prior resolve+outdate-all behavior. SKIPPED/REJECTED still protected (SITES-44646).
- **Suggestion-side subpath enforcement** — final `urlsSuggested` are hard-filtered through `isWithinAuditScope` in both the AI and Bright Data paths, so a broken `/uk/` link is never "fixed" with an out-of-subpath (`/us/`) URL. No-op on non-subpath sites.

Coverage is gated on `urlFrom` (the crawled source page) — the only signal reconstructable, and the correct granularity for a crawl (checked pairs aren't persisted; only broken pairs + crawled pages are).

## Testing

- +9 focused unit tests (sync guard, coverage-gated RESOLVED incl. partial/empty/throwing-manifest cases, subpath drop in the Bright Data path).
- New lines/branches fully covered; existing behavior preserved (RUM-only runs, edited/frozen suggestions).
- `npm test` green for the changed modules; full internal-links suite +9 passing, no regressions.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/src/internal-links/opportunity-suggestions.js
+++ b/src/internal-links/opportunity-suggestions.js
@@ -15,6 +15,49 @@ import { createInternalLinksConfigResolver } from './config.js';
 import { createInternalLinksStepLogger } from './logging.js';
 import { warnOnInvalidSuggestionData } from '../utils/data-access.js';
 import { getMergedAuditInputUrls } from '../utils/audit-input-urls.js';
+import { loadScrapeResultPaths } from './batch-state.js';
+import { normalizeComparableUrl } from './link-key.js';
+import { isWithinAuditScope } from './subpath-filter.js';
+
+/**
+ * Builds the crawl-coverage set for this run: the normalized set of source pages
+ * (urlFrom) that were part of this audit's crawl scope, reconstructed from the
+ * scrape-result-paths manifest. A non-empty set means this was a crawl run, so the
+ * OUTDATED sweep can trust "absence == fix" only for pages in the set (SITES-49911).
+ * Returns null when no manifest exists (RUM/LinkChecker-only runs), so callers fall
+ * back to prior absence-only behavior.
+ *
+ * @param {string} auditId - The audit ID.
+ * @param {Object} loaderContext - Context carrying s3Client/env for the S3 read.
+ * @param {Object} log - Logger instance.
+ * @returns {Promise<Set<string>|null>}
+ */
+async function buildCrawledUrlFromSet(auditId, loaderContext, log) {
+  // No S3 client (e.g. some test contexts / degenerate runs) — do not attempt the
+  // manifest read. Returning null keeps the caller on prior absence-only behavior
+  // without emitting a spurious error log from the S3 layer.
+  if (!loaderContext?.s3Client) {
+    log.debug('Scope guard: no s3Client available; outdating falls back to absence-only');
+    return null;
+  }
+  try {
+    const scrapeResultPaths = await loadScrapeResultPaths(auditId, loaderContext);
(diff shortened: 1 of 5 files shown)


MysticatBot (approved):
Hey @alinarublea,

**Verdict: Approve** - well-scoped bug fix with comprehensive tests and safe fallback behavior.
**Complexity: MEDIUM** - 5 files, coverage-gating logic across three modules.
_Changes: Adds crawl-coverage gating to the broken-internal-links OUTDATED sweep so suggestions are only marked resolved when their source page was actually re-crawled, fixing false-positive outdating on partial crawls (5 files)._

<details>

- suggestion: When all AI suggestions are out-of-scope, `getDefaultSuggestedUrls(site, {})` may produce a generic fallback (e.g. site root) that is technically in-scope but semantically unhelpful to the user. Consider whether dropping the suggestion entirely would be a better UX than offering a placeholder URL. - `src/internal-links/suggestions-generator.js:322`
- suggestion: `buildCrawledUrlFromSet` catches all errors and falls back to absence-only (pre-fix) behavior. This is a safe default for availability, but a persistent S3 permission error would silently disable the scope guard. Consider distinguishing "manifest not found" (expected, info-level) from unexpected errors (warn + metric) so silent degradation is observable. - `src/internal-links/opportunity-suggestions.js:47`
- nit: `isWithinAuditScope` is applied in two places with different empty-result handling: in `opportunity-suggestions.js` an empty result drops the suggestion entirely, while in `suggestions-generator.js` it falls back to `getDefaultSuggestedUrls`. If both paths process the same broken link, the inconsistency could be confusing. - `src/internal-links/opportunity-suggestions.js:329`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 22s | <b>Cost:</b> $6.48 | <b>Commit:</b> <code>97bfd8e6f6261d9835d4aba22bee7819f5bed6d4</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2901, opened 2026-08-17
fix(toc): drop subdomain and cross-domain suggestion URLs (LLMO-6965)

anagarwa:
## Problem

The TOC content-opportunity audit generates suggestions containing URLs from:
1. **Subdomains** of the registered site (e.g. `jobs.wkkellogg.com` when the site is `wkkellogg.com`)
2. **URLs that redirect to subdomains** (e.g. `wkkellogg.com/careers` → `jobs.wkkellogg.com`)
3. **Completely different/foreign domains** (e.g. `shop.middleby.com` getting a suggestion for `middleby.com`)

These suggestions show as "edge deployed" in the LLMO UI, implying optimization was applied — but OAE does not follow redirects, so optimization is never actually applied to these URLs.

Fixes [LLMO-6965](https://jira.corp.adobe.com/browse/LLMO-6965). Scoped to **TOC only** — Summarization and Readability have the same class of bug tracked separately in [LLMO-7069](https://jira.corp.adobe.com/browse/LLMO-7069) and [LLMO-7070](https://jira.corp.adobe.com/browse/LLMO-7070), and are not touched by this PR.

## Fix

Adds a plain hostname-scope guard (`isOnDifferentDomain`, mirroring `isOnDifferentSubdomain` already used in `src/backlinks/handler.js`) at three points in the TOC pipeline:

1. **`getTocInputUrls`** — merged input URLs (customer target URLs, included URLs, Athena agentic URLs, SEO top pages) are filtered to the site's own hostname before anything is scraped.
2. **`validatePageTocFromScrapeJson`** — pages whose scrape resolved (post-redirect) to an out-of-scope hostname are skipped entirely, rather than suggesting either the pre-redirect or off-domain URL.
3. **`generateSuggestions`** — a final guard drops any out-of-scope URL immediately before a suggestion is built, as defense in depth.

Note: deliberately does **not** reuse `isWithinAuditScope`/`filterByAuditScope` from `src/internal-links/subpath-filter.js` — that utility only checks hostname when the site's base URL itself has a subpath, so it would silently let subdomain/cross-domain URLs through for root-domain sites (which is every site affected by this bug: wkkellogg.com, shop.middleby.com).

## Scope

Changes confined to `src/toc/handler.js` and its tests — no shared utilities changed, no impact on other audit types. Does not touch any existing suggestions; remediation of already-published bad TOC suggestions (marking them Outdated) is a separate follow-up.

## Testing

- 6 new test cases covering: input-URL filtering, redirect-based skip, in-domain redirect still processed normally, suggestion-level guard, and malformed-URL fallback behavior.
- 236 tests passing in `test/audits/toc.test.js` + related TOC test files.
- 100% line/branch/statement coverage on `src/toc/*` (repo-enforced).
- Lint clean.

### DIFF
--- a/src/toc/handler.js
+++ b/src/toc/handler.js
@@ -10,7 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
-import { getPrompt } from '@adobe/spacecat-shared-utils';
+import { getPrompt, prependSchema, stripWWW } from '@adobe/spacecat-shared-utils';
 import { Audit, Suggestion } from '@adobe/spacecat-shared-data-access';
 import { AzureOpenAIClient } from '@adobe/spacecat-shared-gpt-client';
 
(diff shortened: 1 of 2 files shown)


src/toc/handler.js
```diff
@@ -58,14 +58,42 @@ function getModeFromData(data, log) {
...
+ * @param {string} siteBaseURL - The site's canonical base URL.
+ * @returns {boolean}
+ */
+function isOnDifferentDomain(url, siteBaseURL) {
+  try {
+    const siteHostname = stripWWW(new URL(prependSchema(siteBaseURL)).hostname).toLowerCase();
+    const urlHostname = stripWWW(new URL(prependSchema(url)).hostname).toLowerCase();
+    return urlHostname !== siteHostname;
+  } catch {
+    return false;
```
MysticatBot, line 79: issue (blocking): The `catch` block returns `false`, meaning any URL that fails to parse is treated as "same domain" (in-scope). This is a fail-open design.

At the `generateSuggestions` guard (the last line of defense), this is arguably acceptable - the test confirms the intent to keep unparseable URLs rather than silently dropping valid suggestions with unusual formatting.

However, at the two earlier guard points (`getTocInputUrls` and `validatePageTocFromScrapeJson`), returning `false` means a malformed URL survives input filtering and proceeds to be scraped and AI-analyzed, consuming resources unnecessarily. If `siteBaseURL` itself is malformed (corrupt site config), the entire domain filter becomes inert across all three guards with no warning logged.

Consider one of:
1. Return `true` here (fail-closed - unparseable URLs are never in-scope) and update the `generateSuggestions` test to expect malformed URLs are filtered out.
2. Accept a `failOpen` parameter so callers can choose the behavior per call site.
3. At minimum, add a `log.warn` in this catch block so malformed-URL bypass is observable in production.

Option 1 is simplest - an unparseable URL cannot be served by OAE regardless, so there is no legitimate reason to generate a suggestion for it.

MysticatBot (changes requested):
Hey @anagarwa,

> ⚠ Degraded review - no spec document was found for this change (searched the PR links, the touched repo's docs, the architecture/guidelines docs, and linked Jira). This review covers code-level quality but could not validate the change against an agreed design, so confidence is reduced. Add a spec link (PR template section 4) and re-request review for a full-confidence pass.

**Verdict: Request changes** - one material concern about the fail-open error handling in the new domain guard function.
**Complexity: MEDIUM** - medium diff, single-service.
_Changes: Adds a hostname-scope guard (isOnDifferentDomain) at three points in the TOC pipeline to prevent generating suggestions for subdomain or cross-domain URLs that OAE cannot serve (2 files)._

## Must fix before merge
1. [Important] Fail-open on URL parse errors allows malformed URLs to bypass domain filtering at input and redirect stages - `src/toc/handler.js:79` (details inline)

<details>

- nit: No test exercises the www/non-www normalization path that stripWWW handles, or mixed-case hostnames. Adding one integration test with www.example.com as base vs example.com as candidate would confirm end-to-end correctness. - `test/audits/toc.test.js`
- suggestion: `isOnDifferentDomain` mirrors `isOnDifferentSubdomain` in `src/backlinks/handler.js`. Consider extracting into a shared utility in `src/utils/` to prevent the two copies from drifting independently. - `src/toc/handler.js:73`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 4m 21s | <b>Cost:</b> $7.09 | <b>Commit:</b> <code>0c32f53460e40d6de78ef0a1a4af2e05587665d6</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2902, opened 2026-08-18
Scope brand-presence data sources to ACCEPTED_REGIONS

rarescheseli:
### Summary of changes

- `offsite-brand-presence-postgrest.js` — query now spans all `ACCEPTED_REGIONS` (`.in('region_code', [...])`) instead of a single hardcoded `'US'`, so GB/CA/AU/IE/NZ brand-presence executions are no longer silently excluded at the DB layer.
- `handler.js` and `offsite-brand-presence-enrichment.js` — both `extractUrlsAndTopics` implementations now log a warning when every row with source data was dropped for region, so an all-region-filtered run is no longer indistinguishable in the logs from a genuine zero-citation run.
- `offsite-brand-presence-semrush.js` — **no region-scoping changes** (region scoping on this path remains deferred to LLMO-6710, unchanged from before this PR). This file's logging was converted from a hand-rolled `log.info/warn/error` + string-prefix pattern to the shared offsite logging convention (`createOffsiteLogger`/`olog`), matching `handler.js` and the other offsite-brand-presence files.
- Added test coverage: PostgreSQL-side region-list plumbing (`mapExecutionsToLegacyBrandPresenceRows` with non-US regions, an empty `regionCodes` array locked in as passed-through rather than silently defaulted, the region filter applied consistently across paginated pages), and a cross-region dedup test on `handler.js` (the same URL cited from two different accepted regions collapses to one URL-store entry, not two).

## Related Issues

Thanks for contributing!

### DIFF
--- a/src/offsite-brand-presence/handler.js
+++ b/src/offsite-brand-presence/handler.js
@@ -272,9 +272,17 @@ function trackTopicUrl(topicMap, topicName, url, category, prompt) {
  */
 function extractUrlsAndTopics(data, allUrls, topicMap, olog, siteHostname, brandTokens) {
   const rows = data.data;
+  let rowsWithSources = 0;
+  let rowsSkippedForRegion = 0;
   for (const row of rows) {
     const sources = row.Sources?.trim();
-    if (!sources || !ACCEPTED_REGIONS.has(row.Region)) {
+    if (!sources) {
+      // eslint-disable-next-line no-continue
+      continue;
+    }
+    rowsWithSources += 1;
+    if (!ACCEPTED_REGIONS.has(row.Region)) {
+      rowsSkippedForRegion += 1;
       // eslint-disable-next-line no-continue
       continue;
     }
@@ -311,6 +319,10 @@ function extractUrlsAndTopics(data, allUrls, topicMap, olog, siteHostname, brand
       }
       /* c8 ignore stop */
     }
+  }
+
+  if (rowsWithSources > 0 && rowsSkippedForRegion === rowsWithSources) {
+    olog.warn('url_extract', `All ${rowsWithSources} row(s) with sources were skipped: region not in ACCEPTED_REGIONS`, { rows: rowsWithSources });
   }
   olog.debug('url_extract', `Found ${allUrls.size} unique source URLs`, { count: allUrls.size });
 }
--- a/src/utils/offsite-brand-presence-postgrest.js
+++ b/src/utils/offsite-brand-presence-postgrest.js
@@ -14,11 +14,11 @@ import { getDateRanges } from '@adobe/spacecat-shared-utils';
 import {
   appendFields, OFFSITE_DOMAIN, OUTCOME, PEER,
 } from './offsite-logging.js';
-import { PROVIDERS } from '../offsite-brand-presence/constants.js';
+import { ACCEPTED_REGIONS, PROVIDERS } from '../offsite-brand-presence/constants.js';
 
 export const EXECUTION_FETCH_BATCH_SIZE = 5000;
 export const MAX_EXECUTION_FETCH_PAGES = 50;
-const DEFAULT_REGION_CODE = 'US';
+const DEFAULT_REGION_CODES = [...ACCEPTED_REGIONS];
 
 // Offsite-only util: keeps its own human component prefix but emits the offsite taxonomy as
 // Splunk-extractable `key=value` tokens. It does not know the audit slug, so `audit` is omitted.
@@ -82,7 +82,7 @@ async function fetchExecutionsWithSources(postgrestClient, {
   startDate,
   endDate,
   models,
-  regionCode = DEFAULT_REGION_CODE,
+  regionCodes = DEFAULT_REGION_CODES,
   log,
 }) {
   const rows = [];
@@ -104,7 +104,7 @@ async function fetchExecutionsWithSources(postgrestClient, {
       .select('id, execution_date, topics, prompt, category_name, region_code, model, brand_presence_sources(source_urls(url))')
       .eq('organization_id', organizationId)
       .eq('site_id', siteId)
-      .eq('region_code', regionCode)
+      .in('region_code', regionCodes)
       .in('model', models)
       .gte('execution_date', startDate)
       .lte('execution_date', endDate)
@@ -168,7 +168,7 @@ export async function loadBrandPresenceDataFromPostgrest({
   organizationId,
   previousWeeks,
   postgrestClient,
-  regionCode = DEFAULT_REGION_CODE,
+  regionCodes = DEFAULT_REGION_CODES,
   log,
 }) {
   if (!siteId || !organizationId || !postgrestClient?.from) {
@@ -190,7 +190,7 @@ export async function loadBrandPresenceDataFromPostgrest({
       startDate,
       endDate,
       models,
-      regionCode,
+      regionCodes,
       log,
     });
 
--- a/src/utils/offsite-brand-presence-postgrest.js
+++ b/src/utils/offsite-brand-presence-postgrest.js
@@ -18,6 +18,7 @@ import { ACCEPTED_REGIONS, PROVIDERS } from '../offsite-brand-presence/constants
 
 export const EXECUTION_FETCH_BATCH_SIZE = 5000;
 export const MAX_EXECUTION_FETCH_PAGES = 50;
+
 const DEFAULT_REGION_CODES = [...ACCEPTED_REGIONS];
 
 // Offsite-only util: keeps its own human component prefix but emits the offsite taxonomy as
--- a/src/utils/offsite-brand-presence-postgrest.js
+++ b/src/utils/offsite-brand-presence-postgrest.js
@@ -182,6 +182,14 @@ export async function loadBrandPresenceDataFromPostgrest({
     return null;
   }
 
+  // Avoid firing a query that's guaranteed to match nothing.
+  if (regionCodes.length === 0) {
+    log?.warn(bp(`No region codes provided for site ${siteId}, skipping PostgREST query`, {
+      event: 'brand_data_load', outcome: OUTCOME.SKIP, peer: PEER.POSTGRES, direction: 'inbound', reason: 'no_region_codes',
+    }));
+    return null;
+  }
+
   const { startDate, endDate } = dateWindow;
 
   try {
--- a/src/utils/offsite-brand-presence-semrush.js
+++ b/src/utils/offsite-brand-presence-semrush.js
@@ -22,13 +22,13 @@ import { getDateWindowForPreviousWeeks } from './offsite-brand-presence-postgres
 import { classifyAndNormalize } from './offsite-brand-presence-enrichment.js';
 import { computeBrandTokens, isExcludedCitedHost } from './offsite-audit-utils.js';
 import {
-  ACCEPTED_REGIONS,
+  createOffsiteLogger, errorField, AUDIT, PEER,
+} from './offsite-logging.js';
+import {
   TOP_CITED_EXCLUDED_DOMAINS,
   YOUTUBE_URL_REGEX,
   REDDIT_URL_REGEX,
 } from '../offsite-brand-presence/constants.js';
-
-const LOG_PREFIX = '[offsite-brand-presence][semrush]';
 
 /**
  * Default spacecat-api-service base URL. Its Elements proxy
@@ -130,16 +130,15 @@ export function buildDomainUrlsUrl({
  *   full page came back (LLMO-6711 shadow-run parity signal — a starved run is visible on
  *   `diagnostics` without grepping logs).
  */
-async function fetchDomainUrls(url, headers, log, siteId, pageSize) {
-  const ctx = { siteId };
+async function fetchDomainUrls(url, headers, olog, pageSize) {
   let response;
   const startedAt = Date.now();
   try {
     response = await fetch(url, { headers, timeout: FETCH_TIMEOUT_MS });
   } catch (error) {
-    log.error(`${LOG_PREFIX} Fetch failed for domain-urls: ${error.message}`, {
-      ...ctx, error: error.message, durationMs: Date.now() - startedAt,
-    });
+    olog.failure('domain_urls_fetch', 'Fetch failed for domain-urls', {
+      peer: PEER.SEMRUSH, direction: 'inbound', durationMs: Date.now() - startedAt, ...errorField(error),
+    }, error);
     return {
       rows: [], ok: false, authFailure: false, truncated: false,
     };
@@ -151,15 +150,15 @@ async function fetchDomainUrls(url, headers, log, siteId, pageSize) {
     // Capture the body so the rejecter is identifiable (api-service requireImsBearer
     // vs Semrush upstream) — the key signal for the LLMO-6709 auth gate.
     const responseBody = await readErrorBodySnippet(response);
-    const logCtx = {
-      ...ctx, status: response.status, responseBody, durationMs,
+    const logFields = {
+      peer: PEER.SEMRUSH, direction: 'inbound', status: response.status, responseBody, durationMs,
     };
     if (authFailure) {
       // Distinct branch so a rejected service token is visible instead of being
       // masked as "Semrush returned nothing" (LLMO-6709 verification).
-      log.error(`${LOG_PREFIX} Service token rejected for domain-urls (HTTP ${response.status}) — verify the IMS service token is authorized by the Semrush proxy (LLMO-6709)`, logCtx);
+      olog.failure('domain_urls_fetch', `Service token rejected for domain-urls (HTTP ${response.status}) — verify the IMS service token is authorized by the Semrush proxy (LLMO-6709)`, logFields);
     } else {
-      log.error(`${LOG_PREFIX} domain-urls returned HTTP ${response.status}`, logCtx);
+      olog.failure('domain_urls_fetch', `domain-urls returned HTTP ${response.status}`, logFields);
     }
     return {
       rows: [], ok: false, authFailure, truncated: false,
@@ -170,9 +169,9 @@ async function fetchDomainUrls(url, headers, log, siteId, pageSize) {
   try {
     body = await response.json();
   } catch (error) {
-    log.error(`${LOG_PREFIX} Could not parse domain-urls response: ${error.message}`, {
-      ...ctx, error: error.message, durationMs: Date.now() - startedAt,
-    });
+    olog.failure('domain_urls_fetch', 'Could not parse domain-urls response', {
+      peer: PEER.SEMRUSH, direction: 'inbound', durationMs: Date.now() - startedAt, ...errorField(error),
+    }, error);
     return {
       rows: [], ok: false, authFailure: false, truncated: false,
     };
@@ -181,31 +180,13 @@ async function fetchDomainUrls(url, headers, log, siteId, pageSize) {
   const raw = Array.isArray(body?.urls) ? body.urls : [];
   const truncated = raw.length >= pageSize;
   if (truncated) {
-    log.warn(`${LOG_PREFIX} domain-urls returned a full page (${raw.length} >= ${pageSize}); response may be truncated`, {
-      ...ctx, rowCount: raw.length, pageSize,
+    olog.warn('domain_urls_fetch', `domain-urls returned a full page (${raw.length} >= ${pageSize}); response may be truncated`, {
+      peer: PEER.SEMRUSH, direction: 'inbound', rowCount: raw.length, pageSize,
     });
   }
   return {
     rows: raw.slice(0, pageSize), ok: true, authFailure: false, truncated,
   };
-}
-
-/**
- * True when a row's `regions` field (comma-joined region codes, e.g. `"US,GB"`, per the
- * `domain-urls` contract) contains at least one code in {@link ACCEPTED_REGIONS} — mirrors
- * the legacy path's region gate (LLMO-6710). An empty/missing value means the endpoint
- * didn't resolve a region for that row; it is treated as "unknown", not "rejected", so a
- * gap in Semrush's region tagging can't silently zero out a run the way a missing `Region`
- * column does on the legacy path.
- *
- * @param {string} [regionsField]
- * @returns {boolean}
- */
-function isAcceptedRegion(regionsField) {
-  if (!regionsField) {
-    return true;
-  }
-  return regionsField.split(',').some((code) => ACCEPTED_REGIONS.has(code.trim()));
 }
 
 /**
@@ -304,6 +285,7 @@ export async function loadCitedUrlsFromSemrush({
   const { log, env } = context;
   const startedAt = Date.now();
   const siteId = site?.getId?.();
+  const olog = createOffsiteLogger(log, { audit: AUDIT.BRAND_PRESENCE, siteId });
   const elapsed = () => Date.now() - startedAt;
   const baseUrl = env?.SPACECAT_API_URI || SPACECAT_API_DEFAULT_BASE_URL;
 
@@ -314,7 +296,9 @@ export async function loadCitedUrlsFromSemrush({
     try {
       await onProgress(text);
     } catch (error) {
-      log.warn(`${LOG_PREFIX} Failed to post Semrush progress update: ${error.message}`, { siteId });
+      olog.warn('brand_data_load', 'Failed to post Semrush progress update', {
+        peer: PEER.SLACK, direction: 'outbound', ...errorField(error),
+      });
     }
   };
   const setDiagnostics = (patch) => {
@@ -323,13 +307,13 @@ export async function loadCitedUrlsFromSemrush({
     }
   };
 
-  log.info(`${LOG_PREFIX} Starting Semrush source attempt`, { siteId, baseUrl });
+  olog.start('brand_data_load', 'Starting Semrush source attempt', { peer: PEER.SEMRUSH, direction: 'inbound', baseUrl });
   await notify(':mag: Starting Semrush URL-Inspector lookup...');
 
   const spaceCatId = site?.getOrganizationId?.();
   if (!spaceCatId) {
-    log.warn(`${LOG_PREFIX} Site has no organization id; skipping Semrush source`, {
-      siteId, durationMs: elapsed(),
+    olog.warn('brand_data_load', 'Site has no organization id; skipping Semrush source', {
+      peer: PEER.SEMRUSH, direction: 'inbound', durationMs: elapsed(), reason: 'no-organization-id',
     });
     await notify(':warning: Site has no organization id — falling back to the legacy source.');
     setDiagnostics({ fallbackReason: 'no-organization-id' });
@@ -341,14 +325,14 @@ export async function loadCitedUrlsFromSemrush({
   const { brand, resolved } = await resolveBrandResultForSite(context, site);
   if (!brand?.brandId) {
     if (resolved) {
-      log.info(`${LOG_PREFIX} No active brand for org ${spaceCatId}; skipping Semrush source`, {
-        siteId, orgId: spaceCatId, durationMs: elapsed(),
+      olog.skip('brand_data_load', `No active brand for org ${spaceCatId}; skipping Semrush source`, {
+        peer: PEER.SEMRUSH, direction: 'inbound', orgId: spaceCatId, durationMs: elapsed(), reason: 'no-active-brand',
       });
       await notify(':information_source: No active brand configured for this org — falling back to the legacy source.');
       setDiagnostics({ fallbackReason: 'no-active-brand' });
     } else {
-      log.warn(`${LOG_PREFIX} Brand resolution failed (transient) for org ${spaceCatId}; using legacy fallback`, {
-        siteId, orgId: spaceCatId, durationMs: elapsed(),
+      olog.warn('brand_data_load', `Brand resolution failed (transient) for org ${spaceCatId}; using legacy fallback`, {
+        peer: PEER.SEMRUSH, direction: 'inbound', orgId: spaceCatId, durationMs: elapsed(), reason: 'brand-resolution-failed',
       });
       await notify(':warning: Brand resolution failed (transient) — falling back to the legacy source.');
       setDiagnostics({ fallbackReason: 'brand-resolution-failed' });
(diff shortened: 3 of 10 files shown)


src/utils/offsite-brand-presence-semrush.js
```diff
@@ -189,6 +190,24 @@ async function fetchDomainUrls(url, headers, log, siteId, pageSize) {
...
+ * column does on the legacy path.
+ *
+ * @param {string} [regionsField]
+ * @returns {boolean}
+ */
+function isAcceptedRegion(regionsField) {
+  if (!regionsField) {
+    return true;
+  }
+  return regionsField.split(',').some((code) => ACCEPTED_REGIONS.has(code.trim()));
```
amitruluimadalina: Case-sensitivity gap: `.trim()` is applied but never `.toUpperCase()`d before `ACCEPTED_REGIONS.has(code.trim())`. `ACCEPTED_REGIONS` is all-uppercase, so a lowercase/mixed-case region code from Semrush (`"us"`, `"Us,Gb"`) would be silently treated as rejected rather than unknown - the same class of silent-data-loss bug LLMO-6710 was about, one level down. No test covers non-uppercase input. Recommend normalizing case before the `Set.has()` check and adding a test for it.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -189,6 +190,24 @@ async function fetchDomainUrls(url, headers, log, siteId, pageSize) {
...
+}
```
amitruluimadalina: This `isAcceptedRegion` treats a missing/empty `regionsField` as accepted (unknown, not rejected), while the legacy paths (`handler.js` / `offsite-brand-presence-enrichment.js`) treat a missing `Region` as rejected via `!ACCEPTED_REGIONS.has(row.Region)`. That divergence is intentional per the PR description, but it is not documented anywhere as a known cross-path inconsistency, and it will directly complicate the LLMO-6711 shadow-run parity comparison referenced in the ADR - a row with unresolved region metadata will show up in one path and not the other with no obvious explanation. Please call this out explicitly in the ADR known-gaps section. Also worth extracting the count-and-warn pattern (duplicated near-identically in three files) into one shared helper so this kind of divergence does not keep happening silently.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -441,12 +460,20 @@ export async function loadCitedUrlsFromSemrush({
...
   const bucketCounts = { 'youtube.com': 0, 'reddit.com': 0, cited: 0 };
+  let classifiedCount = 0;
+  let regionSkippedCount = 0;
   for (const row of result.rows) {
     const bucketed = classifyRow(row, siteHostname, brandTokens);
     if (!bucketed) {
       // eslint-disable-next-line no-continue
       continue;
     }
+    classifiedCount += 1;
```
amitruluimadalina: Region filtering here happens client-side, after `classifyRow` and after the full Semrush `domain-urls` response has already been fetched (API cost/rate-limit already spent) - unlike the Postgres path, which now filters server-side in the query itself. If the Semrush endpoint has no region request parameter this is unavoidable today, but it should be captured explicitly as tracked debt in the ADR's "Configuration / client-convention debt" section rather than just stated as fact in the spec, especially since this is the path meant to grow (LLMO-6844/6818 batching).

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -463,6 +490,12 @@ export async function loadCitedUrlsFromSemrush({
     }
   }
```
amitruluimadalina, line 468: There is no telemetry on how often the "unknown region, kept" branch fires (rows where `regionsField` was falsy and thus accepted by default). The premise of this PR is that a Semrush metadata gap cannot silently zero out a run - but if that gap turns out to be common rather than rare, nobody will notice, and it will also block the LLMO-6711 shadow-run comparison from distinguishing "genuinely in-region" from "unknown, waved through". Recommend adding an `unknownRegionCount` counter alongside `classifiedCount`/`regionSkippedCount` and surfacing it in this log line.

test/utils/offsite-brand-presence-semrush.test.js
```diff
@@ -326,6 +326,70 @@ describe('offsite-brand-presence-semrush', function () {
...
+
+  it('does not warn about region skip when no rows were classified at all', async () => {
+    fetchStub.resolves(okJson({ urls: [] }));
+    await run();
+    expect(warnedWith(/region not in ACCEPTED_REGIONS/)).to.equal(false);
+  });
+
   // --- body / truncation ----------------------------------------------------
 
   it('treats a non-array urls body as empty (no fallback)', async () => {
```
amitruluimadalina, line 331: Test coverage gap: no case exercises whitespace-padded or malformed comma segments (e.g. `"US, GB "`, `"US,"`), and no case exercises lowercase/mixed-case region codes (directly related to the case-sensitivity issue flagged on `isAcceptedRegion` above). Recommend adding these as dedicated unit tests for `isAcceptedRegion` in isolation.

src/offsite-brand-presence/handler.js
```diff
@@ -272,9 +272,17 @@ function trackTopicUrl(topicMap, topicName, url, category, prompt) {
...
+  let rowsSkippedForRegion = 0;
   for (const row of rows) {
     const sources = row.Sources?.trim();
-    if (!sources || !ACCEPTED_REGIONS.has(row.Region)) {
+    if (!sources) {
+      // eslint-disable-next-line no-continue
+      continue;
+    }
+    rowsWithSources += 1;
+    if (!ACCEPTED_REGIONS.has(row.Region)) {
```
amitruluimadalina, line 284: Now that GB/CA/AU/IE/NZ rows can reach this function where only `US` rows ever did before (via the Postgres query change), there is no test proving the same URL cited from two different accepted regions does not get double-counted or unexpectedly merged in `allUrls`/`topicMap` via `trackTopicUrl`. Worth a targeted test with two regions producing the same URL, and confirming `mapExecutionsToLegacyBrandPresenceRows` and other consumers of `loadBrandPresenceDataFromPostgrest` do not implicitly assume region is always US.

amitruluimadalina (changes requested):
Hey @rarescheseli,

## Strengths

- The core fix is correct and closes two real, previously-documented bugs: `src/utils/offsite-brand-presence-postgrest.js:107` moving from `.eq('region_code', regionCode)` (hardcoded `'US'`) to `.in('region_code', regionCodes)` stops silently excluding GB/CA/AU/IE/NZ rows at the DB query level, even though downstream logic already expected them to pass. `src/utils/offsite-brand-presence-semrush.js:204-209`'s `isAcceptedRegion` closes the documented LLMO-6710 gap on the newer Semrush path.
- The "unknown region, don't reject" design for the Semrush path is a deliberate, well-reasoned choice given the failure mode it guards against (a Semrush metadata gap silently zeroing a run), and it's documented in both the function's JSDoc and the spec update.
- The three-file "warn when every row is dropped for region" telemetry (`handler.js:324`, `offsite-brand-presence-enrichment.js:~373`, `offsite-brand-presence-semrush.js:492`) closes a real observability gap - previously an all-region-filtered run was indistinguishable in the logs from a genuine zero-citation run.
- The `regionCode` to `regionCodes` rename was verified safe: the sole caller (`offsite-brand-presence-enrichment.js:391`) never passed the parameter explicitly, so there's no orphaned call site.
- Test coverage of the core new branches is genuine, not just line-count padding - mixed-region rows, missing-`regions` rows, all-rejected rows, and "don't warn when at least one row survives" are all separately asserted.
- Docs (ADR + spec) were updated in the same PR, correctly noting the Semrush request itself still carries no region param (client-side filter only) - good precision.

## Issues

### Important (Should Fix)

1. **Case-sensitivity gap reopens the exact bug class this PR fixes** (`src/utils/offsite-brand-presence-semrush.js:208`). `isAcceptedRegion` does `.trim()` but never normalizes case before `ACCEPTED_REGIONS.has(code.trim())`, and `ACCEPTED_REGIONS` is all-uppercase (`'US'`, `'GB'`, ...). If Semrush ever returns lowercase or mixed-case codes (`"us"`, `"Us,Gb"`), every one of those rows would be silently treated as rejected rather than unknown - the same "metadata gap leads to silent data loss" failure mode LLMO-6710 was about, just moved one level down. No test exercises non-uppercase input. Fix with a `.toUpperCase()` normalization on each code before the `Set.has()` check, and add a test locking in the behavior either way.

2. **Three independent, drifting implementations of the same "count skipped, warn if all skipped" pattern**, with an undocumented behavioral divergence between them. `handler.js:279-286` and `offsite-brand-presence-enrichment.js` (legacy/SharePoint paths) treat a missing region as rejected (`!ACCEPTED_REGIONS.has(row.Region)` is true for `undefined`), while `offsite-brand-presence-semrush.js:204-209` treats a missing region as accepted. That's a real, intentional design choice per this PR's description - but the three implementations that are supposed to encode the same "only ACCEPTED_REGIONS markets" policy don't actually agree on this edge case, and it isn't called out anywhere as a known cross-path inconsistency. This will directly complicate the LLMO-6711 shadow-run parity work referenced in the ADR: a row with unresolved region metadata will show up in one path's output and not the other's, and it won't be obvious why. At minimum, document this explicitly in the ADR's known-gaps section rather than reading as if region scoping is now uniformly closed across sources. Recommend also extracting the shared count-and-warn logic into one helper so future changes (e.g. the telemetry addition below) don't need three synchronized edits.

3. **Semrush region filtering happens client-side, after the API call, not as a request parameter** (`src/utils/offsite-brand-presence-semrush.js:465-475`). Excluded-region rows are still fetched and paid for (API cost/rate-limit already spent) before being discarded - unlike the Postgres path, which now filters server-side in the query itself. If the Semrush `domain-urls` endpoint doesn't support a region parameter, this is currently unavoidable and should be captured explicitly as tracked debt (the spec's section 3.3 states it as a fact, not a cost concern to revisit) - worth a line in the ADR's "Configuration / client-convention debt" section, especially since the Semrush path is the one meant to grow.

4. **No telemetry on how often the "unknown region, kept" branch fires** (`src/utils/offsite-brand-presence-semrush.js:492`). The whole premise of this PR is "a Semrush metadata gap can't silently zero out a run" - but there's currently no visibility into how common that gap actually is. If unresolved regions become the majority case rather than a rare edge, nobody will notice, and the LLMO-6711 shadow-run comparison won't be able to distinguish "genuinely in-region" from "unknown, waved through." Add an `unknownRegionCount` counter alongside the existing `classifiedCount`/`regionSkippedCount` and surface it in the final "Bucketed domain-urls response" log line.

5. **Test coverage gaps around `isAcceptedRegion` and the Postgres region-list plumbing.** No test covers: whitespace-padded or malformed comma segments (e.g. `"US, GB "`, `"US,"`); an empty `regionCodes` array passed into `loadBrandPresenceDataFromPostgrest`/`fetchExecutionsWithSources` (behavior of `.in('region_code', [])` with an empty array is worth locking in, since it's functionally the same "silent zero" failure mode this PR targets on the Semrush side); or that the multi-region `.in()` filter is applied consistently across every paginated fetch, not just the first page.

6. **Cross-region dedup / downstream-consumer risk is unverified.** Now that GB/CA/AU/IE/NZ rows can appear where only `'US'` ever did before, there's no test proving the same URL cited from two different accepted regions doesn't get double-counted or unexpectedly merged in `allUrls`/`topicMap` (`handler.js`'s `trackTopicUrl`). `mapExecutionsToLegacyBrandPresenceRows` and any other consumer of `loadBrandPresenceDataFromPostgrest` that implicitly assumed region is always US wasn't audited in this diff. Worth a targeted test with two regions producing the same URL.

### Minor (Nice to Have)

1. **Unbounded `split(',')` on Semrush's `regions` field** (`src/utils/offsite-brand-presence-semrush.js:208`). No length cap before splitting an externally-sourced string. Low severity given Semrush is a semi-trusted vendor, but a malformed/degenerate response would allocate a large intermediate array per row on every run. A defensive `regionsField.slice(0, 256).split(',')` closes this cheaply.
2. **Stale docstring** (`src/utils/offsite-brand-presence-semrush.js:277`, `loadCitedUrlsFromSemrush`'s JSDoc): still says "Region scoping remains a follow-up (LLMO-6710)" a few lines above the code this PR adds to close that exact gap, in the same file.
3. **Inconsistent warning wording** across the three call sites (`"row(s) with sources were skipped"` vs `"classified row(s) were skipped"`) - harmless today, but makes future Splunk alert correlation across sources harder.
4. `DEFAULT_REGION_CODES = [...ACCEPTED_REGIONS]` (`offsite-brand-presence-postgrest.js:20`) is a snapshot computed once at module load from a frozen Set - safe today, but worth a one-line comment noting the snapshot semantics for future maintainers.

## Recommendations

- Extract the "count total vs. count skipped, warn if skipped equals total and total is greater than zero" pattern into one shared helper used by all three call sites - this is the natural point to also reconcile the missing-region semantic divergence and unify the log-message wording.
- File a fast-follow (or fold into LLMO-6711) to reconcile missing-region-metadata handling across the legacy and Semrush paths before using them for shadow-run parity comparison.
- Confirm a metric/alert exists on the new "all rows skipped for region" warn log so a real Semrush region-tagging outage is caught operationally, not just discoverable by reading logs after the fact.
- Post-deploy, spot-check whether the Postgres query's expanded row volume (now up to 6x, spanning all `ACCEPTED_REGIONS` instead of just `US`) approaches the existing `MAX_EXECUTION_FETCH_PAGES=50` / `EXECUTION_FETCH_BATCH_SIZE=5000` pagination ceiling on the highest-volume sites.

## Assessment

Ready to merge? With fixes
Reasoning: The fix correctly closes both known bugs and is well-tested on the happy paths, but the case-sensitivity gap reopens the same class of silent-data-loss bug this PR exists to fix, and the undocumented cross-path divergence plus test gaps around edge cases (whitespace, empty region lists, cross-region dedup) are worth resolving before this is called "closed" in the ADR - especially since LLMO-6711's shadow-run parity work depends on these paths actually agreeing on edge-case behavior.

Note: CI is currently pending as of this review.

## Next Steps
1. Fix the case-sensitivity gap in `isAcceptedRegion` (item 1) - same bug class as LLMO-6710, cheap to close.
2. Address the cross-path semantic divergence and duplication (items 2, 6) before treating LLMO-6710 as fully closed.
3. Fill the test gaps and telemetry noted in items 3-5.
4. Minor items are optional.

src/utils/offsite-brand-presence-semrush.js
rarescheseli: Reverted — removed all region-scoping work from `offsite-brand-presence-semrush.js` (the file, its unit tests, and the related ADR/spec doc updates) rather than patching the case-sensitivity bug in place. The case-insensitivity gap you flagged (`"us"`/`"Us,Gb"` being silently treated as rejected) is real, but rather than layer a fix on top of a design that's already drawing several other Important-severity concerns (see the sibling comments), we're pulling the whole thing out of this PR and re-scoping it as a proper follow-up under LLMO-6710. This PR now only closes the PostgreSQL-side region bug (the one that actually broke royalenfield.com in production).

src/utils/offsite-brand-presence-semrush.js
rarescheseli: Reverted along with the rest of the Semrush region-scoping code — this comment's cross-path divergence concern is now moot since there's no third path anymore (only handler.js and offsite-brand-presence-enrichment.js still implement the count-and-warn pattern, and they already agree with each other on missing-region semantics — both treat it as rejected). On the "extract a shared helper" suggestion: I left the two remaining call sites as-is for now rather than introducing an abstraction for just two ~6-line blocks — happy to extract one if you'd still like it, but wanted to flag that as a judgment call rather than silently skip it.

src/utils/offsite-brand-presence-semrush.js
rarescheseli: Reverted along with the rest of the Semrush region-scoping code — the client-side-filtering-cost concern no longer applies since this PR no longer touches the Semrush loader's region handling at all. Will be re-evaluated as part of the LLMO-6710 follow-up, where this is worth capturing as tracked debt in the ADR as you suggested.

src/utils/offsite-brand-presence-semrush.js
rarescheseli: Reverted along with the rest of the Semrush region-scoping code, so there's no more "unknown region, kept" branch in this PR to add telemetry to. Good catch regardless — this is exactly the kind of gap the LLMO-6710 follow-up needs to close before that path is trusted for shadow-run parity.

test/utils/offsite-brand-presence-semrush.test.js
rarescheseli: Reverted — this test file's region-scoping additions (including `isAcceptedRegion`) were pulled out entirely along with the source change, so these specific coverage gaps no longer apply. Will resurface with proper coverage of whitespace/malformed segments and case normalization when LLMO-6710 is picked back up.

src/offsite-brand-presence/handler.js
rarescheseli: Fixed — added a targeted test ('merges the same URL cited from two different accepted regions into one URL-store entry') that cites the same normalized YouTube URL from both US and GB rows and asserts it collapses to a single `allUrls` entry and a single `AuditUrl.create` call, not two. Also added coverage on the `mapExecutionsToLegacyBrandPresenceRows`/`loadBrandPresenceDataFromPostgrest` side: a test mapping GB/CA executions through with no US assumption, a test locking in that an empty `regionCodes` array is passed straight to the query rather than silently defaulted, and a test confirming the region filter is applied on every paginated page (not just the first).

rarescheseli:
Thanks for the thorough review — replied inline on each of the 6 comments. Summary of what changed:

**Reverted entirely:** `src/utils/offsite-brand-presence-semrush.js`, its unit test file, and the ADR/spec doc updates describing the LLMO-6710 closure there. The case-sensitivity bug, the cross-path missing-region divergence, the client-side-filtering-cost debt, the missing telemetry, and the test coverage gaps you flagged were all real — rather than patch them one at a time, this PR now leaves the Semrush loader untouched and region scoping on that path stays an open follow-up (LLMO-6710), as it was before this PR.

**Kept and improved:** the PostgreSQL-side fix (`offsite-brand-presence-postgrest.js` now queries `.in('region_code', ACCEPTED_REGIONS)` instead of a hardcoded `'US'`) and the `handler.js`/`offsite-brand-presence-enrichment.js` warn-on-100%-region-skip telemetry — these are the parts that actually fixed the royalenfield.com production bug and aren't in question.

**Also fixed** (from the review, non-Semrush items):
- Added the cross-region dedup test you asked for on `handler.js` (same URL cited from US + GB collapses to one URL-store entry, not two).
- Added a `mapExecutionsToLegacyBrandPresenceRows` test with non-US regions (GB/CA), an empty-`regionCodes`-array test, and a test confirming the region filter applies on every paginated page.
- Added the snapshot-semantics comment on `DEFAULT_REGION_CODES` you suggested.

**Separately (not from this review):** converted all logging in `offsite-brand-presence-semrush.js` to the shared offsite logging convention (`createOffsiteLogger`/`olog`) it was missing before.

All touched files remain at 100% line/branch/statement coverage; full repo suite passes.

MysticatBot (approved):
Hey @rarescheseli,

**Verdict: Approve** - well-scoped fix that closes the production region bug, adds useful telemetry, and migrates Semrush logging to the shared convention.
**Complexity: MEDIUM** - medium diff across a single service boundary with test and doc updates.
_Changes: Expands the PostgREST query from hardcoded US to all ACCEPTED_REGIONS, adds warn-on-100%-region-skip telemetry in handler.js and enrichment.js, and migrates Semrush loader logging to the createOffsiteLogger convention (8 files)._

<details>

- nit: `DEFAULT_REGION_CODES = [...ACCEPTED_REGIONS]` is a module-load snapshot of a frozen Set - safe today, but consider a one-line comment noting the snapshot semantics for future maintainers who might make ACCEPTED_REGIONS dynamic - `src/utils/offsite-brand-presence-postgrest.js:20`
- suggestion: The warn-on-100%-skip pattern in handler.js and enrichment.js is near-identical (~6 lines each). Not worth extracting for just two call sites today, but if a third appears (e.g. Semrush follow-up), that is the threshold to DRY it into a shared helper.
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 6m 19s | <b>Cost:</b> $2.27 | <b>Commit:</b> <code>1e94f1e81748d7043a65a6054bc362ee2b860ba2</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -304,13 +307,13 @@ export async function loadCitedUrlsFromSemrush({
...
 
-  log.info(`${LOG_PREFIX} Starting Semrush source attempt`, { siteId, baseUrl });
+  olog.start('brand_data_load', 'Starting Semrush source attempt', { peer: PEER.SEMRUSH, direction: 'inbound', baseUrl });
   await notify(':mag: Starting Semrush URL-Inspector lookup...');
 
   const spaceCatId = site?.getOrganizationId?.();
   if (!spaceCatId) {
-    log.warn(`${LOG_PREFIX} Site has no organization id; skipping Semrush source`, {
-      siteId, durationMs: elapsed(),
+    olog.warn('brand_data_load', 'Site has no organization id; skipping Semrush source', {
```
amitruluimadalina, line 317: This event says "skipping" but is tagged with the default `outcome=failure` from `createOffsiteLogger`'s `warn()` helper instead of `outcome=skip`. Line 328 in this same file handles the structurally identical case correctly via `olog.skip(...)`. Same applies to line 388 ("Could not derive a date window; skipping Semrush source"), and line 183 ("domain-urls returned a full page...") is a successful-with-caveat response, not a failure at all. Noted for the planned follow-up PR on this file rather than blocking here.

src/utils/offsite-brand-presence-postgrest.js
```diff
@@ -82,7 +83,7 @@ async function fetchExecutionsWithSources(postgrestClient, {
   startDate,
   endDate,
   models,
-  regionCode = DEFAULT_REGION_CODE,
+  regionCodes = DEFAULT_REGION_CODES,
```
amitruluimadalina, line 86: The new test at `test/utils/offsite-brand-presence-postgrest.test.js:338` locks in that an empty `regionCodes` array passes straight through to `.in('region_code', [])` with no warning - accurate as a description of current behavior, but it documents rather than closes the original concern: a caller passing an empty array silently gets zero rows for every region, the same "silent zero" failure class this PR is otherwise trying to eliminate. Worth a guard/warning when `regionCodes` resolves empty. Not blocking - noted for follow-up.

amitruluimadalina (approved):
Hey @rarescheseli,

Re-reviewed the follow-up commit. Good response to the feedback - reverting the Semrush-side region gate rather than patching it in place was the right call, and the docs were reopened cleanly and accurately (no stale "closed" claims left behind).

## Strengths

- Previously flagged case-sensitivity gap (item 1) and the three-way drifting implementation/undocumented divergence (item 2) are resolved by removal: `isAcceptedRegion`, its `ACCEPTED_REGIONS` import, and the region-skip counters are fully gone from `src/utils/offsite-brand-presence-semrush.js` - confirmed by grep against the current file, not just the diff.
- Previously flagged client-side-filtering-cost concern (item 3) and missing telemetry on the "unknown region, kept" branch (item 4) are moot for the same reason - the code they were about no longer exists on this path.
- Previously flagged cross-region dedup risk (item 6) now has a real test (`test/audits/offsite-brand-presence.test.js:622`): two rows citing the same normalized YouTube URL under `US` and `GB` correctly collapse to one `allUrls` entry and one `AuditUrl.create` call.
- Two of the previously flagged test-coverage gaps (item 5) are closed with meaningful assertions on the Postgres path: `test/utils/offsite-brand-presence-postgrest.test.js:338` proves an empty `regionCodes` array is passed straight through rather than silently falling back to the default, and `:354` proves the `region_code` filter is content-checked (not just presence-checked) on every paginated page, which would catch a future query-object-reuse regression.
- The docs (ADR + spec) were reverted precisely - the "closed" paragraph is fully removed, not left half-true, and the known-gaps/rollout-gate language correctly restores "US markets only until region scoping (LLMO-6710) closes."
- The `createOffsiteLogger`/`olog` logging migration in this file is not a new pattern - it already ships in five other files in this codebase (including the sibling `offsite-brand-presence-postgrest.js`), so this closes a pre-existing consistency gap rather than opening new operational surface. The migration is also complete and consistent: no leftover ad hoc `log.info/warn/error` calls remain in the file.
- The logging convention itself defends against log injection: externally-sourced values (like the captured HTTP error response body) are routed through a `fields` argument that quote-escapes and strips control characters, never interpolated raw into the message string.

## Issues

The three items below are all scoped to `offsite-brand-presence-semrush.js` and, per the author, will be addressed in a follow-up PR rather than blocking this one. Flagging them here for visibility and to carry the context forward.

### Noted for follow-up (not blocking this PR)

1. **The `olog` migration introduces a new, concrete telemetry bug: several skip-shaped or purely-informational events are mislabeled `outcome=failure`** (`src/utils/offsite-brand-presence-semrush.js:315`, `:388`, `:183`). `createOffsiteLogger`'s `warn()` helper defaults to `outcome=failure` unless the caller explicitly overrides it. Line 315 ("Site has no organization id; skipping Semrush source") and line 388 ("Could not derive a date window; skipping Semrush source") both say "skipping" but aren't tagged `outcome=skip`, unlike the structurally identical line 328 which correctly calls `olog.skip(...)`. Line 183 ("domain-urls returned a full page...") is a successful response with a caveat, not a failure, and tagging it `outcome=failure` will pollute any `outcome=failure`-based Splunk alert or dashboard. Fix by switching 315 and 388 to `olog.skip(...)` and picking a non-failure outcome for 183.

2. **The empty-`regionCodes`-array design gap is documented by a test but not actually closed** (`test/utils/offsite-brand-presence-postgrest.test.js:338-352`, `src/utils/offsite-brand-presence-postgrest.js:86`). The new test locks in that `regionCodes: []` passes straight through to `.in('region_code', [])` with no warning or validation - accurate as a description of current behavior, but it leaves open the original concern: a caller that ends up passing an empty array gets a silent, unlogged zero-rows-for-every-region result, the same "silent zero" failure class this PR is otherwise trying to eliminate. Low risk today since the only caller never overrides `regionCodes`, but worth a guard/warning when it resolves empty.

3. **No regression-lock test for the Semrush path's current "no region filtering at all" behavior** (`test/utils/offsite-brand-presence-semrush.test.js`). Since `isAcceptedRegion` and every call site were deleted, there's now zero coverage asserting a row from an unaccepted or malformed region still flows through unfiltered on this path. That's the intended current state, but without a pinning test, a future contributor re-adding region logic could silently reintroduce the exact case-sensitivity/malformed-input bugs already flagged and removed here. Recommend one small test along the lines of "does not filter by region (LLMO-6710 not yet implemented on this path)."

### Minor (Nice to Have)

1. The new Postgres test titled `"maps non-US accepted-region executions through without any US assumption"` (`test/utils/offsite-brand-presence-postgrest.test.js:217`) overstates what it verifies - `mapExecutionsToLegacyBrandPresenceRows` is fully region-agnostic and would produce an identical result for any region code, valid or not. Consider retitling to something like "passes non-US region codes through unmodified."
2. The cross-region dedup test (item 6 fix) is real but narrower than it may read - `Region` never enters the `allUrls` map key, so this test can only ever catch a hypothetical future bug where someone starts keying by region. Worth keeping as documentation of the guarantee, just don't read it as proof that cross-region merging was ever a live risk in the shipped code.
3. Minor items from the original review (inconsistent warning wording across call sites, stale docstring phrasing, module-load snapshot comment) remain open at their original Minor severity.

## Recommendations

- Carry items 1-3 above into the planned follow-up PR on `offsite-brand-presence-semrush.js` so the context isn't lost.
- When LLMO-6710 is picked back up, prefer the server-side `.in('region_code', regionCodes)` pattern already proven on the Postgres path over reintroducing a client-side post-filter on Semrush.
- Consider filing/linking a comment on LLMO-6710 capturing the specific design questions this attempt surfaced (case-sensitivity, missing-region semantics, telemetry) so that work isn't lost with the revert.

## Assessment

Ready to merge? Yes
Reasoning: The revert is complete, accurate, and well-tested - architecture and security both gave this a clean pass, and the Postgres-side fix (the actual production bug) is solid. The three remaining items are scoped to the Semrush logging/test surface and are tracked for a follow-up PR rather than blocking this one.

Note: CI is currently passing on this commit.

src/utils/offsite-brand-presence-semrush.js
rarescheseli: Fixed — the two "skipping" cases now explicitly override `outcome: OUTCOME.SKIP` while keeping `warn` log level (both were originally logged at warn intentionally, distinct from the info-level "no active brand" case, so I kept that severity and just corrected the outcome field): `Site has no organization id; skipping Semrush source` and `Could not derive a date window; skipping Semrush source`. The full-page case at line 183 now overrides `outcome: OUTCOME.SUCCESS` since, as you noted, it's a successful response with a caveat, not a failure. `OUTCOME` is now imported from `./offsite-logging.js`.

src/utils/offsite-brand-presence-postgrest.js
rarescheseli: Fixed — added a guard in `loadBrandPresenceDataFromPostgrest`: when `regionCodes.length === 0`, it now logs a warning (`No region codes provided for site ..., skipping PostgREST query`, outcome=skip) and returns `null` immediately, rather than firing a query that's guaranteed to return nothing. Updated the existing test (renamed to 'returns null and warns without querying when regionCodes resolves empty') to assert `postgrestClient.from` is never called and the warning fires, instead of just documenting the pass-through.

rarescheseli:
Addressed the re-review's two inline comments plus the first two Minor items:

- **Outcome mislabeling** (`offsite-brand-presence-semrush.js:183,315,388`) — fixed. The two "skipping" cases now explicitly carry `outcome: OUTCOME.SKIP` (kept at `warn` log level, since that severity was already intentional pre-dating this PR), and the full-page-response case now carries `outcome: OUTCOME.SUCCESS` instead of the default `failure`.
- **Empty `regionCodes` silent-zero gap** (`offsite-brand-presence-postgrest.js`) — fixed. `loadBrandPresenceDataFromPostgrest` now guards on `regionCodes.length === 0`, logs a warning, and returns `null` without querying, instead of silently firing a guaranteed-empty request. Test updated to assert the query is never attempted.
- **Minor #1** — retitled the Postgres test from "maps non-US accepted-region executions through without any US assumption" to "passes non-US region codes through unmodified", since `mapExecutionsToLegacyBrandPresenceRows` is fully region-agnostic and the old title overstated what's verified.
- **Minor #2** — added a comment on the cross-region dedup test clarifying it's a regression-lock/guarantee test (Region never enters the `allUrls` map key), not proof that cross-region duplication was ever a live risk in the shipped code.

Left the remaining "noted for follow-up" item (a pinning test for the Semrush path's current no-region-filtering behavior) and the rest of the original review's Minor items for the planned LLMO-6710 follow-up PR, per your note that they're not blocking here.

All touched files still at 100% coverage; full repo suite passes.

PR #2903, opened 2026-08-18
fix(toc): exclude accessibility skip-navigation lists from hasTocInDom

anagarwa:
## Problem

`hasTocInDom`'s Signal 1 ("a list with 2+ internal anchor links") false-positives on accessibility "skip to content" link lists, which are near-universal WCAG boilerplate in a site's shared header markup:

```html
<ul class="nl-accessibility-links">
  <a href="#main">Skip to main content</a>
  <a href="#secondary-navigation">Skip to navigation</a>
</ul>
```

**Confirmed live on canadiantire.ca**: this exact pattern caused the DOM heuristic to report "TOC present" on **100% of 191 sampled pages** (verified by re-running the audit's own detection logic locally against real scrape data pulled from S3), independently masking genuine TOC gaps site-wide — the audit would never have surfaced a single TOC opportunity for this site even with a healthy scrape pipeline.

The first pass of this fix (English-only) still missed the site's `/fr/` pages, which carry the identical list with localized text ("passer au contenu principal" / "passer à la navigation") — added French phrasing too since the site itself is bilingual.

## Fix

Exclude a list from Signal 1 when **every** internal anchor link in it reads as a skip-navigation link (matched by anchor **text**, not class/id — skip-nav class naming is site/design-system-specific and won't generalize, while "Skip to ..."/"Jump to ..." phrasing is standard a11y copy). A list containing a mix of skip-nav and real section links is still correctly detected as a TOC (covered by a dedicated test).

## Testing

- 5 new `hasTocInDom` unit tests: the reported false positive, a whitespace variant, alternate English phrasing ("Skip navigation" / "Jump to main content"), the French case, and the mixed-list case.
- Full existing `test/audits/toc.test.js` suite passes (204/204 on this branch).
- `eslint` clean.
- Locally re-ran the audit's real detection logic (this fixed `hasTocInDom`, plus the existing heading-extraction/placement utilities) against 191 real `scrape.json` files pulled from `s3://spacecat-prod-scraper` for canadiantire.ca's most recent TOC audit job. Before this fix: 191/191 pages falsely reported "TOC present." After: 1 genuine TOC found (a real `id="toc"` anchor), 186 pages correctly flagged missing, 4 skipped for too few headings.

## Not in scope for this PR

- Only English and French skip-nav phrasing are covered (proven necessary on this bilingual site); other locales aren't handled yet — flagged in a code comment for whoever hits it next.
- A separate, unrelated issue also found during this investigation — intermittent `ECONNRESET`/socket hang-up errors reading scrape.json from S3 (`spacecat-prod-scraper`), seen at scale (~4,600 occurrences/24h across many audits) — is not addressed here and is being raised separately with the platform team.

### DIFF
--- a/src/toc/handler.js
+++ b/src/toc/handler.js
@@ -110,6 +110,40 @@ export const TOPPAGES_CHECK = {
   explanation: 'No URLs found for audit',
 };
 
+// Accessibility "skip navigation" links (e.g. <a href="#main">Skip to main content</a>) are
+// a near-universal WCAG pattern, almost always shipped as a 2+ item list in shared header
+// markup — which satisfies Signal 1 below on every single page of a site, without being a
+// TOC. Confirmed live on canadiantire.ca: `<ul class="nl-accessibility-links">` linking to
+// `#main`/`#secondary-navigation` with text "Skip to main content" / "Skip to navigation"
+// caused hasTocInDom to report true on 100% of 191 sampled pages, masking genuine TOC gaps
+// site-wide. Matched by anchor TEXT rather than by class/id, since the class naming
+// ("nl-accessibility-links" here) is site/design-system-specific and won't generalize, while
+// the "Skip to ..." / "Skip ..." / "Jump to ..." phrasing is standard a11y copy independent
+// of markup conventions.
+//
+// French equivalent included because canadiantire.ca itself is bilingual and re-triggers the
+// exact same false positive on its /fr/ pages via the localized skip-nav copy ("passer au
+// contenu principal" / "passer à la navigation") — confirmed live, same nl-accessibility-links
+// list, just localized text. Other locales are not yet covered; extend this pattern (or move
+// to a per-locale phrase list) if another site surfaces the same gap in a different language.
+const SKIP_NAV_TEXT_RE = /^(skip\s+(to\s+)?(the\s+)?(main\s+)?(content|navigation|nav|footer|search|links?)\b|jump\s+to\s+(the\s+)?(main\s+)?content\b|(passer|aller|sauter)\s+(au|à\s+la|directement\s+au)\s+(contenu(\s+principal)?|navigation)\b)/i;
+
+/**
+ * True when every internal anchor link in the list reads as an accessibility skip-navigation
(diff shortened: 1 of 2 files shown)


MysticatBot (approved):
Hey @anagarwa,

> ⚠ Degraded review - no spec document was found for this change (searched the PR links, the touched repos docs, the architecture/guidelines docs, and linked Jira). This review covers code-level quality but could not validate the change against an agreed design, so confidence is reduced. Add a spec link (PR template section 4) and re-request review for a full-confidence pass.

**Verdict: Approve** - well-scoped fix with solid test coverage; only non-blocking nits on regex tightening.
**Complexity: LOW** - small single-service bug fix, 2 files.
_Changes: Excludes accessibility skip-navigation link lists from the TOC-detection heuristic Signal 1 check, fixing confirmed false positives on bilingual sites (2 files)._

<details>

- nit: Missing `$` anchor on `SKIP_NAV_TEXT_RE` means prefix-only matching - text like "Skip to content - Section Overview" would pass. The `every()` guard makes this extremely unlikely to cause a real false exclusion, but adding `$` (or `)\s*$/i`) tightens the contract at zero cost to legitimate skip-nav links. - `src/toc/handler.js:135`
- nit: "Jump to" branch only accepts `content` as a target, while "Skip to" also allows `navigation|nav|footer|search|links?`. "Jump to navigation" is valid skip-nav copy that would not match. Aligning the two branches would close the gap. - `src/toc/handler.js:135`
- nit: `content\b` will not match "Skip to contents" (plural) because `\b` fails between "t" and "s". Some sites (e.g. gov.uk) use plural form. Changing to `contents?` would cover it. - `src/toc/handler.js:135`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 22s | <b>Cost:</b> $3.59 | <b>Commit:</b> <code>f8a7db272d8a8466e9206c5b5837ca7ece261b25</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2904, opened 2026-08-18
fix(preflight): surface a scrape-failure signal instead of a silent false-clean pass

radhikagpt1208:
When the content-scraper fails to fetch a preflight page (403/DNS/timeout), every DOM/SEO/accessibility check reported a clean pass instead of an error, because the completion message's per-URL FAILED status was received but discarded before reaching the check logic - which then inferred "no issues" purely from a scrape record's absence in S3. Thread that signal through AsyncJobRunner and use it to mark the affected checks with a stable error code from the PreflightError catalog, and skip/short-circuit the accessibility and form-accessibility polls for pages already known to have failed their main scrape.

Fixes SITES-48986 (closed duplicates: SITES-49360, SITES-49871).

## Related Issues

Thanks for contributing!

### DIFF
--- a/test/audits/preflight-canonical.test.js
+++ b/test/audits/preflight-canonical.test.js
@@ -143,7 +143,7 @@ describe('Preflight Canonical Audit', () => {
 
     it('surfaces every canonical tag selector for canonical-tag-multiple, not just the first', async () => {
       // Protects: the whole point of this check is "there is more than one" — surfacing only
-      // one tag's selector would hide the rest from the author (SITES-48646 follow-up).
+      // one tag's selector would hide the rest from the author ( follow-up).
       const ctx = buildContext();
       const rawBody = `<html><head>
         <link rel="canonical" href="${PAGE_URL}">
--- a/src/common/async-job-runner.js
+++ b/src/common/async-job-runner.js
@@ -69,7 +69,7 @@ export class AsyncJobRunner extends StepAudit {
     const { stepNames } = this;
     const { log } = context;
     const {
-      type, jobId, auditContext = {},
+      type, jobId, auditContext = {}, scrapeResults,
     } = message;
     try {
       const job = await this.jobProvider(auditContext.jobId || jobId, context);
@@ -94,7 +94,7 @@ export class AsyncJobRunner extends StepAudit {
       const step = this.getStep(stepName);
 
       const updatedStepContext = {
-        ...context, site, job, type, auditContext,
+        ...context, site, job, type, auditContext, scrapeResults,
       };
 
       updatedStepContext.finalUrl = await this.urlResolver(site, context);
--- a/src/preflight/accessibility.js
+++ b/src/preflight/accessibility.js
@@ -10,7 +10,7 @@
  * governing permissions and limitations under the License.
  */
 
-import { isNonEmptyArray } from '@adobe/spacecat-shared-utils';
+import { isNonEmptyArray, stripTrailingSlash } from '@adobe/spacecat-shared-utils';
 import { DeleteObjectsCommand } from '@aws-sdk/client-s3';
 
 import { saveIntermediateResults, formatStructuredAuditLog } from './utils.js';
@@ -18,6 +18,7 @@ import { sleep } from '../support/utils.js';
 import { accessibilityOpportunitiesMap } from '../accessibility/utils/constants.js';
 import { getObjectFromKey, getObjectKeysUsingPrefix } from '../utils/s3-utils.js';
 import { formatWcagRule } from '../accessibility/utils/generate-individual-opportunities.js';
+import { PreflightError } from './error-constants.js';
 
 export const PREFLIGHT_ACCESSIBILITY = 'accessibility';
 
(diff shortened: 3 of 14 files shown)


MysticatBot (approved):
Hey @radhikagpt1208,

**Verdict: Approve** - solid bug fix with thorough coverage across all preflight handlers.
**Complexity: HIGH** - large diff (966 lines, 14 files); touches shared framework code (AsyncJobRunner) plus uniform changes across all preflight check handlers.
_Changes: Threads the content-scraper per-URL failure signal through AsyncJobRunner into each preflight check, replacing silent false-clean passes with stable error codes from the PreflightError catalog (14 files)._

<details>

- nit: Jira reference stripped from test comment leaving empty parenthetical - `test/audits/preflight-canonical.test.js:540`
- suggestion: Sanitize newlines/tabs in `reason` before logging to prevent log-line confusion in Splunk - `src/preflight/utils.js:63`
- suggestion: Consider making the 401/403 classification regex more lenient (e.g. `/\b40[13]\b/i`) to handle content-scraper format variations - `src/preflight/utils.js:57`
- suggestion: Add a `null` input case to the `buildFailedScrapesMap` unit test alongside `undefined` and `[]` - `test/preflight/utils.test.js:108`

</details>

Note: CI checks are all passing. Codecov confirms 100% coverage on modified lines.

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 12m 31s | <b>Cost:</b> $6.96 | <b>Commit:</b> <code>502dd2c3ef5c535c3996fb79c3b56d7f5d105a20</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

test/audits/preflight-canonical.test.js
```diff
@@ -143,7 +143,7 @@ describe('Preflight Canonical Audit', () => {
 
     it('surfaces every canonical tag selector for canonical-tag-multiple, not just the first', async () => {
       // Protects: the whole point of this check is "there is more than one" — surfacing only
-      // one tag's selector would hide the rest from the author (SITES-48646 follow-up).
+      // one tag's selector would hide the rest from the author ( follow-up).
```
clotton, line 146: Looks like an accidental find/replace stripped the ticket reference here, leaving a malformed `( follow-up)`. Was probably `(SITES-48646 follow-up)` — restore the ref or drop the trailing `( follow-up)` so the comment reads cleanly.

clotton (approved):
Reviewed end-to-end (source + tests + the producer contract in spacecat-content-scraper). This is a solid, well-scoped fix for the false-clean class of bugs (SITES-48986 / 49360 / 49871), and it's safe to merge. Notes below are minor / non-blocking.

**Verified good**
- Producer contract holds: `abstract-handler.js` sends `scrapeResults[].metadata.{url,status,reason}`, and a real 403 is thrown as `HttpError("HTTP 403 error for URL: ...")`, so `reason` carries `HTTP 403` and is correctly classified as `SCRAPE_FORBIDDEN` (PREFLIGHT-101) — the intended "you may not have permission" message is actually reached in prod, not just in tests.
- `stripTrailingSlash` is imported in every file that now uses it (canonical/handler/headings/links/metatags already had it; accessibility/form-accessibility/utils add it), and the map is keyed and looked up with the same normalization on both sides, matching the existing previewUrl matching.
- Both real ambiguity vectors are closed: (a) failed-scrape URLs get `status:'error'` at entry-push time and are filtered out of the a11y/form-a11y scrape request + poll + result lookup, and (b) a genuinely missing result file (poll timeout, URL *not* in `failedScrapes`) is now surfaced as `SCRAPE_TIMEOUT` instead of silent clean.
- Test coverage is thorough (handler, utils classifier, a11y + form-a11y skip/timeout paths).

**Minor / follow-up (non-blocking)**
1. Accidental edit in `preflight-canonical.test.js` (inline) — a ticket ref got stripped, leaving a malformed `( follow-up)`.
2. `buildFailedScrapesMap` only classifies `SCRAPE_FORBIDDEN`/`SCRAPE_TIMEOUT`; a scraper `status:'FAILED'` with no `reason` (the success-but-`statusCode>=400` branch at abstract-handler.js:1216, and any 404/429/5xx) falls back to the generic `SCRAPE_FAILED`. Fine for now — the important thing (no silent clean) is achieved — but worth knowing the nice per-code messaging won't cover those paths.
3. The completion message's `REDIRECT` status isn't treated as a failed scrape, so a preview URL that redirects still leaves no scrape record at the original key and could fall through to a clean result. Likely out of scope here, just flagging.
4. `SCRAPE_ERROR` is documented as "Transient — re-running may succeed," but 401/403 (permission) and 404 usually aren't transient. Only matters if this classification ever drives retry logic.

radhikagpt1208:
Upon discussion ([here](https://cq-dev.slack.com/archives/C08UMQNFB4L/p1787077582468569)), closing this ticket as we are now deferring all V1 work and just focus on V2 going forward. V2 already handles this behavior, so safe to mark this as resolved.

PR #2909, opened 2026-08-24
feat(prerender): reconcile coveredByDomainWide against actual edge state (LLMO-7052)

anuj-adobe:
## Summary
- `coveredByDomainWide` is set immediately and unverified when a domain-wide/path suggestion is deployed (pure pattern match in `spacecat-shared-tokowaka-client`), and the only existing self-heal (`clearStaleCoverageRefs`) only clears it on a *full* domain-wide rollback — never based on a specific URL's own edge truth.
- Adds deploy-date-relative reconciliation in `src/prerender/domain-wide-reconciliation.js`, independent of Problems 1/2 (LLMO-7038/LLMO-6638, still unmerged/blocked on team decisioning):
  - `getDomainWideReconciliationCandidates` — additive daily batch (capped at `DOMAIN_WIDE_RECONCILIATION_BATCH_SIZE=200`) selecting `coveredByDomainWide` suggestions with no post-deploy scrape confirmation yet, appended on top of `DAILY_BATCH_SIZE` in `submitForScraping`.
  - `syncCoveredByDomainWide` — single-fetch/single-save write-back in `processContentAndGenerateOpportunities`, merged with the pre-existing add-side logic (previously `markNewSuggestionsAsCovered` in `handler.js`): covers newly-deployed NEW suggestions and clears the flag off any suggestion this run's scrapes confirm `isDeployedAtEdge: false` — no grace period.
  - Matching is by exact suggestion identity (`normalizePathnameWithQuery`, the same key `buildSuggestionKey` uses for suggestion identity elsewhere in this audit) — each suggestion is reconciled against its own scrape result only, never inferred from a sibling query-param variant's result. An earlier iteration matched by pathname only; changed after confirming with the edge-deploy team that their pattern matcher (`buildUrlMatcher` in `spacecat-shared-tokowaka-client`) also only matches on pathname, which raised the question of whether that pathname-only behavior should itself be revisited on the deploy side (raised separately in #project-tokowaka-scrum-updates) — this reconciliation logic now stays consistent with the rest of the audit's suggestion-identity model instead of assuming that.
  - Explicit guard against a scrape-failure edge case: `status.json` stamps `scrapedAt` on every attempt regardless of success/failure, so eligibility checks `scrapingStatus === 'success'`, not just the timestamp — a bot-blocked/failed scrape never permanently retires a URL from the reconciliation backlog.
- Design discussed and iterated in the [Prerender Audit Enhancement Requests wiki](https://wiki.corp.adobe.com/spaces/~anujmaha/pages/4016155947/Prerender+Audit+Enhancement+Requests), Problem 3.

## Test plan
- [x] `test/audits/prerender/domain-wide-reconciliation.test.js` — 36 tests, 100% lines/branches/functions on the module.
- [x] `test/audits/prerender/behaviour/domain-wide-reconciliation.test.js` — 6 public-entry-point behaviour tests, unchanged since the initial implementation; used to validate the later move/merge refactor was behavior-preserving.
- [x] Full `test/audits/prerender/` suite passes (685/685), no regressions.
- [x] Full repo suite: pre-existing ~9-10 unrelated failures reproduce identically on clean `main` (cited-analysis, CWV, broken-internal-links, product-metatags, sitemap, meta-tags).
- [x] Lint clean.

### DIFF
--- a/src/prerender/domain-wide-reconciliation.js
+++ b/src/prerender/domain-wide-reconciliation.js
@@ -192,9 +192,13 @@ export async function syncCoveredByDomainWide(opportunity, context, successfulCo
     + `${!!domainWideSuggestion}, baseUrl=${baseUrl}`);
 
   // Pathname-only (not query-aware) is intentional here, unlike normalizePathnameWithQuery
-  // used for status.json/candidate dedup elsewhere: edge routing/deployment rules (a
-  // domain-wide /* rule or a path pattern) apply per-pathname, not per query-string variant,
-  // so two query-param variants of the same page are the same deployment unit for this check.
+  // used for status.json/candidate dedup elsewhere: coveredByDomainWide is assigned upstream
+  // by matching /*-suffixed allowedRegexPatterns against pathname only, ignoring the query
+  // string entirely (see buildUrlMatcher in spacecat-shared-tokowaka-client's
+  // src/utils/pattern-utils.js) — a domain-wide or path-level rule covers every query-param
+  // variant of a pathname identically. Reconciling at query-aware granularity would strand
+  // variants like /page?v=2 unreconciled forever unless that exact variant gets scraped,
+  // even though the routing rule that covers it doesn't distinguish query strings either.
   const deployedAtEdgePathnames = new Set();
   const notDeployedPathnames = new Set();
   successfulComparisons.forEach((r) => {
--- a/src/prerender/domain-wide-reconciliation.js
+++ b/src/prerender/domain-wide-reconciliation.js
@@ -10,12 +10,11 @@
  * governing permissions and limitations under the License.
  */
 
-import { Audit } from '@adobe/spacecat-shared-data-access';
+import { Audit, Suggestion } from '@adobe/spacecat-shared-data-access';
 import {
   isDomainWideSuggestionData,
   isPathSuggestionData,
   normalizePathnameWithQuery,
-  readSiteStatusJson,
   toPathname,
 } from './utils/utils.js';
 import { DOMAIN_WIDE_RECONCILIATION_BATCH_SIZE } from './utils/constants.js';
@@ -34,6 +33,19 @@ function isIndividualUrlSuggestionData(data) {
 }
 
 /**
+ * True when a suggestion is the active (non-OUTDATED) domain-wide suggestion with
+ * `edgeDeployed` set. Shared predicate for both the batch-selection lookup (which fetches
+ * its own suggestion list) and the write-back (which reuses an already-fetched list) so
+ * the "what counts as deployed" rule can't drift between them.
+ * @param {Object} s - Suggestion entity
+ * @returns {boolean}
+ */
+function isDeployedDomainWideSuggestion(s) {
+  return s.getStatus() === Suggestion.STATUSES.NEW
+    && isDomainWideSuggestionData(s.getData()) && !!s.getData()?.edgeDeployed;
+}
+
+/**
  * Finds the site's NEW prerender opportunity and, within it, the domain-wide
  * suggestion that currently has `edgeDeployed` set.
  * @param {Object} dataAccess - Data access layer
@@ -47,9 +59,7 @@ async function findDeployedDomainWide(dataAccess, siteId) {
     return null;
   }
   const suggestions = await opportunity.getSuggestions?.() ?? [];
-  const domainWide = suggestions.find(
-    (s) => isDomainWideSuggestionData(s.getData()) && !!s.getData()?.edgeDeployed,
-  );
+  const domainWide = suggestions.find(isDeployedDomainWideSuggestion);
   if (!domainWide) {
     return null;
   }
@@ -92,11 +102,13 @@ function hasPostDeployConfirmation(pageEntry, deployedAtMs) {
  * drops out of this selection once it gets any post-deploy scrape — no bookkeeping
  * needed to track "already handled".
  *
- * @param {Object} context - Audit context (dataAccess, log, s3Client, env)
+ * @param {Object} context - Audit context (dataAccess, log)
  * @param {string} siteId - Site ID
+ * @param {Object} siteStatus - The site's already-read status.json (caller fetches this
+ *   once via readSiteStatusJson and reuses it here rather than re-reading from S3)
  * @returns {Promise<string[]>} URLs to append to this run's scrape batch
  */
-export async function getDomainWideReconciliationCandidates(context, siteId) {
+export async function getDomainWideReconciliationCandidates(context, siteId, siteStatus) {
   const { dataAccess, log } = context;
   const found = await findDeployedDomainWide(dataAccess, siteId);
   if (!found) {
@@ -113,9 +125,8 @@ export async function getDomainWideReconciliationCandidates(context, siteId) {
     return [];
   }
 
-  const siteStatus = await readSiteStatusJson(siteId, context);
   const pageByPathname = new Map(
-    (siteStatus.pages ?? [])
+    (siteStatus?.pages ?? [])
       .filter((p) => p.url)
       .map((p) => [normalizePathnameWithQuery(p.url), p]),
   );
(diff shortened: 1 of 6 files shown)


src/prerender/domain-wide-reconciliation.js
```diff
@@ -0,0 +1,262 @@
...
+
+  const deployedAtEdgePathnames = new Set();
+  const notDeployedPathnames = new Set();
+  successfulComparisons.forEach((r) => {
+    (r.isDeployedAtEdge ? deployedAtEdgePathnames : notDeployedPathnames).add(toPathname(r.url));
+  });
+
+  const toCover = [];
+  if (domainWideSuggestion) {
+    const domainWideSuggestionId = domainWideSuggestion.getId();
```
MysticatBot: issue (blocking): `syncCoveredByDomainWide` builds `deployedAtEdgePathnames` and `notDeployedPathnames` from `successfulComparisons` using `toPathname(r.url)`, which strips query parameters. If two URLs with different query params (e.g., `/page?v=1` and `/page?v=2`) produce opposing `isDeployedAtEdge` values, the same pathname ends up in both sets.

A suggestion that already has `coveredByDomainWide` set and whose pathname appears in both sets will match both `toCover` (line 209: no `coveredByDomainWide` exclusion on per-URL suggestions) and `toClear` (line 240: requires `coveredByDomainWide`). Since `toClear` runs second, it overwrites the `toCover` mutation on the same in-memory object, and both ship in a single `saveMany`. The net result: the flag is cleared even though at least one variant confirmed deployment.

The scenario requires two query-param variants of the same pathname scraped in the same run with opposing results - unlikely but possible.

**Fix:** Before building `toSave`, deduplicate: if a pathname appears in both sets, pick one winner (likely deployed-wins, since a positive confirmation is stronger). Alternatively, exclude pathnames from `notDeployedPathnames` if they also appear in `deployedAtEdgePathnames`.

MysticatBot (changes requested):
Hey @anuj-adobe,

**Verdict: Request changes** - two blocking findings on data correctness and documentation compliance.
**Complexity: HIGH** - large diff (500+ lines, 6 files).
_Changes: Adds deploy-date-relative reconciliation for `coveredByDomainWide` suggestions, extracting batch selection and bidirectional write-back into a new module (6 files)._
Note: Recommend a human read before merge - possible divergence from the documented design (heuristic, flagged by the conventions reviewer). The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] Same pathname can appear in both `deployedAtEdgePathnames` and `notDeployedPathnames` sets, causing last-write-wins data corruption - `src/prerender/domain-wide-reconciliation.js:197` (details inline)
2. [Important] CLAUDE.md requires significant new functionality to have a spec in `/docs/specs/` - wiki discussion exists but no in-repo spec file found among the 18 existing specs in that directory

<details>

- nit: `urlSuggestionsToCover` does not exclude already-covered per-URL suggestions (`!data?.coveredByDomainWide` guard missing), causing redundant writes - `src/prerender/domain-wide-reconciliation.js:209`
- suggestion: Add a comment at `toPathname` usage explaining why pathname-only matching is intentional on the write-back side vs `normalizePathnameWithQuery` on selection side - `src/prerender/domain-wide-reconciliation.js:197`
- suggestion: Guard against `deployedAtMs` being NaN from a malformed `edgeDeployed` value with an early return - `src/prerender/domain-wide-reconciliation.js:66`
- suggestion: Emit a metric in the catch block so persistent `saveMany` failures surface in dashboards rather than only in log streams - `src/prerender/domain-wide-reconciliation.js:253`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 2m 58s | <b>Cost:</b> $7.90 | <b>Commit:</b> <code>e7ab5cf13cf368e967a1d93da5a190b6319d738f</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/prerender/domain-wide-reconciliation.js
anuj-adobe: Fixed in 4288b32c — pathnames confirmed deployed this run now win over not-deployed for the same pathname before the add/remove split is computed, so a suggestion can no longer land in both `toCover` and `toClear` in the same `saveMany`. New test: "does not clear a suggestion when two query-param variants of its pathname disagree this run — the positive confirmation wins".

Also pushed 7d10bd57 addressing the related non-blocking suggestion below (comment explaining why pathname-only matching is intentional here): the grouping isn't arbitrary — `coveredByDomainWide` itself is assigned upstream by matching `/*`-suffixed `allowedRegexPatterns` against pathname only (see `buildUrlMatcher` in `spacecat-shared-tokowaka-client/src/utils/pattern-utils.js`, which does `new URL(url).pathname` and drops the query string before matching). A domain-wide or path-level rule covers every query-param variant of a pathname identically, so reconciling at that same granularity is required for correctness — using a query-aware key would strand variants like `/page?v=2` unreconciled indefinitely even though the rule that covers them doesn't distinguish query strings either.

anuj-adobe:
Addressing the remaining review findings:

**Blocking #1 (pathname collision)** and the related non-blocking comment-clarity suggestion: fixed in 4288b32c and 7d10bd57 — see the inline reply on that thread.

**Nit: redundant re-cover of already-covered suggestions** — fixed in 4288b32c, `urlSuggestionsToCover` now excludes suggestions where `data?.coveredByDomainWide` is already set.

**Nit: NaN guard on `deployedAtMs`** — fixed in 4288b32c, `getDomainWideReconciliationCandidates` now returns `[]` with a `log.warn` when `edgeDeployed` is unparseable.

**Nit: emit a metric in the catch block** — not applying this one. There'''s no existing metrics-emission pattern elsewhere in this audit for this class of non-fatal post-processing failure to follow, and the existing `log.error` is consistent with how similar steps in this file surface failures today. Happy to add one if/when this codebase adopts a metrics convention more broadly.

**Blocking #2 (CLAUDE.md spec-in-`/docs/specs/` requirement)** — the design for this change (problem statement, alternatives considered, and the reconciliation rules) was worked out and reviewed on this Confluence page: https://wiki.corp.adobe.com/spaces/~anujmaha/pages/4016155947/Prerender+Audit+Enhancement+Requests (see "Problem 3" section, LLMO-7052). This change is also self-contained to the prerender audit — new logic lives entirely in `src/prerender/domain-wide-reconciliation.js` and its two call sites in `src/prerender/handler.js`, with no changes to shared framework code, other audits, or cross-audit contracts, which is why we'''re treating the wiki page as the spec of record rather than adding a separate in-repo file. Flagging that explicitly here for reviewer visibility rather than adding a `/docs/specs/` doc.

anuj-adobe:
Follow-up on the pathname-collision fix (4288b32c) and the discussion on the comment I added at the time explaining pathname-only matching (7d10bd57): after further discussion with the team (including confirming with the edge-deploy owner that `buildUrlMatcher` in `spacecat-shared-tokowaka-client` also only matches on pathname), I switched `syncCoveredByDomainWide` from pathname-only matching to `normalizePathnameWithQuery` — commit 8a9742fa6.

**What changed:** `deployedAtEdgeKeys`/`notDeployedKeys` now use the same identity key `buildSuggestionKey` uses for suggestion creation/dedup elsewhere in this audit, so each per-URL suggestion is reconciled against its own scrape result only, never inferred from a sibling query-param variant's result.

**Why:** query params were introduced on the suggestion-identity side specifically to give query variants their own suggestions (the apple.com case), so reconciliation should honor that same identity rather than collapsing back to pathname. Keeping the audit on one consistent key for both suggestion identity and reconciliation seemed better than relying on the deploy side's pathname-only routing behavior, which is itself now a separate open question raised in #project-tokowaka-scrum-updates (whether edge/optimize-at-edge coverage should become query-param-aware too).

**Tradeoff, accepted:** a not-yet-covered per-URL suggestion sharing a pathname with an already-confirmed sibling no longer gets covered until it's scraped in its own right. This isn't a new failure mode — it's the same general risk any suggestion has of falling out of organic/agentic/included ranking (Problem 1 in the linked wiki) — so we're accepting it in exchange for one consistent identity model.

Tests updated: the two tests exercising cross-variant pathname matching were replaced with per-suggestion-identity equivalents; the same-key-collision test now uses a stripped tracking param (`gclid`) instead of two distinct query params, since distinct params no longer collide under this key. PR description updated to match.

MysticatBot (approved):
Hey @anuj-adobe,

**Verdict: Approve** - prior blocking findings addressed, code is well-structured and thoroughly tested.
**Complexity: HIGH** - large diff (500+ lines, 6 files).
_Changes: Adds bidirectional reconciliation for `coveredByDomainWide` against actual edge state, extracting logic into a dedicated module with additive batch selection and single-saveMany write-back (6 files)._

- Pathname collision (same key in both deployed/not-deployed sets) - fixed via `normalizePathnameWithQuery` matching with positive-confirmation-wins semantics.
- Redundant re-cover of already-covered suggestions - guard added (`!data?.coveredByDomainWide`).
- NaN guard on `deployedAtMs` - added with warning log.

<details>

- nit: `setUpdatedBy('system')` called only on `toClear` suggestions, not on `toCover` - if the data-access layer does not set this implicitly on save, covered suggestions will lack an audit trail of who set the flag - `src/prerender/domain-wide-reconciliation.js:240`
- nit: `backlog.sort` relies on `localeCompare` of raw `scrapedAt` strings - works for ISO 8601 but would produce non-deterministic ordering if any status.json entry ever contained a non-ISO date format - `src/prerender/domain-wide-reconciliation.js:151`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 1m 21s | <b>Cost:</b> $6.88 | <b>Commit:</b> <code>8a9742fa6fba8df12ec22b757a3e6510ee9e4682</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

anuj-adobe:
Addressed the two nits from the latest review:

**`setUpdatedBy` missing on `toCover` — fixed, commit b5f806ab3.** Verified this was a real gap, not implicitly handled: `updatedBy` in `spacecat-shared-data-access`'s schema builder has `default: () => 'system'` but no `set` function, so `applyUpdateWatchers` (which runs on every `save`/`saveMany`) never re-stamps it on update — only `updatedAt` gets that treatment, since it has both `default` and `set`. Without the explicit call, `toCover` suggestions would've silently kept whatever `updatedBy` value they had before this automated change, while `toClear` (which already called it) got the correct audit trail. Added the same `s.setUpdatedBy('system')` call to the `toCover` loop, plus assertions in the existing add-side test.

**`localeCompare` on raw `scrapedAt` strings — not applying.** Traced every write path for this field in prerender's status.json (`handler.js:1093`, `1502`, `1529`) — all three always produce `new Date().toISOString()`, so the non-ISO scenario the suggestion describes can't occur given how this file is written today. Treating this the same as the metrics-emission nit from an earlier round: not adding handling for a case the current code can't produce.

src/prerender/domain-wide-reconciliation.js
```diff
@@ -0,0 +1,244 @@
...
+}
+
+/**
+ * True when a suggestion is the active (NEW) domain-wide suggestion with `edgeDeployed` set.
+ * Shared by batch-selection and write-back so "deployed" can't drift between them.
+ * @param {Object} s - Suggestion entity
+ * @returns {boolean}
+ */
+function isDeployedDomainWideSuggestion(s) {
+  return s.getStatus() === Suggestion.STATUSES.NEW
```
sahil9001, line 41: We are not checking `FIXED` here?

src/prerender/domain-wide-reconciliation.js
```diff
@@ -0,0 +1,244 @@
...
+  if (!opportunity) {
+    return null;
+  }
+  const suggestions = await opportunity.getSuggestions?.() ?? [];
+  const domainWide = suggestions.find(isDeployedDomainWideSuggestion);
+  if (!domainWide) {
+    return null;
+  }
+  return { suggestions, domainWide };
+}
```
sahil9001, line 64: Is there a way to pass the pre-calculated suggestion in the function itself rather than fetching it from the DB again?

src/prerender/domain-wide-reconciliation.js
```diff
@@ -0,0 +1,244 @@
...
+ * Save failures are logged and swallowed — non-fatal, doesn't fail the audit run.
+ * @param {Object|null} opportunity - Opportunity entity (no-op if null)
+ * @param {Object} context - Audit context (dataAccess, log, site)
+ * @param {Array} successfulComparisons - This run's non-error comparison results
+ * @returns {Promise<void>}
+ */
+export async function syncCoveredByDomainWide(opportunity, context, successfulComparisons) {
+  const { dataAccess, log, site } = context;
+  const SuggestionDA = dataAccess?.Suggestion;
+  if (!opportunity || typeof opportunity.getSuggestions !== 'function' || !SuggestionDA?.saveMany) {
```
sahil9001: Do we really need these many checks? What is the use of checking saveMany if we know that it already exists?

src/prerender/domain-wide-reconciliation.js
```diff
@@ -0,0 +1,244 @@
...
+  try {
+    await SuggestionDA.saveMany(toSave);
+    if (toClear.length > 0) {
+      log.info(`${LOG_PREFIX} Domain-wide reconciliation: cleared coveredByDomainWide on `
+        + `${toClear.length} suggestion(s) confirmed not deployed at edge this run.`);
+    }
+  } catch (e) {
+    log.error(`${LOG_PREFIX} Failed to save coveredByDomainWide changes for ${toSave.length} `
+      + `suggestion(s): ${e.message}`);
+  }
```
sahil9001, line 279: Can you please break this function into sub functions, currently it's too big of a chunk to go in one.

sahil9001 (changes requested):

src/prerender/domain-wide-reconciliation.js
anuj-adobe: No — a deployed domain-wide suggestion is guaranteed to stay `NEW`, never `FIXED`. There's an existing invariant enforced elsewhere in this file: `detectWrongEdgeDeployedStatus` (`handler.js:153-166`) actively warns if it ever finds a non-NEW suggestion with `edgeDeployed` set, with the comment "edgeDeployed is set when a URL is deployed at the CDN edge, and the suggestion status should not be changed away from NEW after that point." So checking `FIXED` here would only mask that invariant being violated, not handle a legitimate case.

This also mirrors the pre-existing (already-shipped, pre-dating this PR) `getDomainWideSuggestionDeployedAtEdge` check in `handler.js`, which used the exact same `NEW`-only condition — so this isn't a new gap introduced here.

src/prerender/domain-wide-reconciliation.js
anuj-adobe: Not within this call chain, unfortunately — this fetch is the *only* one done in Step 2 (`submitForScraping`). I checked: nothing else in `submitForScraping` reads the Opportunity/Suggestions before this point, so there's no pre-calculated value to pass in.

The broader concern (this vs. the second fetch in `syncCoveredByDomainWide` during Step 3) can't be avoided either, since `submitForScraping` and `processContentAndGenerateOpportunities` are separate Step-Audit steps that run as separate Lambda invocations (potentially far apart in time, dispatched via SQS) — there's no in-memory state to share between them, only what's persisted to S3/DB. Sharing the fetch across steps would mean introducing a new persistence mechanism just for this, which felt like more complexity than the read itself costs.

src/prerender/domain-wide-reconciliation.js
anuj-adobe: On reflection, you're right — reconsidered this. `opportunity.getSuggestions` and `dataAccess.Suggestion.saveMany` are guaranteed by the data-access layer whenever `opportunity` is non-null in production; checking for their existence was guarding against a scenario that can't actually happen, not a real system boundary. That's different from `!opportunity` itself, which is a real, documented no-op case (no active prerender opportunity for the site).

Fixed in ffc7a1305 — dropped both method-existence checks, kept `!opportunity`, and removed the two tests that only existed to exercise the dropped checks (the existing "logs and swallows an error from saveMany without throwing" test already covers the resilience behavior generically). Thanks for pushing on this.

src/prerender/domain-wide-reconciliation.js
anuj-adobe: Done in 05e116b7b — split into three helpers: `partitionComparisonResults` (the deployed/not-deployed key partition), `computeSuggestionsToCover` (add branch), and `computeSuggestionsToClear` (remove branch). `syncCoveredByDomainWide` is now a short orchestrator: guard, fetch, partition, compute both branches, save. No behavior change — all log messages and mutation order preserved, 687 tests passing, 100% coverage maintained.

ssilare-adobe (approved):

PR #2916, opened 2026-08-25
feat(cutover): bow out of legacy canonical / meta-tags / broken-internal-links audits on deliveryConfig.<opp>Engine=blackboard

FentPams:
## What

Generalizes the CWV per-site V1→V2 cutover bow-out (#2825) to three more legacy audits: **canonical**, **meta-tags**, and **broken-internal-links**.

When a site's opportunity is owned by Mystique's blackboard producer cascade — i.e. its per-audit `deliveryConfig.<opp>Engine` is `"blackboard"` — the legacy audit now **bows out**:
- creates no SpaceCat opportunity / suggestion and sends no Mystique message, and
- resolves the pre-existing legacy opportunity and outdates its live (NEW/PENDING) suggestions, via the shared `resolveOpportunityIfNoIssues` helper,

returning the step's normal clean-completion shape. Absent / `"legacy"` / any other value is unchanged (default legacy path).

New shared helper `src/common/delivery-engine.js#isBlackboardEngine(site, engineField)` generalizes `cwv/handler.js#isCwvBlackboardEngine` so any audit can gate on its own engine field. `deliveryConfig` is `type: 'any'`, so **no schema change** is needed.

| Audit | Engine field | Guard location |
|---|---|---|
| canonical | `canonicalEngine` | `processScrapedContent` — gates the CANONICAL opportunity; the separate Elmo `generic-opportunity` is intentionally left running |
| meta-tags (`META_TAGS`) | `metaTagsEngine` | top of `runAuditAndGenerateSuggestions`, before detection / creation / Mystique dispatch. `product-metatags` is **not** gated (no Mystique V2 equivalent) |
| broken-internal-links | `brokenInternalLinksEngine` | top of the opportunity-and-suggestions step, before create + Mystique dispatch |

## Why

Both the legacy audit and Mystique's blackboard cascade write the same SpaceCat `type` opportunity + suggestion rows; if both run for one site they collide (a second null-scoped opportunity, disjoint suggestion keys). A collision-free per-site cutover needs the legacy source to stop writing — the same rationale as #2825 (Spec 009-04 / ADR-0022). Mystique already writes these `<opp>Engine` flags to `deliveryConfig`, but only CWV honored them; this closes the gap for canonical / meta-tags / broken-internal-links.

## Testing

- New `test/common/delivery-engine.test.js` — the helper (true only for `"blackboard"`; degrades to legacy for absent / `"legacy"` / other; per-field independence; null-safe).
- Per-audit bow-out cases in `canonical.test.js`, `metatags.test.js`, and `internal-links/opportunity-suggestions-step.test.js`: assert no opportunity create, no Mystique dispatch, and that the pre-existing legacy opportunity is resolved + its live suggestions outdated.
- All four new test groups pass locally.

> Note: the existing `canonical` / `metatags` suites show some failures on a stale local `node_modules` vs this branch's `main` (`Oppty.SCOPE_TYPES` undefined) — they reproduce on **untouched** tests and are unrelated to this change; `npm ci` in CI resolves them.

## Mystique follow-up (separate repo/PR)

Mystique's cutover panel already writes these flags. Once this deploys, the panel's `audit_honors` for these three opportunities flips `NOT_YET → FULL` (grey → green "stops V1"); until then it honestly reports `inert_pending_spacecat`.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Lucian Felix Ghita", "Dominique Jäggi"]
```

### DIFF
--- a/src/canonical/handler.js
+++ b/src/canonical/handler.js
@@ -27,6 +27,7 @@ import {
 } from '../utils/data-access.js';
 import { getMergedAuditInputUrls } from '../utils/audit-input-urls.js';
 import { convertToOpportunity } from '../common/opportunity.js';
+import { isBlackboardEngine } from '../common/delivery-engine.js';
 import { createOpportunityData, createOpportunityDataForElmo } from './opportunity-data-mapper.js';
 import { CANONICAL_CHECKS } from './constants.js';
 import { getObjectFromKey } from '../utils/s3-utils.js';
@@ -773,8 +774,16 @@ export async function processScrapedContent(context) {
 
   log.info(`[canonical] Generated ${sortedSuggestions.length} canonical suggestions for ${baseURL}`);
 
-  // Create opportunities and sync suggestions
-  if (sortedSuggestions.length > 0) {
+  // Per-site V1 -> V2 cutover (Spec 009-04 / ADR-0022): when the CANONICAL opportunity is
+  // owned by Mystique's blackboard producer cascade (deliveryConfig.canonicalEngine=blackboard),
+  // bow out of the legacy CANONICAL sync and resolve any pre-existing legacy CANONICAL
+  // opportunity so the flip strands no active rows. The Elmo generic-opportunity below is a
+  // separate surface and is intentionally left running.
+  if (isBlackboardEngine(site, 'canonicalEngine')) {
+    log.info(`[canonical] siteId: ${site.getId()} | bowing out of legacy CANONICAL sync — deliveryConfig.canonicalEngine=blackboard (Mystique-owned)`);
+    await resolveOpportunityIfNoIssues(site.getId(), auditType, context.dataAccess, log);
+  } else if (sortedSuggestions.length > 0) {
+    // Create opportunities and sync suggestions
     log.info(`[canonical] Creating canonical opportunity and syncing ${sortedSuggestions.length} suggestions`);
 
     const opportunity = await convertToOpportunity(
--- a/src/common/delivery-engine.js
+++ b/src/common/delivery-engine.js
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2026 Adobe. All rights reserved.
+ * This file is licensed to you under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+ * OF ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ */
+
+/**
+ * Per-site V1 -> V2 cutover: a site's opportunity is owned by the Mystique blackboard
+ * producer cascade — rather than this legacy audit — when its per-audit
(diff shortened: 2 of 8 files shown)


lucianfelix:
Reviewed on Lucian's behalf via Claude Code (companion to mystique#4510).

**Verdict: MERGEABLE.** Helper is correct, gates are wired at the right layers, tests cover the negative cases (no `convertToOpportunity`, no `sqs.sendMessage`, `Opportunity.create` not called). Codecov green. No blocking issues.

## Non-blocking (follow-up ok)
1. **Outdate set narrower than the CWV precedent.** `resolveOpportunityIfNoIssues` (the shared helper reused by all three audits) outdates ONLY `NEW` + `PENDING_VALIDATION`. `resolveLegacyCwvOpportunity` in `src/cwv/handler.js:52-81` outdates everything EXCEPT `OUTDATED` + `FIXED` — so it also catches `IN_PROGRESS` / `SKIPPED` / `ERROR`. Concrete divergence: a live `IN_PROGRESS` legacy suggestion survives the flip on canonical/metaTags/BIL but wouldn't survive on CWV. Product call — please confirm the narrower set is intentional (or align both on one helper).
2. **Canonical gate is placed AFTER expensive detection.** `src/canonical/handler.js:782`. `metatags` gates at the TOP of `runAuditAndGenerateSuggestions` (`src/metatags/handler.js:301`, before autodetect + SSR validation + traffic calc); canonical still runs the full S3 fetch, per-URL HTTP `validateCanonicalRecursively`, and full suggestion assembly BEFORE the gate. Functional behavior is correct; wasted work every scan for bow-out sites. Move to top of `processScrapedContent` in a follow-up.
3. **`isCwvBlackboardEngine` is not retired.** `src/cwv/handler.js:37-39` is now a duplicate of `isBlackboardEngine`. The new helper is strictly more null-safe (`site?.getDeliveryConfig?.()` vs `site.getDeliveryConfig?.()`), so it's a pure win to route CWV through it. Either delete in this PR or file a follow-up — keeping both invites drift.

## Nits
- **Canonical bow-out falls through to Elmo** (`src/canonical/handler.js:782` → line 824-890). PR body says intentional ("Elmo generic-opportunity is intentionally left running"); confirmed. A `// intentional fall-through to Elmo path` marker at the closing `}` would prevent a future refactor from adding an early return.
- **BIL test's outdate assertion is weak** — `bulkUpdateStatus.calledOnce` in `test/audits/internal-links/opportunity-suggestions-step.test.js` doesn't verify the target status. Add `.calledWith(sinon.match.array, 'OUTDATED')` for a stronger regression pin.
- **`context.dataAccess` vs `{ dataAccess }` inconsistency in canonical** — line 721 destructures, line 784 uses `context.dataAccess`. Cosmetic.
- **BIL gate fires before the `!success` check** — this is correct (M owns the opportunity, cleanup should proceed even on a failed audit) but the ordering looks accidental; a comment saying so would help.

## Cross-repo note
Companion PR mystique#4510 writes these flags today. Merge order doesn't matter — the M panel honestly reports `inert_pending_spacecat` (grey dot) until this deploys, then a one-line-per-opp registry flip in M turns those three green.

## Merge plan
Approving. Will merge alongside mystique#4510 once its CI re-runs on the doc-index fix commit.

lucianfelix (approved):
Approved — see summary comment above. No blocking issues; 3 non-blocking + a few nits worth a follow-up. Helper is correct, gates wired at the right layers, negative-case tests present, codecov green.

PR #2922, opened 2026-08-27
feat(brand-claims): llmo-7177 semrush-aware trigger + ingest_source/domain on event

rachel-kittens:
## Summary
Makes the `brand-claims` audit decide **DRS-sheet vs Semrush per site** and stamps `ingest_source` on the existing `BRAND_PRESENCE_SHEET_WRITTEN` event, so Mystique can source the right feed. Same queue, additive/backward-compatible. Implements LLMO-7177.

Consumer side of this contract: **experience-platform/mystique#4642** (LLMO-7178).

## What it does
- **Per-site decision** in `src/brand-claims/handler.js`, reusing the offsite helpers: `resolveSemrushEntitlement` (fail-closed, reason codes) + `resolveEnableSemrush` (`BRAND_CLAIMS_SEMRUSH_ENABLED` env + per-run Slack `enableSemrush` override, override wins).
- **Semrush branch** (enabled **and** entitled): skips the DRS S3-sheet lookup; emits `ingest_source: "semrush_feed"`, replay-stable `week`/`year` (an explicit window from `auditContext.messageData` wins, else the current ISO week — so a retry/DLQ redrive across the UTC week boundary re-emits the same bucket), `cadence` (`weekly` default), and `null` `sheet_date`/`s3_bucket`/`s3_key`. Runs **before** the `brandSlug` guard, so a brand whose name sanitizes to empty can still take the feed path.
- **DRS branch** (default / disabled / non-entitled / inconclusive check): unchanged behavior, now stamped with explicit `ingest_source: "brand_presence_s3"`. A Semrush hiccup can never zero out a site that has a DRS sheet.

## No `domain` on the event (spec #248 correction)
An earlier revision stamped a registrable `domain`. That's been **removed**: `domain` is a per-*market* concept (each market has its own domain; the brand has none), and the downstream Serenity proxy resolves workspace/project/market from `organization_id` + `brand_id` — both already on the event. A brand-primary-site domain would mis-bind multi-market brands (mysticat-architecture#248). So the Semrush payload carries **no `domain`**; identity is `org_id` + `brand_id`.

## Event contract
Additive `ingest_source` (optional, defaults to `brand_presence_s3` when absent, so a pre-7177 consumer is unaffected); both branches now stamp it explicitly.

## Tests
Entitlement-gate matrix (entitled / flag-disabled / no-workspace / transient / no-client → DRS fallback), Semrush payload shape (no sheet, no `domain`, `ingest_source: semrush_feed`), per-run cadence, Slack-override precedence, replay-stable `week`/`year` (frozen clock, explicit-window override), and the DRS deep-equal (with the additive `ingest_source`). 100% coverage gate green.

## Notes
- **Dormant until migration:** the Semrush branch only fires with `BRAND_CLAIMS_SEMRUSH_ENABLED` on (default off) or a Slack override, **and** an entitled brand. Flag off → every brand takes the unchanged DRS path.
- Docs: `docs/specs/2026-08-11-brand-claims-scheduled-trigger.md` + `docs/decisions/006-brand-claims-drs-vs-semrush-source.md` updated.
- Live Semrush runs remain gated on the Semrush-side deliverables (LLMO-6836 auth + serenity endpoints); this ticket is buildable/testable now.

### DIFF
--- a/src/brand-claims/handler.js
+++ b/src/brand-claims/handler.js
@@ -15,48 +15,84 @@ import { ListObjectsV2Command } from '@aws-sdk/client-s3';
 import { ok } from '@adobe/spacecat-shared-http-utils';
 import { hasText, isoCalendarWeek } from '@adobe/spacecat-shared-utils';
 import { resolveSemrushEntitlement } from '../utils/semrush-entitlement.js';
-import { resolveEnableSemrush } from '../utils/offsite-audit-utils.js';
+import { resolveEnableSemrush, toApexHost } from '../utils/offsite-audit-utils.js';
 import { createOffsiteLogger, AUDIT } from '../utils/offsite-logging.js';
 
 const BP_PLATFORM = 'chatgpt_free';
 
 /**
  * `ingest_source` values on the `BRAND_PRESENCE_SHEET_WRITTEN` event. Additive contract
- * (LLMO-7177): the field is OPTIONAL and defaults to `brand_presence_s3` when absent, so
- * the DRS branch stays byte-identical to the pre-7177 event and only the Semrush branch
- * stamps a value. Mystique reads this to source the right feed (the DRS BP `.xlsx` in S3
- * vs the Semrush URL-Inspector feed keyed by `domain`).
+ * (LLMO-7177): the field is OPTIONAL and defaults to `brand_presence_s3` when absent (so a
+ * pre-7177 consumer is unaffected). BOTH branches now stamp it explicitly — the DRS branch
+ * with `brand_presence_s3`, the Semrush branch with `semrush_feed` — so Mystique can read
+ * it to source the right feed (the DRS BP `.xlsx` in S3 vs the Semrush URL-Inspector feed
+ * keyed by `domain`) without relying on the absent-means-DRS default.
  */
 const INGEST_SOURCE_DRS = 'brand_presence_s3';
 const INGEST_SOURCE_SEMRUSH = 'semrush_feed';
 
 /**
- * Cadence stamped on a Semrush-branch event when the brand config carries none. Claims runs
- * on a weekly cadence (see `isMondayPartition`), so `weekly` is the safe default; a future
- * per-brand cadence column on `brands` is honored automatically via `brand.cadence`.
+ * Cadence stamped on the Semrush-branch event. Claims runs on a weekly cadence (see
+ * `isMondayPartition`), and the `brands` projection (`getBrandForSite`) selects no cadence
+ * column — there is none today — so the value is a fixed default. When a per-brand `cadence`
+ * column is added to `brands`, extend that projection AND read it here (do not read
+ * `brand.cadence` today: it is always `undefined`).
  */
 const BRAND_CLAIMS_DEFAULT_CADENCE = 'weekly';
 
 /**
- * Derives the registrable domain for the Semrush feed from the site's base URL — the
- * www-stripped hostname, matching the `siteHostname` convention the offsite Semrush loader
- * (`offsite-brand-presence-semrush.js`) already uses for the same brand. Returns `null` when
- * no base URL is configured or it cannot be parsed (the caller then falls back to the DRS
- * path, since a `semrush_feed` event requires a `domain`).
+ * Coerces a Slack-delivered `week`/`year` override to a bounded positive integer, or
+ * `undefined` when absent/invalid. Slack custom args arrive as strings, so numeric strings
+ * are accepted alongside real numbers; anything outside the sane bound is rejected (treated
+ * as "not supplied") rather than trusted.
+ *
+ * @param {number|string} raw
+ * @param {number} min
+ * @param {number} max
+ * @returns {number|undefined}
+ */
+function toBoundedInt(raw, min, max) {
+  if (raw === undefined || raw === null || raw === '') {
+    return undefined;
+  }
+  const n = Number(raw);
+  if (!Number.isInteger(n) || n < min || n > max) {
+    return undefined;
+  }
+  return n;
+}
+
+/**
+ * Resolves the `(week, year)` the Semrush run is FOR. Prefers an explicit run window from
+ * `auditContext.messageData` (`week`+`year`, tri-state like `enableSemrush`: both must be
+ * present and valid to override), so a retry / DLQ redrive that crosses the UTC ISO-week
+ * boundary re-emits the SAME bucket for the same run rather than drifting to a new week.
+ * Falls back to the current ISO week when no valid explicit window is supplied.
+ *
+ * @param {object} [auditContext] - `message.auditContext`
+ * @returns {{week: number, year: number}}
+ */
+function resolveRunWeekYear(auditContext) {
+  const md = auditContext?.messageData ?? {};
+  const week = toBoundedInt(md.week, 1, 53);
+  const year = toBoundedInt(md.year, 2000, 9999);
+  if (week !== undefined && year !== undefined) {
+    return { week, year };
+  }
+  return isoCalendarWeek(new Date());
+}
+
+/**
+ * Derives the registrable domain for the Semrush feed from the site's base URL, reusing the
+ * shared `toApexHost` (`offsite-audit-utils.js`) so the eventual PSL upgrade stays a
+ * one-site change. Returns `''` when no base URL is configured or it cannot be parsed (the
+ * caller then falls back to the DRS path, since a `semrush_feed` event requires a `domain`).
  *
  * @param {object} site - Site model (`getBaseURL()`).
- * @returns {string|null}
+ * @returns {string}
  */
 function registrableDomainFromSite(site) {
-  const baseURL = site?.getBaseURL?.();
-  if (!hasText(baseURL)) {
-    return null;
-  }
-  try {
-    return new URL(baseURL).hostname.replace(/^www\./, '');
-  } catch {
-    return null;
-  }
+  return toApexHost(site?.getBaseURL?.());
 }
 const SHEET_FILENAME_RE = /-w(\d{1,2})-(\d{4})(?:-(\d{6}))?\.xlsx$/i;
 const KEY_DATE_RE = /\/(\d{4})\/(\d{2})\/(\d{2})\//;
@@ -247,18 +283,16 @@ export default async function brandClaimsHandler(message, context) {
   }
 
   const brandSlug = sanitizePathComponent(brand.name);
-  if (!brandSlug) {
-    log.warn(`brand-claims: brand name "${brand.name}" (${brand.id}) sanitizes to an empty S3 path component — cannot look up its sheet`);
-    return ok();
-  }
 
   // Per-site DRS-vs-Semrush decision (LLMO-7177). Default is the DRS Brand Presence sheet;
   // Semrush is chosen only when it is enabled for this run AND the brand is entitled to it.
   // The per-run Slack override (`enableSemrush`) takes precedence over the env flag, mirroring
   // offsite-brand-presence's `resolveEnableSemrush` pattern. Anything else — disabled,
   // non-entitled, or an inconclusive entitlement check — falls back to the unchanged DRS path,
-  // so a Semrush hiccup can never zero out a site that has a DRS sheet.
-  const olog = createOffsiteLogger(log, { audit: AUDIT.BRAND_PRESENCE, siteId: resolvedSiteId });
+  // so a Semrush hiccup can never zero out a site that has a DRS sheet. This runs BEFORE the
+  // `brandSlug` guard: the Semrush feed is keyed by `domain`, not the S3 brand-slug path
+  // component, so a brand whose name sanitizes to empty can still run on the feed path.
+  const olog = createOffsiteLogger(log, { audit: AUDIT.BRAND_CLAIMS, siteId: resolvedSiteId });
   const enableSemrushOverride = resolveEnableSemrush(message?.auditContext ?? {}, olog);
   const semrushEnabled = enableSemrushOverride
     ?? (env?.BRAND_CLAIMS_SEMRUSH_ENABLED === 'true');
@@ -274,8 +308,10 @@ export default async function brandClaimsHandler(message, context) {
         // key the feed, so fall back to the DRS sheet rather than emit an unroutable event.
         log.warn(`brand-claims: brand ${brand.id} is Semrush-entitled but site ${resolvedSiteId} has no parseable base URL for a registrable domain — falling back to the DRS sheet`);
       } else {
-        const { week, year } = isoCalendarWeek(new Date());
-        const cadence = brand.cadence || BRAND_CLAIMS_DEFAULT_CADENCE;
+        // Replay-stable window: an explicit `week`/`year` (Slack override / redrive) is used
+        // verbatim, else the current ISO week — so a DLQ redrive across the week boundary does
+        // not re-bucket the same run.
+        const { week, year } = resolveRunWeekYear(message?.auditContext);
         const event = {
           event_type: 'BRAND_PRESENCE_SHEET_WRITTEN',
           schema_version: 1,
@@ -285,7 +321,7 @@ export default async function brandClaimsHandler(message, context) {
           site_id: resolvedSiteId,
           week,
           year,
-          cadence,
+          cadence: BRAND_CLAIMS_DEFAULT_CADENCE,
           sheet_date: null,
           platform: BP_PLATFORM,
           s3_bucket: null,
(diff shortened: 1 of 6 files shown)


src/brand-claims/handler.js
```diff
@@ -210,6 +252,62 @@ export default async function brandClaimsHandler(message, context) {
...
+    });
+    if (entitlement.entitled) {
+      const domain = registrableDomainFromSite(site);
+      if (!hasText(domain)) {
+        // A semrush_feed event REQUIRES a domain; without a parseable base URL we cannot
+        // key the feed, so fall back to the DRS sheet rather than emit an unroutable event.
+        log.warn(`brand-claims: brand ${brand.id} is Semrush-entitled but site ${resolvedSiteId} has no parseable base URL for a registrable domain — falling back to the DRS sheet`);
+      } else {
+        const { week, year } = isoCalendarWeek(new Date());
+        const cadence = brand.cadence || BRAND_CLAIMS_DEFAULT_CADENCE;
```
cwjwisse: issue (blocking): [Important] `brand.cadence` can never be populated, so this branch is dead and the test that covers it cannot fail.

`brand` comes only from `getBrandForSite`, whose PostgREST projection is `.select('id, name, brand_claims_enabled')` (`src/brand-claims/handler.js:86`). PostgREST returns only the projected columns, so `brand.cadence` is `undefined` for every row the real query can return, today and equally after a `cadence` column is added to `brands`. The emitted `cadence` is therefore always `weekly`.

Why it matters: the comment at `src/brand-claims/handler.js:34-36` promises that "a future per-brand cadence column on `brands` is honored automatically via `brand.cadence`", and a green test backs that promise, so whoever adds the column will ship it believing the wiring exists. The test `honors a per-brand cadence from the brand config on the semrush event` (`test/audits/brand-claims.test.js:460-475`) passes only because the fake builder records and discards the column list (`test/audits/brand-claims.test.js:50`) while `selectResult` is stubbed with `cadence: 'daily'`. The same file pins the real projection at `test/audits/brand-claims.test.js:333`, so the two tests assert contradictory things about whether `cadence` can reach the handler. `cadence` is a cross-service field on the new contract, which is where a false green is most expensive.

How to fix: either add `cadence` to the projection at `src/brand-claims/handler.js:86` and update the assertion at `test/audits/brand-claims.test.js:333`, or drop the `brand.cadence ||` hook plus its test, stamp `BRAND_CLAIMS_DEFAULT_CADENCE` directly, and reword the comment to say the select list must be extended when the column lands.

src/brand-claims/handler.js
```diff
@@ -210,6 +252,62 @@ export default async function brandClaimsHandler(message, context) {
...
+      orgId: organizationId, brandId: brand.id,
+    });
+    if (entitlement.entitled) {
+      const domain = registrableDomainFromSite(site);
+      if (!hasText(domain)) {
+        // A semrush_feed event REQUIRES a domain; without a parseable base URL we cannot
+        // key the feed, so fall back to the DRS sheet rather than emit an unroutable event.
+        log.warn(`brand-claims: brand ${brand.id} is Semrush-entitled but site ${resolvedSiteId} has no parseable base URL for a registrable domain — falling back to the DRS sheet`);
+      } else {
+        const { week, year } = isoCalendarWeek(new Date());
```
cwjwisse: issue (blocking): [Important] Semrush `week`/`year` are derived from the wall clock, so the same logical run can emit two different weeks.

The DRS branch derives them from immutable sheet metadata (`week: sheet.week, year: sheet.year`, `src/brand-claims/handler.js:325-326`, parsed from the sheet filename), so redelivering the same SQS message emits an identical event. The Semrush branch uses `isoCalendarWeek(new Date())`, so a retry, DLQ redrive, or manual re-run that lands on the other side of the UTC ISO-week boundary stamps a different `(week, year)` for the same run and creates a second bucket. It also means a Semrush-path run for any week other than "now" is inexpressible, so there is no backfill or replay path on the new source.

The test has the same property: it recomputes `isoCalendarWeek(new Date())` after the await and feeds both values into the deep-equal (`test/audits/brand-claims.test.js:436-457`), so the two most rollout-sensitive fields on the new event are the only ones not pinned to literals, and the assertion can flake across the Sunday-to-Monday UTC boundary.

How to fix: derive the Semrush `week`/`year` from an explicit run window rather than `new Date()` (for example an optional `week`/`year` from `auditContext.messageData`, defaulting to the current ISO week, using the tri-state override pattern this file already imports), and freeze the clock in the test (`sandbox.useFakeTimers`) so the assertion pins literal values.

src/brand-claims/handler.js
```diff
@@ -210,6 +252,62 @@ export default async function brandClaimsHandler(message, context) {
...
+          brand: brandSlug,
+          site_id: resolvedSiteId,
+          week,
+          year,
+          cadence,
+          sheet_date: null,
+          platform: BP_PLATFORM,
+          s3_bucket: null,
+          s3_key: null,
+          ingest_source: INGEST_SOURCE_SEMRUSH,
```
cwjwisse: issue (blocking): [Important] The in-repo spec for this handler is marked Implemented and still describes the single-branch behavior this PR replaces.

`docs/specs/2026-08-11-brand-claims-scheduled-trigger.md` carries `Status: Implemented` and is the design of record for this handler. Two of its numbered run steps are now wrong:

- Step 4 ("Run") says the handler builds the S3 prefix `{siteId}/{brandSlug}/analytics/chatgpt_free/` and selects the latest sheet. The Semrush branch skips the S3 lookup entirely, which the PR's own test asserts (`expect(s3Client.send).to.not.have.been.called`).
- Step 5 ("Emit") enumerates the event payload as `organization_id`, `brand_id`, `brand`, `site_id`, `week`, `year`, `cadence: "weekly"`, `sheet_date`, `platform`, `s3_bucket`, `s3_key`. The event now also carries `ingest_source` on both branches and `domain` on the Semrush branch, and `sheet_date` / `s3_bucket` / `s3_key` are null on that branch.

The spec is also silent on the new per-site source decision and the `BRAND_CLAIMS_SEMRUSH_ENABLED` gate. `CLAUDE.md` states "Specs should reflect the implemented state once work is complete" and requires ADRs for "technical or architectural decisions with long-term impact"; the analogous offsite Semrush source swap shipped both (`docs/specs/2026-08-05-offsite-cited-urls-semrush-source.md`, `docs/decisions/002-offsite-cited-urls-semrush-source.md`). This is the artifact a reader consults to learn what the event looks like, and it is in this repo, so it is fixable on this branch.

How to fix: update the spec's steps 4 and 5 for the two-branch behavior and the two new fields, and add an ADR for the DRS-vs-Semrush source decision, or state in the spec why an ADR is not warranted here.

cwjwisse (changes requested):
Hey @rachel-kittens,

**Verdict: Request changes** - the decision structure and fail-to-DRS fallbacks are sound, but the per-brand cadence hook cannot fire, a new test certifies that it can, and the repo's implemented-status spec for this handler is left describing the old single-branch behavior.
**Complexity: MEDIUM** - 345 changed lines across 2 files, single service, no high-risk path touched.
_Changes: makes `brand-claims` pick DRS-sheet vs Semrush per site and stamps `ingest_source`/`domain` on the existing `BRAND_PRESENCE_SHEET_WRITTEN` event (2 files)._
Note: Recommend a human read before merge - possible divergence from the documented design (heuristic, flagged by the conventions reviewer): this PR changes the event contract that `docs/specs/2026-08-11-brand-claims-scheduled-trigger.md` (Status: Implemented) pins, and the related architecture spec is still `status: proposed` and does not mention the new fields. The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Important] `brand.cadence` is never returned by the brands query, so the per-brand cadence branch is dead and its test asserts an unreachable path - `src/brand-claims/handler.js:278` (details inline)
2. [Important] Semrush `week`/`year` come from the wall clock at emit time, so the event is not replay-stable and its test assertion is clock-derived - `src/brand-claims/handler.js:277` (details inline)
3. [Important] `docs/specs/2026-08-11-brand-claims-scheduled-trigger.md` is marked Status: Implemented and pins both the run steps and the emitted payload this PR changes, and it is not updated here - `src/brand-claims/handler.js:293` (details inline)

<details>

- nit: the contract comment says the DRS branch "stays byte-identical to the pre-7177 event and only the Semrush branch stamps a value", but the DRS branch now stamps `ingest_source` too - `src/brand-claims/handler.js:24-28` vs `src/brand-claims/handler.js:332`
- nit: the `!brandSlug` guard now also blocks the Semrush branch, which needs no S3 path component, while its message still says "cannot look up its sheet" - `src/brand-claims/handler.js:250-252`
- nit: `olog` is bound to `AUDIT.BRAND_PRESENCE`, so brand-claims override warnings emit as `audit=brand-presence` and land in the offsite-brand-presence taxonomy slot - `src/brand-claims/handler.js:261`
- suggestion: reuse the exported `toApexHost` (`src/utils/offsite-audit-utils.js:222`) instead of a fourth inline www-strip, so the declared PSL follow-up stays a one-site change - `src/brand-claims/handler.js:50-60`
- suggestion: add coverage for a non-www subdomain base URL and for an invalid `enableSemrush` value; both are new input classes with no test - `test/audits/brand-claims.test.js:388`, `test/audits/brand-claims.test.js:477`
</details>

cwjwisse:
Thanks Rachel. One contract issue surfaced in the `#tmp-claims-semrush-elements` thread: the configured-prompt feed Brand Claims needs is **not** the ai-visibility surface (that's Semrush's global catalog, keyed by domain) — it's projectengine (`/serenity/prompts`) + Elements, keyed by **`workspaceId`+`projectId`** (endpoint map: serenity-docs#421).

This handler stamps `{ ingest_source: 'semrush_feed', domain: registrableDomainFromSite(site) }` on `BRAND_PRESENCE_SHEET_WRITTEN`, and the mystique consumer (#4642) keys the feed off that `domain`. `domain` alone can't address a customer's configured prompts — the event needs to carry the workspace/project identity (or enough for the consumer to resolve it).

Since this PR is base=`main` (not stacked on the consumer), the event-schema change has to land in lockstep with #4642's storer — producer and consumer must agree on the new key before the flag is enabled anywhere. The entitlement gate, cadence, and DRS-default behavior are all unaffected; it's specifically the Semrush-branch payload key that changes.

cwjwisse:
**Correcting my earlier comment on this PR — the event needs no new field after all.**

Above I said `domain` alone can't address a customer's configured prompts and that *"the event needs to carry the workspace/project identity."* That was wrong, and since the correction landed on the consumer PR in a different repo it never made it back here. Flagging it now because it changes what this PR has to do.

**There is no brand-level `project_id` to stamp.** A Semrush project is one `(geoTargetId, languageCode)` *market* slice, not a brand — `brand_to_semrush_projects` is keyed `(brand_id, semrush_location_id, language)`, one row per market. A brand maps to a *workspace*; its markets map to projects. So the audit worker could not resolve a single `project_id` for a brand even in principle.

**The event already carries what the consumer needs.** The Serenity proxy is keyed on `organization_id` + `brand_id`, both already on `BRAND_PRESENCE_SHEET_WRITTEN`, and it resolves the workspace/project hop itself. The two hops stay separate: `orgId` + `brandId` on the Adobe→proxy hop, `workspaceId` + `projectId` on the proxy→Semrush hop only.

**So the only change worth making here is *removing* `domain`** from the Semrush branch — no field is added, and there is no lockstep schema coordination with the consumer. That also removes an untrusted input from the path, which matters more now that reads run under a broadly-entitled credential.

Separately, the Semrush-side data dependency I flagged as blocking is **resolved**: the answer↔sources record is reconstructable by joining `141adc88` to `404fb017` on `(project_id, prompt, model, date)`, after measuring that at most one execution exists per tuple. Details in [adobe/mysticat-architecture#248](https://github.com/adobe/mysticat-architecture/pull/248) §1.3. Nothing in this PR is gated on it.

cwjwisse:
Correction to my comment above — **the `domain` removal is already done on this branch**, and I missed it.

`1ac0bee5` (2026-09-01) dropped it from the Semrush payload, two days before I commented. Verified at head: neither event literal carries a `domain` key (`src/brand-claims/handler.js:299-320` for the Semrush branch, `:345-362` for DRS), the reasoning is recorded inline, and there is a negative assertion pinning it at `test/audits/brand-claims.test.js:460`. The inline comment states the rationale more precisely than my own did:

> No `domain`: it is a per-market concept (each market has its own domain, the brand has none) … resolved from `org_id` + `brand_id` downstream

So the substance of my comment stands but the ask does not: **there is nothing left to change here on the event contract.** Apologies for the noise.

The rest of that comment is still accurate — no brand-level `project_id` exists, the event needs no new field, and the Semrush-side data dependency is resolved via the `141adc88`↔`404fb017` join.

One mechanical note unrelated to the above: this PR currently reports `mergeable: CONFLICTING` against `main` and will need a merge before it can land.

cwjwisse:
Ran the same multi-agent review pass over this PR. **No blocking findings** — I verified the four things I most wanted to check and they all hold:

- **`domain` is gone** from both event literals (`handler.js:299-320`, `:345-362`), pinned by the negative assertion at `test/audits/brand-claims.test.js:460`.
- **Entitlement gating is fail-closed** and correctly ordered — the `brand_claims_enabled` opt-out runs before the source decision, both `entitled:false` and `resolved:false` fall through to DRS, and the Slack override can only *request* entitlement, never bypass it.
- **Cadence** stamps a constant `weekly`; the dead `brand.cadence` read from the prior review is gone and guarded by a test.
- **DRS default path preserved** — flag default off, strict `=== 'true'`, and the only payload delta is the added `ingest_source`, pinned by the pre-existing deep-equal at `:294`.

Two things worth a decision, both about that one added field:

**1. The `ingest_source` stamp ships live and unflagged on the DRS path.** `handler.js:359` adds it to every DRS event on merge — not behind `BRAND_CLAIMS_SEMRUSH_ENABLED` — so 100% of sites start emitting a new field immediately, against whatever consumer is currently deployed. The body's "dormant until migration" note does not cover this path.

That is safe only if the deployed `BrandPresenceSheetWrittenEvent` tolerates unknown fields. I could not confirm it from here — `model_config` sits outside the diff hunks. **[unverified]** But the mystique codebase uses `ConfigDict(extra="forbid")` in 28 places, so it is not a hypothetical: under `forbid`, every DRS event poison-drops until the consumer deploys.

Two ways to close it: confirm `model_config` on that model and state the deploy order in the body (producer-first is only safe under `ignore`/`allow`), or stamp `ingest_source` **only on the Semrush branch** and let absent mean DRS — which the contract at `handler.js:24-31` already defines, and which also resolves the next point.

**2. The explicit `brand_presence_s3` token makes the consumer's `CLAIMS_INGEST_SOURCE` inert for event-driven runs.** The companion consumer PR is explicit that `None` must mean "no per-message override", because *"defaulting to a real token here would make an omitted field indistinguishable from an explicit `brand_presence_s3` and silently override a deployment's env default."* Sending an explicit token on every DRS event is exactly that override.

Both choices are defensible — this may well be intended. The gap is that ADR 006 Decision 4 argues only the upside ("so the consumer never has to infer it") and records no consequence for the env default. One line in Consequences stating the env fallback is intentionally bypassed for event-driven runs would settle it.

A few nits, lowest value first, take or leave:

- **PR title still advertises `domain`**, removed in `1ac0bee5`. Kodiak squash makes the title the release commit message, so the stale claim lands in `CHANGELOG.md` permanently.
- **Entitlement reason literals drift**: tests and ADR 006 use `flag_disabled`/`no_workspace`/`check_failed`/`no_client`, but `semrush-entitlement.js:83-90` exports the *hyphenated* forms, and that is what gets logged. An operator grepping Splunk for the ADR's underscored value finds nothing. Importing `SEMRUSH_ENTITLEMENT_REASONS` in the test would enforce it — which is what that module's own "so a future rename can't silently drift" comment intends.
- **`toBoundedInt` accepts non-numeric types** (`handler.js:55-63`): `Number(true) === 1`, so `{ week: true }` silently emits `week: 1`.
- **`year` bound is wider than the consumer's** — 9999 here vs `le=2100` downstream, so an override in that range is accepted and then rejected downstream, with a success log on this side.
- **Dead `domain` scaffolding in tests** (`:388`, `:401`, `:412`, `:657` — `BASE_URL`/`getBaseURL`); the handler reads no base URL at head.
- **Structured logging half-adopted**: the olog is built at `handler.js:285` but the two decision-outcome lines at `:323`/`:329` are free-text with interpolated ids, including raw `brand.name` — against `offsite-logging.js:26-30`. Those lines are the only visibility for a silent Semrush→DRS degrade.

Also confirmed, not flagged: CI and `codecov/patch` are green at head, so the coverage-gate claim is CI-backed; ADR placement and numbering follow the repo conventions; and the change matches the design of record, which requires dropping `domain` and adding no identity field.

PR #2935, opened 2026-09-02
fix(cwv): gate CWV suggestions on a minimum RUM sample count (SITES-50756)

tkotthakota-adobe:
## Problem

Vonage repeatedly reported ASO CWV LCP values wildly inconsistent with Google PageSpeed Insights / CrUX, eroding confidence in the tool ([SITES-50756](https://jira.corp.adobe.com/browse/SITES-50756)):

- `/unified-communications/campaigns/vbc-long-cse/` — ASO **31.3s LCP (Critical)** vs PSI field desktop **1s (passing)**
- `/contact-centers/campaigns/open-cti/` — ASO **11.3s LCP**

## Root cause — small-sample p75

Not a measurement bug: a **p75 computed from a handful of RUM samples** is dominated by a single outlier navigation (a bot, a DataDome bot-mitigation challenge, a cold-cache hit). Pulling the live RUM the audit itself queries confirmed it:

| Page | Weighted PV | Bucket | Desktop LCP (p75) | **Actual samples** |
|---|---|---|---|---|
| `…/vbc-long-cse/` | 600 | top-N | 31,340 ms 🔴 | **n = 2** |
| `…/open-cti/` | 2,100 | top-N | 11,250 ms 🔴 | **n = 1** |
| `/` (homepage) | 38,200 | homepage | 2,548 ms | n = 129 |

The audit's page-selection buckets (homepage, top-N-by-pageviews, ≥7000/wk threshold) had **no minimum-sample floor** on the homepage/top-N buckets, and `hasFailingMetrics` only checked `p75 > threshold` — never how many measurements produced that p75. PSI/CrUX don't show these because they gate on a minimum sample count and average over 28 days; ASO reacts faster and so ate the noise.

A pageviews floor alone is insufficient — `open-cti` has 2,100 weighted pageviews but only **1** desktop LCP sample (weighted pageviews count every navigation; `lcpCount` only counts navigations that emitted a `cwv-lcp` checkpoint). The correct lever is the **per-metric measurement count**, which the RUM client already returns (`lcpCount`/`clsCount`/`inpCount`).

## Change

- Introduce `MIN_METRIC_SAMPLES = 10`. A threshold breach counts as failing only when that metric's p75 is backed by ≥ 10 measurements, evaluated **per metric per device** (`hasFailingMetrics` / `filterToFailingDeviceMetrics`).
- **Homepage is exempt** (always reported, per product decision) — its suggestions waive the floor.
- Pages observed this run only with too few samples are also excluded from the `scrapedUrlsSet` OUTDATE coverage set, so a sparse week never resolves a real prior issue ([SITES-48436](https://jira.corp.adobe.com/browse/SITES-48436) intent preserved).
- Export `isHomepage` from `cwv-audit-result.js` for reuse.

With `n ≥ 10`, both Vonage false Criticals are suppressed while the 129-sample homepage and other well-sampled pages are unaffected.

## Testing

- TDD: added unit tests for `hasFailingMetrics` (floor enforced / waived), `filterToFailingDeviceMetrics`, `hasReliableSamples`, `MIN_METRIC_SAMPLES`; and integration tests (low-sample page → no suggestion; homepage low-sample → still suggested; sparse page's prior suggestion not OUTDATED).
- `src/cwv/opportunity-sync.js` and `src/cwv/cwv-audit-result.js` are at **100% coverage**.
- Existing fixtures updated so their intentionally-failing pages clear the floor.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
scope: single-repo
relatedPRs: []
rationale: "Adds a per-metric minimum RUM sample-count floor so noisy low-sample p75 metrics no longer spawn false Critical CWV suggestions; confined to the cwv audit handler. Worst case is a few CWV suggestions suppressed or reshown — bounded, reversible by redeploy, and recovered by the next weekly re-audit."
recommendations: "MIN_METRIC_SAMPLES is a single tunable constant; adjust if the floor proves too strict/lax against production RUM."
backout: "Redeploy the previous release; CWV suggestions self-heal on the next weekly audit."
```


tkotthakota-adobe:
https://github.com/adobe/spacecat-audit-worker/pull/2935

tathagat2241 (approved):
## Review: fix(cwv): gate CWV suggestions on a minimum RUM sample count (SITES-50756)

Bugfix (+329/−32): adds a per-metric-per-device minimum RUM sample floor (`MIN_METRIC_SAMPLES = 10`) so a p75 computed from a handful of samples (a bot, a bot-mitigation challenge, a cold-cache hit) no longer spawns false Critical CWV suggestions — the Vonage 31.3s-LCP-from-n=2 problem. I reviewed the gate logic, the homepage/group handling, and the OUTDATE-coverage change.

### Blockers
- None.

### Should-fix
- None.

### Nits / worth confirming
- **The OUTDATE guard is page-level while the NEW-suggestion gate is per-metric.** `hasReliableSamples` returns true if *any* metric on *any* device clears the floor, so a page reliable on LCP (n=129) but sparse on INP (n=2) enters `scrapedUrlsSet`. If that page's prior suggestion concerns the sparse metric, it could be aged out to OUTDATED even though that metric has no reliable signal this run — the per-metric preservation intent of SITES-48436 isn't fully mirrored at OUTDATE granularity. It's bounded and self-healing (a wrongly-aged suggestion resurfaces on the next weekly audit, exactly as the CM block says), and CWV suggestions are per-URL so the practical blast radius is small — but worth confirming against `syncSuggestions`' matching granularity if per-metric fidelity matters.
- **`MIN_METRIC_SAMPLES = 10` is applied uniformly to single pages and groups.** For a group aggregating many pages, 10 *total* samples is a low bar (real groups clear it easily, so no over-suppression), and for a single page it's the noise floor — reasonable, and documented as a single tunable. Just noting one constant serves two population sizes.

### Strengths
- **Correct root-cause lever.** It gates on the per-metric *measurement count* (`lcpCount`/`clsCount`/`inpCount`), not pageviews — and the PR correctly diagnoses why pageviews alone were insufficient (weighted pageviews count every navigation; `lcpCount` counts only navigations that emitted a `cwv-lcp` checkpoint — hence open-cti's 2,100 PV but n=1 LCP). Verified the lookup: `deviceMetrics[`${metric}Count`] || 0`, so a missing count → 0 → suppressed when enforced.
- **Clean, DRY refactor.** A single `isDeviceMetricFailing` helper drives `hasFailingMetrics`, `filterToFailingDeviceMetrics`, and (via `metricSampleCount`) `hasReliableSamples`, so the threshold + floor semantics are identical and per-metric-per-device everywhere. The threshold boundary is unchanged from the original (`> THRESHOLDS[metric]`).
- **Groups are safely gated — I checked the risk and cleared it.** My concern was that gating groups (only the homepage is exempt) would silently suppress *all* group suggestions if group entries lacked counts; the fixture confirms `type: "group"` entries carry real per-metric counts (`lcpCount: 8`, `clsCount: 7`, …), so the floor discriminates rather than blanket-drops.
- **Homepage exemption + OUTDATE preservation threaded consistently.** `enforceMinSamplesFor = (entry) => !isHomepage(entry, baseURL)` is applied to both the failing filter and `scrapedUrlsSet`, matching the "homepage always reported" product decision; and low-sample pages are excluded from the OUTDATE coverage set so a sparse week can't resolve a real prior issue, preserving SITES-48436 intent. `isHomepage` is exported cleanly for reuse.
- **Excellent test coverage.** Unit tests for the floor enforced/waived and device-row stripping, plus integration tests for low-sample→no-suggestion, homepage-low-sample→still-suggested, and low-sample-prior→not-OUTDATED; fixtures updated so intentionally-failing pages clear the floor; 100% coverage on both touched source files. Conservative, accurate CM block (bounded, reversible, self-heals next weekly audit).

### Tests
Comprehensive and behavior-focused — I read the test list and it covers every branch of the new logic. I couldn't execute the suite, so the coverage figures are trusted; the logic and fixture shape I verified directly.

### Verdict
**Approve.** Correct, well-scoped fix that targets the right lever (per-metric sample count, not pageviews), handles homepage/group/OUTDATE consistently, and is thoroughly tested. The only thing I'd note is the page-level-vs-per-metric granularity of the OUTDATE guard — bounded and self-healing today, worth a glance if you later want per-metric preservation fidelity.

tkotthakota-adobe:
Thanks for the review. Addressed the two nits in `dae863af`:

**1 — OUTDATE guard page-level vs per-metric gate.** Confirmed and fixed. The gap was real: `hasReliableSamples` put a page in `scrapedUrlsSet` if *any* metric cleared the floor, so a page reliable on LCP (n=129) but sub-floor on INP (n=2) could age out a prior INP suggestion on an n=2 non-signal — the per-metric preservation from SITES-48436, not mirrored at OUTDATE granularity. Fix:
- `buildCwvOutdateCoverage` builds a per-URL reliable-metrics map and an `isWithinCoverage` predicate that OUTDATEs a suggestion only when the page cleared the floor on **every metric that suggestion's own `issues[]` concern**. Homepage waives the floor (unchanged); group/pattern rows and legacy suggestions without per-metric `issues[]` keep the existing page-level behaviour.
- `syncSuggestions` didn't forward `isWithinCoverage` to `handleOutdatedSuggestions` (only `scrapedUrlsSet`), so I added the forward — additive, defaults to `null`, so every other caller's `scrapedUrlsSet` guard is unchanged.
- Unit tests cover every branch of both new helpers; the three existing OUTDATE tests now exercise the per-metric path. `opportunity-sync.js` stays at 100% coverage.

**2 — `MIN_METRIC_SAMPLES = 10` uniform for pages and groups.** Left as-is, intentionally: as you noted, real groups clear 10 total samples easily (no over-suppression) and 10 is the right single-page noise floor, and it's one documented tunable. Kept as one constant.

Two pre-existing "creates a new opportunity object" tests fail on my local checkout (they also fail on the unmodified branch head and are green in CI), so those are an env artifact on my side, not from this change.

ramboz (approved):
## Summary
Adds a per-metric minimum RUM sample floor (`MIN_METRIC_SAMPLES = 10`) so a p75 built from 1-2 samples can't spawn a false Critical CWV suggestion, with the homepage exempted and the OUTDATE-coverage logic updated so a sparse week can't wrongly resolve a real prior issue. This is a second pass on top of tathagat2241's review, which already went deep on the gate logic, groups, and OUTDATE coverage and flagged two nits. Ready to merge.

## Strengths
- The follow-up commit (`dae863af`) does what the PR description claims. `buildCwvOutdateCoverage` in [opportunity-sync.js:170](src/cwv/opportunity-sync.js) now OUTDATEs a suggestion only when every metric its own `issues[]` reference was reliably measured this run, not just "any metric on the page cleared the floor". I traced this end to end: `isWithinCoverage` is threaded through `syncSuggestions` into `handleOutdatedSuggestions` in [data-access.js:290](src/utils/data-access.js), where it correctly takes precedence over `scrapedUrlsSet` (that precedence already existed for the broken-internal-links case from SITES-49911, so this PR only had to plumb the new predicate through, not build the mechanism).
- `MIN_METRIC_SAMPLES` uniform-for-pages-and-groups nit: left as-is with a clear one-line rationale in the reply. Reasonable, not worth relitigating.
- Test coverage for the fix is real, not just line coverage. `test/audits/cwv.test.js` has a dedicated `buildCwvOutdateCoverage` suite that exercises the exact scenario from the nit (LCP reliable, INP sub-floor, prior INP suggestion must survive) plus group rows, legacy suggestions with no per-metric `issues[]`, and a missing-cwv-array defensive case.
- `isHomepage` export and `enforceMinSamplesFor = (entry) => !isHomepage(entry, baseURL)` is a clean, minimal way to thread the exemption through both the failing-metrics filter and the coverage builder without duplicating the homepage check.

## Out of Scope, Worth Tracking
- Running `test/audits/cwv.test.js` against this branch locally, 22 of 91 tests fail with `TypeError: Cannot read properties of undefined (reading 'SITE')` inside `convertToOpportunity` (src/common/opportunity.js:112/131). I reran the same file against unmodified `main` with the same node_modules and got 19 failures with the identical error, so this is a pre-existing local/env issue (likely a stale dependency in the shared node_modules), not something this PR introduced. The 3 new tests this PR adds that touch `syncOpportunitiesAndSuggestions` hit the same pre-existing wall locally, but Codecov and the CI-green state on the PR say the real CI run isn't affected. Worth someone eventually chasing why this whole test file can't run clean locally, since it's silently swallowing real assertions for anyone testing outside CI.

## Verdict
**Ready to merge:** Yes
**Reasoning:** The fix is correct, the follow-up commit genuinely closes both nits from the first review, and the new logic has direct unit coverage for the exact edge case it was written for.

*Reviewed with Claude Code.*

PR #2937, opened 2026-09-02
feat(brand-claims): forward enrichment mode on the ready-signal (LLMO-7312)

akshaymagapu:
## What

The `brand-claims` audit now forwards a run **`mode`** onto the `BRAND_PRESENCE_SHEET_WRITTEN` ready-signal. `mode: "enrich"` tells mystique's BP consumer to run the cache-safe **off-site opportunity link refresh** (force only `d_claim_citation_opportunity_matches` + `d_claims_delivered`; every expensive claims/LLM fact stays cached) instead of a full claims run.

Middle leg of a 3-repo change for [LLMO-7312](https://jira.corp.adobe.com/browse/LLMO-7312):
- adobe/spacecat-api-service — dedicated Slack command emits `mode: enrich` (in progress)
- **this PR** — spacecat-audit-worker: forward `mode` on the ready-signal
- experience-platform/mystique#4804 — BP consumer runs the enrichment (already open; must land first)

## How

- `readRunMode(message)` reads the mode from the audit message: folded into `data` (a JSON string or object), or an explicit top-level `mode` / `auditContext.mode`. Anything other than `"enrich"` is a normal full run.
- Stamps `mode` on the emitted event.
- Like `on_demand`, `mode: "enrich"` bypasses the per-brand enable gate here so an explicit operator refresh is emitted even for a brand that isn't on the weekly schedule.

## Testing

```
32 passing
```

- 5 new tests: mode via `data` JSON string, mode as top-level field, default `full`, malformed/unknown → `full`, and enrich bypassing the enable gate for a disabled brand.
- Updated the happy-path deep-equal to include `mode: 'full'`.
- `eslint` clean on the handler.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
rationale: "Adds a mode flag forwarded on an internal event; defaults to full run, easily reverted."
```

### DIFF
--- a/src/brand-claims/handler.js
+++ b/src/brand-claims/handler.js
@@ -260,23 +260,30 @@ export default async function brandClaimsHandler(message, context) {
   // Best-effort: the run has already been triggered, so a persistence failure must
   // not fail the handler (fail-open matches the api-service cooldown's own miss
   // handling — a missed record just means the next request isn't throttled).
-  try {
-    await Audit.create({
-      siteId: resolvedSiteId,
-      isLive: site.getIsLive(),
-      auditedAt: new Date().toISOString(),
-      auditType: 'brand-claims',
-      auditResult: {
-        onDemand: onDemand === true,
-        week: sheet.week,
-        year: sheet.year,
-        sheetDate: sheet.sheetDate,
-        sheetKey: sheet.key,
-      },
-      fullAuditRef: sheet.key,
-    });
-  } catch (err) {
-    log.warn(`brand-claims: failed to persist brand-claims audit for site ${resolvedSiteId} (run already triggered): ${err.message}`);
+  //
+  // Enrichment runs (LLMO-7312) are skipped: an off-site opportunity link refresh
+  // is not a claims "run", so it must NOT advance `auditedAt` — doing so would
+  // reset the on-demand cooldown (blocking a subsequent full run) and flip the
+  // elmo-ui cooldown state. The cooldown/UI track full claims runs only.
+  if (!isEnrich) {
+    try {
+      await Audit.create({
+        siteId: resolvedSiteId,
+        isLive: site.getIsLive(),
+        auditedAt: new Date().toISOString(),
+        auditType: 'brand-claims',
+        auditResult: {
+          onDemand: onDemand === true,
+          week: sheet.week,
+          year: sheet.year,
+          sheetDate: sheet.sheetDate,
+          sheetKey: sheet.key,
+        },
+        fullAuditRef: sheet.key,
+      });
+    } catch (err) {
+      log.warn(`brand-claims: failed to persist brand-claims audit for site ${resolvedSiteId} (run already triggered): ${err.message}`);
+    }
   }
 
   return ok();
--- a/src/brand-claims/handler.js
+++ b/src/brand-claims/handler.js
@@ -147,29 +147,15 @@ async function findLatestSheet(s3Client, bucket, prefix, log) {
 // Publishes the ready-signal for the brand's latest BP sheet, but only for brands whose
 // brand_claims_enabled gate is on. Enabling/disabling the gate is done out of band (the
 // api-service enable-brand-claims Slack command) and acts as the per-site opt-in list.
-// Reads the run mode from the audit message (LLMO-7312). api-service's run-audit
-// folds `mode` into the audit `data` (a JSON string, or object); an explicit
-// top-level `mode` or `auditContext.mode` is also honored. Anything other than
-// "enrich" is a normal full run.
-function readRunMode(message) {
-  let data = message?.data;
-  if (typeof data === 'string') {
-    try {
-      data = JSON.parse(data);
-    } catch {
-      data = null;
-    }
-  }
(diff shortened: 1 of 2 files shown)


src/brand-claims/handler.js
```diff
@@ -236,6 +242,9 @@ export default async function brandClaimsHandler(message, context) {
     s3_bucket: drsBpBucket,
     s3_key: sheet.key,
     on_demand: onDemand === true,
+    // Only stamp `mode` for enrichment; a full run omits it entirely so its event
+    // stays identical to the DRS weekly emit (mystique treats absent mode as full).
+    ...(isEnrich ? { mode: 'enrich' } : {}),
```
MysticatBot, line 247: issue (blocking): The enrichment `mode` is stamped on the outbound SQS event here but not on the audit row written by `Audit.create` below (~line 259). The comment block at ~line 251 explains that row supports the api-service 7-day on-demand cooldown via `getLatestAuditByAuditType("brand-claims")`. An enrichment run writes a new audit row with `onDemand: false`, which may reset the cooldown timer and block a subsequent on-demand full run.

Two options:

1. Skip `Audit.create` for enrichment runs (if the cooldown should not apply to enrichments).
2. Stamp `mode` in `auditResult` so the api-service can filter enrichment rows:

```js
auditResult: {
  onDemand: onDemand === true,
  ...(isEnrich ? { mode: 'enrich' } : {}),
  week: sheet.week,
  // ...
},
```

Option 1 is simpler if the cooldown is the only consumer of that row.

MysticatBot (changes requested):
Hey @akshaymagapu,

**Verdict: Request changes** - one blocking finding on audit-row side effects.
**Complexity: LOW** - small diff, single handler.
_Changes: Forwards a `mode` field onto the brand-claims ready-signal so mystique can run a cache-safe enrichment instead of a full claims run (2 files)._

## Must fix before merge
1. [Important] Enrichment runs write an audit row that may reset the on-demand 7-day cooldown - `src/brand-claims/handler.js:247` (details inline)

<details>

- nit: PR description references `readRunMode(message)` with `data`/`auditContext.mode` fallbacks that do not exist in the diff - update the description to match the actual top-level destructuring
- suggestion: `mode` comparison is case-sensitive; `'ENRICH'` or `'Enrich'` silently becomes a full run - consider normalizing with `mode?.toLowerCase()` - `src/brand-claims/handler.js:159`
- suggestion: add a test for `onDemand: true` combined with `mode: 'enrich'` to verify both fields appear on the emitted event - `test/audits/brand-claims.test.js:227`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 5m 36s | <b>Cost:</b> $3.99 | <b>Commit:</b> <code>503ae24737596bf5386da8e5ecee1781e9ed7d9d</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/brand-claims/handler.js
akshaymagapu: Good catch — went with Option 1 in `c` (latest push): `Audit.create` is now guarded on `!isEnrich`, so an enrichment run no longer writes a `brand-claims` audit row. That keeps `auditedAt` tracking full claims runs only, so an enrich can't reset the api-service on-demand cooldown or flip the elmo-ui state. Added a test asserting `Audit.create` is not called for `mode: 'enrich'` while the ready-signal is still emitted.

MysticatBot (approved):
Hey @akshaymagapu,

**Verdict: Approve** - prior blocking finding addressed; no new blocking issues.
**Complexity: LOW** - small diff, single handler.
_Changes: Forwards a `mode` field onto the brand-claims ready-signal so mystique can run a cache-safe enrichment instead of a full claims run, and skips audit-row persistence for enrichment runs to preserve the on-demand cooldown (2 files)._

<details>

- suggestion: `mode` comparison is case-sensitive - `'ENRICH'` or `'Enrich'` from an upstream caller silently degrades to a full run. Consider normalizing: `const isEnrich = mode?.toLowerCase() === 'enrich';` - `src/brand-claims/handler.js:156`
- suggestion: add a test for `{ onDemand: true, mode: 'enrich' }` to lock in the interaction between the two bypass flags (both stamped on the event, audit row still skipped) - `test/audits/brand-claims.test.js`
- suggestion: when `mode` is present but not `'enrich'`, a `log.info` would help operators detect misconfigured callers rather than silently falling through to a full run - `src/brand-claims/handler.js:157`
</details>

- Audit-row cooldown reset: `Audit.create` is now guarded with `!isEnrich`, so an enrichment run no longer writes a brand-claims audit row or advances `auditedAt`.


---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 8m 36s | <b>Cost:</b> $3.38 | <b>Commit:</b> <code>209b8a4dbff59e1800b51fb1b57e7c7ebfda7dc8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2929, opened 2026-08-31
Mark autofixed vulnerabilities FIXED only after a rescan confirms the fix

andresbott:
## Summary
The security-vulnerabilities autofix marked a Suggestion `FIXED` the moment the PR opened — so customers saw "Fixed" before anything merged, and the Suggestion read `FIXED` while its FixEntity was still `PENDING` (an internally inconsistent state). This reworks the verify side so a vuln reaches `FIXED` only when a rescan of the deployed environment confirms it's gone (or a human overrides).

## What
- On a rescan, a disappeared **autofixed (`IN_PROGRESS`)** suggestion is marked `FIXED` and its PR-open `PENDING` FixEntity is **promoted to `DEPLOYED` in place** — preserving the PR-url provenance and avoiding a duplicate entity (a fresh `DEPLOYED` is created only when no `PENDING` exists, so status never outruns a FixEntity).
- A disappeared **`NEW`** suggestion (the customer bumped the dependency themselves) ages to `OUTDATED`, not `FIXED`, so ASO doesn't take credit for fixes it didn't make.
- The all-clear ("no vulnerabilities found") branch now routes through the same path instead of blanket-`OUTDATE`-ing everything, so a just-fixed vuln becomes `FIXED` and already-`FIXED` suggestions are preserved; the opportunity still resolves.
- vulns is treated as author-only: the deployed-environment scan **is** the production signal, so `DEPLOYED` is terminal and the separate publish step is skipped.
- Two opt-in hooks are added to the shared reconcile helper (`IN_PROGRESS` candidacy + in-place FixEntity promotion); the backlinks path passes neither and is unchanged.

## Notes
- Implements Phase 2 + Phase 3 of the sec-vulns suggestion-lifecycle plan (locked decisions D1–D4). Phase 1 (autofix-worker dropping the optimistic `FIXED` write) is already merged.

PR #2939, opened 2026-09-03
feat(vulnerabilities): FixEntity-aware vuln lifecycle reconcile (self-fix + regression restart)

bottadobe:
## Summary

Security-vulnerabilities marked a Suggestion `FIXED` as soon as the autofix PR opened, while its FixEntity was still only `PENDING`. So the customer saw "Fixed" for a vuln that was still live in their environment, the two records contradicted each other, and nothing ever reacted to a vuln coming back or to a customer fixing it themselves.

The audit worker now reconciles the Suggestion/FixEntity pair against every deployed-environment scan, so `FIXED` only means the scan no longer finds the vuln.

## What

`reconcileVulnSuggestions` runs first and owns the FIXED-side decisions, joining each suggestion to its FixEntity. The existing `syncSuggestions` then only creates brand-new findings and refreshes data — it makes no status calls of its own.

The cases that do something:

| When | Suggestion | Fix entity |
|---|---|---|
| **Customer self-fix** — a `NEW` finding disappears | `NEW` → `FIXED` | create `DEPLOYED`, stamped `origin: customer-self-fix` |
| **Unvalidated finding drops out** — a `PENDING_VALIDATION` finding disappears | `PENDING_VALIDATION` → `OUTDATED` | untouched (never a confirmed fix) |
| **Rescan confirms the fix** — a `FIXED` suggestion's vuln is gone | stays `FIXED` | `PENDING` → `DEPLOYED` (+ `deployedAt`) |
| **Regression** — a fixed vuln shows up again | stays `FIXED`, kept as history, **plus** a fresh `NEW`/`PENDING_VALIDATION` finding | stays `DEPLOYED` |

On a regression we don't rewrite the old record — the original `FIXED` suggestion and its `DEPLOYED` fix are left exactly as they are, and a brand-new finding restarts the lifecycle. It's idempotent: if there's already an active finding for that vuln we don't open another, so it won't stack up one per scan while the vuln lingers.

Everything else — an asserted fix while the vuln is still present, an already-verified fix, a finding with no active fix — is deliberately left untouched.

## Notes

- All the logic lives in the vuln handler; the shared sync engine (`data-access.js`) is unchanged from `main`.
- The base sync is pinned to a status-neutral merge (`() => null`) so reconcile stays the only thing that moves a vuln's status — in particular a still-present `FIXED` suggestion (a regression) is left alone instead of being flipped back by the default merge.
- Bumps `@adobe/spacecat-shared-data-access` to pick up `FixEntity.ORIGINS.CUSTOMER_SELF_FIX` (#1911).
- Supersedes #2929 (the earlier in-progress-based approach on this branch).

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Alejandro Moratinos", "Andres Bott", "Ramon Bisswanger", "Andrei Dulvac", "Dominique Jäggi"]
rationale: "Changes vuln status reconciliation logic (reporting/tracking only, not an access-control boundary); a wrong status is self-correcting on the next scan by design."
```


docs/decisions/006-security-vulnerabilities-rescan-lifecycle.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Decision
+
+The audit worker (`src/vulnerabilities/handler.js`) is the reconciler. A single
+pass, `reconcileVulnSuggestions`, owns **every** vuln Suggestion status
+transition against each scan, joining each suggestion to its `FixEntity`. It runs
+**before** the base `syncSuggestions` (which then only creates brand-new findings
+and refreshes present ones, with a status-neutral merge). The decision table:
+
+1. **Customer self-fix.** An open finding (`NEW`/`PENDING_VALIDATION`) that
```
bisswanger: this is great - and nice for reporting :-).
I guess we would just need something on the UI to display that accordingly in the Deployed view

docs/decisions/006-security-vulnerabilities-rescan-lifecycle.md
```diff
@@ -0,0 +1,117 @@
...
+   `FixEntity.ORIGINS` value), so a self-fix is distinguishable from the automated
+   pipeline (`spacecat`).
+2. **Rescan confirms an asserted fix.** `FIXED + PENDING`, vuln **gone** → promote
+   `PENDING → DEPLOYED` (+ stamp `deployedAt`); the suggestion stays `FIXED`.
+3. **Regression.** `FIXED + DEPLOYED`, vuln **back** → **archive** the old
+   suggestion to `OUTDATED` and **open a fresh** `NEW` (or `PENDING_VALIDATION`)
+   finding for the live vuln; roll the fix `DEPLOYED → ROLLED_BACK`. The `FIXED`
+   record is never mutated in place — history stays clean.
+4. **Fresh assertion, vuln still present** → wait (no change).
+5. **Stale assertion.** `FIXED + PENDING`, vuln still present past **30 days**
```
bisswanger: if we change it to OUTDATED, it also disappears from the "Deployed" view - so there is no history anymore to the customer that someone ever pressed the "deploy" button.

should we rather leave the suggestion in FIXED and just have it "FAILED" with a reason?
Then users can see it in the ASO UI accordingly

docs/decisions/006-security-vulnerabilities-rescan-lifecycle.md
```diff
@@ -0,0 +1,117 @@
...
+and refreshes present ones, with a status-neutral merge). The decision table:
+
+1. **Customer self-fix.** An open finding (`NEW`/`PENDING_VALIDATION`) that
+   disappears with **no** FixEntity → Suggestion `FIXED` **and create** a
+   `DEPLOYED` FixEntity stamped `origin = customer-self-fix` (a new shared
+   `FixEntity.ORIGINS` value), so a self-fix is distinguishable from the automated
+   pipeline (`spacecat`).
+2. **Rescan confirms an asserted fix.** `FIXED + PENDING`, vuln **gone** → promote
+   `PENDING → DEPLOYED` (+ stamp `deployedAt`); the suggestion stays `FIXED`.
+3. **Regression.** `FIXED + DEPLOYED`, vuln **back** → **archive** the old
```
bisswanger: same as below - OUTDATED hides it from the customer.
So traces etc are gone from the UI

I would rather leave the old one "untouched" and just raise a new suggestion

bisswanger: also this contradicts a bit to down where DEPLOYED is terminal - and it's not listed as a valid status transition there either

docs/decisions/006-security-vulnerabilities-rescan-lifecycle.md
```diff
@@ -0,0 +1,117 @@
...
+
+`FIXED→OUTDATED`, `NEW→FIXED`, `PENDING→DEPLOYED`, `DEPLOYED→ROLLED_BACK`, and
+`PENDING→FAILED` are all legal transitions, and `STATUS_TRANSITION_ENFORCEMENT`
+defaults to `warn`. The one shared change is additive: the new
+`FixEntity.ORIGINS.CUSTOMER_SELF_FIX` value (`@adobe/spacecat-shared-data-access`
+≥ 4.28.0).
+
+## Consequences
+
+- `FIXED` means the vuln is actually gone from the deployed environment — verified
```
bisswanger: hmm no - FIXED rather means that the user started processing a fix - while PENDING it's not gone yet

Amoratinos (approved):

docs/decisions/006-security-vulnerabilities-rescan-lifecycle.md
dulvac: I'd call it  "origin:external-fix" instead of "origin: customer-self-fix."

andresbott: how important is this? the constant is already merged upstream

andresbott:
I never intended to commit the docs folder, removed now

PR #2940, opened 2026-09-03
fix(cdn-analysis): isolate aggregation by site host to prevent cross-site data bleed

claudiaboldis:
## Summary

Sites that share a single org-level BYOCDN bucket (bucket naming is per-org, not per-site — see `resolveCdnBucketName`) currently aggregate **identical** raw CDN log data into their own `aggregated_logs_<siteId>` / `aggregated_referral_logs_<siteId>` tables, because `insert-aggregated.sql` / `insert-aggregated-referral.sql` never filter on the row's own host — the site's base-URL host is only ever used to *exclude self-referrals*, never as a positive filter on which rows belong to the site being processed.

## Confirmed in production (LLMO-7337)

`firefly.adobe.com` and `community.adobe.com` share the exact same `cdnBucketConfig` (org `9E1005A551ED61CA0A490D45AdobeOrg`, `byocdn-cloudfront`). Downloaded and read the actual aggregated Parquet output for firefly's `siteId` across three dates spanning ~2.5 months — **100% of rows in every sample had `host = community.adobe.com`**, zero rows for firefly's own host:

| Date | Rows sampled | `host` field |
|---|---|---|
| 2026-06-20 hr03 | 67,627 | 100% `community.adobe.com` |
| 2026-08-15 hr04 | 30,168 | 100% `community.adobe.com` |
| 2026-09-02 hr09 | 21,076 | 100% `community.adobe.com` |

This is why firefly's Agentic Traffic dashboard has shown zero data since 2026-06-18: `cdn-logs-report`'s daily export correctly applies a host filter for `firefly.adobe.com`, and since every aggregated row actually belongs to `community.adobe.com`, it correctly finds nothing — the bug is upstream, in aggregation, not in the report/export layer.

## Fix

- `src/cdn-analysis/handler.js`: compute a site-scoped `hostPattern` (same `(www.)?` tolerance already used by `buildSiteFilters` at the report layer), reusing the existing `escapeForAthenaRegexLiteral` helper. Pass it to both `insert-aggregated` and `insert-aggregated-referral` template params (previously `insert-aggregated-referral` didn't receive any host value at all).
- All 7 CDN provider SQL templates (CloudFront, Akamai, Cloudflare, Fastly, FrontDoor, Imperva, Other), both `insert-aggregated.sql` and `insert-aggregated-referral.sql`:
  - Added `AND REGEXP_LIKE(<raw-host-column>, '{{hostPattern}}')` to restrict aggregation to the target site's own rows.
  - Referral files' `hosts` CTE (used to build the first-party-host set for self-referral exclusion) is scoped the same way — previously it collected every host sharing the raw path, so a legitimate referral *from* another site sharing the same bucket could have been wrongly treated as a self-referral and dropped.

## Impact / follow-ups

- This is likely not isolated to firefly/community — any pair of sites sharing an org-level BYOCDN bucket hits the same bug. Worth an audit of sites with duplicate `cdnBucketConfig.bucketName` values.
- No historical backfill included here — once this ships, a `backfill-llmo` re-run for affected sites should populate correct data going forward; the contaminated historical rows for the affected window aren't recoverable (the raw data was never wrong, just mis-attributed at aggregation time, so a re-run against the same raw retention window — if still available — would fix it).
- Full trail of the investigation that found this: [LLMO-7337](https://jira.corp.adobe.com/browse/LLMO-7337).

## Test plan

- [x] `npm run lint` — clean (aside from a pre-existing local `core.autocrlf` CRLF artifact unrelated to this change)
- [x] `npm test` — full suite passing, 100% statements/branches/functions/lines maintained
- [ ] Team review of the SQL changes across all 7 providers — flagging as draft since this touches every CDN ingestion path and I'd like eyes on the host-matching regex behavior (particularly FrontDoor's compound `COALESCE(...)` expression and Imperva's `s_computername`) before merge


src/cdn-analysis/handler.js
```diff
@@ -294,6 +294,12 @@ export async function processCdnLogs(auditUrl, context, site, auditContext) {
 
   const { year, month, day, hour } = getHourParts(auditContext);
   const { host } = new URL(site.getBaseURL());
+  // Multiple sites can share one org-level CDN bucket/raw-log path (bucket naming is
+  // per-org, not per-site - see resolveCdnBucketName). Without a positive host match,
+  // aggregation for this site would ingest every site's traffic sharing that path.
+  // Mirrors the (www.)? tolerance used by buildSiteFilters at the report layer.
+  const rootHost = escapeForAthenaRegexLiteral(host.replace(/^www\./, ''));
+  const hostPattern = `(?i)^(www\\.)?${rootHost}$`;
```
akshaymagapu: host is not always baseurl in cdn logs, there can be custom hosts that we set via cdnlogsfilters config.

this can fail for lot of sites

akshaymagapu (changes requested):
knowing host at cdn analysis is tricky and not always base url
filtering to baseurl - we can lose data. please be cautious with this change

src/cdn-analysis/handler.js
akshaymagapu: the reporting part will filter the actual site based on custom filter or default filter of host
https://github.com/adobe/spacecat-audit-worker/blob/main/src/utils/cdn-utils.js#L552

this was decided as lot of custom host values can exist and we dont know upfront the actual host value in raw data

claudiaboldis: Good catch — reworked to call `buildSiteFilters(site.getConfig()?.getLlmoCdnlogsFilter(), site)` directly instead of a baseURL-only pattern, so custom `cdnlogsFilter` overrides are respected the same way the report layer already handles them. Pushed in 728961bb.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

claudiaboldis: Right — reusing `buildSiteFilters` from `cdn-utils.js` directly (same function `query-builder.js` calls for reporting) instead of reimplementing host matching. Since its output references the aggregated table's column names (`host`, `x_forwarded_host`, etc.), each provider's raw-to-aggregated query is now a two-level SELECT: an inner subquery aliasing the raw per-provider columns to those generic names, then an outer query applying the site filter against them — so aggregation can never diverge from what the report layer would keep. Pushed in 728961bb.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/cdn-analysis/handler.js
```diff
@@ -294,6 +295,13 @@ export async function processCdnLogs(auditUrl, context, site, auditContext) {
 
   const { year, month, day, hour } = getHourParts(auditContext);
   const { host } = new URL(site.getBaseURL());
+  // Multiple sites can share one org-level CDN bucket/raw-log path (bucket naming is
+  // per-org, not per-site - see resolveCdnBucketName). Without a positive host match,
+  // aggregation for this site would ingest every site's traffic sharing that path.
+  // Reuse buildSiteFilters - the same host-matching logic the report layer already
+  // uses - so aggregation honors per-site cdnlogsFilter overrides instead of assuming
+  // raw host always equals the base URL (raw host values are not predictable upfront).
+  const siteFilterClause = buildSiteFilters(site.getConfig()?.getLlmoCdnlogsFilter(), site);
```
akshaymagapu: one concern on this approach.

The design intent of the CDN log filters is that they're adjustable at any time. Today the flow is:

Customer sends raw CDN logs, we aggregate everything into the aggregated bucket (host-agnostic).
cdn-logs-report applies buildSiteFilters at report time to produce what the user actually sees.

The value of applying the host filter at the report layer is recoverability. If a report comes back empty and the customer or AOE complains, we can inspect, see the host value doesn't match what we expected, update the filter, and just rerun the report. The aggregated bucket still holds all the data, so the fix is cheap and works retroactively.

With this change, the host filter moves up into aggregation, so the aggregated bucket only ever contains rows that matched the filter at aggregation time. If the filter was wrong, that data is no longer in the aggregated bucket, and reprocessing depends on the raw logs still being within retention. So the fix only works fast-forward, and we lose the ability to correct past reports by adjusting the filter.

akshaymagapu (changes requested):

akshaymagapu:
One possible middle ground: gate the aggregation-layer host filter on whether the host filter actually matches anything.

Before applying the filter, run a single Athena query against the data with the site's host filter. If any row comes back, the filter is good and we can safely apply host filtering at the aggregation layer. If zero rows come back, don't apply host filtering, keep everything, so we don't lock in a wrong or not-yet-known host before it has ever produced data.

I'd keep this in Athena rather than reading from postgres. The aggregation path runs with high concurrency, and adding a postgres read per run would put unnecessary load on the connection pool. An Athena probe keeps the check in the same layer we're already querying and avoids that.

That way we get the cross-site isolation for the sites where the host is confirmed working, without sacrificing recoverability for sites still being onboarded or debugged.

wdyt?

src/cdn-analysis/handler.js
claudiaboldis: That's a real cost I hadn't weighed enough — thanks for spelling it out. Implemented your middle-ground proposal in 41c9d544: added `check-site-filter.sql` per provider, run via `athenaClient.query` right before the INSERT. If it finds a matching row, the real `siteFilterClause` gets applied; if it finds nothing, we pass `'TRUE'` instead so aggregation stays host-agnostic (today's behavior) until a future run confirms the host. Recoverability for not-yet-confirmed sites is preserved, and we still get isolation once a site's host is known-good.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

claudiaboldis:
Implemented as described in 41c9d544 — thanks for the concrete design, this is a much better tradeoff than what I had.

- Added `check-site-filter.sql` per provider (7 files): the same raw-column-to-generic-name aliasing as `insert-aggregated.sql`'s inner subquery, but `SELECT 1 ... WHERE {{siteFilterClause}} LIMIT 1` instead of the full aggregation.
- In `handler.js`, this runs via `athenaClient.query` (not Postgres) right before building the INSERT queries, once per provider per run.
- If the probe returns a row, `siteFilterClause` (the real `buildSiteFilters` output) is used as-is. If it returns nothing, we substitute the literal `'TRUE'` so the WHERE clause becomes a no-op and aggregation stays host-agnostic — identical to today's behavior — until a later run confirms a match.
- Added two tests covering both branches, plus a log line (`site filter matched no raw rows ... aggregating without host filter`) so the fallback is visible in ops when it happens.

This keeps recoverability for sites still being onboarded/debugged while giving real isolation once a site's host is confirmed working — happy to adjust further if you spot an issue with the probe query shape itself.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

akshaymagapu (changes requested):
Hey @claudiaboldis,

Thanks for the deep investigation on this one - the root-cause writeup (firefly's table being 100% community.adobe.com) and the inline rationale comments are genuinely great. The approach is sound; a few things to address before merge.

### Strengths
- Reusing `buildSiteFilters` so aggregation honors the same allowlisted, VULN-37491-escaped host logic the report layer uses is the right call - one predicate, no divergence.
- `insert-aggregated.sql` (non-referral) is correctly restructured in all 7 providers: the inner subquery projects the full normalized column set before `WHERE {{siteFilterClause}}`, so any filter key resolves. The probe subqueries are likewise complete.
- INSERT column order is preserved across providers, so the aggregated schema isn't corrupted.
- Both probe branches are covered and the suite passes at 100%.

### Critical

**C1. `insert-aggregated-referral.sql` throws `COLUMN_NOT_FOUND` for path-based base URLs and custom `cdnlogsFilter` keys - all 7 providers.**
The referral `hosts` CTE projects only `host` + `x_forwarded_host`, then applies `WHERE {{siteFilterClause}}`. But `buildSiteFilters` (`cdn-utils.js:552-587`) emits `REGEXP_LIKE(url, ...)` whenever a base URL has a path (`:568`), and `referer`/`user_agent`/`cdn_provider` for custom filters (`ALLOWED_FILTER_KEYS`, `:526`). None of those columns exist in the `hosts` CTE. The `base`/`referrals_raw` CTE is also short: akamai/cloudflare/cloudfront/imperva alias `referer_raw` (not `referer`) and omit `cdn_provider`; fastly/other/frontdoor filter via `SELECT *` over raw names (`request_user_agent`, `properties.referer`) that don't match the canonical keys.
The probe exposes all 6 columns, so it returns `filterConfirmed = true`, and the very next query - the referral insert - throws. The two inserts are awaited with no try/catch, so this fails the whole provider iteration on every run and every retry. It's a regression: before this PR the referral insert had no site filter and no column dependency.
Example: a site `https://example.com/blog` sharing a bucket -> clause is `(...host...) AND REGEXP_LIKE(url, '^/?blog(?:/|$)')` -> probe passes -> `hosts` CTE has no `url` -> Athena throws.
Fix: scope the `hosts` CTE filter to host-only columns (its only job is building the first-party host set, so it doesn't need the full clause), and expose the full canonical column set in `base`/`referrals_raw` before applying the clause. Add a static check that every column a generated `siteFilterClause` references is present in the CTE it lands in - that catches this class without a live Athena.

**C2. The probe has no error handling.**
It's a new Athena query ahead of the inserts with no try/catch, and the loop body isn't wrapped. A transient throttle/timeout now aborts the whole aggregation for that site/hour - including inserts that would have succeeded - a failure mode that didn't exist before. A persistent failure in any one `check-site-filter.sql` fails that provider fleet-wide. Wrap the probe and degrade to a safe default on failure.

### Important

**I1. No kill switch.**
Unlike the adjacent retry feature (`CDN_ANALYSIS_RETRY_ENABLED`), this has no env gate, so it goes live for every site on the next run - on a data path that's hard to rebuild past raw-log retention. Add a flag (e.g. `CDN_AGGREGATION_HOST_FILTER_ENABLED`, default off) and roll out dev -> stage -> a few sites -> fleet, so C1 or any surprise can be flipped off in seconds instead of an emergency redeploy.

**I2. The fallback is silent and mutes an existing alarm.**
The unfiltered path only logs `log.info`. Worse, the existing "raw logs present but aggregation empty" warning at `handler.js:544` can't fire when the fallback keeps rows. Given the original firefly failure was 11 weeks of silent zeros, add a `log.warn` + metric on the fallback and an alert like "site in fallback for N runs straight."

**I3. Tests are green but don't exercise the broken path.**
Every fixture uses a root-domain site with no custom filter - the one shape that happens to work in the buggy referral SQL, so C1 is invisible to CI. Add a path-based-URL fixture and a custom-filter fixture (`referer`/`user_agent`/`cdn_provider`) through the referral insert.

### Minor
- **Referer GROUP BY change** in `insert-aggregated.sql` (cloudfront/fastly/frontdoor/other) is verified safe - the agentic report does `SUM(count)` and doesn't group by referer, so totals are unchanged. Worth a one-line note in the PR description.
- **Imperva `x_forwarded_host`** changed from constant `''` to `COALESCE(s_computername, '')` - alters stored output; call it out or keep `''` outside the filter.
- **`SELECT *, ... AS x_forwarded_host`** (fastly/frontdoor) collides if the raw schema ever adds a `host`/`x_forwarded_host` column; prefer explicit column lists.
- **`filterConfirmed = probeRows.length > 0`** assumes the client returns data rows only (no header row) - pin that with a comment.
- **Probe test mock** returns `[{host:'example.com'}]` but the real `SELECT 1 ... LIMIT 1` yields `[{_col0:1}]`; harmless, but misleading.
- **No test asserts the probe runs once per provider** - add a call-count assertion to an existing multi-provider test.

### Assessment
Not ready to merge. Two blockers: C1 (a provider-wide runtime break the trivial fixtures hide) and C2 (an unguarded new failure surface). I1/I2/I3 are worth landing alongside. Nice work on the investigation - the mechanism and comments are solid; it's the edges that need tightening.

claudiaboldis:
Thanks for the thorough review — addressed all of it in 41a8ef4f.

**C1 (blocker) — fixed.** You were right: the referral `hosts` CTE only ever had `host`/`x_forwarded_host`, so a path-scoped base URL or a custom `cdnlogsFilter` on `user_agent`/`referer`/`cdn_provider` would have thrown `COLUMN_NOT_FOUND`. Restructured all 7 providers around one shared `raw_normalized` CTE that exposes the full canonical column set (matching `ALLOWED_FILTER_KEYS`) before either `hosts` or `base` applies `siteFilterClause` — so any shape the clause takes resolves. Also split `url` (path-only, matching what the path regex in `buildSiteFilters` actually expects) from `url_with_query` (kept for the existing utm-extraction logic) — a few providers' raw url field mixes path+query in one column, which would have broken path matching silently even after fixing the missing-column issue. Added two regression tests that exercise the referral insert end-to-end: a path-scoped base URL, and a custom filter on `user_agent`.

**C2 (blocker) — fixed.** Wrapped the probe in try/catch; a failure now logs `log.warn` and degrades to the unfiltered fallback for that run instead of aborting the provider's inserts. Added a test for this.

**I1 — added.** `CDN_AGGREGATION_HOST_FILTER_ENABLED` (default off). Off means `siteFilterClause` is the `'TRUE'` tautology unconditionally and the probe never runs — a real no-op, so nothing changes for any site until it's explicitly turned on. Added a test asserting zero probe calls and unfiltered SQL when it's unset.

**I2 — done.** The "matched no rows" fallback is now `log.warn` (was `info`), called out in a comment as the one signal ops has that a site is running unfiltered, since the fallback also mutes the existing "aggregation empty" alarm for that site.

**I3 — added** the two C1 regression tests above, plus a probe-failure test and a kill-switch-off test. Didn't build out full metrics/alerting for "N runs in fallback straight" — that's cross-run state tracking, felt like a bigger, separable follow-up rather than something to fold into this PR; happy to file it if you'd rather track it explicitly.

**Minor items:**
- fastly/frontdoor referral now use explicit column lists instead of `SELECT *`.
- Left the imperva `x_forwarded_host` GROUP BY behavior as-is per your "call it out" option — added a comment noting it's pre-existing, not introduced here (stored value is unchanged).
- Commented the probe's no-header-row assumption.
- Fixed the probe test mock to look like a real `SELECT 1 ... LIMIT 1` result (`{ _col0: 1 }`) instead of a named column.
- Added the call-count assertion (scoped to the probe's own log description, since `athenaClient.query` is also used by `shouldRecreateTable`'s schema checks).

Referer GROUP BY note — agreed, no action needed, thanks for verifying it against the report's `SUM(count)`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #2943, opened 2026-09-04
feat(brand-claims): delay the on-demand claims ready-signal by 850s (LLMO-7263)

akshaymagapu:
## What

On-demand Brand Claims runs (LLMO-7263) are fired by api-service alongside their prerequisite audits (`offsite-brand-presence`, `wikipedia-analysis` — see adobe/spacecat-api-service#3208). Delay **only** the on-demand ready-signal published to the BP-sheet-ready queue so those inputs land in mystique before it runs claims.

- `mysticat-bp-sheet-ready` is a standard (non-FIFO) queue, so per-message `DelaySeconds` applies; 850s stays under the SQS 900s hard cap.
- Weekly and enrich runs publish immediately (delay 0).

## Coordination

No deploy-order dependency with the api-service change (if this ships first, a claims run with no prerequisites is just delayed 850s for nothing; if api-service ships first, claims runs without the delay).

## Testing

- `test/audits/brand-claims.test.js` — asserts `DelaySeconds=850` on the on-demand ready-signal and `0` on a weekly run. **33 passing**, handler eslint clean.

### DIFF
--- a/src/brand-claims/handler.js
+++ b/src/brand-claims/handler.js
@@ -19,6 +19,12 @@ const BP_PLATFORM = 'chatgpt_free';
 const SHEET_FILENAME_RE = /-w(\d{1,2})-(\d{4})(?:-(\d{6}))?\.xlsx$/i;
 const KEY_DATE_RE = /\/(\d{4})\/(\d{2})\/(\d{2})\//;
 const MAX_LISTING_PAGES = 10;
+// On-demand claims runs (LLMO-7263) are fired by api-service alongside their
+// prerequisite audits (offsite-brand-presence, wikipedia-analysis). Delay the claims
+// ready-signal so those inputs land in mystique before it runs claims. The
+// mysticat-bp-sheet-ready queue is a standard (non-FIFO) queue, so per-message
+// DelaySeconds applies; 850s stays under the SQS 900s hard cap.
+const ON_DEMAND_READY_DELAY_SECONDS = 850;
 
 // Must match DRS's sanitize_path_component byte-for-byte (BP consumer re-lists on this value).
 export function sanitizePathComponent(component) {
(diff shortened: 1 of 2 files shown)


src/brand-claims/handler.js
```diff
@@ -249,7 +255,11 @@ export default async function brandClaimsHandler(message, context) {
     batch_id: null,
   };
 
-  await sqs.sendMessage(queueUrl, event);
+  // Delay only the on-demand ready-signal so the prerequisite audits complete first;
+  // weekly/enrich runs publish immediately. (sendMessage args: queue, body, msgGroupId,
+  // delaySeconds.)
+  const readyDelaySeconds = onDemand === true ? ON_DEMAND_READY_DELAY_SECONDS : 0;
```
MysticatBot, line 261: issue (blocking): **`sqs.sendMessage` does not support `DelaySeconds` - the delay is silently discarded.**

The shared `SQS.sendMessage` in `@adobe/spacecat-shared-utils/src/sqs.js` has this signature:

```js
async sendMessage(
  queueUrl,
  message,
  messageGroupId = undefined,
  messageDeduplicationId = undefined,
)
```

The 4th parameter is `messageDeduplicationId`, not `delaySeconds`. The class never sets `DelaySeconds` on the `SendMessageCommand` params. Since `mysticat-bp-sheet-ready` is a standard (non-FIFO) queue, the FIFO block is skipped entirely, and the value `850` passed here is silently ignored. No delay is applied to the message.

The inline comment on lines 257-259 states `(sendMessage args: queue, body, msgGroupId, delaySeconds.)` - this is incorrect and masks the bug.

The tests pass because `sqs.sendMessage` is a Sinon stub that records arguments without exercising the real implementation.

**Fix options:**
1. Extend `SQS.sendMessage` in `spacecat-shared` to accept and forward `delaySeconds` to `SendMessageCommand`. Then update this call to use the correct parameter position.
2. If the shared library change would delay this, construct the `SendMessageCommand` directly in this handler with `{ DelaySeconds: 850, ... }`, bypassing the wrapper for this call.

MysticatBot (changes requested):
Hey @akshaymagapu,

**Verdict: Request changes** - the SQS delay is silently dropped because the shared sendMessage wrapper does not support DelaySeconds.
**Complexity: LOW** - small diff, single service, 2 files.
_Changes: adds an 850-second SQS message delay for on-demand brand-claims ready-signals so prerequisite audits complete first (2 files)._

## Must fix before merge

1. [Critical] `sqs.sendMessage` 4th argument is `messageDeduplicationId`, not `delaySeconds` - the 850s delay is silently discarded - `src/brand-claims/handler.js:261` (details inline)

<details>

- suggestion: add a static test assertion `expect(ON_DEMAND_READY_DELAY_SECONDS).to.be.at.most(900)` so a future bump past the SQS hard cap fails at test time rather than in production - `test/audits/brand-claims.test.js`
</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 10m 13s | <b>Cost:</b> $4.35 | <b>Commit:</b> <code>b8978e43eaaaddf6f71f10d177f3825835070a04</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2947, opened 2026-09-07
fix(prerender): outdate suggestions for URLs that start redirecting

anuj-adobe:
## Summary
- RCV/prerender audits couldn't outdate a stale suggestion once its URL starts redirecting: the generic outdating mechanism (`handleOutdatedSuggestions` in `src/utils/data-access.js`) requires a URL to be present in `scrapedUrlsSet` to be reconsidered, but redirected URLs never entered that set (they never enter `comparisonResults`, since `getScrapeResultPaths` only returns `COMPLETE`-status URLs).
- `getScrapeJobStats` now tags `ScrapeUrl` `REDIRECT`-status entries distinctly (`scrapingStatus: 'redirect'` instead of the generic `'failed'`) and returns their URLs (`redirectedUrls`), which get unioned into `scrapedUrlsSet` so the existing outdating guard picks them up — no changes needed to the shared `syncSuggestions`/`handleOutdatedSuggestions` helper itself.
- Adds `urlsRedirected` to the per-run metrics log line and to `status.json`, mirroring the existing `scrapeForbiddenCount` pattern — previously there was zero redirect-specific observability anywhere in this audit.
- Scope: only `REDIRECT`-status URLs are unioned in (not other `FAILED` reasons like 403/404/5xx) — 403s already have separate handling (`createScrapeForbiddenOpportunity`), and other failure reasons can be transient, so they're intentionally left out of auto-outdating for now.

## Context
Follow-up to the content-scraper fix for [LLMO-6423](https://jira.corp.adobe.com/browse/LLMO-6423) ([spacecat-content-scraper#937](https://github.com/adobe/spacecat-content-scraper/pull/937)), which broadened redirect detection so that any non-canonical redirect (path or host change — only bare-domain↔www and http→https are still allowed through) is treated as a scrape failure, not just cross-domain redirects. This surfaced a gap where previously-suggested pages that now redirect would keep a stale, no-longer-verifiable suggestion around indefinitely.

## Jira
https://jira.corp.adobe.com/browse/LLMO-6423

## Test plan
- [x] `npm run test:spec -- test/audits/prerender/**` — 574 passing
- [x] `npm test` (full suite) — 9945 passing; 1 unrelated pre-existing flaky timeout in `test/audits/llmo-referral-traffic-daily/handler.test.js` (no code overlap with this change, reproduces in isolation too)
- [x] `npm run lint` — clean
- [x] 100% line/branch/statement coverage on `src/prerender` maintained

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Nitin Gupta", "Dominique Jäggi"]
rationale: "Extends suggestion-outdating logic to redirected URLs; a stale or wrongly-outdated suggestion is recoverable on the next audit run."
```

### DIFF
--- a/src/prerender/handler.js
+++ b/src/prerender/handler.js
@@ -1244,6 +1244,9 @@ export async function uploadStatusSummaryToS3(auditUrl, auditData, context) {
     const scrapeForbiddenCount = mergedPages.filter(
       (p) => p.scrapeError?.statusCode === 403,
     ).length;
+    const urlsRedirected = mergedPages.filter(
+      (p) => p.scrapingStatus === 'redirect',
+    ).length;
 
     const statusSummary = {
       baseUrl: auditUrl,
@@ -1258,6 +1261,7 @@ export async function uploadStatusSummaryToS3(auditUrl, auditData, context) {
       scrapeForbidden: auditResult.scrapeForbidden ?? false,
       scrapeForbiddenCount,
       scrapeForbiddenSince: auditResult.scrapeForbiddenSince ?? existingStatus.scrapeForbiddenSince,
+      urlsRedirected,
       lastAuditSuccess: auditResult.lastAuditSuccess !== false,
       pages: mergedPages,
     };
@@ -1289,7 +1293,7 @@ export async function uploadStatusSummaryToS3(auditUrl, auditData, context) {
  * @param {number} urlsToCheckLength - Fallback count when ScrapeUrl is unavailable
  * @param {Object} context - Audit context with dataAccess, s3Client, env, log
  * @returns {Promise<{urlsSubmittedForScraping: number, scrapeForbiddenCount: number,
- *   scrapeForbidden: boolean, missingPages: Object[]}>}
+ *   scrapeForbidden: boolean, missingPages: Object[], redirectedUrls: string[]}>}
  */
 export async function getScrapeJobStats(
   scrapeJobId,
@@ -1308,6 +1312,7 @@ export async function getScrapeJobStats(
       scrapeForbiddenCount: 0,
       missingPages: [],
       submittedUrlSet: null,
+      redirectedUrls: [],
     };
   }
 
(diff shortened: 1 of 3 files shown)


MysticatBot (approved):
Hey @anuj-adobe,

**Verdict: Approve** - clean bug fix with appropriate test coverage and no blocking issues.
**Complexity: LOW** - small diff, single service.
_Changes: Fixes redirect-aware outdating in prerender audits so URLs that start redirecting have their stale suggestions marked OUTDATED, and adds redirect count metrics to the scraping log line and status.json (3 files)._

<details>

- nit: Optional chaining `su.getStatus?.()` silently produces `undefined` when the method is absent, causing the feature to do nothing for that URL with no log trace. Consider dropping the `?.` to fail loud, or adding a debug log when status is undefined on a non-complete ScrapeUrl. - `src/prerender/handler.js:1350`
- nit: Test 1 ("pre-fix gap") assertion is conditionally weak - if `bulkUpdateStatus` is never called at all, the test passes silently. Asserting explicitly on `bulkUpdateStatus.notCalled` or restructuring to verify the outdating path ran but excluded the redirect URL would be more precise. - `test/audits/prerender/behaviour/redirect-outdating.test.js:269`
- nit: `redirectedUrls` are unioned into `scrapedUrlsSet` unconditionally, bypassing the `isDeployedAtEdge` filter applied to `successfulComparisons`. Safe today because `handleOutdatedSuggestions` independently guards edge-deployed suggestions (validated by test 3), but the asymmetry could surprise a future reader. - `src/prerender/handler.js:1508`
- suggestion: Consider a focused unit test for `getScrapeJobStats` with a mix of REDIRECT and other-status ScrapeUrls, asserting the returned `redirectedUrls` array directly. The integration test covers this end-to-end, but a targeted assertion would lock down the extraction logic. - `src/prerender/handler.js:1362`

</details>

---

<sub><b>Skill:</b> pr-review | <b>Model:</b> us.anthropic.claude-opus-4-6-v1[1m] | <b>Duration:</b> 3m 44s | <b>Cost:</b> $6.83 | <b>Commit:</b> <code>3909ea6a0dc949193fbb68ee6c7e1c8be6d984e8</code><br>If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

nit23uec (approved):

PR #2951, opened 2026-09-07
feat: migrate Content AI to content sources

duynguyen:
## Summary

- replace legacy Content AI configuration/index APIs with acquisition content-source discovery and search
- persist the resolved source `name` on the site so discovery runs only when needed
- update FAQ readiness checks, tests, and the migration spec for optional `name`/`index` coexistence

## Dependency

Pins `@adobe/spacecat-shared-data-access` 4.30.0, which includes the content AI config `name` support and merge semantics from adobe/spacecat-shared#1917. A contract test exercises the persistence path with the real installed `Config` implementation.

## Validation

- `npx mocha --require test/setup-env.js --reporter dot test/utils/content-ai.test.js test/audits/faqs/utils.test.js test/audits/faqs/handler.test.js test/audits/llmo-customer-analysis.test.js` (137 passing)
- `npx eslint src/utils/content-ai.js test/utils/content-ai.test.js`
- real Content AI source discovery and one-result FAST search succeeded for Lovesac (`lovesac-com-generative`)

Full local FAQ completion was not run because the SharePoint app secret available during validation was expired.

### DIFF
--- a/src/utils/content-ai.js
+++ b/src/utils/content-ai.js
@@ -11,6 +11,53 @@
  */
 
 import { ImsClient } from '@adobe/spacecat-shared-ims-client';
+import { Config } from '@adobe/spacecat-shared-data-access/src/models/site/config.js';
+
+function normalizeSiteUrl(value) {
+  if (!value) {
+    return null;
+  }
+
+  try {
+    const url = new URL(value);
+    const hostname = url.hostname.toLowerCase().replace(/^www\./, '');
+    const pathname = url.pathname.replace(/\/+$/, '');
+    return `${hostname}${url.port ? `:${url.port}` : ''}${pathname}${url.search}`;
+  } catch {
+    return value;
+  }
+}
+
+function findContentSourceForSite(contentSources, site) {
+  const overrideBaseURL = site.getConfig()?.getFetchConfig()?.overrideBaseURL;
+  const siteUrls = new Set(
+    [overrideBaseURL, site.getBaseURL()].filter(Boolean).map(normalizeSiteUrl),
+  );
+
+  return contentSources.find((source) => (
+    siteUrls.has(normalizeSiteUrl(source.acquisitionConfig?.baseUrl))
+  ));
+}
+
+async function parseResponse(response, action) {
+  if (!response.ok) {
+    let detail;
+    try {
+      const problem = await response.json();
+      detail = problem.detail;
+    } catch {
+      // Use the HTTP status text when the response is not problem JSON.
+    }
+
+    const error = new Error(
+      `${action}: ${response.status} ${detail || response.statusText}`.trim(),
+    );
+    error.status = response.status;
+    throw error;
+  }
+
+  return response.status === 204 ? null : response.json();
+}
 
 /**
  * Calculates a weekly cron schedule set to run one hour from now.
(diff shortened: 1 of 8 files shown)


src/utils/content-ai.js
```diff
@@ -76,181 +123,148 @@ export class ContentAIClient {
...
-      cursor = json.cursor;
+      const response = await fetch(url, { headers: this.getHeaders() });
+      // eslint-disable-next-line no-await-in-loop
+      const page = await parseResponse(
+        response,
+        'Failed to list Content AI acquisition sources',
+      );
+      contentSources.push(...(page?.items || []));
+      cursor = page?.cursor;
     } while (cursor);
```
cwjwisse, line 154: **[MEDIUM] Bound externally driven pagination**

If Content AI repeats a non-empty cursor, this loop continues until the Lambda timeout. A stubbed reproduction confirmed repeated requests with no internal termination guard. Add seen-cursor detection plus a page/deadline limit; ideally also bound each fetch with an abort timeout. Please include repeated-cursor and excessive-page regression tests.

---
*Reviewed with ❤️ by Pi 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

src/utils/content-ai.js
```diff
@@ -76,181 +123,148 @@ export class ContentAIClient {
...
-        'Content-Type': 'application/json',
-        Authorization: this.getAuthHeader(),
-      },
-      body: JSON.stringify(requestBody),
-    });
-
-    return searchResponse;
+  static async persistContentSourceName(site, name) {
+    const siteConfig = site.getConfig();
+    siteConfig.updateContentAiConfig({ name });
```
cwjwisse, line 161: **[HIGH] The installed data-access version does not provide this mutator**

This branch still pins `@adobe/spacecat-shared-data-access` 4.28.0 (`package.json:94` and the lockfile), where `updateContentAiConfig` is absent. A direct run against the installed dependency produced `TypeError: siteConfig.updateContentAiConfig is not a function`. Therefore first-time source discovery fails at persistence and FAQ readiness reports the source unavailable.

Please merge/publish `spacecat-shared#1917`, bump both dependency files here, and add a contract test using the real installed `Config` rather than mocking this method.

---
*Reviewed with ❤️ by Pi 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

cwjwisse (changes requested):
## Scout PR Review: feat: migrate Content AI to content sources

### Summary
This cleanly moves FAQ readiness from retired configuration/index APIs to acquisition-source discovery, persisted source names, and named-source vector search. The implementation and tests are strong, but the exact branch cannot persist a discovered source because it still installs data-access 4.28.0, before the required mutator exists.

### Validation
- Full suite: 9,956 passed, 4 pending, 0 failed
- Lint and `git diff --check`: passed
- Direct installed-package probe reproduced the missing-method failure
- Direct pagination probe reproduced the repeated-cursor loop

### Verdict
**Request changes**

Merge and publish `spacecat-shared#1917`, pin the resulting data-access release here, add a real installed-package contract test, and bound source pagination.

---
*Reviewed with ❤️ by Pi 🤖 and [Scout](https://github.com/Adobe-AIFoundations/scout)*

baranovskyalexandr:
Clean rewrite onto the content-source APIs — approve with one caveat.

**Merge-ordering dependency (please enforce):** `persistContentSourceName` calls `siteConfig.updateContentAiConfig({ name })`, which only exists after adobe/spacecat-shared#1917 ships. Since the shared-data-access version is intentionally not bumped here, this must not deploy before #1917 is released and a follow-up bumps the dependency — otherwise `updateContentAiConfig` is undefined at runtime and every FAQ resolution throws.

Minor, non-blocking:
- `validateContentAI` now relies on `searchContentSource` throwing on non-OK (via `parseResponse`) rather than inspecting `.ok`. The catch returns the resolved name with `isSearchWorking: false` → handler maps to "search endpoint validation failed", which is correct. Just confirm `/content-sources/search` returns non-2xx (not a 200 with empty results) when a source isn't searchable, else a broken source reads as healthy.
- `normalizeSiteUrl` includes `url.search` in the identity key. Fine for real source baseUrls, but a stray query string on either side would break matching.
- Verified the removed client methods (`getConfigurations`, `runSemanticSearch`, `getConfigurationForSite`, `createConfiguration`) have no other consumers on main; the dropped content-ai mock in `llmo-customer-analysis.test.js` is dead-mock cleanup. 👍

baranovskyalexandr (approved):

src/utils/content-ai.js
duynguyen: Thanks for flagging this. The Content AI pagination contract defines `cursor` as non-null while another page exists and `null` on the final page, so this client intentionally follows that contract. A repeated non-empty cursor would be an API contract violation; I would prefer to surface and fix that in Content AI rather than add defensive cursor/page limits here that could mask the upstream defect. A per-request timeout is a separate resilience concern, but I do not think repeated-cursor handling belongs in this migration.

src/utils/content-ai.js
cwjwisse, line 161: issue (blocking): This call is not available in the dependency installed by this branch. The PR still pins `@adobe/spacecat-shared-data-access` 4.28.0, while the required `updateContentAiConfig` mutator is introduced by the still-open companion PR adobe/spacecat-shared#1917. On a first-time source discovery, this throws before `site.save()`, so FAQ validation reports the source as unavailable instead of persisting it. Merge and publish the companion, bump this repo's package and lockfile to that release, and add a contract test against the installed Config implementation.

cwjwisse (dismissed):
Hey @duynguyen,

**Verdict: Request changes** - the migration still cannot persist a discovered source with the dependency installed by this branch.
**Complexity: HIGH** - large cross-cutting migration with source discovery, persistence, search, and FAQ behavior changes.
_Changes: Replaces legacy Content AI configuration and index usage with acquisition-source discovery, persistence, and named vector search, including focused tests and a migration spec (8 files)._
Note: Recommend a human read before merge - this change introduces a new Content AI integration contract and includes the migration specification. The bot review is a complement to, not a replacement for, a human read here.

## Must fix before merge
1. [Critical] Ship and pin the data-access mutator required for first-time source persistence - `src/utils/content-ai.js:161` (details inline)

<details>

- suggestion: Add a real-installed-Config contract test for the persistence path so dependency drift cannot make first-time discovery fail silently - `src/utils/content-ai.js:161`
- suggestion: Even though the documented API contract ends pagination with a null cursor, detect a repeated non-empty cursor and fail fast instead of allowing a malformed upstream response to run until the worker timeout - `src/utils/content-ai.js:154`
</details>

<details>
<summary>Test quality snapshot</summary>

- New code: 182 LOC across 3 files; new tests: 317 LOC across 4 files (ratio 1.74)
- Test cases: 33 new it() blocks; max 28 in `test/utils/content-ai.test.js`
- Assertion density: 1.42 per test; mock density: 0.00 per test
- Fired signals: assertion-thin: 1.42 assertions per test is below the 1.5 preflight floor
</details>

Note: CI checks are currently passing for build, deploy, Codecov patch, CLA, Kodiak, and configuration protection; several duplicate workflow jobs are skipped.

src/utils/content-ai.js
duynguyen: Fixed in 211c6842. The PR now pins `@adobe/spacecat-shared-data-access` 4.30.0, and the focused suite includes a contract test that calls `persistContentSourceName` with the real installed `Config` implementation and verifies that adding `name` preserves the existing `index`. The four affected suites pass with 137 tests.

duynguyen: Fixed in 211c6842. The PR now pins `@adobe/spacecat-shared-data-access` 4.30.0, and the focused suite includes a contract test that calls `persistContentSourceName` with the real installed `Config` implementation and verifies that adding `name` preserves the existing `index`. The four affected suites pass with 137 tests.

PR #2953, opened 2026-09-08
fix: Sitemap audit uses overrideBaseURL when present

mikeweis:
* If the site config specifies an override URL for the base URL, use the override base URL to find the initial set of critical sitemap files

## Related Issues
- [SITES-51218](https://jira.corp.adobe.com/browse/SITES-51218)

Thanks for contributing!

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Jesus Sanchez", "oleksii iatsiuk", "Dominique Jäggi"]
rationale: "Small, scoped fix to sitemap base-URL resolution; recoverable via redeploy and re-run of the audit."
```

### DIFF
--- a/src/sitemap/handler.js
+++ b/src/sitemap/handler.js
@@ -55,7 +55,7 @@ import { convertToOpportunity } from '../common/opportunity.js';
 import { createOpportunityData } from './opportunity-data-mapper.js';
 
 const auditType = Audit.AUDIT_TYPES.SITEMAP;
-const HANDLER_VERSION = 2; // manually update as needed
+const HANDLER_VERSION = 3; // manually update as needed
 
 // HTTP status codes explicitly handled by this audit for suggestion generation
 const TRACKED_STATUS_CODES = Object.freeze([...REDIRECT_STATUSES, 404]);
--- a/src/sitemap/handler.js
+++ b/src/sitemap/handler.js
@@ -29,6 +29,7 @@
 
 import {
   isArray,
+  isValidUrl,
 } from '@adobe/spacecat-shared-utils';
 import { Audit } from '@adobe/spacecat-shared-data-access';
 import {
@@ -356,13 +357,25 @@ export async function findSitemap(inputUrl, log) {
 /**
  * Main audit runner function
  */
-export async function sitemapAuditRunner(baseURL, context) {
+export async function sitemapAuditRunner(baseURL, context, site) {
   const { log } = context;
   const startTime = process.hrtime();
 
   log.info(`Starting sitemap audit v${HANDLER_VERSION}-${COMMON_VERSION} for ${baseURL}`);
 
-  const auditResult = await findSitemap(baseURL, log);
+  let urlToProbe = baseURL;
+  const overrideBaseURL = site?.getConfig?.()?.getFetchConfig?.()?.overrideBaseURL;
+  if (overrideBaseURL) {
+    if (isValidUrl(overrideBaseURL)) {
+      /* c8 ignore next */
+      log.info(`Sitemap: using overrideBaseURL ${overrideBaseURL} instead of baseURL ${baseURL} for site ${site.getId()}`);
+      urlToProbe = overrideBaseURL;
+    } else {
+      log.warn(`Sitemap: ignoring overrideBaseURL "${overrideBaseURL}" for site ${site.getId()} — missing http/https protocol; using baseURL ${baseURL} instead`);
+    }
+  }
+
+  const auditResult = await findSitemap(urlToProbe, log);
 
   const endTime = process.hrtime(startTime);
   const elapsedSeconds = endTime[0] + endTime[1] / 1e9;
--- a/test/audits/sitemap.test.js
+++ b/test/audits/sitemap.test.js
@@ -279,6 +279,88 @@ describe('Sitemap Audit', () => {
         url,
       });
     });
+
+    it('probes overrideBaseURL instead of baseURL when it is a valid http(s) URL', async () => {
+      const overrideBaseURL = 'https://override-domain.adobe';
+      const site = {
+        getId: sandbox.stub().returns('site-123'),
+        getConfig: sandbox.stub().returns({
+          getFetchConfig: sandbox.stub().returns({ overrideBaseURL }),
+        }),
+      };
+
+      nock(overrideBaseURL).get('/robots.txt').reply(404);
+
+      const result = await sitemapAuditRunner(url, context, site);
+      expect(result).to.eql({
+        auditResult: {
+          reasons: [
+            {
+              error: ERROR_CODES.CANNOT_READ_ROBOTS,
+              value: `Fetch error for ${overrideBaseURL}/robots.txt Status: 404`,
+            },
+          ],
+          url: overrideBaseURL,
+        },
+        fullAuditRef: url,
+        url,
+      });
+    });
+
+    it('falls back to baseURL when overrideBaseURL is missing a http/https protocol', async () => {
+      const site = {
+        getId: sandbox.stub().returns('site-123'),
+        getConfig: sandbox.stub().returns({
+          getFetchConfig: sandbox.stub().returns({ overrideBaseURL: 'override-domain.adobe' }),
+        }),
+      };
+
+      nock(url).get('/robots.txt').reply(404);
+
+      const result = await sitemapAuditRunner(url, context, site);
+      expect(result).to.eql({
+        auditResult: {
+          reasons: [
+            {
+              error: ERROR_CODES.CANNOT_READ_ROBOTS,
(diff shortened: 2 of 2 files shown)


oiatsiuk (commented):
A few things worth a look before merging, focused mostly on where `overrideBaseURL` ends up being visible downstream:

- The override-URL resolution (`overrideBaseURL` → validate → fallback) is added inline here, but the actual root cause is that `noopUrlResolver` in `src/common/base-audit.js` never consults `overrideBaseURL` at all. Every `RunnerAudit`-based handler goes through that resolver, so a fix there would apply to all of them instead of needing to be re-copied into each audit that wants override support (as has already happened ad hoc in health-check, broken-links-guidance, page-citability, page-intent, backlinks, and agentic-urls, with inconsistent validation between them).

Inline notes below on a couple of correctness issues and a reuse opportunity in this file specifically.

src/sitemap/handler.js
```diff
@@ -356,13 +357,25 @@ export async function findSitemap(inputUrl, log) {
...
+  let urlToProbe = baseURL;
+  const overrideBaseURL = site?.getConfig?.()?.getFetchConfig?.()?.overrideBaseURL;
+  if (overrideBaseURL) {
+    if (isValidUrl(overrideBaseURL)) {
+      /* c8 ignore next */
+      log.info(`Sitemap: using overrideBaseURL ${overrideBaseURL} instead of baseURL ${baseURL} for site ${site.getId()}`);
+      urlToProbe = overrideBaseURL;
+    } else {
+      log.warn(`Sitemap: ignoring overrideBaseURL "${overrideBaseURL}" for site ${site.getId()} — missing http/https protocol; using baseURL ${baseURL} instead`);
+    }
```
oiatsiuk, line 375: Two related issues stem from probing `urlToProbe` here while the rest of the function still uses `baseURL`:

1. **Inconsistent audit record**: the returned `auditResult` (built by `findSitemap(urlToProbe, ...)` → `toPersistedAuditResult`) sets `auditResult.url = urlToProbe` (the override host when active), but the outer return a few lines down still sets `fullAuditRef: baseURL` and `url: baseURL`. The same persisted Audit record ends up with two different "site URL" values depending on which field you read — your own new test (`probes overrideBaseURL instead of baseURL...`) asserts exactly this split as expected.
2. **Leak into customer-facing suggestions**: sitemap/page URLs discovered here (via `getPagesWithIssues`, `buildErrorSuggestionFromReason`) are sourced from whatever host was actually probed. When `overrideBaseURL` is an internal/staging alias, that alias can end up verbatim in the `sitemapUrl`/`recommendedAction`/`pageUrl` fields shown to the customer in the resulting opportunity, instead of their real domain. Worth double-checking this is the intended behavior for the override use case.

src/sitemap/handler.js
```diff
@@ -356,13 +357,25 @@ export async function findSitemap(inputUrl, log) {
...
+export async function sitemapAuditRunner(baseURL, context, site) {
   const { log } = context;
   const startTime = process.hrtime();
 
   log.info(`Starting sitemap audit v${HANDLER_VERSION}-${COMMON_VERSION} for ${baseURL}`);
 
-  const auditResult = await findSitemap(baseURL, log);
+  let urlToProbe = baseURL;
+  const overrideBaseURL = site?.getConfig?.()?.getFetchConfig?.()?.overrideBaseURL;
+  if (overrideBaseURL) {
```
oiatsiuk, line 368: `src/health-check/handler.js` already exports `getEffectiveBaseURL(site)` doing this exact same `getConfig().getFetchConfig().overrideBaseURL` + `isValidUrl` + fallback logic. Could this reuse/import that helper (or a shared version moved to `src/common/`) instead of reimplementing it here? Otherwise a future fix to the validation rule (e.g. rejecting private IPs) has to be applied in both places to stay in sync.

src/sitemap/handler.js
```diff
@@ -356,13 +357,25 @@ export async function findSitemap(inputUrl, log) {
...
-export async function sitemapAuditRunner(baseURL, context) {
+export async function sitemapAuditRunner(baseURL, context, site) {
   const { log } = context;
   const startTime = process.hrtime();
 
   log.info(`Starting sitemap audit v${HANDLER_VERSION}-${COMMON_VERSION} for ${baseURL}`);
 
-  const auditResult = await findSitemap(baseURL, log);
+  let urlToProbe = baseURL;
+  const overrideBaseURL = site?.getConfig?.()?.getFetchConfig?.()?.overrideBaseURL;
```
oiatsiuk, line 367: This `/* c8 ignore next */` excludes the `log.info` success line from coverage, but the new test `probes overrideBaseURL instead of baseURL when it is a valid http(s) URL` exercises exactly this branch, so the line is actually covered. The ignore comment looks unnecessary here and just masks true coverage reporting for this line.

src/sitemap/handler.js
```diff
@@ -356,13 +357,25 @@ export async function findSitemap(inputUrl, log) {
 /**
  * Main audit runner function
  */
-export async function sitemapAuditRunner(baseURL, context) {
+export async function sitemapAuditRunner(baseURL, context, site) {
   const { log } = context;
   const startTime = process.hrtime();
```
oiatsiuk, line 363: Minor: this could be a single derived expression instead of a `let` reassigned across the nested if/else, e.g. `const urlToProbe = isValidUrl(overrideBaseURL) ? overrideBaseURL : baseURL;` with the logging as a separate step — matching the simpler pattern already used for the same concept in `page-citability/handler.js` / `page-intent/handler.js`.

src/sitemap/handler.js
```diff
@@ -356,13 +357,25 @@ export async function findSitemap(inputUrl, log) {
...
+      urlToProbe = overrideBaseURL;
+    } else {
+      log.warn(`Sitemap: ignoring overrideBaseURL "${overrideBaseURL}" for site ${site.getId()} — missing http/https protocol; using baseURL ${baseURL} instead`);
+    }
+  }
+
+  const auditResult = await findSitemap(urlToProbe, log);
 
   const endTime = process.hrtime(startTime);
   const elapsedSeconds = endTime[0] + endTime[1] / 1e9;
```
oiatsiuk, line 381: This completion log (and the "Starting sitemap audit..." log above) still reference `baseURL` even when `urlToProbe` (the override) is what was actually probed. Only the one-time override-info log in between reveals the real target, which makes it harder to correlate a slow/failed run with the right host from the completion line alone.

jsanchez07 (approved):

src/sitemap/handler.js
mikeweis: The Sitemap audit -- very generally speaking -- ignores logging lines from its unit testing so that they can be enhanced / tuned in the future without changing any unit tests (that are really designed to validate logic, without trying to rely on logging to do that logic validation).

src/sitemap/handler.js
mikeweis: * The "Starting" and "completed" log lines deliberately reference the same URL -- in this case `baseURL` -- for consistency and to easily correlate this audit run versus some other site's audit run.
* In real life, the `overrideBaseURL` is the `www.` subdomain added to the base URL.  Thus, there should be no real confusion when correlating back to the right host.

src/sitemap/handler.js
mikeweis: Good point.  I thought the `let` was more obvious to the human reader of what changed.  Will consider collapsing these statements in a future enhancement to this code.  Thanks! 👍

src/sitemap/handler.js
mikeweis: These audits in `spacecat-audit-worker` repo are eventually getting migrated to the `mystique` (Mysticat) repo, and they are being re-written & re-architected for that environment.  I wanted to keep this particular change isolated.  But nonetheless an excellent point. 👍

src/sitemap/handler.js
mikeweis: 1. This is an excellent point, and I will capture this explicitly in the related ticket as a possible concern.
2. On the Production environment -- which is customer facing -- the `overrideBaseURL` is typically the `www.` subdomain added to the `baseURL`.  Example:
  * baseURL = `https://example.com`
  * overrideBaseURL = `https://www.example.com`

mikeweis:
> ... the actual root cause is that noopUrlResolver in src/common/base-audit.js never consults overrideBaseURL at all ...

True, but the purpose of the `noopUrlResolver` is to pass the `baseURL` exactly as is, without any modifications.  ... When each of these existing audits -- currently in `spacecat-audit-worker` repo -- get migrated into the `mystique` (Mysticat) repo, the use of the `overrideBaseURL` needs to be revisited.  The Mysticat function `eds_content_origin( ... )` seems to have the "use overrideBaseURL" logic.

PR #2958, opened 2026-09-10
feat(offsite-brand-presence): retrieve Semrush URLs via S2S session-token flow (LLMO-6709)

tb-adbe:
## What changed

Switches the offsite-brand-presence Semrush **URL retrieval** from a raw IMS-service-token call to the **S2S consumer session-token** flow now supported by api-service ([adobe/spacecat-api-service#3217](https://github.com/adobe/spacecat-api-service/pull/3217)).

All changes are isolated to [`src/utils/offsite-brand-presence-semrush.js`](https://github.com/adobe/spacecat-audit-worker/blob/claude/audit-worker-semrush-s2s-38be8e/src/utils/offsite-brand-presence-semrush.js) and its test. Bucketing, the entitlement gate, the `OFFSITE_BRAND_PRESENCE_SEMRUSH_ENABLED` gate, and the fallback contract are unchanged.

### Before → after (auth)

| | Before | After |
|---|---|---|
| Token | raw IMS **service** token (`getServiceAccessToken`, authorization_code, default client) | **S2S consumer** IMS token (`getServiceAccessTokenV3`, client_credentials, `SEMRUSH_S2S_*`) → exchanged for a customer-scoped **session token** |
| Exchange | none — sent straight to the data route | `POST /api/v1/auth/s2s/login` with `{ imsOrgId }` → 15-min session token whose `tenants` claim authorizes the read |
| Host | ASO (`spacecat.experiencecloud.live`) | **LLMO** (`llmo.experiencecloud.live`) — `x-product` is host-driven at the edge |
| `x-promise-token` | forwarded | dropped (session token is self-contained) |

New graceful-fallback diagnostics reasons: `no_ims_org_id`, `session_token_failed`. Reuses the existing `getImsOrgId` helper for the customer IMS org id.

## Why

The old path minted a raw IMS **service** token, which carries no `tenants`/org membership, so api-service's IMS path rejects it (401/403) — the source of the `LLMO-6709` "service token rejected" fallbacks. PR #3217 authorizes S2S consumers for the brand-presence routes via the existing `hasAccess(organization)` check; this change adopts that flow and should close LLMO-6709.

## Reviewer notes

- **Fail-safe:** a missing/misconfigured S2S setup (`SEMRUSH_S2S_*`, unresolved IMS org id, login failure) falls back to the legacy source rather than failing the audit.
- **Route already S2S-enabled:** `.../url-inspector/domain-urls` requires `brand:read` in api-service's `required-capabilities.js` — no producer change needed.
- **Scope:** URL retrieval only. Fetching prompts *from Semrush* (for the cited/youtube/reddit audits) is deliberately out of scope and tracked as follow-up work.
- **Coverage:** loader module remains at 100% lines/branches/statements; 58 loader tests + full handler suite pass; lint clean.

### Infra required before this is live (ops)

- Register audit-worker as an **S2S consumer** with `brand:read` (multi-org scope), producing client_credentials creds.
- Env vars: `SEMRUSH_S2S_IMS_HOST/CLIENT_ID/CLIENT_SECRET/CLIENT_SCOPE`, `LLMO_API_BASE_URL` (default `https://llmo.experiencecloud.live`), optional `LLMO_S2S_LOGIN_URL` for non-prod path prefixes.
- Producer-side Vault (from #3217): `SEMRUSH_ADMIN_ELEMENT_API_KEY` + `SEO_API_BASE_URL` per env.

## Related Issues

LLMO-6709

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```

### DIFF
--- a/docs/decisions/002-offsite-cited-urls-semrush-source.md
+++ b/docs/decisions/002-offsite-cited-urls-semrush-source.md
@@ -50,15 +50,19 @@ involves several non-obvious trade-offs, so it warrants an ADR alongside the spe
    host — the legacy top-cited gate, now applied per URL rather than per domain rollup.
    Citations are clamped (`Math.max(0, …)`) and zero-citation URLs are dropped.
 5. **Auth = IMS service bearer via v2 `getServiceAccessToken()` (`authorization_code`
-   grant), default IMS client.** This is the same S2S path `commerce-product-enrichments`,
-   `vulnerabilities` and `permissions` use. v3 `getServiceAccessTokenV3()`
-   (`client_credentials`) was tried first but returns IMS **`400 unauthorized_client`** —
-   the worker's default IMS client is only provisioned for `authorization_code`; only a
-   dedicated integration (content-ai's `CONTENTAI_*`) is registered for `client_credentials`.
-   The token is forwarded unchanged to Semrush; no `x-promise-token` for a service caller.
-   **Open risk (LLMO-6709):** the proxy is designed around a real *user* IMS token, so whether
-   Semrush accepts the worker's *service* token is unverified — the flag stays off until
-   confirmed end-to-end (use `enableSemrush:true` on one canary run, per Decision 7, to test).
+   grant), default IMS client.** ⚠️ **Superseded by Decision 9 (LLMO-6709 resolved).** This
+   originally minted a raw IMS *service* token and forwarded it unchanged to Semrush. That
+   token carries no `tenants`/org membership, so api-service's IMS path rejects it (401/403) —
+   the `domain-urls-auth-failed` fallbacks. The auth mechanism is now the S2S consumer
+   session-token flow; see Decision 9. Retained for history:
+   > This is the same S2S path `commerce-product-enrichments`, `vulnerabilities` and
+   > `permissions` use. v3 `getServiceAccessTokenV3()` (`client_credentials`) was tried first
+   > but returns IMS **`400 unauthorized_client`** — the worker's default IMS client is only
+   > provisioned for `authorization_code`; only a dedicated integration (content-ai's
+   > `CONTENTAI_*`) is registered for `client_credentials`. The token is forwarded unchanged
+   > to Semrush; no `x-promise-token` for a service caller.
+   > **Open risk (LLMO-6709):** the proxy is designed around a real *user* IMS token, so
+   > whether Semrush accepts the worker's *service* token is unverified.
 6. **`PAGE_SIZE` is a fixed constant (1000).** The response is sorted by
    citations globally across every host, so a low-citation bucket can be starved by too
    small a page — a generous page is cheap since it's one request either way. 1000 is the
@@ -236,6 +240,37 @@ involves several non-obvious trade-offs, so it warrants an ADR alongside the spe
    checked against, since the sibling repo can ship the exact mechanism being ruled out
    before the review round closes.
 
+9. **Auth = S2S consumer session-token flow (LLMO-6709, supersedes Decision 5).**
+   api-service ([adobe/spacecat-api-service#3217](https://github.com/adobe/spacecat-api-service/pull/3217))
+   exposes the Semrush-backed brand-presence routes (`domain-urls` included) to **S2S
+   consumers** — authorized by the existing `hasAccess(organization)` check against the
+   caller's session-token `tenants` claim — rather than to raw IMS service tokens. The loader
+   now does a 3-leg exchange:
+   1. resolve the **customer IMS org id** (threaded in from the handler, which already
+      resolves it for DRS scraping — falling back to `getImsOrgId(site, …)` only when a
+      caller omits it, so the session token is never scoped to a different org than the run);
+   2. mint the **consumer's own** IMS token via `getServiceAccessTokenV3()`
+      (`client_credentials`) using dedicated `SEMRUSH_S2S_*` credentials (the same shape
+      content-ai's `CONTENTAI_*` uses; `IMS_CLIENT_CODE` is set to a placeholder because
+      `createFrom` validates it as required but the `client_credentials` grant never sends it);
+   3. exchange it at `POST /api/v1/auth/s2s/login` (`{ imsOrgId }`) for a 15-min,
+      customer-scoped session token, then present that as `Bearer` on the `domain-urls` call.
+   Both the login and data calls must hit the **LLMO host** (`LLMO_API_BASE_URL`, default
+   `llmo.experiencecloud.live`; `LLMO_S2S_LOGIN_URL` overrides the login URL for non-prod
+   path prefixes) because the Fastly edge sets `x-product` from the host. `x-promise-token`
+   forwarding is dropped (the session token is self-contained). The session token is cached
+   module-level keyed by `imsOrgId` (TTL under 15 min), so a warm container reuses it across
+   invocations for the same customer org, collapsing the mint+login to zero network calls.
+   New fallback reasons: `no_ims_org_id`, `ims_token_failed`, `session_token_auth_failed`
+   (401/403 — a static config problem: missing `brand:read` or wrong tenant), and
+   `session_token_failed` (transient/other). The login endpoint's error body is captured in
+   the structured log (`responseBody`) but never forwarded to Slack. **Infra required:**
+   register the worker as an S2S consumer with `brand:read` (multi-org scope), set
+   `SEMRUSH_S2S_*` + `LLMO_API_BASE_URL`, and — producer-side — `SEMRUSH_ADMIN_ELEMENT_API_KEY`
+   / `SEO_API_BASE_URL` in Vault per PR #3217. Until an org has a resolvable `imsOrgId`,
+   Semrush is skipped (`no_ims_org_id`) and the run uses the legacy source — consistent with
+   DRS scraping, which already requires `imsOrgId`.
+
 ## Consequences
 
 - Enabling the flag can never silently zero out offsite (fallback), but the fallback
@@ -267,23 +302,28 @@ override (`enableSemrush`, Decision 7) is the intended tool for step 1: run a si
 request against the real Semrush proxy on one site and confirm the auth path works
 end-to-end before flipping the env var fleet-wide.
 
-1. **Auth/authz verified (LLMO-6709).** Confirm the worker's **service** IMS token is
-   accepted by the Semrush proxy end-to-end. Confirm `tracingFetch` does not emit the
-   `Authorization` header into traces/spans (service-bearer leak).
+1. **Auth/authz verified (LLMO-6709) — now the S2S consumer session-token flow (Decision 9).**
+   Confirm the worker is registered as an S2S consumer with `brand:read`, `SEMRUSH_S2S_*` +
+   `LLMO_API_BASE_URL` are set, and the producer-side Vault keys are provisioned; then confirm
+   the login exchange + `domain-urls` read succeed end-to-end on one canary org. Confirm
+   `tracingFetch` does not emit the `Authorization` header (IMS token or session token) into
+   traces/spans.
 2. **`dataSource` shipped** (this PR) — so parity can be measured.
 3. **Shadow-run parity on a canary site (LLMO-6711)** — top-70 overlap per bucket vs legacy.
 4. **Fleet enable** per environment — **US markets only** until region scoping (LLMO-6710) closes.
 
 ## Configuration / client-convention debt (to resolve before the 3rd caller)
 
-This loader reads `SPACECAT_API_URI` and mints an **IMS service bearer**; `brand-resolver.js`
-reads `SPACECAT_API_BASE_URL` and uses **`x-api-key`** — for the *same* spacecat-api-service.
-The IMS scheme is justified here (the Elements proxy must forward a bearer to Semrush), but
-**two base-URL env vars + two auth schemes** can drift (one caller repointed to stage/prod, the
-other not → IMS traffic to an untrusted issuer → 401 → silent legacy fallback). Documented as
-debt: pick one base-URL env var and one api-service client convention (a shared helper) **before
-the `cited-domains` follow-up adds a third caller** that copies whichever it finds first. A
-tracking ticket will be filed before this debt is resolved; not required to close this PR.
+This loader reads `LLMO_API_BASE_URL` and uses the **S2S consumer session-token** flow
+(Decision 9); `brand-resolver.js` reads `SPACECAT_API_BASE_URL` and uses **`x-api-key`** — for
+the *same* spacecat-api-service. The session-token scheme is required here (only S2S consumers
+may read the Semrush-backed routes), but **two base-URL env vars + two auth schemes** can drift
+(one caller repointed to stage/prod, the other not → traffic to the wrong host/issuer → 401 →
+silent legacy fallback). Documented as debt: pick one base-URL env var and one api-service
+client convention (a shared helper) **before the next S2S-backed source adds a third caller**
+that copies whichever it finds first — the org-id → IMS-token → session-token sequence in
+particular is a good extraction candidate the moment a second consumer needs it. A tracking
+ticket will be filed before this debt is resolved; not required to close this PR.
 
 ## Alternatives Considered
 
--- a/src/utils/offsite-brand-presence-semrush.js
+++ b/src/utils/offsite-brand-presence-semrush.js
@@ -75,6 +75,41 @@ const FETCH_TIMEOUT_MS = 10_000;
 const ERROR_BODY_SNIPPET_MAX = 500;
 
 /**
+ * S2S session-token cache TTL. The api-service session token lives 15 min; we cache for a
+ * shorter window so a cached token is never handed to a data call at (or near) expiry. The
+ * cache is module-level, so a warm Lambda container reuses it across audit invocations —
+ * collapsing the IMS mint + login exchange to zero network calls for a brand whose customer
+ * org was minted recently.
+ */
+const SESSION_TOKEN_TTL_MS = 10 * 60 * 1000;
+
+/**
+ * Module-level session-token cache, keyed by customer `imsOrgId` (the scope the token is
+ * minted for). Value: `{ token, expiresAt }`. Not keyed by brand/site — one customer-scoped
+ * session token authorizes every brand under that org.
+ */
+const sessionTokenCache = new Map();
+
+/**
+ * @param {string} imsOrgId
+ * @param {number} nowMs
+ * @returns {string|null} the cached, still-valid session token for this org, or null.
+ */
+function getCachedSessionToken(imsOrgId, nowMs) {
+  const entry = sessionTokenCache.get(imsOrgId);
+  return entry && entry.expiresAt > nowMs ? entry.token : null;
+}
+
+/**
+ * @param {string} imsOrgId
+ * @param {string} token
+ * @param {number} nowMs
+ */
+function cacheSessionToken(imsOrgId, token, nowMs) {
+  sessionTokenCache.set(imsOrgId, { token, expiresAt: nowMs + SESSION_TOKEN_TTL_MS });
+}
+
+/**
  * Reads a non-2xx response body as a short diagnostic snippet. Never throws.
  *
  * @param {Response} response
@@ -104,11 +139,13 @@ async function readErrorBodySnippet(response) {
  * The scheme is normalised to `Bearer` (IMS may return `token_type: "bearer"` lowercase,
  * which a strict `startsWith('Bearer ')` parser upstream would reject).
  *
+ * Returns the full `Bearer <token>` header string (not a bare token) — named accordingly.
+ *
  * @param {object} context - Lambda context (env + log).
  * @returns {Promise<string>} e.g. "Bearer eyJ...".
  * @throws {Error} when the token response has no access_token.
  */
-async function getImsAccessToken(context) {
+async function getImsAuthorizationHeader(context) {
   const { env } = context;
   const imsClient = ImsClient.createFrom({
     ...context,
@@ -118,6 +155,11 @@ async function getImsAccessToken(context) {
       IMS_CLIENT_ID: env?.SEMRUSH_S2S_CLIENT_ID,
       IMS_CLIENT_SECRET: env?.SEMRUSH_S2S_CLIENT_SECRET,
       IMS_SCOPE: env?.SEMRUSH_S2S_CLIENT_SCOPE,
+      // `ImsClient.createFrom` validates clientCode as required, but the client_credentials
+      // grant (`getServiceAccessTokenV3`) never sends it. Set it explicitly so a stripped
+      // env (or the default authorization_code client being retired) can't surface as a
+      // confusing `createFrom` throw masquerading as a token-mint failure.
+      IMS_CLIENT_CODE: env?.SEMRUSH_S2S_CLIENT_CODE || 'unused-for-client-credentials',
     },
   });
   const token = await imsClient.getServiceAccessTokenV3();
@@ -135,12 +177,20 @@ async function getImsAccessToken(context) {
  * consumer. Both this call and the subsequent data call must hit the LLMO host (see
  * `LLMO_API_DEFAULT_BASE_URL`).
  *
+ * On a non-2xx response the thrown error carries `.status` and `.responseBody` so the
+ * caller can (a) split an authz denial (401/403 — a static config problem retries won't
+ * fix) from a transient/other failure, and (b) log the full body while keeping it OUT of
+ * any user-facing (Slack) message — the login endpoint's error body is untrusted upstream
+ * content. The error `.message` deliberately holds only the status, never the body.
+ *
  * @param {object} params
  * @param {string} params.loginUrl - Fully-qualified S2S login URL.
- * @param {string} params.imsAuthorization - `Bearer <ims-token>` from `getImsAccessToken`.
+ * @param {string} params.imsAuthorization - `Bearer <ims-token>` from
+ *   `getImsAuthorizationHeader`.
  * @param {string} params.imsOrgId - Target customer IMS org id (e.g. `...@AdobeOrg`).
  * @returns {Promise<string>} the session token (raw JWT, no scheme prefix).
- * @throws {Error} on network error, non-2xx, unparseable body, or a missing sessionToken.
+ * @throws {Error} on network error, non-2xx (with `.status`/`.responseBody`), unparseable
+ *   body, or a missing sessionToken.
  */
 async function exchangeForSessionToken({ loginUrl, imsAuthorization, imsOrgId }) {
   const response = await fetch(loginUrl, {
@@ -154,8 +204,10 @@ async function exchangeForSessionToken({ loginUrl, imsAuthorization, imsOrgId })
     timeout: FETCH_TIMEOUT_MS,
   });
   if (!response.ok) {
-    const responseBody = await readErrorBodySnippet(response);
-    throw new Error(`s2s login returned ${response.status}${responseBody ? `: ${responseBody}` : ''}`);
+    const error = new Error(`s2s login returned ${response.status}`);
+    error.status = response.status;
+    error.responseBody = await readErrorBodySnippet(response);
+    throw error;
   }
   let body;
   try {
@@ -336,6 +388,11 @@ function classifyRow(row, siteHostname, brandTokens) {
  * @param {object} params.site - Site model (`getOrganizationId()`).
  * @param {Array<{week:number, year:number}>} params.previousWeeks
  * @param {object} params.context - Lambda context (env, log, dataAccess).
+ * @param {string} [params.imsOrgId] - Customer IMS org id (`...@AdobeOrg`) the session token
+ *   is scoped to. The handler already resolves this (for DRS scraping too), so it threads it
+ *   in to avoid a second, independently-fetched lookup that could scope the session token to
+ *   a different org than the rest of the run. Falls back to `getImsOrgId(site, ...)` only when
+ *   the caller omits it.
  * @param {string} [params.siteHostname] - www-stripped site hostname for owned-URL filtering.
  * @param {function(string): Promise<*>} [params.onProgress] - Optional best-effort progress
  *   callback (e.g. a Slack thread reply), invoked with a short human-readable status string at
@@ -345,8 +402,8 @@ function classifyRow(row, siteHostname, brandTokens) {
  * @param {object} [params.diagnostics] - Optional out-param, mutated in place. On a null
  *   return, set to `{ fallbackReason }` with a specific code (`no_organization_id`,
  *   `no_active_brand`, `brand_resolution_failed`, `not_entitled`, `entitlement_check_failed`,
- *   `no_date_window`, `no_ims_org_id`, `ims_token_failed`, `session_token_failed`,
- *   `domain_urls_auth_failed`, or `domain_urls_failed`).
+ *   `no_date_window`, `no_ims_org_id`, `ims_token_failed`, `session_token_auth_failed`,
+ *   `session_token_failed`, `domain_urls_auth_failed`, or `domain_urls_failed`).
  *   The two entitlement reasons additionally set `entitlementReason` to the granular cause
  *   from `resolveSemrushEntitlement` (`flag_disabled` | `no_workspace` | `no_client` |
  *   `check_failed`) — `fallbackReason` alone cannot distinguish a confirmed non-entitlement
@@ -356,7 +413,7 @@ function classifyRow(row, siteHostname, brandTokens) {
  * @returns {Promise<Map<string, {count:number, domain:string|null}> | null>}
  */
 export async function loadCitedUrlsFromSemrush({
-  site, previousWeeks, context, siteHostname, onProgress, diagnostics,
+  site, previousWeeks, context, imsOrgId: providedImsOrgId, siteHostname, onProgress, diagnostics,
 }) {
   const { log, env } = context;
   const startedAt = Date.now();
@@ -481,8 +538,10 @@ export async function loadCitedUrlsFromSemrush({
 
   // Leg 1: the customer IMS org id (…@AdobeOrg) — distinct from spaceCatId (the SpaceCat
   // org UUID used in the route path). This is what the session token's `tenants` claim,
-  // and thus api-service's hasAccess(organization), is keyed on.
-  const imsOrgId = await getImsOrgId(site, context.dataAccess || {}, log);
+  // and thus api-service's hasAccess(organization), is keyed on. The handler already
+  // resolves this (and reuses it for DRS scraping), so prefer the threaded value; the
+  // lookup is only a fallback for callers that don't provide it.
+  const imsOrgId = providedImsOrgId || await getImsOrgId(site, context.dataAccess || {}, log);
   if (!imsOrgId) {
     olog.warn('data_acquisition_bp_data_semrush_read', 'Could not resolve customer IMS org id; skipping Semrush source', {
       peer: PEER.SEMRUSH, direction: 'inbound', orgId: spaceCatId, brandId: brand.brandId, durationMs: elapsed(), reason: 'no_ims_org_id', outcome: OUTCOME.DEGRADED,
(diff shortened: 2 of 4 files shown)


src/utils/offsite-brand-presence-semrush.js
```diff
@@ -77,31 +90,85 @@ async function readErrorBodySnippet(response) {
...
-  const imsClient = ImsClient.createFrom(context);
-  const token = await imsClient.getServiceAccessToken();
+async function getImsAccessToken(context) {
+  const { env } = context;
+  const imsClient = ImsClient.createFrom({
+    ...context,
+    env: {
+      ...env,
+      IMS_HOST: env?.SEMRUSH_S2S_IMS_HOST,
+      IMS_CLIENT_ID: env?.SEMRUSH_S2S_CLIENT_ID,
```
amitruluimadalina, line 192: **Important:** This env override never sets `IMS_CLIENT_CODE` - it silently falls through from the spread `...env` (the default authorization_code client's value). The real `ImsClient.createFrom` throws if `clientCode` is falsy, even though `getServiceAccessTokenV3()` never reads it. The sinon stub in tests never runs this validation, so a missing `IMS_CLIENT_CODE` in some environment would surface as a confusing `ims_token_failed`, masking a config gap. Consider an explicit placeholder with a comment noting V3 ignores it, or a contract test against the real `createFrom`.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -77,31 +90,85 @@ async function readErrorBodySnippet(response) {
...
+ * which a strict `startsWith('Bearer ')` parser upstream would reject).
  *
  * @param {object} context - Lambda context (env + log).
  * @returns {Promise<string>} e.g. "Bearer eyJ...".
  * @throws {Error} when the token response has no access_token.
  */
-async function getAuthorizationHeader(context) {
-  const imsClient = ImsClient.createFrom(context);
-  const token = await imsClient.getServiceAccessToken();
+async function getImsAccessToken(context) {
```
amitruluimadalina: **Important:** No caching here (or in `exchangeForSessionToken` below) - a brand-new `ImsClient` is minted and a brand-new session token exchanged on every single call, with no reuse across sites/brands in the same invocation. This roughly triples auth-related round trips per audited brand versus the pre-PR flow. Recommend a session-token cache keyed by `imsOrgId` with a TTL under the 15-minute expiry.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -401,9 +471,31 @@ export async function loadCitedUrlsFromSemrush({
...
+  // are rejected by the IMS path). So we:
+  //   1. resolve the customer's IMS org id (the session token must be scoped to it),
+  //   2. mint the consumer's own IMS token (client_credentials),
+  //   3. exchange it for a 15-min session token whose `tenants` claim names that org.
+  // The domain-urls call then presents that session token.
+
+  // Leg 1: the customer IMS org id (…@AdobeOrg) — distinct from spaceCatId (the SpaceCat
+  // org UUID used in the route path). This is what the session token's `tenants` claim,
+  // and thus api-service's hasAccess(organization), is keyed on.
+  const imsOrgId = await getImsOrgId(site, context.dataAccess || {}, log);
```
amitruluimadalina: **Important:** This re-derives `imsOrgId` via a second DB lookup, but the caller (`src/offsite-brand-presence/handler.js:822-823`) already resolves the same value via `site.getOrganization()`/`getImsOrgId()` before calling this loader, and reuses it later for DRS scraping. Two independent resolution paths for the same value is wasted work and a latent divergence risk. Recommend threading `imsOrgId` through as a parameter from the handler instead.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -401,9 +471,31 @@ export async function loadCitedUrlsFromSemrush({
...
+  if (!imsOrgId) {
+    olog.warn('data_acquisition_bp_data_semrush_read', 'Could not resolve customer IMS org id; skipping Semrush source', {
+      peer: PEER.SEMRUSH, direction: 'inbound', orgId: spaceCatId, brandId: brand.brandId, durationMs: elapsed(), reason: 'no_ims_org_id', outcome: OUTCOME.DEGRADED,
+    });
+    await notify(':x: Could not resolve the customer IMS org id — falling back to the legacy source.');
+    setDiagnostics({ fallbackReason: 'no_ims_org_id' });
```
amitruluimadalina, line 587: **Important:** Before this PR, Semrush retrieval was never gated on IMS-org resolution. Now any site whose Organization lacks a resolvable `imsOrgId` permanently falls back to legacy for Semrush, even though nothing about the site/brand changed. Reasonable given the session token must be org-scoped, but worth a fleet-wide check of how many currently-entitled brands have no resolvable `imsOrgId` before this ships broadly - this fallback reason will look identical to 'everything is fine, just degraded'.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -77,31 +90,85 @@ async function readErrorBodySnippet(response) {
...
+      Authorization: imsAuthorization,
+      'Content-Type': 'application/json',
+      Accept: 'application/json',
+    },
+    body: JSON.stringify({ imsOrgId }),
+    timeout: FETCH_TIMEOUT_MS,
+  });
+  if (!response.ok) {
+    const responseBody = await readErrorBodySnippet(response);
+    throw new Error(`s2s login returned ${response.status}${responseBody ? `: ${responseBody}` : ''}`);
```
amitruluimadalina: **Important:** Every non-2xx status here (401, 403, 400, 5xx) is collapsed into one generic error and `session_token_failed` reason. This file's own `fetchDomainUrls` explicitly splits 401/403 into a distinct `authFailure` branch a few lines away, precisely so a rejected token is identifiable instead of looking like 'nothing came back'. This PR reintroduces that ambiguity for the new leg - a 403 (config problem, retries won't help) and a 5xx (transient) currently produce the same signal. Recommend mirroring the `authFailure` split here.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -413,16 +505,27 @@ export async function loadCitedUrlsFromSemrush({
...
+  // Leg 3: exchange it for a customer-scoped SpaceCat session token via the LLMO host.
+  const loginUrl = env?.LLMO_S2S_LOGIN_URL || `${baseUrl}${S2S_LOGIN_DEFAULT_PATH}`;
+  let sessionToken;
+  try {
+    sessionToken = await exchangeForSessionToken({ loginUrl, imsAuthorization, imsOrgId });
+  } catch (error) {
+    olog.warn('data_acquisition_bp_data_semrush_read', 'Failed to exchange for an S2S session token', {
+      peer: PEER.SEMRUSH, direction: 'inbound', orgId: spaceCatId, brandId: brand.brandId, durationMs: elapsed(), reason: 'session_token_failed', outcome: OUTCOME.DEGRADED, ...errorField(error),
+    }, error);
+    await notify(`:x: Failed to obtain an S2S session token (\`${error.message}\`) — falling back to the legacy source.`);
```
amitruluimadalina: **Important:** Up to 500 raw characters of the login endpoint's own response body get forwarded verbatim into this Slack-facing message, unbounded by content shape (only by length), bypassing the logging taxonomy's sanitization. The sibling `ims_token_failed` path's error text is bounded to a well-known OAuth error shape from the IMS client; this new LLMO login endpoint has no such guarantee. Recommend keeping the full body in the logged fields (as today) but forwarding only the status code to `notify()` for this leg.

amitruluimadalina (changes requested):
Hey @tb-adbe,

## Strengths

- The 3-leg auth flow (resolve IMS org id -> mint S2S consumer IMS token -> exchange for session token -> call domain-urls) is sequenced correctly behind the existing entitlement gate, so the new network calls only fire for brands already confirmed Semrush-entitled.
- It follows the repo's existing dedicated client_credentials pattern (the same shape `content-ai.js` uses with `CONTENTAI_*`) rather than inventing a new S2S wiring convention.
- `imsOrgId` (customer identifier) and `spaceCatId` (SpaceCat org UUID) are kept correctly distinct in telemetry - the customer IMS org id is never written to a log field, only passed into the login request body. No token value (IMS token or session token) is ever logged.
- Test coverage of the new failure surface is thorough: missing org id, IMS mint failure, missing access_token, non-2xx login with and without a body, network error during login, unparseable login JSON, and missing sessionToken are all covered, each asserted against a specific fallback reason.
- Dropping `x-promise-token` forwarding is a correct simplification now that the session token is self-contained.
- No new dependencies are introduced.

## Issues

### Important (Should Fix)

1. **src/utils/offsite-brand-presence-semrush.js:111-122** - `getImsAccessToken`'s env override never sets `IMS_CLIENT_CODE`; it silently falls through from the spread `...env`, i.e. the default authorization_code client's value. The real `@adobe/spacecat-shared-ims-client` `ImsClient.createFrom` throws if `clientCode` is falsy, even though `getServiceAccessTokenV3()` (the client_credentials grant used here) never reads it. The test suite can't catch this because `imsCreateFrom` is a bare sinon stub that never runs the real `createFrom` validation. If `IMS_CLIENT_CODE` is ever absent from the environment (a stripped-down env, or the legacy default client being retired once every consumer migrates to dedicated S2S clients), this throws a confusing, unrelated error that gets swallowed into `reason: 'ims_token_failed'`, masking a config gap as a token-mint failure. Fix: explicitly set a placeholder (with a comment noting V3 ignores it) or add a contract test against the real `ImsClient.createFrom` with only `SEMRUSH_S2S_*` set.

2. **src/utils/offsite-brand-presence-semrush.js:111 and :145** - No caching of either the IMS client_credentials token or the 15-minute session token. `getImsAccessToken` constructs a brand-new `ImsClient` on every call, so the shared client's own in-instance memoization never gets a chance to hit - every audited site/brand mints a fresh IMS token and a fresh session token. This adds a second network round trip (the session exchange) on top of the existing one (IMS mint), roughly tripling auth-related round trips per audited brand with no reuse across brands/sites processed in the same invocation or warm container. At fleet scale (this audit runs per site) this is a real, avoidable multiplier on both latency and load against IMS and the new login endpoint. Recommend a session-token cache keyed by `imsOrgId` with a TTL comfortably under the 15-minute expiry.

3. **src/utils/offsite-brand-presence-semrush.js:485** - This re-derives `imsOrgId` via `getImsOrgId(site, context.dataAccess || {}, log)`, but the caller (`src/offsite-brand-presence/handler.js:822-823`) already resolves the same value via `site.getOrganization()` / `organization.getImsOrgId()` before calling this loader, and reuses it later for DRS scraping. This is wasted work (a second DB lookup for a value already in scope) and a duplication risk: two independent code paths computing the 'same' value could diverge (stale cache, different fetch semantics), scoping the Semrush session token to a different org than the one used elsewhere in the same run without anyone noticing until an auth failure. Fix: thread `imsOrgId` through as a parameter from the handler, falling back to `getImsOrgId` only if the caller omits it.

4. **src/utils/offsite-brand-presence-semrush.js:156-158** - `exchangeForSessionToken` treats every non-2xx response (401 invalid/expired IMS token, 403 IMS token valid but consumer lacks `brand:read` for this org, 400 bad `imsOrgId`, 5xx login-service outage) identically - one generic error, one `fallbackReason: 'session_token_failed'`. This file's own `fetchDomainUrls` (a few lines away) explicitly splits 401/403 into a distinct `authFailure` branch with its own log message and diagnostic, precisely so a rejected token is identifiable instead of looking like 'Semrush returned nothing'. This PR reintroduces that same ambiguity one hop earlier, for the new leg. A 403 here is a static configuration problem that retries will never fix, while a 401 or 5xx might be transient - currently these are indistinguishable in the fallback reason. Fix: mirror the `authFailure` split from `fetchDomainUrls`, surfacing the status class in diagnostics or a distinct fallback reason per class.

5. **src/utils/offsite-brand-presence-semrush.js:517** - On a login failure, up to 500 raw characters of the login endpoint's own response body are interpolated verbatim into a Slack-facing `notify()` message, bypassing the logging taxonomy's sanitization entirely. This is a materially different disclosure surface than the sibling `ims_token_failed` path, whose error text comes from the IMS client and is bounded to a well-known OAuth error shape (e.g. `invalid_scope`). The new LLMO login endpoint (owned by the companion PR) has no such guarantee - if it is ever misconfigured to echo a stack trace, headers, or partial token in its error body, that content lands directly in Slack rather than only in structured logs. Fix: keep the full body in the logged fields as today, but truncate or generalize what is forwarded to `notify()` for this leg to just the status code, matching the discipline already applied to the IMS-mint leg.

6. **docs/decisions/002-offsite-cited-urls-semrush-source.md** - This ADR (Decision 5, and the section listing 'Open risk (LLMO-6709)') still documents the old `authorization_code` / `getServiceAccessToken()` flow and explicitly calls the auth question this PR resolves an unresolved open risk. It is not updated in this PR, so the design record will actively contradict the shipped code the moment this merges. Recommend a same-PR or same-day amendment (e.g. a new decision entry superseding Decision 5) so the ADR and the code stay in sync.

7. **src/utils/offsite-brand-presence-semrush.js:485-491** - Before this PR, Semrush retrieval was never gated on IMS-org resolution. Now, any site whose `Organization` row lacks a resolvable `imsOrgId` permanently falls back to the legacy source for Semrush, even though nothing about the site or brand itself changed. This is a reasonable trade-off given the session token must be org-scoped, but it is a new, silent, permanent-until-data-fixed degradation path that looks identical to 'everything is fine, just degraded' in diagnostics. Worth a fleet-wide sanity check (how many currently-entitled brands have no resolvable `imsOrgId`) before this ships broadly.

### Minor (Nice to Have)

- `getImsAccessToken` (renamed from `getAuthorizationHeader`) returns a full `Bearer <token>` header string, not a bare access token - the old name was more accurate; consider `getImsAuthorizationHeader` to avoid a future caller assuming they get a raw token.
- `getImsOrgId` already logs its own warning when it cannot resolve an org/IMS-org id, and the loader immediately logs a second, differently-worded warning for the same event - not incorrect, but noisy for anyone counting failures by log line.
- `LLMO_API_BASE_URL` and `LLMO_S2S_LOGIN_URL` are two independently-overridable env vars that both default toward production LLMO infrastructure. A dev/CI deployment that forgets to set both could silently attempt to exchange a dev-issued credential against production infrastructure rather than failing closed toward a non-prod default.
- The session token is used purely as an opaque string - there is no local check that it is even a well-formed JWT, or that its `tenants` claim matches the requested `imsOrgId`, before it is handed to the data call. Defensible since api-service is the real enforcement point, but a one-line decode-and-warn (not a hard block) would catch a login-endpoint bug locally instead of via another round of `domain_urls_auth_failed` triage.

## Recommendations

- Extract the org-id -> IMS-token -> session-token sequence into a small shared helper the moment a second S2S consumer needs it (the PR description suggests more S2S-backed sources are coming through this same api-service pattern).
- Add IMS/session token caching keyed by `imsOrgId`, with a TTL safety margin under the stated 15-minute expiry.
- Add a contract test against the real `ImsClient.createFrom` (not the sinon stub) with only `SEMRUSH_S2S_*` set, to close the `IMS_CLIENT_CODE` gap before it is discovered in production.
- Update `docs/decisions/002-offsite-cited-urls-semrush-source.md` (and its companion spec doc) in this PR or a same-day fast-follow.
- Before merge, explicitly confirm the `SEMRUSH_S2S_*` secrets, the S2S consumer registration, and the `LLMO_API_BASE_URL` / `LLMO_S2S_LOGIN_URL` env vars called out in this PR's own 'Infra required before this is live' section are actually provisioned for the target environment - otherwise this fix silently becomes a no-op with a relabeled failure reason.

## Assessment

Ready to merge? With fixes.
Reasoning: The move to an S2S-backed session-token flow is the right direction and is well-tested and consistent with the one precedent already in this repo (content-ai's client_credentials pattern) - nothing here argues against that goal. The issues above are all about hardening this specific implementation: an untested `IMS_CLIENT_CODE` dependency that could silently break token minting in some environments, a 2-3x amplification of auth-endpoint load with no caching, a redundant org-id lookup that duplicates work the caller already did, a diagnostic regression on the new auth hop relative to this file's own established pattern, and an unbounded error-body forward into Slack. None are architectural blockers, but items 1-5 should be resolved before this reaches fleet-wide cadence.

## Next Steps

1. Address the IMS_CLIENT_CODE gap, the missing token caching, and the redundant imsOrgId lookup first (items 1-3).
2. Restore the auth/authz failure distinction on the session-exchange leg and bound the Slack-forwarded error text (items 4-5).
3. Update the ADR and confirm infra provisioning (items 6 and the recommendations) before or alongside merge.
4. Minor items are optional.

tb-adbe:
Thanks for the thorough review @amitruluimadalina — great catches. Addressed in `37bbf6ab`.

### Important items

1. **`IMS_CLIENT_CODE` gap** — ✅ Fixed. `getImsAuthorizationHeader` now sets `IMS_CLIENT_CODE: env?.SEMRUSH_S2S_CLIENT_CODE || 'unused-for-client-credentials'` explicitly (with a comment: `createFrom` validates it as required, but the `client_credentials` grant never sends it), so a stripped env can't surface a confusing `createFrom` throw masked as `ims_token_failed`. Added two tests (default placeholder + `SEMRUSH_S2S_CLIENT_CODE` override).

2. **No token caching** — ✅ Added. Module-level session-token cache keyed by `imsOrgId`, TTL 10 min (under the 15-min server expiry). A warm container reuses it across invocations for the same customer org, collapsing the IMS mint + login exchange to zero network calls. Only successes are cached; expiry re-mints. Added cache-hit and TTL-expiry tests (fake timers).

3. **Redundant `imsOrgId` lookup** — ✅ Fixed. Added an `imsOrgId` param; the handler threads in the value it already resolves at `handler.js:822-823` (and reuses for DRS), so the loader no longer does an independent, differently-sourced (`Organization.findById` vs `site.getOrganization()`) re-fetch. Falls back to `getImsOrgId` only when a caller omits it — removes the redundant DB call and the org-divergence risk you flagged.

4. **Undifferentiated `session_token_failed`** — ✅ Fixed. `exchangeForSessionToken` now throws with `.status`/`.responseBody`, and the orchestrator splits 401/403 → `session_token_auth_failed` (static config problem) vs `session_token_failed` (transient/other), mirroring `fetchDomainUrls`' `authFailure` branch.

5. **Unbounded error body → Slack** — ✅ Fixed. The login endpoint's error body now stays in the structured `responseBody` log field only; the Slack `notify()` forwards just the HTTP status. Test asserts the body is logged but absent from the Slack message.

6. **Stale ADR 002** — ✅ Updated in this PR. Decision 5 is marked superseded; new **Decision 9** documents the S2S session-token flow; the LLMO-6709 runbook gate and the config-debt section are refreshed for the new mechanism.

7. **New `imsOrgId` gating = silent permanent fallback** — Acknowledged as a design trade-off (the session token must be org-scoped). Worth noting it's not a *new* class of degradation: DRS scraping already skips when `imsOrgId` is absent (`handler.js:521`), so such an org already produces no scraped content regardless of Semrush — and the fallback is surfaced distinctly as `no_ims_org_id` (`outcome=degraded`, not a silent skip). I'll run the fleet sanity check (currently-entitled brands with no resolvable `imsOrgId`) before fleet-wide enablement.

### Minor

- **Rename** `getImsAccessToken` → `getImsAuthorizationHeader` (it returns a full `Bearer <token>`) — ✅ done.
- **Double warning** — effectively resolved by item 3: the handler-threaded path no longer calls `getImsOrgId`, so the loader's structured warn is the only one on that path.
- **Two env vars defaulting to prod** — the code default *is* intentionally prod (`LLMO_API_DEFAULT_BASE_URL`); documented the non-prod override (`LLMO_S2S_LOGIN_URL`) in Decision 9. Leaving fail-closed-to-prod as-is unless you feel strongly.
- **Local JWT/tenants sanity check** — deferring; api-service is the enforcement point and a malformed token surfaces as `session_token_auth_failed`/`domain_urls_auth_failed`. Happy to add a decode-and-warn if you'd prefer.

### Recommendations

- **Shared helper extraction** — noted in ADR 002's config-debt section as the extraction candidate (org-id → IMS-token → session-token) for the moment a second S2S consumer lands.
- **Infra provisioning** — tracked in the PR's "Infra required" section; the S2S consumer registration + `SEMRUSH_S2S_*` / `LLMO_API_BASE_URL` + producer-side Vault keys will be confirmed before enablement (the fix fails safe to legacy until then).

Verification: lint clean, loader module at 100% coverage, 172 tests passing (loader + handler).

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -401,28 +528,95 @@ export async function loadCitedUrlsFromSemrush({
...
+        error,
+      );
+      // Forward only the status to Slack — the login endpoint's error body is untrusted
+      // upstream content and must never be echoed into a user-facing channel (it stays in
+      // the structured `responseBody` log field above).
+      await notify(`:x: Failed to obtain an S2S session token${error.status ? ` (HTTP ${error.status})` : ''} — falling back to the legacy source.`);
+      setDiagnostics({ fallbackReason: reason });
+      return null;
+    }
+    cacheSessionToken(imsOrgId, sessionToken, nowMs);
```
amitruluimadalina: **Important:** The cache is populated here on a successful exchange, but nothing on the `domain-urls` auth-failure path (the `result.authFailure` branch further down, which sets `domain_urls_auth_failed`) evicts this entry. Before caching was introduced, every call minted fresh so an auth rejection self-healed on the next invocation. Now, if the cached token was revoked or the org's grant changed after being cached, every subsequent call for this org within the remaining TTL keeps replaying the same invalid token and keeps failing, instead of re-minting. Recommend evicting `sessionTokenCache` for this `imsOrgId` on that auth-failure branch (or retrying once with a fresh token) before falling back, plus a regression test covering it.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -54,13 +68,47 @@ export const PAGE_SIZE = 1000;
...
+ * org was minted recently.
+ */
+const SESSION_TOKEN_TTL_MS = 10 * 60 * 1000;
+
+/**
+ * Module-level session-token cache, keyed by customer `imsOrgId` (the scope the token is
+ * minted for). Value: `{ token, expiresAt }`. Not keyed by brand/site — one customer-scoped
+ * session token authorizes every brand under that org.
+ */
+const sessionTokenCache = new Map();
```
amitruluimadalina, line 107: **Important:** This Map has no eviction beyond per-key overwrite - entries for orgs that stop being scanned accumulate for the life of a warm container with no size cap or sweep. Small per-entry cost, but unbounded module-level state like this tends to surface later as an unattributed slow leak. Recommend a light eviction path (sweep on write, or an LRU cap) before this caching pattern gets copied into other loaders.

docs/decisions/002-offsite-cited-urls-semrush-source.md
```diff
@@ -50,15 +50,19 @@ involves several non-obvious trade-offs, so it warrants an ADR alongside the spe
...
-   grant), default IMS client.** This is the same S2S path `commerce-product-enrichments`,
-   `vulnerabilities` and `permissions` use. v3 `getServiceAccessTokenV3()`
-   (`client_credentials`) was tried first but returns IMS **`400 unauthorized_client`** —
-   the worker's default IMS client is only provisioned for `authorization_code`; only a
-   dedicated integration (content-ai's `CONTENTAI_*`) is registered for `client_credentials`.
-   The token is forwarded unchanged to Semrush; no `x-promise-token` for a service caller.
-   **Open risk (LLMO-6709):** the proxy is designed around a real *user* IMS token, so whether
-   Semrush accepts the worker's *service* token is unverified — the flag stays off until
-   confirmed end-to-end (use `enableSemrush:true` on one canary run, per Decision 7, to test).
+   grant), default IMS client.** ⚠️ **Superseded by Decision 9 (LLMO-6709 resolved).** This
```
amitruluimadalina, line 53: **Important:** Leaving Decision 5 in place with a superseded banner and the original prose quoted back verbatim (rather than editing it down to a short historical note, or moving it to a dedicated superseded-decisions appendix) works for one supersession but doesn't scale - a reader has to hold two decisions in their head to understand current behavior. Worth settling a repo-wide convention here before a third supersession makes this ADR harder to parse than the code it describes. Not blocking this PR.

amitruluimadalina (changes requested):
Hey @tb-adbe,

Re-reviewed the one new commit (37bbf6a) against the seven previously-flagged issues.

## Strengths

- Previously flagged item 1 (untested IMS_CLIENT_CODE dependency) is now addressed: `getImsAuthorizationHeader` explicitly sets `IMS_CLIENT_CODE` with a placeholder and a comment explaining why `createFrom` needs it even though the client_credentials grant never sends it, with two new tests covering the default and the override.
- Previously flagged item 2 (no token caching) is now addressed: a module-level session-token cache keyed by `imsOrgId`, TTL 10 minutes under the server's 15-minute expiry, collapses the mint plus login exchange to zero network calls on a warm-container hit for a recently-seen org. Covered by a cache-hit test and a TTL-expiry re-mint test using fake timers.
- Previously flagged item 3 (redundant imsOrgId lookup) is now addressed: the handler threads its already-resolved `imsOrgId` into the loader, which only falls back to its own lookup when the caller omits it - confirmed there is exactly one resolution point per run now, with a dedicated test asserting the fallback lookup is skipped when the value is threaded.
- Previously flagged item 4 (auth vs authz conflation on the session-exchange leg) is now addressed: the thrown error carries `.status`, and 401/403 is classified into a distinct `session_token_auth_failed` reason mirroring the existing `fetchDomainUrls` `authFailure` pattern, with tests covering 401, 403, and a non-auth 500 distinctly.
- Previously flagged item 5 (unbounded response body forwarded to Slack) is now addressed: the full body stays in the structured log fields only; the Slack-facing message now carries only the HTTP status. A test directly asserts a secret-looking deny body reaches the log but never any Slack-bound call.
- Previously flagged item 6 (stale ADR) is now addressed: a new Decision 9 documents the S2S flow end to end, and Decision 5 plus the exit-criteria and configuration-debt sections are explicitly marked superseded rather than left as stale current-state claims.
- Previously flagged item 7 (permanent legacy fallback for sites with an unresolvable imsOrgId) is unchanged, which is expected since it is an operational/fleet-data concern rather than something this diff could address - the ADR is upfront about it and notes it is consistent with the existing DRS-scraping constraint.
- The `getImsAccessToken` -> `getImsAuthorizationHeader` rename and the naming/doc cleanup that goes with it is a genuine clarity improvement.

## Issues

### Important (Should Fix)

1. **src/utils/offsite-brand-presence-semrush.js:610 (and the domain-urls auth-failure branch around line 630-637)** - The new session-token cache is populated on a successful exchange but is never invalidated when the cached token is later rejected by the actual `domain-urls` call (a 401/403, `result.authFailure`). Before this commit, every call minted fresh, so an auth rejection self-healed on the very next invocation. Now, if a session token is revoked or an org's grant changes after being cached, every call for that `imsOrgId` within the remaining TTL window (up to 10 minutes) keeps replaying the same now-invalid cached token, keeps failing the same way, and keeps falling back to legacy - across potentially many audit runs for that org - instead of self-healing. This is a regression introduced specifically by the new caching mechanism, with no test coverage for this path. Fix: on `result.authFailure` at the domain-urls call site, evict the cache entry for `imsOrgId` (or retry once with a freshly-minted token) before falling back, and add a regression test that seeds the cache, forces an authFailure, and asserts the next call re-mints rather than reusing the poisoned token.

2. **src/utils/offsite-brand-presence-semrush.js:91** - `sessionTokenCache` is a plain Map with no eviction beyond a per-key overwrite; expired or one-off entries for orgs that stop being scanned are never removed and accumulate for the life of a warm container. The per-entry cost is small, but unbounded, monotonically-growing module state in a Lambda fleet is exactly the shape of issue that surfaces months later as a slow leak nobody can attribute quickly. Recommend a light eviction path (sweep expired entries opportunistically on write, or cap the map to a bounded LRU size) before this pattern is copied into other loaders, especially since this file's own ADR now flags the org-id to IMS-token to session-token sequence as an extraction candidate for a future second S2S consumer.

3. **docs/decisions/002-offsite-cited-urls-semrush-source.md:53 and :243** - Superseding Decision 5 by leaving it in place with a banner and a verbatim quoted blockquote, rather than editing it down or moving it to a short superseded-decisions note, works for a single supersession but does not scale: a reader now has to hold two decisions in their head (skim Decision 5's banner, jump to Decision 9, cross-reference which risk/consequence sections were rewritten) to understand current behavior. Not a blocker for this PR, but worth settling a repo-wide convention (edit in place with a one-line historical note, or a dedicated superseded-decisions appendix) before a third supersession makes this ADR harder to parse than the code it describes.

### Minor (Nice to Have)

- The 10-minute session-token cache TTL has no environment override, unlike `LLMO_S2S_LOGIN_URL`/`LLMO_API_BASE_URL`. If the login endpoint's actual token lifetime ever changes, there is no operational lever to adjust the cache window without a code deploy.
- `getCachedSessionToken`/`cacheSessionToken` have no defensive guard against a falsy `imsOrgId` polluting the cache key space; not currently reachable given the existing guard above the caching logic, but worth a comment given the cache is process-global and shared across all callers of this module.
- Carried forward from the prior review and not addressed in this commit (as expected, not the stated scope): non-prod environments still default toward production LLMO infrastructure unless both `LLMO_API_BASE_URL` and `LLMO_S2S_LOGIN_URL` are explicitly overridden, and the returned session token is still trusted as an opaque string with no local well-formedness or claims check before use.

## Recommendations

- Add the cache-eviction-on-authz-failure regression test alongside the fix for issue 1.
- Consider logging a distinct signal when a cache hit is immediately followed by a domain_urls_auth_failed, to distinguish "our cache went stale mid-window" from "the consumer registration is actually broken" without manual correlation.
- Given the ADR's own note that this org-id to IMS-token to session-token sequence is an extraction candidate for the next S2S-backed caller, consider factoring the cache helpers and the exchange sequence into a shared helper now, so the eviction-on-failure fix and the growth bound are not dropped again by a future copy-paste.

## Assessment

Ready to merge? With fixes.
Reasoning: All seven previously-flagged issues are genuinely and correctly resolved, not just superficially, and the fixes are well-tested. The new caching mechanism this commit introduces, however, carries its own new gap: a stale or revoked cached token is never evicted on a downstream auth rejection, which can turn a single auth blip into a sustained multi-minute outage window for an org that the pre-caching code would have self-healed immediately. That gap, together with the unbounded cache growth, should be closed before this reaches fleet-wide cadence. The ADR-structure note is not blocking.

## Next Steps

1. Add cache eviction on a domain-urls auth failure, with a regression test (item 1).
2. Add a growth bound or sweep to the session-token cache (item 2).
3. Settle the ADR supersession convention before or alongside merge (item 3); minor items are optional.

tb-adbe:
Thanks for the fast re-review @amitruluimadalina. Addressed the caching gaps in `314148f6`.

### Important

1. **No cache eviction on downstream auth rejection (regression)** — ✅ Fixed. On a `domain-urls` `authFailure` (401/403) the loader now evicts the cache entry for that `imsOrgId` before falling back, so a revoked/rotated token can't be replayed for the rest of its TTL — the next run re-mints, restoring the pre-caching self-heal-on-next-invocation. I went with **eviction** rather than in-run retry-with-fresh-token: eviction exactly restores the pre-caching behavior (that run fell back too), whereas retry would be a new behavior + a second round of network calls. Per your recommendation, I also added a `wasCachedToken` log field on the auth-failure fallback so ops can tell a mid-window staleness (`wasCachedToken=true`, self-heals next run) from a genuinely broken registration (`wasCachedToken=false`, a freshly-minted token rejected). Regression test seeds the cache, forces a 401, and asserts the next call re-mints; a companion test asserts a non-auth 500 does **not** evict.

2. **Unbounded cache growth** — ✅ Fixed. `cacheSessionToken` now opportunistically sweeps expired entries on each write, so the map is bounded by the number of distinct orgs seen within one TTL window rather than the life of the warm container. Test covers the sweep keeping other orgs' unexpired entries while dropping expired ones. (Went with sweep-on-write over an LRU cap — simpler, and the natural bound fits a session-token cache; happy to add a hard cap if you'd prefer a firmer ceiling.)

3. **ADR supersession convention** — Acknowledged, and agree it doesn't scale past one supersession. Leaving it out of this PR as you noted it's non-blocking; I'll raise a repo-wide convention (edit-in-place-with-note vs. a superseded-decisions appendix) separately so it's settled before a third supersession rather than decided ad hoc here.

### Minor

- **Cache TTL env override** — ✅ Added `SEMRUSH_S2S_SESSION_TTL_MS` (invalid/non-positive values fall back to the 10-min default), closing the asymmetry with `LLMO_S2S_LOGIN_URL`/`LLMO_API_BASE_URL`. Tested (override honored; invalid values fall back).
- **Falsy `imsOrgId` cache-key guard** — Added a comment on the cache: it's only ever reached with a truthy `imsOrgId` (the `no_ims_org_id` guard returns first), so the key space can't be polluted. Left as a documented invariant rather than a redundant runtime guard.
- **Carried-forward minors** (prod-default host, opaque-token trust) — unchanged, as you flagged they're out of this PR's scope.

### Recommendations

- **Distinct stale-cache signal** — ✅ Done via the `wasCachedToken` field above.
- **Extract shared helper (cache + exchange sequence)** — Deferred, but noted in ADR 002's config-debt section as the extraction candidate so the eviction + growth-bound logic here isn't lost to a future copy-paste; I'll do it when the second S2S consumer lands.

Verification: lint clean, loader module at 100% coverage, 69 loader tests + full handler suite passing.

AndreiAlexandruParaschiv (approved):

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -435,8 +670,21 @@ export async function loadCitedUrlsFromSemrush({
 
   const result = await fetchDomainUrls(url, headers, olog, PAGE_SIZE);
   if (!result.ok) {
+    if (result.authFailure) {
+      // The session token was rejected by the data call — evict it so a revoked/rotated
+      // token can't be replayed from the cache for the rest of its TTL; the next run
+      // re-mints. `wasCachedToken` lets ops tell a mid-window staleness (self-heals next
+      // run) from a genuinely broken registration (a freshly-minted token rejected).
+      evictSessionToken(imsOrgId);
```
amitruluimadalina, line 678: Minor: eviction fires identically on 401 and 403 here, but these are different failure classes (401 = the token itself is invalid; 403 = the token is valid but the bearer lacks sufficient scope/grant, RFC 6750 section 3.1). For a durable 403 (the grant was actually revoked or never existed for this org), evicting means every subsequent invocation now pays a full mint + login-exchange round trip instead of failing cheaply from cache once per TTL window - not a regression below the original pre-caching behavior, but a good fast-follow candidate: evict only on 401, or add a short backoff for repeated 403s on the same org.

amitruluimadalina (approved):
Hey @tb-adbe,

Re-reviewed the one new commit (314148f) against the two Important issues from the last round.

## Strengths

- Previously flagged item 1 (cached session token never invalidated on a downstream auth rejection) is now correctly fixed: `evictSessionToken(imsOrgId)` is called when the data call's `result.authFailure` is true, deleting the cache entry so the next invocation re-mints. Traced end to end against the new tests (eviction-on-401 across three sequential calls, no-eviction-on-500 confirming the scope is auth-only) and the fix is complete, not superficial - it correctly evicts both a stale cache hit and a freshly-minted-then-rejected token.
- Previously flagged item 2 (unbounded cache growth) is now correctly fixed: `cacheSessionToken` sweeps expired entries from the map on every write, bounding growth to distinct orgs live within one TTL window rather than distinct orgs seen over a container's entire lifetime. The sweep-during-iteration pattern is safe, and the multi-org test confirms an unexpired entry for a different org survives the sweep.
- The new `SEMRUSH_S2S_SESSION_TTL_MS` override and its fail-safe fallback (rejecting NaN, zero, and negative values) are correctly implemented and tested, as is the documented `imsOrgId` cache-key guard.
- Test coverage for all of the new logic is thorough and each test genuinely exercises the claimed scenario via call-count assertions rather than just return values.

## Issues

### Minor (Nice to Have)

- **src/utils/offsite-brand-presence-semrush.js:678** - Eviction now fires identically on both 401 and 403, but these are not the same failure class (RFC 6750 section 3.1: 401 means the token itself is invalid, 403 means the token is valid but the bearer lacks sufficient scope/grant). For a 401 this fix is exactly correct. For a durable 403 (the S2S consumer's brand:read grant was actually revoked or never existed for this org), evicting buys nothing - the next mint plus exchange produces a token that fails the same way, and it gets evicted again, so a persistently-misconfigured org pays a full three-leg round trip on every subsequent invocation instead of failing cheaply from cache once per TTL window. Worth noting this is not a regression below the original pre-caching behavior (every call always minted fresh before this feature existed), just a reversion to that baseline for this one narrow, already-broken-registration case - not blocking, but a good candidate for a fast-follow: evict only on 401, or add a short backoff for repeated 403s on the same org.
- **src/utils/offsite-brand-presence-semrush.js:97** - `resolveSessionTtlMs` has no upper bound. An operator setting `SEMRUSH_S2S_SESSION_TTL_MS` above the server's real ~15-minute session-token lifetime would defeat the reason the cache TTL is kept under that expiry in the first place. Consider clamping the override to a sane value relative to the documented server lifetime.
- The eviction path is keyed only by `imsOrgId`, not by the specific token instance that failed - if this file ever batches multiple brands under the same org concurrently within one invocation, an interleaved race could have one call's eviction remove a different, still-valid token freshly cached by a concurrent call for the same org. Self-healing and low severity given the current single-call-per-invocation usage.
- No test currently pins `wasCachedToken=false` for the freshly-minted-then-rejected case, and the invalid-TTL-override test only proves resolution does not throw, not that the fallback value is exactly the 10-minute default. Neither is required, both would close the loop given the stated intent of the diagnostic field.

## Recommendations

- Consider splitting the eviction policy (401 evict-and-remint immediately, 403 leave cached for the remainder of its TTL or add a backoff) as a fast-follow, so a durable authorization gap does not generate continuous retry pressure against IMS and the login endpoint for its full duration.
- If left as-is for now, consider a metric or alert on repeated `wasCachedToken:false` plus `authFailure` for the same `imsOrgId`, so a broken S2S grant surfaces as an incident rather than generating background IMS load until someone greps logs.
- Consider clamping the TTL override to a bound relative to the server's real session lifetime.

## Assessment

Ready to merge? Yes.
Reasoning: Both Important issues from the last round are genuinely and completely fixed, with no new correctness bugs in the fix mechanism itself, and this is well tested. The 401/403 eviction-policy note is a real but narrow-precondition, self-healing operational tradeoff, not worse than the original pre-caching baseline, and does not need to block this PR - worth a fast-follow rather than a blocking fix.

PR #2970, opened 2026-09-14
chore(offsite-brand-presence): clean up S2S debugging, restore page size, add optional overrides (LLMO-6709)

tb-adbe:
## What changed

Now that Semrush S2S access works end-to-end, this removes the debugging-only additions and the temporary limit added while diagnosing it — keeping the pieces that pay off beyond this incident.

### Removed
- **Restored `PAGE_SIZE` 50 → 1000** — back to the server clamp / ADR 002 Decision 6 value, so a low-citation bucket (reddit/youtube on a press-heavy site) isn't starved by the shared citations-sorted page. (The 50 cap was a debugging-time limit.)
- **`imsHostHasScheme` / `imsScopeHasSpaces` flags** — they only ever flagged the two specific setup mistakes (scheme in `SEMRUSH_S2S_IMS_HOST`, whitespace in the scope), both now fixed.
- **`requestUrl` on the domain-urls start (`Querying…`) and success (`Bucketed…`) logs** — verbose on every run; kept on the **failure** logs where it aids diagnosis.

### Kept (useful beyond debugging)
- **`/api/v1` gateway prefix** — the actual fix (#2967).
- **60s data-call timeout** (separate from the 10s login timeout) — `domain-urls` legitimately runs longer than 10s.
- **`imsConfigDiagnostics`** on `ims_token_failed` (failure-only: `imsHost`/`imsClientId`/`imsScope`/`hasClientSecret`) — pays off on future cred rotation / new-env config drift.
- **Granted consumer-identity log** (`Obtained S2S session token` with `consumerClientId`/`tenantCount`) — low-volume audit line mirroring api-service's `[s2s] granted`.
- **`requestUrl` on failure logs.**

### Docs
- ADR 002: added **Decision 10** documenting the `/api/v1` gateway prefix and the 60s data-call timeout. (Decision 6's page size is 1000 again, so it's consistent.)

## Reviewer notes
- No secrets/token values are logged (unchanged).
- Loader module remains at **100%** coverage; 78 loader + 108 handler tests pass; lint clean.
- Behavioral change to watch: with `PAGE_SIZE` back to 1000 the response is larger/slower than the 50 it ran with during rollout — the 60s timeout covers it, but worth a sanity check that runs still complete comfortably.

## Related Issues

LLMO-6709

## Added: optional env overrides (defaults unchanged)

Two optional knobs so the `domain-urls` call can be tuned per environment without a code deploy — both fail-safe to their defaults on an absent/invalid value:

- `OFFSITE_SEMRUSH_TIMEOUT_MS` — data-call timeout (default **60s**; login exchange stays 10s).
- `OFFSITE_SEMRUSH_PAGE_SIZE` — requested page size (default/ceiling **1000**; a positive integer clamped to `[1, 1000]`).

ADR 002 Decisions 6 and 10 updated to note the overrides.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


src/utils/offsite-brand-presence-semrush.js
```diff
@@ -55,15 +55,28 @@ export const LLMO_API_DEFAULT_BASE_URL = 'https://llmo.experiencecloud.live';
...
+ * @param {object} [env]
+ * @returns {number}
  */
-export const PAGE_SIZE = 50;
+function resolvePageSize(env) {
+  const override = Number(env?.OFFSITE_SEMRUSH_PAGE_SIZE);
+  if (!Number.isFinite(override) || override <= 0) {
+    return PAGE_SIZE;
+  }
+  return Math.min(Math.floor(override), PAGE_SIZE);
```
amitruluimadalina: **Important:** `resolvePageSize` has no lower-bound clamp. For a fractional override like `SEMRUSH_DOMAIN_URLS_PAGE_SIZE='0.5'`, `override = 0.5` passes the `Number.isFinite(override) || override <= 0` check, then `Math.floor(0.5) = 0`, so `pageSize` becomes `0`. That is sent verbatim as `pageSize=0` in the query string, and the truncation check becomes `0 >= 0 = true`, so the run looks like a successful call that legitimately found zero cited URLs rather than a config error. Fix: clamp the lower bound too, e.g. `Math.min(Math.max(Math.floor(override), 1), PAGE_SIZE)`, and add a test for a fractional override.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -72,12 +85,25 @@ export const PAGE_SIZE = 50;
...
+
+/**
+ * Resolves the offsite Semrush data-request timeout from env, ignoring a non-numeric or
+ * non-positive override (fail-safe to the default rather than a 0/NaN timeout).
+ *
+ * @param {object} [env]
+ * @returns {number} timeout in ms.
+ */
+function resolveSemrushTimeoutMs(env) {
+  const override = Number(env?.OFFSITE_SEMRUSH_TIMEOUT_MS);
```
amitruluimadalina, line 116: **Important:** `resolveDomainUrlsTimeoutMs` has no upper-bound clamp, unlike its sibling `resolvePageSize`. Nothing stops `SEMRUSH_DOMAIN_URLS_TIMEOUT_MS` from being set to an arbitrarily large value (e.g. mid-incident). The 60s default exists as headroom against the Lambda's 900s budget so this call can fail gracefully into the existing degrade paths; an unbounded override defeats that and risks a hard Lambda kill instead of a clean degrade. Recommend a ceiling constant tied to the 900s budget, mirroring the treatment `resolvePageSize` already gets.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -223,23 +249,16 @@ async function getImsAuthorizationHeader(context) {
...
 function imsConfigDiagnostics(env) {
-  const imsHost = env?.SEMRUSH_S2S_IMS_HOST;
-  const imsScope = env?.SEMRUSH_S2S_CLIENT_SCOPE;
   return {
-    imsHost,
+    imsHost: env?.SEMRUSH_S2S_IMS_HOST,
     imsClientId: env?.SEMRUSH_S2S_CLIENT_ID,
-    imsScope,
+    imsScope: env?.SEMRUSH_S2S_CLIENT_SCOPE,
     hasClientSecret: Boolean(env?.SEMRUSH_S2S_CLIENT_SECRET),
```
amitruluimadalina, line 280: **Important:** removing `imsHostHasScheme`/`imsScopeHasSpaces` trades away cheap, real diagnostic value for a class of config mistake that is likely to recur, not a one-off. These flags exist because IMS host/scope values are typically sourced from Vault/Helm values across multiple environments and get hand-edited; a scheme-prefixed host or whitespace-in-scope is exactly the kind of mistake a different environment or a future engineer reintroduces later. "Both now fixed" is true of the current values, not of the mistake class. The flags cost nothing at runtime (pure derived booleans, emitted only when true). Recommend keeping them, or at minimum preserving the derivation as a small exported helper so it is a one-line re-add rather than reconstructing the regex from git history when this recurs.

src/utils/offsite-brand-presence-semrush.js
```diff
@@ -811,7 +832,6 @@ export async function loadCitedUrlsFromSemrush({
     direction: 'inbound',
     orgId: spaceCatId,
```
amitruluimadalina, line 854: **Important:** dropping `requestUrl` from this success (`Bucketed...`) log removes the only way to correlate this line with the start line or a later failure, with no compensating field added. Association now relies on timestamp proximity plus `orgId`/`brandId`, which degrades multi-tenant log correlation under concurrent invocations for different brands of the same org. Worth checking whether the `olog` helper already stamps a run-scoped correlation id before accepting this as sufficient.

amitruluimadalina (changes requested):
Hey @tb-adbe,

## Strengths

- The new `resolvePageSize`/`resolveDomainUrlsTimeoutMs` env overrides follow the existing fail-safe-clamp pattern already used by `resolveSessionTtlMs` in this file - parse, validate `Number.isFinite && > 0`, fall back to the module default. Consistent, not a bespoke one-off.
- `resolvePageSize` correctly clamps to the server's own ceiling (`Math.min(..., PAGE_SIZE)`), and the resolved `pageSize`/`timeoutMs` are threaded through to `fetchDomainUrls` rather than closing over the module constants, so the existing truncation check (`raw.length >= pageSize`) still works against the overridden value, not the old default.
- Failure-path observability is preserved: `requestUrl` was only dropped from the start/success logs; all three failure branches in `fetchDomainUrls` (fetch failure, non-2xx, auth-rejected) still carry it, and this is directly exercised by a test.
- The `imsHostHasScheme`/`imsScopeHasSpaces` removal, the `PAGE_SIZE` restoration, and the `Bucketed...` log change are all cleanly done: no orphaned test assertions or other references to the removed flags remain anywhere in the repo, and the `PAGE_SIZE` restoration is regression-protected by a test that checks both the exported constant and the wire value.
- New override behavior (valid, invalid, and clamp-to-ceiling cases) gets dedicated tests for both `SEMRUSH_DOMAIN_URLS_TIMEOUT_MS` and `SEMRUSH_DOMAIN_URLS_PAGE_SIZE`.
- ADR 002 Decision 10 is substantively correct and matches the shipped code and the companion PR (`/api/v1` prefix, 60s data-call timeout) rather than drifting from it.

## Issues

### Important (Should Fix)

1. **`resolvePageSize` has no lower-bound clamp, so a fractional override silently produces `pageSize=0`.** (`src/utils/offsite-brand-presence-semrush.js:78`)
   For `SEMRUSH_DOMAIN_URLS_PAGE_SIZE='0.5'`: `override = 0.5` is finite and `> 0`, so it skips the fallback, then `Math.floor(0.5) = 0`, giving `pageSize = 0`. That's sent verbatim as `pageSize=0` in the query string, and the existing truncation check becomes `0 >= 0 = true`, so the run looks like a successful Semrush call that legitimately found zero cited URLs, rather than a config error. Only reachable via Lambda env misconfiguration, not request input, but it's a real gap the current tests don't cover (only `'0'`, `'5000'`, `'200'` are tested, not a sub-1 fractional value). Fix: clamp the lower bound too, e.g. `Math.min(Math.max(Math.floor(override), 1), PAGE_SIZE)`, and add a test for a fractional override.

2. **`resolveDomainUrlsTimeoutMs` has no upper-bound clamp, unlike its sibling `resolvePageSize`.** (`src/utils/offsite-brand-presence-semrush.js:104`)
   Nothing stops `SEMRUSH_DOMAIN_URLS_TIMEOUT_MS` from being set to an arbitrarily large value (e.g. mid-incident, someone reasoning "just give it more time"). The whole point of the 60s default was headroom against the Lambda's 900s budget so this call can fail gracefully into the existing degrade paths; an unbounded override defeats that and risks a hard Lambda kill instead of a clean degrade. Recommend a ceiling constant tied to the 900s budget, mirroring the treatment `resolvePageSize` already gets.

3. **Removing `imsHostHasScheme`/`imsScopeHasSpaces` trades away cheap, real diagnostic value for a class of config mistake that is likely to recur, not a one-off.** (`src/utils/offsite-brand-presence-semrush.js:255-262`)
   These flags exist because IMS host/scope values are typically sourced from Vault/Helm values across multiple environments and get hand-edited; a scheme-prefixed host or whitespace-in-scope is exactly the kind of mistake a different environment or a future engineer reintroduces later (new customer IMS org, scope rotation, new-env onboarding). "Both now fixed" is true of the current values, not of the mistake class - there's no validation preventing recurrence. The flags cost nothing at runtime (pure derived booleans, emitted only when true). Recommend keeping them, or at minimum preserving the derivation as a small exported helper (e.g. `classifyImsConfigMistake`) so it's a one-line re-add rather than reconstructing the regex from git history when this recurs.

4. **Dropping `requestUrl` from the success (`Bucketed...`) log removes the only way to correlate that line with its start line or a later failure, with no compensating field added.** (`src/utils/offsite-brand-presence-semrush.js:829-833`)
   Once `requestUrl` is gone from both the start and success logs, association relies on timestamp proximity plus `orgId`/`brandId`, which degrades multi-tenant log correlation in Splunk/CloudWatch under concurrent invocations for different brands of the same org. Worth checking whether the `olog` helper already stamps a run-scoped correlation id before accepting this as sufficient; if not, consider adding one.

5. **No test covers the interaction between a downward `SEMRUSH_DOMAIN_URLS_PAGE_SIZE` override and the truncation/`diagnostics.truncated` signal.** (`test/utils/offsite-brand-presence-semrush.test.js`)
   The truncation check is correctly wired to use the resolved `pageSize`, not the module constant, today - but there's no test asserting `diagnostics.truncated === true` when a full page comes back against an overridden (non-1000) ceiling. A future refactor that accidentally reverts the truncation check to close over the `PAGE_SIZE` constant instead of the local parameter would silently break the starvation-detection signal for any environment using the override, with nothing in the current suite catching it. Recommend adding one test: run with a low page-size override, feed a full page of rows, assert `diagnostics.truncated === true`.

6. **The PR description covers only the first commit; the second commit's new env knobs (`SEMRUSH_DOMAIN_URLS_TIMEOUT_MS`, `SEMRUSH_DOMAIN_URLS_PAGE_SIZE`) aren't mentioned.** Given this repo's own convention of documenting every override in both the ADR and the PR body, and that these are genuinely new runtime-configurable behavior (not pure cleanup), the description (and arguably the "chore" framing) should be updated to describe what's actually shipping, so a future incident responder or reviewer working from the PR record alone knows these knobs exist.

### Minor (Nice to Have)

- `resolveDomainUrlsTimeoutMs` accepts non-integer values (e.g. `'1.5'`) without complaint; not a correctness issue on its own, but related to the point below.
- Three near-identical numeric env-override resolvers now exist in this file (`resolveSessionTtlMs`, `resolveDomainUrlsTimeoutMs`, `resolvePageSize`) with copy-pasted validation logic. Worth a shared `resolvePositiveIntEnvOverride(env, key, default, { min, max })` helper before a fourth knob shows up.
- ADR 002 Decision 6 could note explicitly that `PAGE_SIZE=1000` is now the value that runs in production traffic (not just "the ceiling"), since the original decision described 1000 as aspirational/max-only.
- The JSDoc on the truncation return type still describes truncation relative to `PAGE_SIZE`; now that page size is overridable, it should say "the requested/effective page size" so a future reader debugging a truncation report with an override in play isn't misled.
- `docs/specs/2026-08-05-offsite-cited-urls-semrush-source.md` (not touched by this PR) still describes `PAGE_SIZE` as "a fixed constant" and cites a stale 10s timeout; this PR makes that doc slightly more wrong. Worth a follow-up pass since ADR 002 and the spec doc now disagree.
- No dedicated pure-function test block exists for `resolvePageSize`/`resolveDomainUrlsTimeoutMs` edge cases (fractional, negative) - current coverage is via the integration path only, which is real coverage but makes intent less explicit.

## Recommendations

- Add a shared numeric-override resolver helper to replace the three near-duplicate functions before a fourth env-tunable knob is added.
- Log the resolved `pageSize`/`timeoutMs` once at cold start so an operator debugging a starved run doesn't need Helm/shell access to the running Lambda's env to know whether an override is active.
- Confirm the deploy repo's Helm values naming doesn't collide with or diverge from the new `SEMRUSH_DOMAIN_URLS_*` env var names.
- Do the sanity check the PR's own reviewer notes flag: confirm the 60s timeout has comfortable headroom against a real max-size (1000-row) Semrush response before this rides to prod, since PAGE_SIZE=1000 is now the default rather than the 50 used during rollout.

## Assessment

Ready to merge? With fixes.
Reasoning: nothing here is externally exploitable (all the flagged inputs are Lambda environment configuration, not request-derived), the "no secrets logged" claim holds up against the full file, and the core restoration/ADR-documentation work is sound and consistent with existing patterns. The `resolvePageSize` lower-bound clamp (issue 1) is the one item worth treating as a real correctness bug to block on - it's cheap to fix and can otherwise produce a silent, hard-to-diagnose "successful" empty result in production. The rest are solid catches worth addressing but not blockers on their own.

## Next Steps
1. Fix the `resolvePageSize` lower-bound clamp and add the missing fractional-override test (issue 1).
2. Add the upper-bound clamp on `resolveDomainUrlsTimeoutMs` (issue 2).
3. Decide on issue 3 (keep or preserve the IMS gotcha flags) and update the PR description for issue 6.
4. The remaining important items (4, 5) and all minor items are good to fold in but non-blocking on their own.

amitruluimadalina:
One clarification on my review above: the timeout-clamp comment on `src/utils/offsite-brand-presence-semrush.js:104` was written against the prior commit's naming (`resolveDomainUrlsTimeoutMs` / `SEMRUSH_DOMAIN_URLS_TIMEOUT_MS`). After the follow-up rename commit these are `resolveSemrushTimeoutMs` / `OFFSITE_SEMRUSH_TIMEOUT_MS`, but the finding itself is unchanged: that function still has no upper-bound clamp, unlike `resolvePageSize`. The other three inline comments (lines 78, 261, 833) aren't affected by the rename.

tb-adbe:
Thanks @amitruluimadalina — great review. Addressed in `f80d9aea`. (Note the env vars were renamed to `OFFSITE_SEMRUSH_*` in the commit before this review, so the fixes use those names.)

### Important

1. **`resolvePageSize` lower-bound clamp** — ✅ Fixed. Now `Math.min(Math.max(Math.floor(override), 1), PAGE_SIZE)`, so a fractional `0.5` clamps up to `1` instead of flooring to `0` (which would have made an empty response look like a legitimate zero-URL run). Added a fractional-override test.

2. **`resolveSemrushTimeoutMs` upper clamp** — ✅ Fixed. A valid override is clamped to a **2-minute ceiling** (`SEMRUSH_TIMEOUT_MAX_MS`), so it can't approach the 900s Lambda budget and turn a clean degrade into a hard kill. Added a clamp-to-ceiling test; ADR 002 Decision 10 notes it.

3. **Re-add `imsHostHasScheme` / `imsScopeHasSpaces`** — ✅ Done. Restored both derived flags (pure booleans, emitted only when true on `ims_token_failed`) — you're right that it's a recurring hand-edited-config mistake class, not a one-off, and they cost nothing at runtime. Re-added the two tests. Logging only, no behavior change.

4. **`requestUrl` dropped from start/success hurts correlation** — Respectfully leaving as-is. The platform `logWrapper` already stamps a per-invocation **`traceId`** on every line (visible in the incident's own Splunk output: `[traceId=1-6aa400ce-…]`; `offsite-logging.js` references it too). Every log line of a run — start, `Obtained S2S session token`, `Bucketed…`, and all failure branches — shares that `traceId`, and each brand runs in its own invocation with its own `traceId`, so concurrent multi-tenant correlation holds without `requestUrl`. It remains on the failure branches where the exact URL is diagnostically useful; on the happy path it was redundant with `traceId` and verbose, which is why it was dropped.

5. **Truncation test against an overridden page size** — ✅ Added. Runs with a low `OFFSITE_SEMRUSH_PAGE_SIZE` override, feeds a full page, and asserts `diagnostics.truncated === true` — guarding the truncation check from a future refactor that closes over the `PAGE_SIZE` constant instead of the resolved value.

6. **PR description missing the env knobs** — Already addressed before this review: the description has an "optional env overrides" section documenting both `OFFSITE_SEMRUSH_TIMEOUT_MS` and `OFFSITE_SEMRUSH_PAGE_SIZE`, and the title was updated to "…add optional overrides."

### Minor / recommendations
- **Shared `resolvePositiveIntEnvOverride` helper** for the three near-duplicate resolvers — good call; deferring to a small follow-up so this PR stays focused (happy to do it now if you'd prefer).
- **ADR Decision 6 / JSDoc "effective page size" / spec-doc staleness** — will fold the doc-accuracy tweaks into the follow-up; the spec doc (`2026-08-05-…`) is a separate file worth its own pass.
- **Ops:** confirmed the naming is `OFFSITE_SEMRUSH_*` (not `SEMRUSH_DOMAIN_URLS_*`) for the Helm check. And agreed on the **60s-vs-1000-row** sanity check — since the default is now 1000 (we rolled out on 50), I'll verify a full-size response returns comfortably under 60s before this merges.

Loader module remains at 100% coverage; lint clean.

amitruluimadalina (approved):
Hey @tb-adbe,

Re-reviewed after the follow-up commit (`f80d9aea`) addressing the prior review.

## Strengths

Five of the six prior Important findings are correctly and completely fixed, each with a targeted regression test:

- **Page-size lower clamp (previously finding 1):** `resolvePageSize` now does `Math.min(Math.max(Math.floor(override), 1), PAGE_SIZE)`, closing the fractional-override path that could silently send `pageSize=0`. New test covers `OFFSITE_SEMRUSH_PAGE_SIZE='0.5'` -> `pageSize=1`.
- **Timeout upper-bound clamp (previously finding 2):** new `SEMRUSH_TIMEOUT_MAX_MS` (2 minutes) constant, well-reasoned against the 900s Lambda budget, with `resolveSemrushTimeoutMs` now clamping via `Math.min(override, SEMRUSH_TIMEOUT_MAX_MS)`. Test confirms a 10-minute override clamps to 120000ms.
- **IMS diagnostic flags restored (previously finding 3):** `imsHostHasScheme`/`imsScopeHasSpaces` are back, and improved over the original version - conditionally spread so they only appear in the log line when true, rather than emitted as `false` noise. Three tests cover the absent, scheme-in-host, and whitespace-in-scope cases.
- **Truncation-vs-override test (previously finding 5):** new test confirms `diagnostics.truncated === true` still fires correctly against an overridden (non-1000) page size, using the exact scenario recommended - a full page at the overridden ceiling.
- **PR description updated (previously finding 6):** the PR body now documents both `OFFSITE_SEMRUSH_TIMEOUT_MS` and `OFFSITE_SEMRUSH_PAGE_SIZE`, their defaults, and clamp behavior; ADR 002 is updated consistently.

No new issues were introduced by this commit - each clamp is minimal, correctly ordered relative to the existing guards, and backed by a test that exercises the actual regression scenario rather than a superficial happy path.

## Issues

### Minor (Nice to Have)

- **The `requestUrl` success-log correlation gap (previously finding 4) is still open.** This commit didn't touch the `Bucketed...`/`Querying...` log lines, and the PR body still frames the original removal as final rather than a reviewed trade-off. It's low-risk - `orgId`/`brandId` still give coarse correlation, and this was never a correctness or security concern - but it would be good to either add a lightweight correlation field or add one sentence in the PR body noting this was deliberately left as-is, so a future reader doesn't have to re-derive that it was a considered decision rather than an oversight.
- With a second clamped resolver and a second ceiling constant (`SEMRUSH_TIMEOUT_MAX_MS`) now in the file alongside `resolvePageSize`, extracting a shared `resolveClampedIntEnvOverride(env, key, { min, max, default })` helper is a more worthwhile fast-follow than before - not blocking, but cheaper to do now than after a fourth knob shows up.
- The JSDoc describing truncation is still slightly stale relative to the resolved `pageSize` parameter it actually checks against (carried over from the prior round, not touched by this commit).

## Recommendations

- Fast-follow on the shared numeric-override-resolver helper while there are only two implementations to unify.
- Either close finding 4 with a small correlation-field addition, or add the one-line acknowledgment to the PR body so it reads as a decision rather than an omission.

## Assessment

Ready to merge? Yes.
Reasoning: all correctness-relevant findings from the prior review (page-size clamp, timeout ceiling, diagnostic flags, truncation test coverage, PR documentation) are resolved cleanly and are well-tested. The one carried-forward item is a documented, low-risk observability trade-off, not a defect, and doesn't block merge on its own.

AndreiAlexandruParaschiv (approved):

PR #2971, opened 2026-09-14
fix(offsite-brand-presence): enableSemrush:true falls back to legacy on failure (LLMO-6709)

tb-adbe:
## What changed

A per-run Slack `enableSemrush:true` run now **falls back to the legacy source on a Semrush failure**, exactly like the `OFFSITE_BRAND_PRESENCE_SEMRUSH_ENABLED=true` env flag — instead of **hard-stopping** (`success:false`, no legacy).

## Why

The hard-stop was a rollout safeguard: while validating the Semrush S2S path, a forced `enableSemrush:true` canary run should surface a failure loudly rather than silently mask it with legacy. Now that S2S works in production, that asymmetry is no longer wanted — a Slack-triggered run should degrade gracefully to legacy just like the fleet path, so a Semrush hiccup never zeroes out offsite for that run.

### Before → after

| Enabled via | Semrush technical failure — before | after |
|---|---|---|
| Slack `enableSemrush:true` | ❌ hard stop (`success:false`, no legacy) | ✅ **falls back to legacy** |
| Env `OFFSITE_BRAND_PRESENCE_SEMRUSH_ENABLED=true` | ✅ falls back to legacy | ✅ falls back to legacy (unchanged) |

Unchanged: entitlement skips (`not_entitled` / `entitlement_check_failed`) already fell back and still do (logged at `outcome=skip` vs a technical failure's `outcome=degraded`); a genuinely-empty-but-successful Semrush result (0 URLs) is still a normal zero-URL Semrush run, not a failure.

## Changes

- Removed the hard-stop branch and `hardStopOnFailure` in `offsite-brand-presence/handler.js`; the failure path always sets `dataSource:'legacy'` + `fallbackReason` and continues to legacy.
- ADR 002: updated **Decision 1** and **Decision 7** to record the always-fallback behavior (and note the dropped hard-stop for history).
- Updated the handler tests that asserted the hard-stop to assert legacy fallback.

## Reviewer notes

- `handler.js` remains at **100%** coverage; offsite test suites pass (230 across loader + handler + drs-status).
- Independent of the in-flight cleanup PR #2970 (that one touches the loader + Decisions 6/10; this touches the handler + Decisions 1/7 — non-overlapping, no conflict).

## Related Issues

LLMO-6709

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
```


docs/decisions/002-offsite-cited-urls-semrush-source.md
```diff
@@ -15,19 +15,17 @@ involves several non-obvious trade-offs, so it warrants an ADR alongside the spe
 
 ## Decision
```
amitruluimadalina, line 17: Decisions 8, 8b, and 8c further down in this same file still say entitlement-skip reasons are "exempted from the Decision 1 hard-stop" and reference "the handler's hard-stop-exemption check." Since this rewrite removes hard-stop entirely, those later decisions now describe a dangling concept and need reconciling in this PR (along with docs/specs/2026-08-05-offsite-cited-urls-semrush-source.md:143, which still lists the removed hard-stop as a live acceptance criterion).

test/audits/offsite-brand-presence.test.js
```diff
@@ -324,25 +324,27 @@ describe('Offsite Brand Presence Handler', function () {
       );
     });
 
-    it('hard-stops (success:false, no legacy) when an enableSemrush:true run fails', async () => {
-      // Forced on via the Slack override — a failure must be visible, not masked.
+    it('falls back to legacy when an enableSemrush:true run fails (same as the env flag)', async () => {
```
amitruluimadalina, line 327: No test in this file covers the double-failure scenario this PR is meant to guard against: enableSemrush:true forces a genuine technical failure (not an entitlement skip) AND the legacy fallback also returns no URLs. Today that run reports success:true with all-zero urlCounts and only a fallbackReason field as evidence anything went wrong. Worth a test pinning down the expected diagnostic signal for that case.

amitruluimadalina (changes requested):
Hey @tb-adbe,

## Strengths

- The removal is clean and narrowly scoped: `hardStopOnFailure`, the dedicated `olog.failure(...)` call, and the `:x: ... stopping` Slack message are deleted with no orphaned references, unused variables, or lint fallout in `src/offsite-brand-presence/handler.js`. The entitlement-skip vs. technical-failure distinction (`isEntitlementSkip` -> `outcome=skip` vs `outcome=degraded`) is preserved verbatim.
- The rewritten tests in `test/audits/offsite-brand-presence.test.js` are meaningfully stronger than a flipped boolean: the primary case now asserts `mockLoadBrandPresenceData` was actually called and that `urlCounts['youtube.com']` reflects real legacy data flowing through end to end.
- The ADR (`docs/decisions/002-offsite-cited-urls-semrush-source.md`) is updated in the same PR for Decisions 1 and 7, keeping code and documented contract in sync for the parts that were touched - good discipline, just incomplete (see Issues).
- Slack visibility for genuine technical failures is not actually lost by this change: `loadCitedUrlsFromSemrush` (`src/utils/offsite-brand-presence-semrush.js`) already posts a distinct, often `:x:`-prefixed `notify()` message on every one of its failure branches (IMS token failure, `domain-urls` request failure, etc.) before returning `null`, wired through `onProgress` -> `postMessageOptional` in the handler. Removing the handler's own duplicate hard-stop message does not create a silent-failure gap in the Slack thread.
- No sibling handler (`cited-analysis`, `youtube-analysis`, `reddit-analysis`) implemented an equivalent hard-stop, so this PR doesn't leave a sibling handler's failure semantics inconsistent with this one.

## Issues

### Important (Should Fix)

1. **The ADR is now internally self-contradictory, and a companion spec doc still describes the removed behavior as current.** `docs/decisions/002-offsite-cited-urls-semrush-source.md` Decision 1 was rewritten to say a failure "ALWAYS falls back to legacy," but Decisions 8, 8b, and 8c (lines 112, 135, 140, 142, 144 in the base file, untouched by this PR) still say entitlement-skip reasons are "exempted from the Decision 1 hard-stop," reference "the handler's hard-stop-exemption check," and are titled "`entitlement-check-failed` stays exempted from hard-stop." Since hard-stop no longer exists for any reason, "exemption from hard-stop" is now a dangling concept that will confuse the next reader of the same file. Separately, `docs/specs/2026-08-05-offsite-cited-urls-semrush-source.md:143` still lists as a live acceptance criterion: "`enableSemrush:true` (Slack override) + Semrush fails => **hard stop** (`success:false`, `dataSource:'semrush'`, no fallback)" - this directly contradicts the new behavior and wasn't touched by the diff. Fix: sweep both docs for "hard stop"/"hard-stop" and either strike the obsolete framing or mark it explicitly as historical, following the same "(History: ...)" pattern already used in the Decision 1 rewrite.

2. **Stale comments in two untouched source files still describe a mechanism that no longer exists.** `src/utils/semrush-entitlement.js:55` and `:71` document `SEMRUSH_ENTITLEMENT_SKIP_REASONS` as feeding "the handler's hard-stop-exemption check," and `src/utils/offsite-brand-presence-semrush.js:610` calls `fallbackReason` "the coarse, contract-level signal the handler's hard-stop [check consumes]." The `hardStopOnFailure` variable and its branch no longer exist anywhere in the handler. These comments will mislead a future engineer who greps for "hard-stop" while trying to understand these exported values. Fix: update both comments to describe the current contract (entitlement skip vs. technical failure, both now falling back to legacy).

3. **Test titles retained from before this change still frame a distinction that no longer exists.** `test/audits/offsite-brand-presence.test.js:351` ("falls back to legacy when an ENV-enabled Semrush run fails **(no hard stop)**"), `:494`, and `:526` ("**(never hard-stops)** ... even with enableSemrush:true") were written when there was a real env-flag-vs-override distinction to call out. Now every failure path behaves identically regardless of how Semrush was enabled, so these titles read as if something is still being specially exempted. Fix: drop the parenthetical or reword to state the env flag and the override now behave identically on failure.

4. **No test covers the double-failure scenario this PR's own premise is meant to protect against.** When `enableSemrush:true` forces a genuine technical failure (not an entitlement skip) and the legacy fallback *also* returns no URLs for that site/window, the run now reports `success:true` with all-zero `urlCounts` and only a `fallbackReason` field as evidence anything went wrong - no test in the diff exercises this specific combination. Given the PR description's stated goal is "a Semrush hiccup never zeroes out offsite for that run," this exact edge case (offsite legitimately ends up zeroed anyway, just via double-failure rather than a hard stop) deserves a test that pins down what diagnostic signal is expected to survive.

## Recommendations

- Multiple reviewers independently raised the same rollout-validation concern: the hard-stop was the only mechanism that made an `enableSemrush:true` canary failure loud rather than just logged. Now that it's gone, a canary failure surfaces only as `outcome=degraded` in logs with `success:true` in the result. Worth confirming with whoever owns the LLMO-6709 rollout validation that this is sufficient for any remaining testing windows, or adding a lightweight metric/alert on that specific log line as a replacement signal - not necessarily in this PR, but worth tracking.
- Do a repo-wide `grep -ri "hard.stop"` across `docs/` and `src/` before merge to catch any remaining references beyond the ones enumerated above.

## Assessment

Reasoning: The behavioral change itself is correct, narrowly scoped, and well-tested for the core path, with no security or correctness defects found. It should not merge as-is because it leaves the ADR self-contradictory, a companion spec doc actively wrong, and stale comments in two other files describing a mechanism that no longer exists - all fixable in a follow-up pass to the same PR.

## Next Steps

1. Reconcile the ADR (Decisions 8/8b/8c) and the companion spec doc with the new always-fallback behavior.
2. Update the stale "hard-stop" comments in `semrush-entitlement.js` and `offsite-brand-presence-semrush.js`, and the outdated test titles.
3. Add a test for the double-failure (Semrush technical failure + empty legacy) scenario, or explicitly document why it's out of scope.

tb-adbe:
Thanks @amitruluimadalina — all four addressed in `14d2bb22`.

### Important

1. **ADR self-contradiction + spec doc still describing hard-stop** — ✅ Fixed.
   - ADR 002: reworded Decision 8's "exempted from the Decision 1 hard-stop" to the current contract (both entitlement reasons fall back to legacy like any failure, but log at `outcome=skip` vs a technical failure's `outcome=degraded`), fixed the "hard-stop-exemption check/contract" phrasing in 8b, and marked **8c as Historical** (the `enableSemrush:true` hard-stop was removed entirely — nothing hard-stops now), following the same `(History: …)` pattern as the Decision 1 rewrite.
   - Spec `docs/specs/2026-08-05-…:143`: acceptance criterion changed from "**hard stop** (`success:false`, no fallback)" to "falls back to legacy, same as the env flag," with a historical note.

2. **Stale "hard-stop" comments in untouched source** — ✅ Fixed in `semrush-entitlement.js` (the two `SEMRUSH_*_REASON(S)` doc comments), the loader (`offsite-brand-presence-semrush.js` `fallbackReason` comment), and — same class, thanks for prompting the sweep — `test/utils/semrush-entitlement.test.js:397`. All now describe the skip-vs-degraded logging contract.

3. **Stale test titles** — ✅ Reworded the three: "…(no hard stop)" → "…(same as the override)", and dropped "(never hard-stops)" from the two entitlement-skip titles (they still assert the fall-back, which is the meaningful part).

4. **Double-failure test** — ✅ Added: `enableSemrush:true` + a genuine technical failure (`domain_urls_failed`) + empty legacy → asserts `success:true`, `dataSource:'legacy'`, `fallbackReason` preserved, `urlCounts` zero, and no error-level log. This pins exactly the surviving diagnostic signal for the case where offsite legitimately ends up zeroed (gracefully, not via hard-stop). (There was already an ENV-var variant of this; this adds the Slack-override variant the premise is about.)

### Recommendations
- **Loud-canary signal** — acknowledged; this is the deliberate trade-off of the PR (fallback over hard-stop). Confirming with the LLMO-6709 rollout owner that `outcome=degraded` logs + `fallbackReason` are sufficient for remaining test windows, and will track a metric/alert on that log line as a follow-up if they want a louder replacement — out of scope for this PR.
- **Repo-wide `grep -ri "hard.stop"`** — done; the only remaining hits are the intentional historical notes above and one unrelated pagination comment (`offsite-brand-presence-postgrest.js:211`, "doesn't page as a hard stop").

Docs/comments/tests only — no behavior change. Affected suites green (215 across handler + entitlement + loader).

AndreiAlexandruParaschiv (approved):

amitruluimadalina (approved):
Hey @tb-adbe,

## Strengths

Previously flagged issues now addressed:

- The ADR (`docs/decisions/002-offsite-cited-urls-semrush-source.md`) is now internally consistent: Decision 1's "always falls back to legacy" and Decisions 8/8b/8c no longer describe two different worlds, with 8c explicitly marked historical and pointing back to Decision 1. Verified against the PR-branch `handler.js` that `hardStopOnFailure` and the old `success:false` early-return genuinely no longer exist anywhere in the file.
- The companion spec doc (`docs/specs/2026-08-05-offsite-cited-urls-semrush-source.md:143`) was corrected from a live "hard stop" acceptance criterion to "falls back to legacy," with a historical note.
- The stale comments in `src/utils/semrush-entitlement.js:55,71` and `src/utils/offsite-brand-presence-semrush.js:610` now describe the current skip-vs-degraded contract, verified consistent with the actual `outcome: OUTCOME.SKIP` vs `outcome: OUTCOME.DEGRADED` branching in the handler.
- All three stale test titles ("(no hard stop)" / "(never hard-stops)") were reworded to match current behavior, with no assertions changed underneath.
- The double-failure scenario (technical failure + empty legacy under `enableSemrush:true`) is now covered by a new test with meaningful assertions (`success:true`, `dataSource:'legacy'`, `fallbackReason` set, `urlCounts` zeroed, legacy loader called once) that would genuinely catch a regression in this combination.
- The "docs+test only, no behavior change" framing holds up under inspection: every touched line in the two source files is a comment; no exported symbol, conditional, or return value changed.
- No secrets, tokens, or sensitive data introduced anywhere in the diff.

## Issues

### Minor (Nice to Have)

1. **The new double-failure test's `expect(log.error).to.not.have.been.called` assertion is vacuous.** `test/audits/offsite-brand-presence.test.js` (new test, near line 479) - tracing `handler.js` end to end shows `log.error` is only ever emitted via `.failure(...)` calls, and there is no `.failure(...)` call anywhere on the Semrush-fallback code path. This means the assertion would pass identically whether the code correctly falls back or partially regresses toward the old hard-stop behavior - it doesn't discriminate the thing its inline comment ("no hard stop / failure-level log") claims to verify. The scenario is still meaningfully covered by the other assertions in the same test (`success`, `dataSource`, `fallbackReason`, `urlCounts`). Recommend dropping this line, or replacing it with an assertion on the actual `log.warn`/`outcome: OUTCOME.DEGRADED` call the handler does make on this path.
2. ADR Decision 8c's "Historical" framing preserves the entire original argument under a disclaimer rather than compressing it into a short historical note - for a decision record meant to be a durable artifact, a future reader still has to mentally subtract out reasoning about a mechanism that no longer exists. Not blocking, but this is the same class of accretion that produced the original stale-doc finding.
3. Minor terminology drift between two adjacent comment blocks in `src/utils/semrush-entitlement.js` ("the handler's fallback-logging check" vs "log an entitlement skip") describing the same concept with different phrasing.
4. The new double-failure test doesn't assert `mockLoadCitedUrlsFromSemrush` was called once, unlike the symmetric assertion it makes for `loadBrandPresenceData` - would make the "Semrush was genuinely attempted and failed" premise explicit rather than implicit in the mock wiring.

## Recommendations

- Tighten or remove the non-discriminating `log.error` assertion in the new test in a small follow-up, since it currently implies coverage it doesn't provide.
- Consider condensing ADR Decision 8c to a short historical footnote in a future pass, to avoid the same stale-doc pattern recurring.

## Assessment

Ready to merge? **Yes**
Reasoning: All four previously-flagged Important issues are genuinely resolved and independently verified against the real PR-branch code (not just the diff text), the new test meaningfully closes the coverage gap it targets, and the change is confirmed docs/comments/tests-only with no behavior or security impact. The remaining items are cosmetic follow-ups, not blockers.

PR #2972, opened 2026-09-14
feat(url-prompts): Semrush url-prompts enrichment over S2S (LLMO-6712)

tb-adbe:
## What

Rebases the WIP url-prompts enrichment from #2872 onto current `main` and rewires its auth/routing to the **S2S session-token flow** that landed since that PR was cut (#2965 / #2970 / #2971). Enriches the cited / youtube / reddit analysis handlers with per-URL prompts from the Semrush-backed Serenity `url-prompts` endpoint, gated on `enableSemrush`.

Supersedes #2872 (that PR was built on the pre-S2S raw-IMS-token design and no longer loads against `main`).

## Why the rebase wasn't mechanical

#2872's loader imported `SPACECAT_API_DEFAULT_BASE_URL` and `getAuthorizationHeader` from `offsite-brand-presence-semrush.js` — both **deleted** when that module moved to the S2S session-token model. The feature had to be re-plumbed onto the new auth, not just merged.

## Changes

**`offsite-brand-presence-semrush.js`** — three new shared exports so both loaders use one auth path and **one session-token cache** (no double-minting):
- `resolveApiBaseUrl(env)` — LLMO host + `/api/v1` prefix in one place.
- `getS2sSessionAuthorization({ context, imsOrgId })` — IMS mint → customer-scoped session-token exchange, cached by `imsOrgId`; throws with a `.reason` code.
- `evictS2sSessionToken(imsOrgId)` — drop a stale cached token on a data-call 401/403.

**`url-prompts-semrush.js`** — replaced the removed raw-IMS auth with the S2S flow; resolves the customer `imsOrgId` via `getImsOrgId`; builds requests on the LLMO base+prefix (fixes the Varnish-400 routing); dropped the obsolete `x-promise-token`; bounded the per-URL fan-out (up to 50) with a concurrency pool of 5; evicts the session token on 401/403. Still best-effort (never throws — enrichment must not fail the audit).

**Handlers (cited / youtube / reddit)** — kept `main`'s structured `olog` logging; preserved the feature wiring (`enableSemrush` plumbing, `isUrlFromSemrush` tagging, prompt attachment). `url-topic-enrichment.js` no longer clobbers Semrush-sourced prompts with legacy topic prompts.

## Contract verified against api-service `main` (v1.810.0)

- Route `GET /v2/orgs/:spaceCatId/brands/:brandId/serenity/brand-presence/url-inspector/url-prompts` → `elementsController.listUrlPrompts` — same controller family + S2S capability (`llmo/can_view`) as the already-working `domain-urls`.
- Same `authorizeBrandSubWorkspace` gate as `domain-urls` → the existing S2S token authorizes this identically.
- Request params (`url`, `startDate`/`endDate`, `platform=all`) and response shape (`{ prompts: [{ prompt, ... }] }`) match our call exactly.

## Testing

- 100% coverage on both source files and the three handlers; lint clean.
- Full affected suites green (loader, url-prompts, three handlers).

## Rollout notes

- S2S consumer registration already covers the `url-prompts` endpoint (confirmed).
- Service layer (`getUrlPrompts`) is flagged POC in api-service (unit tests deferred) but the endpoint is live-verified — worth a canary via `enableSemrush:true` before broad enablement.

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["@amitruluimadalina"]
```


amitruluimadalina (changes requested):
Hey @tb-adbe,

## Strengths

- Best-effort isolation is applied consistently: `fetchUrlPrompts` in `src/utils/url-prompts-semrush.js` never throws (network/non-2xx/parse errors all resolve to empty results), and the loader is capped, timeout-bounded, and concurrency-limited (pool of 5 over up to 50 URLs) so it can't burst the LLMO edge.
- The hardstop behavior actually matches the ADR's claim: `buildSemrushDebugHaltResult` (`src/utils/offsite-audit-utils.js`) returns `success:false`, and the Mystique dispatch post-processor short-circuits on falsy `success`, so nothing is sent to Mystique when `enableSemrushWithHardstop` fires.
- The new `enableSemrushWithHardstop` flag is threaded consistently and completely across all the hops that already carry `enableSemrush` (three analysis handlers, `requestOffsiteScrape`, `triggerAnalysisAudits`, `scheduleDrsStatusPoll`), with no positional-argument misalignment across call sites or tests.
- Token scoping is correct: the session-token cache in `src/utils/offsite-brand-presence-semrush.js` is keyed by the customer's `imsOrgId` (not `spaceCatId`), so one customer's requests can never receive another customer's cached token, and eviction on 401/403 is scoped to the same key with no cross-tenant risk.
- No credential leakage: the bearer token lives only in the `Authorization` header, never in a logged URL or query string, and `buildUrlPromptsUrl` correctly URL-encodes the audit-sourced `url` value so there is no query-injection or SSRF vector (the request host is always the trusted, pinned LLMO base URL).
- Test coverage on the new files is strong: `loadUrlPromptsFromSemrush`'s five early-return guards, `fetchUrlPrompts`'s per-outcome branches (network error, 401/403, other non-2xx, JSON parse failure, non-array body, prompt cap, falsy-row filtering), and token eviction (fires on 401/403, doesn't fire on network/other errors) are all individually tested.
- No new dependencies introduced, and no secrets or credentials in the new ADR (`docs/decisions/006-offsite-url-prompts-enrichment.md`), which is itself a clear, well-reasoned document (alternatives considered, consequences stated).

## Issues

### Important (Should Fix)

1. **The "never fails the audit" guarantee is not structurally enforced, and is untested for the failure case.** `src/utils/url-prompts-semrush.js` in `loadUrlPromptsFromSemrush`: the calls to `resolveBrandResultForSite`, `getDateWindowForPreviousWeeks(getPreviousWeeks())`, and `getImsOrgId` are all unguarded (no try/catch) - only the session-token call and the per-URL fetches are protected. If any of these three helpers ever rejects or throws (a data-access hiccup, brand-resolver error), the exception propagates out of `fetchStoreData` and fails the whole analysis audit, directly contradicting both the ADR ("must never fail... enrichment is optional metadata") and this file's own docstring. No test forces any of these three calls to reject, so the gap is invisible in the 100%-coverage claim (an unguarded await has no uncovered branch). Wrap the brand/date/IMS-org resolution in one top-level try/catch that logs and returns an empty `Map`, and add rejection tests for each.

2. **`src/utils/url-topic-enrichment.js` (~line 101-102) suppresses legacy topic prompts more broadly than intended, and ships with no dedicated test.** Every candidate URL (up to 50) is tagged `isUrlFromSemrush: true` whenever the flag is on, regardless of whether Semrush actually returned any prompts for it (per the handlers' own "assume Semrush for every URL" comment). The new guard `!urlItem.isUrlFromSemrush && match.prompts.size > 0` then drops legacy topic prompts for every one of those tagged URLs, including ones where Semrush returned nothing - a net prompt-coverage regression versus today's behavior, not just "don't clobber when both match." `test/utils/url-topic-enrichment.test.js` has no changes in this PR, so neither branch of the new guard is tested. Gate on actual prompt presence (e.g. `!urlItem.prompts?.length`) instead of the source tag, and add tests for both branches.

3. **The ADR's "one auth path" claim does not hold - `domain-urls` was not refactored onto the new shared function, and the sharing is unverified by any test.** `src/utils/offsite-brand-presence-semrush.js` exports the new `getS2sSessionAuthorization` for `url-prompts-semrush.js` to consume, but the existing `domain-urls` loader (`loadCitedUrlsFromSemrush`) still runs its own inline mint-exchange-cache-classify sequence rather than calling the new wrapper. Only the cache primitives, `resolveApiBaseUrl`, and the timeout are actually shared; the auth orchestration and 401/403 classification now live in two places that can silently drift. There's also no test confirming the headline "no double-mint" behavior: that a token cached via one entry point is reused by the other for the same `imsOrgId`. Refactor `domain-urls` to call `getS2sSessionAuthorization` (preserving its Slack-notify/consumer-claims logging via an injected hook), or correct the ADR's claim; add a cross-loader cache-reuse regression test. Given there are now two consumers, also consider extracting the shared S2S auth/cache/base-URL logic out of `offsite-brand-presence-semrush.js` (a module named for a single feature) into its own module (e.g. `offsite-s2s-auth.js`) so the boundary matches its actual scope.

4. **Third-party prompt strings are capped in count but not in length or type at the ingestion boundary.** `src/utils/url-prompts-semrush.js` in `fetchUrlPrompts`: `rows.map((row) => row?.prompt).filter(Boolean).slice(0, MAX_URL_PROMPTS)` caps the *count* of prompts per URL but not the *length* or *type* of each value - a non-string `row.prompt` (number/object) survives `Boolean()` and is stored/forwarded verbatim, and an oversized string from a misbehaving upstream isn't truncated at this boundary (only later, defensively, in the SQS payload path). Since `prompts` are third-party content that flows into stored audit data and downstream Mystique dispatch, coerce to string and cap length at ingestion (e.g. ~2-4 KB), rather than relying on the SQS byte-budget as an incidental backstop.

5. **The debug hardstop pattern (reporting as a failed audit) is accumulating as an operational habit without a tracked removal plan.** `buildSemrushDebugHaltResult` in `src/utils/offsite-audit-utils.js` is the second time this "report success:false to represent a debug skip" pattern ships (ADR 002 removed an earlier one). It pollutes audit-success dashboards/alerting unless every consumer special-cases `reason=semrush_debug_halt`, and because the flag threads through the self-heal loop, one flagged trigger fans a "failed audit" signal across cited/youtube/reddit simultaneously. If the audit framework supports a genuine non-failure terminal/skipped state, prefer that; otherwise file a tracked ticket now for the removal the ADR already promises, rather than leaving it aspirational.

6. **Hardstop tests don't assert the negative - that Mystique dispatch is actually skipped.** In each of `test/audits/cited-analysis/handler.test.js`, `reddit-analysis/handler.test.js`, and `youtube-analysis/handler.test.js`, the hardstop test cases assert the returned `success:false`/`reason`, but none assert that the Mystique/SQS dispatch call is never made after that return. Since the entire value of the hardstop is "nothing downstream runs," add that negative assertion so a future refactor that accidentally moves the return can't silently start dispatching.

### Minor (Nice to Have)

- JSDoc mismatch: the `@param {boolean} [enableSemrush]` doc comment on `fetchStoreData` in all three handlers (`cited-analysis/handler.js`, `reddit-analysis/handler.js`, `youtube-analysis/handler.js`) doesn't match the actual parameter name `runSemrush`.
- The ~18-line enrichment block and the hardstop `if` block are copy-pasted verbatim into all three analysis handlers; consider extracting a shared `enrichUrlsWithSemrushPrompts(...)` helper into `offsite-audit-utils.js` so future fixes (e.g. issue 2 above) don't need three edits.
- Enrichment always fans out against `MYSTIQUE_URLS_LIMIT` (50) regardless of a smaller per-run `forwardedUrlLimit`/`urlLimit` override, so a run configured for fewer URLs still issues up to 50 token-bearing requests and can enrich URLs that are never dispatched.
- Observability asymmetry: `getS2sSessionAuthorization` in `offsite-brand-presence-semrush.js` doesn't log a "token granted" line with decoded consumer claims the way the existing `domain-urls` inline path does, and it doesn't expose whether a returned token came from cache - both signals `domain-urls` already relies on for triage. Consider folding that logging into the shared wrapper along with the refactor in Important item 3.
- `MAX_URL_PROMPTS = 5` (`url-prompts-semrush.js`) is a hardcoded cap with no env override, unlike the sibling timeout knob.
- Minor inefficiency: `fetchUrlPrompts` maps and filters the full upstream `prompts` array before slicing to `MAX_URL_PROMPTS`; capping first would avoid unnecessary work on an oversized response.
- A URL entry missing `.url` produces a literal `url=undefined` query value in `buildUrlPromptsUrl`; harmless given encoding, but a `filter(Boolean)` guard would be cleaner.
- `mapWithConcurrency`'s order-preservation is asserted by a test that uses a url-keyed result map, so the actual ordering guarantee isn't exercised; and the "both `enableSemrush` and `enableSemrushWithHardstop` set simultaneously" combination isn't tested in the handler suites, though the boolean-OR logic makes it low risk.
- Bare `catch {}` blocks in `fetchUrlPrompts` swallow all errors into a generic `error` category with no captured context, which is defensible given the aggregate-only logging design but makes a systemic failure (e.g., a header-building bug) indistinguishable from a normal per-URL miss.
- No overall wall-clock budget on the enrichment phase: at concurrency 5 over up to 50 URLs, a slow (non-adversarial) upstream could consume up to 10 sequential request-timeout windows, a meaningful fraction of the Lambda's execution budget, before the rest of the audit runs.

## Recommendations

- Prioritize issue 1 (structural best-effort guarantee) and issue 3 (auth path consolidation) before this graduates from the current per-run debug-flag canary to broader `enableSemrush` enablement - both are foundational claims the ADR makes that don't currently hold.
- Add a note to ADR 006 documenting that turning on enrichment changes prompt provenance for the whole candidate set (not just Semrush-matched URLs), once issue 2 is resolved either way.
- Consider consolidating the growing set of hand-threaded per-run debug flags (`enableSemrush`, `enableSemrushWithHardstop`, `enableBrandProfile`, `urlLimit`) into a single overrides object passed through the handler chain, so adding the next flag is a field addition rather than a six-call-site signature change.

## Assessment

Ready to merge? With fixes
Reasoning: The core best-effort design, token scoping, and hardstop mechanics are sound and well tested for the happy path, but the "never fails the audit" and "one shared auth path" claims that the ADR and code comments make are not actually true today, and both are foundational enough (and easy enough to silently break further) that they should be fixed before this is enabled beyond a per-run debug canary.

## Next Steps

1. Address the enrichment resilience gap (issue 1) and the auth-path consolidation (issue 3) first, since both are foundational correctness claims.
2. Resolve the topic-prompt suppression regression (issue 2) and the third-party string validation gap (issue 4).
3. Decide on the debug-hardstop operational pattern (issue 5) and add the missing negative test (issue 6).
4. Minor items are optional and can follow in a fast-follow.

tb-adbe:
Thanks for the thorough review @amitruluimadalina — addressed in `17af94e4`. Per-item:

### Important

1. **Best-effort not structurally enforced** ✅ Fixed. `loadUrlPromptsFromSemrush` now wraps the brand / date-window / IMS-org resolution (not just the token + per-URL fetches) in a top-level try/catch → logs `reason=prerequisites_failed` and returns an empty `Map`, so a *thrown* prerequisite error can no longer propagate out of `fetchStoreData` and fail the audit. Added rejection tests for `resolveBrandResultForSite` and `getImsOrgId`.

2. **`url-topic-enrichment` over-suppression** ✅ Fixed. The guard now gates on actual prompt presence (`!urlItem.prompts?.length`) instead of the `isUrlFromSemrush` tag, so a candidate URL that Semrush returned nothing for keeps its legacy topic prompts — no coverage regression. Added both-branch tests, and an ADR note (§4b) on provenance.

3. **"One auth path" not true** ✅ Fixed (consolidated, not just re-worded). `getS2sSessionAuthorization` now returns `{ authorization, sessionToken, fromCache }`, and `loadCitedUrlsFromSemrush` (domain-urls) was refactored onto it — its richer telemetry (Slack notify, decoded consumer-claims success line gated on `fromCache`, data-call eviction) stays in the caller. Added a cross-loader cache-reuse regression test proving no double-mint across entry points. (Extracting to a standalone `offsite-s2s-auth.js` module is a reasonable follow-up; kept in place here to bound the diff.)

4. **Prompt strings not length/type-capped at ingestion** ✅ Fixed. `fetchUrlPrompts` now coerces each row to a non-empty string and truncates to `MAX_PROMPT_LENGTH` (4 KB), capped-first. Added tests (non-string dropped, oversized truncated).

5. **Debug-hardstop pattern / removal plan** ✅ Documented. Removal ships as a **follow-up cleanup PR under LLMO-6712** (same story) once url-prompts is confirmed working across all audits — ADR 006 Consequences now states this explicitly, and notes the framework has no first-class "skipped" terminal state (why the halt reuses `success:false`).

6. **Hardstop tests don't assert dispatch skipped** ✅ Fixed. Each hardstop test now also asserts `context.sqs.sendMessage` is never called after the short-circuit.

### Minor
- ✅ Fixed the `@param enableSemrush` → `runSemrush` JSDoc on `fetchStoreData` (all three handlers).
- ✅ Wall-clock: url-prompts default timeout is now **30s** (still tunable via the shared `OFFSITE_SEMRUSH_TIMEOUT_MS`; domain-urls stays 60s) to bound the sequential-batch worst case.
- Deferred (fast-follow, noted): shared `enrichUrlsWithSemrushPrompts` helper to dedupe the 3 handlers; `getS2sSessionAuthorization` cache-hit/consumer-claims logging; `MAX_URL_PROMPTS` env knob; the `urlLimit`-vs-50 fan-out; and the order-preservation / both-flags-set test gaps — happy to pick these up in the cleanup PR.

100% coverage on all touched source files; lint clean; 538 affected tests pass.

tb-adbe:
Follow-up: the remaining "minor / nice-to-have" items from the review are now also addressed (commits `3cb77284` + `5c0b722e`):

- **Shared enrichment helper** — the 3 analysis handlers now call one `enrichUrlsWithSemrushPrompts(...)` instead of a copy-pasted block.
- **urlLimit-scoped fan-out** — enrichment is capped at the run's resolved Mystique `urlLimit`, so a run scoped to fewer URLs no longer issues token-bearing requests for URLs dropped before dispatch.
- **`MAX_URL_PROMPTS` env knob** — `OFFSITE_URL_PROMPTS_MAX` (clamped `[1,50]`), same knob shape as the timeout.
- **Observability in the shared wrapper** — `getS2sSessionAuthorization` returns `{ authorization, sessionToken, fromCache }`; url-prompts now logs the decoded consumer identity + cache-hit state like domain-urls.
- **Order-preservation + both-flags-set** test gaps filled.
- **Standalone `offsite-s2s-auth.js` module** — the S2S auth/cache/base-URL logic is extracted out of `offsite-brand-presence-semrush.js` into its own module (both loaders depend on it), with its own dedicated test; the domain-urls loader test now mocks the auth module. Pure structural move, no behavior change.

100% coverage on all touched source; lint clean; full affected suites green.

amitruluimadalina (changes requested):
Hey @tb-adbe,

## Strengths

- Previously flagged: the "never fails the audit" guarantee, the topic-prompt suppression regression, the untyped/uncapped third-party prompt strings, the untracked debug-hardstop pattern, and the missing hardstop negative test are all still resolved and unaffected by the latest commits.
- Previously flagged: the "one auth path" claim is now not just true but cleanly architected. `src/utils/offsite-s2s-auth.js` is a genuinely self-contained module (only imports the IMS client and tracing fetch, no cycle) that both `domain-urls` and `url-prompts` now depend on for auth, cache, and base-URL resolution, with the non-auth `resolveSemrushTimeoutMs` correctly left behind in the feature module. This directly and fully addresses the module-boundary recommendation from the previous round.
- The move preserves every behavioral invariant verified: cache key (`imsOrgId`), TTL and its override, the three `.reason` classification codes, eviction scoping (401/403 only), and no token leakage into logs.

## Issues

### Important (Should Fix)

1. **No overall wall-clock budget on the enrichment phase (carried from last round, unchanged).** The per-request timeout was lowered to 30s, but there's still no aggregate deadline around the concurrency-5 fan-out over up to 50 URLs, so a stalling upstream can still consume roughly 300s in the worst case.

2. **Per-run debug flags remain hand-threaded through six-plus call sites (carried from last round, unchanged).** Not targeted by either of the latest commits.

### Minor (Nice to Have)

- **The auth-module extraction dropped five IMS-mint-leg unit tests without recreating them.** `test/utils/offsite-brand-presence-semrush.test.js` previously asserted the `SEMRUSH_S2S_*` to `IMS_*` env-var mapping, the `IMS_CLIENT_CODE` placeholder default and its `SEMRUSH_S2S_CLIENT_CODE` override (the defensive workaround at `src/utils/offsite-s2s-auth.js` around line 177-181), the login request's method/headers/body shape, and the lowercase-`token_type` Bearer-normalization path (line 188). All five were deleted in this commit and none were reproduced in the new `test/utils/offsite-s2s-auth.test.js`. The runtime code is unchanged and correct today, but a future env-mapping typo or a regression in the Bearer-normalization defense would now pass CI silently. Worth restoring these five assertions in `offsite-s2s-auth.test.js`, where `imsCreateFrom` is already stubbed.
- **The cross-loader "no double-mint across entry points" regression test was deleted and not replaced anywhere.** The prior round added a test proving a token minted by `domain-urls` is reused (not re-minted) by `getS2sSessionAuthorization` for the same `imsOrgId`. That specific test was removed as part of this refactor, and nothing today exercises the real shared cache across both real consumers - `offsite-s2s-auth.test.js` only tests the module in isolation, and both loader test suites now mock it. The extraction makes this scenario structurally safer (one module, one cache), so this is lower risk than before, but the explicit assertion that both entry points actually route through it is gone. Worth adding one integration test that leaves `offsite-s2s-auth.js` unmocked for both consumers and asserts a single login call across them.
- The same "pure structural move, no behavior change" commit actually changed a log field's key and value: the `data_acquisition_bp_data_semrush_read` start-log field was renamed from `baseUrl` (host root only) to `apiBaseUrl` (host plus gateway prefix) - verified directly in the diff. Any Splunk query or dashboard keyed on the old `baseUrl` field for that event will silently stop matching.
- The `session_token_auth_failed` classification's 401 branch is only exercised indirectly (via the loader), never as a direct unit test the way the 403 case is.
- A previously-present "omits tenantCount when tenants are absent" assertion for `decodeS2sConsumerClaims` wasn't carried over to the new test file.
- `url-prompts-semrush.js` still imports the non-auth `resolveSemrushTimeoutMs` from the domain-urls-named module; fine for now, worth promoting to a small shared config module if a third Semrush loader appears.
- Carried from last round, all still open: the hardstop `if` block remains duplicated across the three handlers (tracked for removal under LLMO-6712); non-2xx response bodies in `fetchUrlPrompts` are still left undrained; there's still no explicit ops-alerting note for sustained auth failure; url-prompts enrichment still isn't gated on the same Semrush entitlement check `domain-urls` performs; a missing `.url` entry still produces a literal `url=undefined`; the `isUrlFromSemrush` tag still conflates "was a candidate" with "was enriched"; `getDateWindowForPreviousWeeks` is still the one prerequisite call not directly tested as a thrower; the concurrency order-preservation test still asserts against a url-keyed map rather than array position; and the `enableSemrushWithHardstop: undefined` to key-absent contract is still only implicitly proven.

## Recommendations

- Restoring the two dropped sets of tests is good hygiene before this settles, even though neither blocks correctness today - both close gaps in coverage that a refactor introduced.
- Consider reverting or explicitly calling out the `baseUrl` to `apiBaseUrl` log-field change in the commit body, since "pure structural move" commits are exactly the ones reviewers and dashboards trust not to alter observable behavior.

## Assessment

Ready to merge? With fixes
Reasoning: The architectural work across this whole review cycle is now complete and sound - every original correctness finding is resolved, and the module-boundary cleanup specifically requested last round landed cleanly. What remains are two lower-urgency items carried from the previous round (wall-clock budget, flag threading) plus test-restoration housekeeping from the latest refactor.

## Next Steps

1. Address the wall-clock budget and debug-flag consolidation.
2. Restore the dropped tests and address the log-field change when convenient.
3. Remaining minor items are optional.

tb-adbe:
Thanks @amitruluimadalina — the housekeeping items are addressed in `4f5c314b`.

### Done
- **Restored the 5 dropped IMS-mint assertions** in `offsite-s2s-auth.test.js`: `SEMRUSH_S2S_*`→`IMS_*` env mapping, the `IMS_CLIENT_CODE` placeholder default + `SEMRUSH_S2S_CLIENT_CODE` override, the login request method/headers/body shape, and the Bearer-scheme normalization. Also added a **direct 401** classification test and the **`decodeS2sConsumerClaims` "no tenants → no tenantCount"** assertion.
- **Reverted the log-field change** — good catch, it was live on `main`. The `data_acquisition_bp_data_semrush_read` start log is back to `baseUrl` (host root), so dashboards keyed on it keep matching. (The extraction had inadvertently switched it to `apiBaseUrl`.)
- **Drain non-2xx bodies** in `fetchUrlPrompts` (best-effort) so the connection releases.
- **`.url` guard** — url-prompts skips entries with no usable url (no `url=undefined` request) and returns empty when none are usable; added a direct `getDateWindowForPreviousWeeks`-throws prerequisites test.
- **ADR 006**: ops-alerting note for sustained `session_token_auth_failed`.

### Cross-loader "no double-mint" test
Documented rather than re-added: at runtime both loaders import `getS2sSessionAuthorization` from the one `offsite-s2s-auth.js`, so they share a single module-level cache — the guarantee is now **structural**. A cross-loader esmock test can't actually prove it (esmock gives each `esmock()` call its own module instance/cache), so it's asserted on the shared implementation both consumers route through; I added a comment in the auth test pointing at this.

### Intentionally deferred (your prior calls / cleanup PR)
- **Wall-clock budget** — this was a deliberate decision last round (dropped the per-request default to 30s instead of adding an aggregate deadline); happy to add a phase deadline if you'd prefer it, but it's not an accidental gap.
- **Debug-flag threading consolidation** and the **hardstop removal** — both slated for the LLMO-6712 cleanup PR (the hardstop is temporary by design).

100% coverage on `offsite-s2s-auth.js` / `offsite-brand-presence-semrush.js` / `url-prompts-semrush.js`; lint clean; 555 affected tests pass.

amitruluimadalina (approved):
Hey @tb-adbe,

## Strengths

- All test-restoration claims in the new commit (`4f5c314b`) verified correct by direct inspection of assertions, not just names: the 5 restored IMS-mint tests, the direct 401 classification test, and the "omits tenantCount" test in `test/utils/offsite-s2s-auth.test.js` all assert exactly what the commit message claims.
- The `baseUrl` vs `apiBaseUrl` log-field revert (`src/utils/offsite-brand-presence-semrush.js:574-575`) is genuinely fixed - confirmed the start-log field is `baseUrl` (host root) again, matching what ops dashboards key on.
- `fetchUrlPrompts`'s non-2xx body-drain and the "skip entries without a usable url" guard (`src/utils/url-prompts-semrush.js:166-172`, `329-338`) are both correctly implemented and covered for their stated branches (drain-resolves, drain-rejects, mixed usable/unusable, all-unusable).
- The `offsite-s2s-auth.js` module extraction is a clean, cycle-free boundary (only imports `ImsClient` + `tracingFetch`) - the module-boundary recommendation from the previous round landed as promised.
- Credential hygiene throughout remains sound: bearer tokens never appear in logs/URLs, `imsConfigDiagnostics` never returns secret material, third-party prompt strings are type-checked and length-capped at ingestion before storage.
- Previously flagged: the "never fails the audit" guarantee, the topic-prompt suppression regression, the untyped/uncapped third-party prompt strings, the untracked debug-hardstop pattern, the missing hardstop negative test, and the "one auth path" module-boundary concern are all still resolved and unaffected by the latest commit.

## Issues

### Minor (Nice to Have)

1. `src/utils/url-prompts-semrush.js:170` - the non-2xx response body drain (`await Promise.resolve(response.text?.()).catch(() => {})`) has no size cap, unlike `readErrorBodySnippet` (`offsite-s2s-auth.js`, capped at `ERROR_BODY_SNIPPET_MAX`) used for the same class of call elsewhere in this codebase. This is an internal LLMO/Serenity proxy response rather than raw attacker-controlled input, so it is defense-in-depth rather than a live risk - worth reusing the capped reader when convenient.
2. `src/utils/offsite-s2s-auth.js:342-370` - `getS2sSessionAuthorization` has no in-flight request coalescing, so two concurrent calls for the same `imsOrgId` could each see a cache miss and mint independently. No current call site actually invokes the two loaders concurrently for the same org, so this is a latent gap rather than an active bug - worth hardening (cache the in-flight promise, not just the resolved token) before any future caller introduces concurrent access.
3. `test/utils/offsite-s2s-auth.test.js:140-144` - the cross-loader "no double-mint across entry points" regression test asked for in the previous round was replaced with an explanatory comment rather than restored. The comment's reasoning (a naive esmock test can't prove it) is accurate, but a test importing both loaders' real code without esmocking `offsite-s2s-auth.js` itself would still be buildable and would close the gap.
4. `test/utils/offsite-brand-presence-semrush.test.js` - the reverted `baseUrl`/`apiBaseUrl` log-field fix (a "pure structural move" that silently renamed an ops-dashboard-keyed field) has no regression test asserting the start-log payload's field name, so the same class of regression could reoccur on a future refactor with nothing to catch it.
5. `src/utils/url-prompts-semrush.js:379-389` - the aggregate summary log has no `authFailures` count (401/403 is folded into generic `non2xx`), unlike the auth-observability parity the docstring claims. Would make a future auth-scope mismatch between `domain-urls` and `url-prompts` harder to diagnose from logs alone.
6. No overall wall-clock budget on the enrichment fan-out in `url-prompts-semrush.js` - concurrency-5 over up to 50 URLs with only a per-request 30s timeout can consume roughly 300s worst case, alongside the DRS scrape round-trip already sharing the same Lambda's 900s budget. Worth an aggregate deadline before this graduates past the debug canary.
7. The per-run debug flags (`enableSemrush`, `enableSemrushWithHardstop`, `enableBrandProfile`, `urlLimit`) are hand-threaded through 6+ call sites each (`requestOffsiteScrape` is now up to 9 positional args). Since the ADR commits to removing `enableSemrushWithHardstop` once url-prompts is validated, this is a maintainability nice-to-have rather than a blocker - collapsing these into one overrides object would help if another flag is added before that removal lands.
8. `src/utils/offsite-brand-presence-semrush.js:574` still re-derives the LLMO host root inline instead of reusing it from `resolveApiBaseUrl` - the second time this file's host-resolution logic has been a source of drift.
9. `src/utils/url-prompts-semrush.js:336` - the "no usable url" early return is silent (no `olog.warn`), unlike every other prerequisite-failure branch in the same function.
10. `docs/decisions/006-offsite-url-prompts-enrichment.md` overclaims the double-mint guarantee as "covered by a cross-loader regression test" when the newest commit's own test comment says the opposite (structural, not tested) - worth a one-line wording fix.
11. `src/utils/url-prompts-semrush.js:291-338` - the S2S mint and brand/date resolution run before filtering out unusable URLs, so an all-unusable-URL run still performs a full mint for zero data calls.
12. `src/utils/url-prompts-semrush.js:329` - `filter(({ url }) => url)` would throw on a malformed non-object entry rather than skip it, which is a small gap against the comment's own stated "defense in depth" intent.

## Recommendations

- None of the above block this from shipping behind the current debug-hardstop canary. Worth picking off items 1-6 as a quick follow-up before broader `enableSemrush` enablement, since a couple of them (the missing regression tests in items 3 and 4) are exactly the kind of gap that let a real bug (the log-field rename) slip through once already.

## Assessment

Ready to merge? Yes
Reasoning: The new commit delivered everything its message claimed - all restored tests assert what they say, and both small behavior fixes (body drain, url filtering) are correct. Everything left open is test-coverage debt, observability nicety, or forward-looking hardening, with no active defect in any currently-reachable code path.

## Next Steps

1. Minor items are optional and can follow in a fast-follow PR under LLMO-6712.

PR #2976, opened 2026-09-15
fix(preflight): preserve completed readability guidance (SITES-51778)

bhellema:
## Summary

- preserve an asynchronously completed Preflight `AsyncJob` during the outer handler final persistence
- prevent intermediate audit-result saves from overwriting enriched readability output after guidance completes
- add deterministic regressions for both persistence race windows

## Root cause

The readability handler queued Mystique guidance and returned `{ processing: true }`. If all guidance callbacks completed before the outer Preflight handler final write, the callback persisted enriched readability results and marked the job `COMPLETED`, but the outer handler then replaced that result with its stale local snapshot and reset the status to `IN_PROGRESS`.

## Fix

Both final and intermediate Preflight persistence reload the job and skip stale result writes when async guidance has already marked it `COMPLETED`. Normal in-progress and synchronous completion behavior remains unchanged.

## Jira

[SITES-51778](https://jira.corp.adobe.com/browse/SITES-51778)

## Validation

- `npx mocha test/audits/preflight.test.js` — 152 passing
- focused race regressions — 2 passing
- `npx eslint src/preflight/handler.js src/preflight/utils.js`
- `git diff --check`

## Change Management

```yaml
cm-assessment: v1
changeType: standard
impact: unnoticeable
risk: minor
changeApprovedBy: ["Radhika Gupta"]
```


src/preflight/handler.js
```diff
@@ -457,17 +457,21 @@ export const preflightAudit = async (context) => {
...
+      } else {
+        jobEntity.setStatus(AsyncJob.Status.COMPLETED);
+        jobEntity.setEndedAt(new Date().toISOString());
+      }
+      await jobEntity.save();
     }
-    await jobEntity.save();
   } catch (error) {
     log.error(`[preflight-audit] site: ${site.getId()}, job: ${jobId}, step: ${step}. Error during preflight audit.`, error);
     const jobEntity = await AsyncJobEntity.findById(jobId);
```
radhikagpt1208, line 477: This fix guards the two *success* paths (final persistence above, and `saveIntermediateResults` in `utils.js`) against overwriting a job that async readability guidance already completed — but this `catch` block isn't guarded the same way.

If an unrelated error occurs anywhere earlier in `preflightAudit` (e.g. a different handler throws) at the same moment the readability guidance callback races in and marks the job `COMPLETED` with its enriched result, the next lines here will unconditionally `setStatus(FAILED)` → `save()`, discarding the completed/enriched result and reproducing the exact race this PR sets out to fix — just on the failure path instead of the success path.

Suggest applying the same `jobEntity.getStatus() === AsyncJob.Status.COMPLETED` guard here (skip the FAILED overwrite, just log) before `setStatus(FAILED)`/`save()`. Also seems worth a regression test analogous to `preserves a job completed by guidance before final persistence`, but for the error path.

src/readability/preflight/guidance-handler.js
```diff
@@ -38,8 +49,90 @@ function mapMystiqueSuggestionsToOpportunityFormat(mystiquesuggestions) {
...
+    throw new Error('[readability-suggest guidance]: Missing S3 context for response accumulation');
+  }
+
+  const responsePrefix = `${READABILITY_PREFLIGHT_RESPONSE_PREFIX}/${auditId}/`;
+  const responseKey = `${responsePrefix}${encodeURIComponent(messageId)}.json`;
+
+  await s3Client.send(new PutObjectCommand({
+    Bucket: bucketName,
+    Key: responseKey,
+    Body: JSON.stringify({ messageId, mappedSuggestions }),
```
radhikagpt1208, line 71: These per-response objects (and the `.lock` completion object) are only ever deleted on the successful completion path (the `DeleteObjectsCommand` at the end). Any job that never reaches `mystiqueResponsesReceived >= mystiqueResponsesExpected` — a crashed/abandoned callback, an expired `AsyncJob`, or a Mystique response that simply never arrives (the exact failure mode this whole readability saga has been about) — leaves `preflight-readability/responses/**` and `preflight-readability/completions/**` orphaned in the bucket indefinitely.

Could we confirm the `S3_MYSTIQUE_BUCKET_NAME` bucket has an S3 Lifecycle rule that expires these two prefixes (say after a few days)? If not, this will slowly accumulate cost and clutter. A lifecycle policy is also a cheap safety net that bounds the blast radius of any future stuck-job bug.

src/readability/preflight/guidance-handler.js
```diff
@@ -175,14 +253,42 @@ export default async function handler(message, context) {
...
+        }));
+        const claimedAt = Date.parse(await claim.Body.transformToString());
+        if (Number.isFinite(claimedAt) && Date.now() - claimedAt > 5 * 60 * 1000) {
+          await s3Client.send(new DeleteObjectCommand({
+            Bucket: env.S3_MYSTIQUE_BUCKET_NAME,
+            Key: completionKey,
+          }));
+          throw new Error(`[readability-suggest guidance]: Removed stale completion claim for AsyncJob ${auditId}; retrying`);
+        }
+        throw new Error(`[readability-suggest guidance]: Completion is already in progress for AsyncJob ${auditId}`);
```
radhikagpt1208, line 279: The completion-claim lock's correctness depends on the queue redelivering this message. When a concurrent callback holds the claim, this path `throw`s (`Completion is already in progress`) rather than completing — and similarly the stale-claim branch throws `Removed stale completion claim; retrying`. Both rely on the SQS consumer treating the throw as a processing failure and redriving the message so the completion is retried later. If it isn't redelivered, the responses are all in S3 but the `AsyncJob` is never marked `COMPLETED` — i.e. stuck `IN_PROGRESS`.

Can we confirm:
1. This handler's queue has a redrive policy (or the throw reliably triggers redelivery), so a contended/stale-claim throw is actually retried rather than dropped, and
2. The visibility timeout is longer than the time between acquiring the claim and the final `save()` — otherwise the 5-minute staleness check could race against a legitimately in-flight completion and prematurely delete a valid claim.

If the queue doesn't redrive on throw, the contended callback needs to reschedule itself explicitly instead of just throwing.

src/readability/preflight/guidance-handler.js
```diff
@@ -175,14 +253,42 @@ export default async function handler(message, context) {
   const allResponsesReceived = updatedReadabilityMetadata.mystiqueResponsesReceived
     >= updatedReadabilityMetadata.mystiqueResponsesExpected;
   if (allResponsesReceived && updatedReadabilityMetadata.mystiqueResponsesExpected > 0) {
```
radhikagpt1208, line 255: Bigger-picture question on the delivery model, since this PR is reworking exactly this path.

Today readability suggestions are **all-or-nothing**: intermediate callbacks only buffer their response to S3, and the job's visible `result` / `status` are written exactly once — here, when `mystiqueResponsesReceived >= mystiqueResponsesExpected`. Until the final response lands, the MFE polls and sees every readability opportunity still in `processing` with no `aiSuggestion`; then the whole set appears at once. (This isn't new — `main` behaves the same way — but this PR is the natural place it's being touched.)

The downside is that a **single** lost or errored Mystique response withholds the *entire* readability result, not just that one paragraph — which is precisely the "stuck `IN_PROGRESS` forever" symptom this whole saga has been chasing. A progressive model — merge each response into the result as it arrives, and complete on a count **or** a timeout — would degrade gracefully: one missing response costs one suggestion instead of all of them, and the job still terminates.

is moving to progressive/streaming delivery (incremental `setResult` per response + a completion timeout) something you'd want to fold into this PR while we're in here, or is it better scoped as a follow-up? I lean follow-up given the blast radius, but wanted your call since the buffering rewrite here already gets us most of the way there.

radhikagpt1208 (approved):

src/readability/preflight/guidance-handler.js
bhellema: There is a 365 day retention on the bucket, I check this out last night.

src/readability/preflight/guidance-handler.js
bhellema: Agreed this is a real resilience improvement, but it should be addressed in V2 rather than folded into this race-fix PR. Progressive delivery needs a concurrency-safe result merge, explicit partial-result semantics, MFE validation, and a durable timeout reconciler because the callback handler cannot finalize a job when no callback arrives. This PR will remain scoped to preventing completed/enriched jobs from being regressed or overwritten; V2 should add progressive response delivery and bounded completion so one missing response does not hold back all successful suggestions.

PR #2984, opened 2026-09-16
fix(metatags): validate Genvar suggestion length against audit ideal range (SITES-42023)

ckkovac:
## Problem

[SITES-42023](https://jira.corp.adobe.com/browse/SITES-42023) — the Genvar auto-suggest response was persisted verbatim, so a meta description shorter than the audit's ideal minimum (e.g. **134 chars** vs the required **140–160**) was shipped and then immediately re-flagged as "Description too short" on the next Preflight run. Reported by a customer (Dover Corporation) via the ASO Preflight team.

## Change

In `src/metatags/metatags-auto-suggest.js`:

- Added `isSuggestionLengthValid(tagName, suggestion)` mirroring `seo-checks.checkForTagsLength` — a suggestion clears the audit only when its length is in the ideal range (`idealMinLength..idealMaxLength`: description 140–160, title 40–60, h1 ≤70).
- After the Genvar call, any suggestion outside that range is **logged as a `log.warn`** (for Splunk visibility) but the best-effort suggestion is still shipped.

## Why validate-and-warn (not retry)

Genvar runs at **temperature 0** (deterministic), so re-requesting an out-of-range suggestion returns the identical text — a retry could never produce a different, in-range result and would only add latency/cost. This PR is therefore a **backstop + observability signal**, not the fix on its own.

## The actual fix lives in the Genvar service

The durable fix tightens the Genvar service's own validation bands so its existing regenerate-with-feedback loop fires. That's a separate PR against `OneAdobe/aem-genai-variations-appbuilder` (tracked in SITES-42023) — this PR's `log.warn` measures how often that path is needed.

## Verification

- `metatagsAutoSuggest` tests pass (27), including new validate/warn cases; lint clean.
- **100%** line/branch/statement/function coverage on the changed file.

## Related

- SITES-42023 (this bug)
- SITES-51838 (parked Mystique/blackboard-path durable fix)
- SITES-51837 (Concern 2 — content-quality validation, cross-opportunity)


ckkovac:
Durable fix (the Genvar service side) is now open: OneAdobe/aem-genai-variations-appbuilder#155 — it tightens the service's own validation bands so its existing regenerate loop fires. This PR is the audit-worker backstop + `log.warn` observability; #155 is what actually fixes the customer issue.

clotton (approved):
**Reviewed — LGTM ✅** (approving)

Verified the core correctness claim: `isSuggestionLengthValid` faithfully mirrors `seo-checks.checkForTagsLength`'s "healthy" band, so a shipped suggestion warns exactly when the next audit would re-flag it:

- **description** → valid iff `140 ≤ len ≤ 160` (idealMin/idealMax) ✓
- **title** → valid iff `40 ≤ len ≤ 60` ✓
- **h1** → no `idealMinLength`, so the `?? 1` fallback enforces "non-empty" with an upper bound of `70`, which matches the audit's effective h1 rule (empty → issue; `> idealMaxLength` → issue) ✓

Other things I checked:
- The `log.warn` is purely additive — the best-effort suggestion is still shipped unconditionally, so there's no behavior change to the persisted output. Non-breaking. 👍
- `tagIssueData.aiSuggestion.length` in the warn is safe (guarded by the truthy `aiSuggestion` check above it).
- Extracting `generateGenvarSuggestions` is a clean no-op refactor; error handling/throw semantics are preserved.
- Rationale for validate-and-warn over retry (Genvar at temperature 0 → deterministic) is sound; agree a retry here would only add latency/cost.
- CI green (build, codecov/patch), and the added tests cover in-range / below / above / undefined and the h1 no-lower-bound case.

Non-blocking notes (no change required):
1. The warn fires once per out-of-range tag; on a large site that could be a bit chatty in Splunk — but that's the intended signal for measuring how often the Genvar-service fix (SITES-42023) is needed, so 👍.
2. Since this PR is explicitly the backstop/observability half, worth making sure the durable `OneAdobe/aem-genai-variations-appbuilder#155` band-tightening lands before we lean on the `log.warn` volume as the success metric.

Thanks for the clear PR description and the linked follow-ups (SITES-51838 / SITES-51837).

PR #2992, opened 2026-09-17
revert(offsite): drop extra YouTube query params, keep only video id (LLMO-6712)

tb-adbe:
## Context

#2983 preserved every YouTube URL query param byte-identically, on the theory that Semrush's `url-prompts` `CBF_source` exact-match needed them. Root-cause investigation since (production testing + a direct `domain-urls`/`url-prompts` debug script) showed that theory was wrong:

- A direct A/B test against a live, data-rich brand showed `url-prompts` returns prompts **with or without** extra params (`t=`, `list=`, `vl=`, …) — they don't determine a match/no-match split, just which exact citation-instance's prompt set comes back.
- The actual root cause of the reported "0 prompts" was **URL store staleness**: `youtube-analysis` (and cited/reddit) enrich whatever URLs are already sitting in the URL store (sorted by `createdAt` desc); an existing URL is never refreshed on subsequent `offsite-brand-presence` runs (`if (existingUrlSet.has(entry.url)) { return entry.url; }` — a pure skip). If citation trends have shifted since a URL was first stored, its citations may no longer fall in `url-prompts`' rolling ~2-week query window, even though the URL is well-formed. This is a *data-freshness* gap, not a param-stripping bug — confirmed live: triggering `offsite-brand-presence` (with `enableSemrush`) to refresh the store, then re-running `youtube-analysis`, went from `urlsWithPrompts=0` to `urlsWithPrompts=37/50`.

## Change

This reverts #2983 (`git revert c6705037`, clean, no conflicts — nothing since has touched these files). Restores the #2981 behavior: `normalizeYoutubeUrl` keeps only `v=<id>` for watch URLs (drops `t=`, `list=`, tracking tokens, etc.) and strips all query params for other YouTube URLs (`youtu.be`, shorts) — same as before, just simpler/cleaner stored URLs, since we've now confirmed param preservation wasn't required for correctness.

The `watch`-vs-`youtu.be` **form** itself is still preserved (not rewritten) — that part of #2981/#2983 was correct and unaffected by this revert; dedup-by-video-id in the loader still reconciles the two forms.

## Testing

430 tests passing across the affected suites; lint clean. Test expectations are restored to #2981's exact values (this revert also cleanly undoes #2983's test changes).


tb-adbe:
Superseded by #2993 (merged) — same revert, landed via GitHub's native revert button. Closing.

